PR #5, opened 2017-07-01
started on tests

chriscoey:
lots of tests to add
can also add integer conic tests etc with Pajarito later

will fail tests for now

### DIFF
--- a/test/contconic.jl
+++ b/test/contconic.jl
@@ -1,7 +1,7 @@
 # Continuous conic problems
 
-function contconictest(solver::AbstractSolver, eps=Base.rtoldefault(Float64))
-    @testset "Set up model, variables, constraint, objective" begin
+function contconictest(solver::AbstractSolver, ε=Base.rtoldefault(Float64))
+    @testset "Linear model" begin
         # linear conic problem
         # min -3x - 2y - 4z
         # st    x +  y +  z == 3
@@ -21,9 +21,7 @@ function contconictest(solver::AbstractSolver, eps=Base.rtoldefault(Float64))
 
         setattribute!(m, Sense(), MinSense)
         setobjective!(m, 1, 0, v, [-3 -2, -4])
-    end
 
-    @testset "Solve model, check results" begin
         optimize!(m)
 
         @test cangetattribute(m, TerminationStatus())
@@ -35,16 +33,16 @@ function contconictest(solver::AbstractSolver, eps=Base.rtoldefault(Float64))
         @test getattribute(m, DualStatus()) == FeasiblePoint
 
         @test cangetattribute(m, ObjectiveValue())
-        @test isapprox(getattribute(m, ObjectiveValue()), -11, atol=eps)
+        @test getattribute(m, ObjectiveValue()) ≈ -11 atol=ε
 
         @test cangetattribute(m, VariablePrimal(), v)
-        @test isapprox(getattribute(m, VariablePrimal(), v), [1, 0, 2], atol=eps)
+        @test getattribute(m, VariablePrimal(), v) ≈ [1, 0, 2] atol=ε
 
         @test cangetattribute(m, ConstraintDual(), c)
-        @test isapprox(getattribute(m, ConstraintDual(), c), [3, 1], atol=eps)
+        @test getattribute(m, ConstraintDual(), c) ≈ [3, 1] atol=ε
 
         # TODO var dual and con primal
     end
 
-    # TODO more test sets here
+    # TODO more models
 end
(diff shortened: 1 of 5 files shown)


chriscoey:
@mlubin should we export everything? (then we don't have to prefix all the functions in test)

mlubin:
@jrevels, any guidance with the exporting question? My understanding is that it's preferable not to export anything at all, but it would be nice for the tests to get everything in scope.

test/contconic.jl
```diff
@@ -0,0 +1,50 @@
...
+        vc = addconstraint!(m, v, NonNegative(3))
+        c = addconstraint!(m, [3, 2], -[1 1 1; 0 1 1], Zero(2))
+        @test getattribute(m, ConstraintCount()) == 4
+
+        setattribute!(m, Sense(), MinSense)
+        setobjective!(m, 1, 0, v, [-3 -2, -4])
+    end
+
+    @testset "Solve model, check results" begin
+        optimize!(m)
```
mlubin: I don't think ``m`` will be in scope here.

test/contlinear.jl
```diff
@@ -0,0 +1,98 @@
...
+        # min -x
+        # st   x + y <= 1   (x + y - 1 ∈ NonPositive)
+        #       x, y >= 0   (x, y ∈ NonNegative)
+
+        m = Model(solver)
+
+        v = addvariables!(m, 2)
+        @test getattribute(m, VariableCount()) == 2
+
+        @test cansetattribute(m, VariableLowerBound())
```
mlubin: We don't do ``VariableLowerBound`` anymore. We should add a variablewise constraint with the GreaterThan set.

test/runtests.jl
```diff
@@ -1,5 +1,17 @@
-using MathOptInterface
-using Base.Test
+using MathOptInterface, Base.Test
+# TODO using solvers?
 
-# write your own tests here
-@test 1 == 1
+include("contlinear.jl")
+@testset "Continuous linear problems"
+    contlineartest(GLPKSolverLP())
```
mlubin: No reason to list solvers here until they implement MOI

test/contconic.jl
```diff
@@ -0,0 +1,50 @@
...
+        @test cangetattribute(m, TerminationStatus())
+        @test getattribute(m, TerminationStatus()) == Success
+
+        @test cangetattribute(m, PrimalStatus())
+        @test getattribute(m, PrimalStatus()) == FeasiblePoint
+        @test cangetattribute(m, DualStatus())
+        @test getattribute(m, DualStatus()) == FeasiblePoint
+
+        @test cangetattribute(m, ObjectiveValue())
+        @test isapprox(getattribute(m, ObjectiveValue()), -11, atol=eps)
```
joaquimg: What if we use `@test getattribute(m, ObjectiveValue()) ≈ -11 atol=ε`?

chriscoey: yup that seems nicer

mlubin: We were using ``isapprox`` previously only for compatibility with 0.5.

chriscoey:
OK i fixed those comments except for prefixing the functions, as the issue of whether to export is unresolved

mlubin:
Let's not export things for now. Just put ``const MOI = MathOptInterface`` at the top of the tests and use the ``MOI.`` prefix.

mlubin:
Also, tests need to be passing on travis before we can merge, otherwise the docs won't update.

joaquimg:
Should we have a minimal TODO list to track what needs to accomplished before starting to work on solver APIs?

jrevels:
> Just put const MOI = MathOptInterface at the top of the tests and use the MOI. prefix.

This kind of thing is generally my preference. Alternatively, you can just put the list of bindings you were going to export as the list of things you're `using` in the test namespace. Definitely feels like a bad idea to export everything.

chriscoey:
Thanks Jarrett!

chriscoey:
not updated for new functions. @mlubin perhaps you could update with how you intend the function syntax to be used then merge

test/contlinear.jl
```diff
@@ -7,22 +7,22 @@ function contlineartest(solver::MOI.AbstractSolver, ε=Base.rtoldefault(Float64)
         # st   x + y <= 1   (x + y - 1 ∈ NonPositive)
         #       x, y >= 0   (x, y ∈ MOI.NonNegative)
```
joaquimg: Which one are we using in comments? I vote for not using `MOI.` in comments

mlubin: Yeah, we don't need MOI there

chriscoey:
My bad, that was find and replace

PR #7, opened 2017-07-01
change model to instance, separate solver and instance attributes

### LINKED ISSUE #6: separate model and solver attributes
only 3 overlap, and they don't really make sense (per discussion with @mlubin). so separate these attributes.

also from discussion with @mlubin: define an interface for asking a solver whether a group of "types" of constraints is supported together. and also a way to ask the solver, given an existing model that has already been solved, what types of constraints can be added to it. EDIT moved to #8

there is confusion on my part on what a model is and what a solver is. I think perhaps the terminology could be revisited to make it clearer.

chriscoey:
fixes #6

### DIFF
--- a/src/MathOptInterface.jl
+++ b/src/MathOptInterface.jl
@@ -5,43 +5,43 @@ module MathOptInterface
 
 Abstract supertype for "solver" objects.
 A solver is a lightweight object used for selecting solvers and parameters.
-It does not store any instance data.
+It does not store any solver instance data.
 """
 abstract type AbstractSolver end
 
 """
-    AbstractInstance
+    AbstractSolverInstance
 
 Abstract supertype which represents a solver's in-memory representation of an optimization problem.
 """
-abstract type AbstractInstance end
+abstract type AbstractSolverInstance end
 
 """
-    Instance(solver::AbstractSolver)
+    SolverInstance(solver::AbstractSolver)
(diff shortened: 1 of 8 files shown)


mlubin:
I'm leaning towards ``SolverInstance`` for clarity.

joaquimg:
I like 'Instance', but if you need clarity why not 'ModelInstance'?
It seems reasonable, since the one writes a model to describe a mathematical program and when specific numbers are set then we have a instance of a model.
'SolverInstance' might be confuse with 'Solver'.

mlubin:
The important point to communicate is that the MOI "model" should be thought of as the solver's representation of the instance loaded into its API. It's not just the problem data itself, independent of a solver.

chriscoey:
OK I'll change it to `SolverInstance`

src/MathOptInterface.jl
```diff
@@ -5,43 +5,43 @@ module MathOptInterface
...
-    optimize!(m::AbstractModel)
+    optimize!(m::AbstractSolverInstance)
 
 Start the solution procedure.
 """
 function optimize! end
 
 """
-    freemodel!(m::AbstractModel)
+    freesolver instance!(m::AbstractSolverInstance)
```
mlubin: this method name need to be fixed

PR #10, opened 2017-07-03
reordered apireference.md to clarify basis statuses

chriscoey:


src/MathOptInterface.jl
```diff
@@ -31,7 +31,7 @@ Start the solution procedure.
 function optimize! end
 
 """
-    freesolver instance!(m::AbstractSolverInstance)
+    freesolverinstance!(m::AbstractSolverInstance)
```
mlubin: I'd vote to rename this to ``free!``

chriscoey: i was thinking the same

chriscoey: changing it

docs/src/apireference.md
```diff
@@ -96,47 +96,18 @@ The value of the attribute is of type `ResultStatusCode`.
 ResultStatusCode
 ```
 
+## Variables and Constraints
+
 ### Basis Status
 
-[what? BasisStatus]
+The `BasisStatus` attribute of a variable or constraint indicates its presence in a basis.
```
mlubin: There's more than one way in which a variable/constraint may be *not* in the basis (i.e., at an upper or lower bound), so this wording is a bit unclear.

chriscoey: reasonable

chriscoey: fixed but i still think more thought needs to go into these. the basis status description in attributes.jl is not super clear. eg what if it's not an optimal primal solution, but rather a ray

Issue #9, opened 2017-07-01
an Instance type


mlubin:
It would be nice if we had a data type to store all the data for an instance in MOI format independent of a solver API. This will make it more natural to talk about performing transformations, and we could also revive ``loadproblem!`` which could be more efficient for throwing everything to a ``SolverInstance`` at once.

Should the data structure just be a dictionary with vectors of variablewise, affine, and quadratic constraints (by type)? Do we care about type stability?

mlubin:
@ccoffrin @odow @blegat @joaquimg @chriscoey @joehuchette @jrevels

joehuchette:
Type stability might be meaningful here. On the other hand, a dictionary would be nice, since it would be more natural to add new constraint types.

joaquimg:
You are thinking about keeping all the instance data in the struct, including matrix coefficients?
I so I would be a little worried about keeping the same info in 3 places (JuMP, MOI and Solver).
But if all the data you mean all the references I think its a nice idea.
I prefer not to have a dictionary, arrays are so much faster... In my opinion, MOI should be lightweight and JuMP should make everything pretty.
If type stability is important here we should care about it so that MOI has minimal overhead.

mlubin:
> You are thinking about keeping all the instance data in the struct, including matrix coefficients?

Yes, everything.

JuMP will need an internal data structure to represent an MOI instance. I'm proposing to make this into a standard, documented data structure at the MOI level instead of at the JuMP level, because it will be useful outside of just JuMP. This data structure will replace ``linconstr``, ``quadconstr``, ``sosconstr``, etc., so there will still only be two copies of the model.

joaquimg:
> so there will still only be two copies of the model

This is good.

> throwing everything to a `SolverInstance` at once

I like it.

Do you have a sketch?

Do you have more pros and cons on the top of the head?

mlubin:
I think a reasonable interface for this data structure would be the existing getters and setters like ``addconstraint!``, ``getconstraintaffine``, ``getconstraintquadratic``, etc. so that we have the ability to easily modify the internal representation later on.

Conceptually we'll have the following high-level objects:
- ``Solver``: the existing solver objects like ``GurobiSolver()``
- ``SolverInstance``: an instance whose data belongs to the solver API, supports ``optimize!``, getting solutions back, etc.
[...]

PR #12, opened 2017-07-03
introduce a function concept

mlubin:
@blegat's comment [here](https://github.com/JuliaOpt/MathOptInterface.jl/issues/9#issuecomment-312516818) exposes the fact that we currently have 6 types of functions with different representations, between the combinations of variablewise, affine, quadratic, and scalar-valued, and vector-valued. Instead of enumerating all of these combinations in ``addconstraint!``, I think it makes sense to encode these with function objects. 

With this, the signatures of ``addconstraint!`` and ``setobjective!`` become:
```julia
addconstraint!(m::AbstractSolverInstance, func::F, set::S)::ConstraintReference{F,S}
setobjective!(m::AbstractSolverInstance, func::F, N::Int=1)
```
where `func` is a simple `struct` with the corresponding data.

``func`` could be, e.g.:
```julia
"""
    ScalarVariablewiseFunction(variable)

The function that extracts the scalar variable referenced by `variable`, a `VariableReference`.
This function would naturally be used for single variable bounds or integrality constraints.
"""
struct ScalarVariablewiseFunction <: AbstractFunction
    variable::VariableReference
end

"""
    VectorQuadraticFunction{T}(affine_variables, affine_coefficients, quadratic_rowvariables, quadratic_colvariables, quadratic_coefficients, constant)

The vector-valued quadratic function with i`th` component ("output index") defined as ``\\frac{1}{2}x^TQ_ix + a_i^T x + b_i``, where:
* ``a_i`` is a sparse vector specified in tuple form by the subset of `affine_variables, affine_coefficients` for the indices `k` where `affine_outputindex[k] == i`.
* ``b_i`` is a scalar specified by `constant[i]`
* ``Q_i`` is a symmetric matrix is specified in triplet form by the subset of `quadratic_rowvariables, quadratic_colvariables, quadratic_coefficients` for the indices `k` where `quadratic_outputindex[k] == i`

Duplicate indices in ``a_i`` or ``Q_i`` are accepted, and the corresponding coefficients are summed together. "Mirrored" indices `(q,r)` and `(r,q)` (where `r` and `q` are `VariableReferences`) are considered duplicates; only one need be specified.
"""
struct VectorQuadraticFunction{T} <: AbstractFunction
    affine_outputindex::Vector{Int}
    affine_variables::Vector{VariableReference}
    affine_coefficients::Vector{T}
    quadratic_outputindex::Vector{Int}
    quadratic_rowvariables::Vector{VariableReference}
    quadratic_colvariables::Vector{VariableReference}
    quadratic_coefficients::Vector{T}
    constant::Vector{T}
end
```

It's still a TODO to rethink the modification and query routines under this function concept, but I believe we'll be able to unify querying and modification across constraints and objectives.

I think this change is already justified as a simplification of the current interface as it is, but in addition to that it also makes everything more extensible for other kinds of functions as requested by @chriscoey.

@joehuchette @ccoffrin @odow @joaquimg

### DIFF
--- a/docs/src/apireference.md
+++ b/docs/src/apireference.md
@@ -40,7 +40,7 @@ ReturnsDuals
 SupportsAddConstraintAfterSolve
 SupportsDeleteConstraint
 SupportsAddVariableAfterSolve
-SupportsQuadraticObjective
+SupportsObjective
 SupportsConicThroughQuadratic
 ```
 
@@ -62,9 +62,7 @@ List of solver instance attributes
 RawSolver
 Sense
 NumberOfVariables
-NumberOfVariablewiseConstraints
-NumberOfAffineConstraints
-NumberOfQuadraticConstraints
+NumberOfConstraints
 ResultCount
 ObjectiveValue
 ObjectiveBound
@@ -135,9 +133,7 @@ VariableBasisStatus
 Constraint references and functions for adding, modifying, and removing constraints.
 
 ```@docs
-VariablewiseConstraintReference
-AffineConstraintReference
-QuadraticConstraintReference
+ConstraintReference
 candelete(::AbstractSolverInstance,::ConstraintReference)
 isvalid(::AbstractSolverInstance,::ConstraintReference)
 delete!(::AbstractSolverInstance,::ConstraintReference)
@@ -159,7 +155,20 @@ ConstraintDual
 ConstraintBasisStatus
 ```
 
-### Sets
+## Functions
+
+List of recognized functions.
+```@docs
+AbstractFunction
+ScalarVariablewiseFunction
+VectorVariablewiseFunction
+ScalarAffineFunction
+VectorAffineFunction
+ScalarQuadraticFunction
+VectorQuadraticFunction
+```
+
+## Sets
 
 List of recognized sets.
 
--- a/docs/src/apireference.md
+++ b/docs/src/apireference.md
@@ -31,6 +31,7 @@ setattribute!
 
 ```@docs
 AbstractSolver
+supportsproblem
 ```
 
 List of solver attributes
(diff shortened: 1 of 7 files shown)


joaquimg:
I like it!
People usually add constraints after the complete constraint is well defined. However, it is very common to query and modify very specific parts of a constraint.

Having AbstractFunctions would also be nice for the NLP interface right?

mlubin:
> Having AbstractFunctions would also be nice for the NLP interface right?

It's not obvious how to best deal with functions defined by callbacks, but yes, in principle we can use this same interface for NLP.

mlubin:
Updated with a new concept of ``AbstractFunctionModification``, so that you can do things like:
```julia
modifyobjective!(m, ScalarConstantChange(10.0))
modifyconstraint!(m, c, ScalarConstantChange(10.0))
```
We no longer need to enumerate all of the possible ways to modify an objective or constraint in ``modifyobjective!`` and ``modifyconstraint!``. Instead, we maintain a list of function modification types. This seems much simpler to me.

I also removed the concept of multiple objectives because the alternative would be passing through objective references everywhere. Nobody has actually requested multiple objective support, nor do I have any particular motivation to support it, so let's simplify things for now.

ccoffrin:
I have applications that are multi-objective, however, I usually address them at the JuMP level, not MOI.  So I am in favor of removing this, unless, some solvers have specialized native support detailing with multi-objective.  In that case, I would put it on the wish list for v0.2 of MOI.

mlubin:
It's fair to leave multiobjective on the wish list for MOI (Gurobi supports two objectives), but it feels like one step too far right now.

mlubin:
This PR is mostly coherent now, so I'll merge soon barring major objections.

PR #15, opened 2017-07-07
add FeasibilitySense and make sense required in setobjective!

### LINKED ISSUE #14: default value for Sense?
Currently we haven't defined what happens when the `Sense` attribute is never set. Should we define it to be minimization by default or make it required to provide a sense? I'm leaning towards the latter.

mlubin:


mlubin:
Merging quickly so that we can proceed with other updates. Still open for discussions on this point.

PR #17, opened 2017-07-07
more work on the manual

mlubin:
Worth looking at the ~`mixintopt`~ `solveintegerlinear` code example to see if we're happy with the final result. ~There's a small awkwardness in that we have the ``LessThan`` and ``GreatThan`` sets but no set representing equality with a constant, so for equality constraints we have to move the right-hand side into the constant of the affine expression.~ I added the ``EqualTo`` set for consistency.


mlubin:
Merging but still open for comments. I'm going to start working on tests.

Issue #23, opened 2017-07-12
Tests to be ported from MPB


joaquimg:
List of tests to be ported from MPB

**linproginterface.jl**
- [x] 1 "Testing LP interface with $solver"
    - [x] a "Basic interface" 
    - [x] b "addvar! interface"  
    - [x] c "setconstrLB! and setconstrUB!"
        now: "Modify constants in Nonnegative and Nonpositive"
        and: "Modify GreaterThan and LessThan sets as linear constraints"
    - [x] d "Issue #40 from Gurobi.jl"
    - [x] e "Change coeffs, del constr, del var"
- [ ] 2 "Testing LP interface extra with $solver"
    - [ ] a "setconstrLB! and setcontrUB!"
    - [ ] b "setvarLB! and setvarUB!"

**linprog.jl**
- [ ] 1 "Testing linprog and subfunctions with $solver"
    - [ ] a "LP1"
    - [ ] b "LP2"
    - [x] c "LP3 infeasible"
    - [x] d "LP4 unbounded"
    - [x] e "LP4 unbounded with unique ray"

**mixintprog.jl**
- [x] 1 "Testing mixintprog with $solver"
    now: "Knapsack Model"

**quadprog.jl**
- [x] 1 "Testing quadprog with $solver"
    - [x] a "QP1"
    - [x] b "QP2"
- [ ] 2 "Testing QP duals with $solver"
    - [ ] a "QP1"
    - [ ] b "QP2"
- [x] 3 "Testing SOCP interface with $solver"

**conicinterface.jl**
- [x] 1 "Testing linear problems through conic interface with $solver"
    - [x] a "LIN1"
    - [x] b "LIN1A"
    - [x] c "LIN2"
    - [x] d "LIN2A"
    - [x] e "LIN3 infeasible"
    - [x] f "LIN4 infeasible"
- [x] 2 "Testing SOC problems through conic interface with $solver"
    - [x] a "SOC1"
    - [x] b "SOC1A"
    - [x] c "SOC2"
    - [x] d "SOC2A"
    - [x] e "SOC3 infeasible"
    - [x] f "SOC4"
- [x] 3 "Testing SOCRotated problems through conic interface with $solver"
    - [x] a "SOCRotated1"
    - [x] b "SOCRotated1A"
    - [x] c "SOCRotated2 infeasible"
- [x] 4 "Testing SOCINT problems through conic interface with $solver"
[...]

PR #25, opened 2017-07-13
Get started in contquadratic tests

joaquimg:


joaquimg:
Start porting quadratic tests detailed in #23

test/contquadratic.jl
```diff
@@ -0,0 +1,219 @@
...
+            v = MOI.addvariables!(m, 3)
+            @test MOI.getattribute(m, MOI.VariableCount()) == 3
+
+            c1 = MOI.addconstraint!(m, MOI.ScalarAffineFunction(v, [1.0,2.0,3.0], 0.0), MOI.GreaterThan(4.0))
+            @test MOI.getattribute(m, MOI.ConstraintCount()) == 1
+
+            c2 = MOI.addconstraint!(m, MOI.ScalarAffineFunction([v[1],v[2]], [1.0,1.0], 0.0), MOI.GreaterThan(1.0))
+            @test MOI.getattribute(m, MOI.ConstraintCount()) == 2
+
+            MOI.setobjective!(m, MOI.MinSense, MOI.ScalarQuadraticFunction(v, [0.0,0.0,0.0],[v[1], v[1], v[2], v[2], v[3]], [v[1], v[2], v[2], v[3], v[3]], [2.0, 1.0, 2.0, 1.0, 2.0], 0.0))
```
mlubin: Are the off-diagonals scaled correctly? Shouldn't it be ``[2.0, 0.5, 2.0, 0.5, 2.0]``? These are coefficients in Q and the objective is (1/2).

test/contquadratic.jl
```diff
@@ -0,0 +1,219 @@
...
+            @test MOI.getattribute(m, MOI.Sense()) == MOI.MinSense
+
+            if MOI.cangetattribute(m, MOI.ObjectiveFunction())
+                obj = MPI.getattribute(m, MOI.ObjectiveFunction())
```
mlubin: We need to document the structure we expect for returned functions. It's probably not reasonable to expect the quadratic function to be returned in the exact same order as was given for input. We could possibly impose CSC format with increasing indices within a column, or don't impose anything and make comparisons after normalizing.

test/contquadratic.jl
```diff
@@ -0,0 +1,219 @@
...
+                @test obj.affine_variables == v
+                @test obj.affine_coefficients == [0.0,0.0,0.0]
+                @test obj.quadratic_rowvariables == [v[1], v[1], v[2], v[2], v[3]]
+                @test obj.quadratic_colvariables == [v[1], v[2], v[2], v[3], v[3]]
+                @test obj.quadratic_coefficients == [2.0, 1.0, 2.0, 1.0, 2.0]
+                @test obj.constant == 0.0
+            end
+
+            if MOI.cangetattribute(m, MOI.ConstraintFunction(), c1)
+                aff = MOI.getattribute(m, MOI.ConstraintFunction(), c1)
```
mlubin: We probably need a similar discussion here. Are affine functions returned in the exact same order that you provided them? I didn't think about this when I was writing some of the linear tests.

test/contquadratic.jl
mlubin: See https://github.com/JuliaOpt/MathProgBase.jl/blob/bdaaa744d0ef43e95dfd6b2f3eea9c8fbd6e2a9d/test/quadprog.jl#L19

test/contquadratic.jl
```diff
@@ -0,0 +1,219 @@
...
+            v = MOI.addvariables!(m, 3)
+            @test MOI.getattribute(m, MOI.VariableCount()) == 3
+
+            c1 = MOI.addconstraint!(m, MOI.ScalarAffineFunction(v, [1.0,2.0,3.0], 0.0), MOI.GreaterThan(4.0))
+            @test MOI.getattribute(m, MOI.ConstraintCount()) == 1
+
+            c2 = MOI.addconstraint!(m, MOI.ScalarAffineFunction([v[1],v[2]], [1.0,1.0], 0.0), MOI.GreaterThan(1.0))
+            @test MOI.getattribute(m, MOI.ConstraintCount()) == 2
+
+            MOI.setobjective!(m, MOI.MinSense, MOI.ScalarQuadraticFunction(v, [0.0,0.0,0.0],[v[1], v[1], v[1], v[2], v[2], v[3], v[3]], [v[1], v[2], v[2], v[2], v[3], v[3], v[3]], [2.0, 0.5, 0.5, 2.0, 1.0, 1.0, 1.0], 0.0))
```
mlubin: Same comment about the scaling of the off-diagonal terms

test/contquadratic.jl
```diff
@@ -0,0 +1,219 @@
...
+
+            @test MOI.supportsproblem(solver, MOI.ScalarAffineFunction, [(MOI.VectorAffineFunction{Float64},MOI.NonPositive),(MOI.ScalarQuadraticFunction{Float64},MOI.LessThan)])
+
+            m = MOI.SolverInstance(solver)
+
+            x = MOI.addvariable!(m)
+            y = MOI.addvariable!(m)
+            @test MOI.getattribute(m, MOI.VariableCount()) == 2
+
+            c1 = MOI.addconstraint!(m, MOI.VectorAffineFunction([1,2], [x,y,x,y],[-1.0,1.0,1.0,1.0], [0.0,0.0]), MOI.NonPositive(2))
```
mlubin: Shouldn't this be ``NonNegative``?

test/contquadratic.jl
joaquimg: I like CSC format with increasing indices.
But we are returning Variable references right?
Are we writing our own Vector of triplets(or triple of vectors)?

test/contquadratic.jl
joaquimg: I think affine functions should be returned in increasing order also...

test/contquadratic.jl
joaquimg: Yes, `NonNegative`

test/contquadratic.jl
joaquimg: On this example, we are ignoring duplicates or adding them?

mlubin: Which are the duplicates here? Specifying (i,j,v) also specifies (j,i,v).

test/contquadratic.jl
mlubin: There's no real ordering on the variable references. Let me throw something together.

test/contquadratic.jl
joaquimg: the example you posted is:
```julia

rows = [1, 2, 2, 2, 3, 3, 3]
cols = [1, 1, 1, 2, 2, 3, 3]
vals = [2.0, 0.5, 0.5, 2.0, 1.0, 1.0, 1.0]
```
Here `(2,1,0.5)` appears twice. Same for `(3,3,1.0)`. They are being added right?

The triplets I wrote are:
```julia
rows = [1,1,2,2,3]
cols = [1,2,2,3,3]
vals = [2.0, 1.0, 2.0, 1.0, 2.0]
```

mlubin: Good point. So is the objective in the comment ``x^2 + 0.5xy + y^2 + 0.5yz + z^2`` correct?

joaquimg: Lest see, my understanding is that the values:
```julia
rows = [1,1,2,2,3]
cols = [1,2,2,3,3]
vals = [2.0, 1.0, 2.0, 1.0, 2.0]
```
imply in the Matrix:
```julia
Q = [[2.0, 1.0, 0.0],
     [1.0, 2.0, 1.0],
     [0.0, 1.0, 2.0]]
```
And thus the objective should be: `x^2 + xy + y^2 + yz + z^2`, right?

mlubin: Yes, that seems right

test/contquadratic.jl
```diff
@@ -0,0 +1,329 @@
...
+
+            @test MOI.supportsproblem(solver, MOI.ScalarAffineFunction, [(MOI.VectorAffineFunction{Float64},MOI.NonNegative),(MOI.ScalarQuadraticFunction{Float64},MOI.LessThan)])
+
+            m = MOI.SolverInstance(solver)
+
+            x = MOI.addvariable!(m)
+            y = MOI.addvariable!(m)
+            @test MOI.getattribute(m, MOI.VariableCount()) == 2
+
+            c1 = MOI.addconstraint!(m, MOI.VectorAffineFunction([1,2], [x,y,x,y],[-1.0,1.0,1.0,1.0], [0.0,0.0]), MOI.NonNegative(2))
```
mlubin: The set is now called `Nonnegative`

mlubin:
We haven't yet defined our duality conventions for QPs, so these tests need some discussion.

joaquimg:
Should I comment the getting duals part? So we can merge.
Or its better to just wait for the discussion?
I could also get the manual on QCQP duals started here by following Boyd's Convex Optimization...

mlubin:
Yes, you can comment out the duals for now

test/contquadratic.jl
```diff
@@ -0,0 +1,331 @@
...
+function contquadratictest(solver::MOI.AbstractSolver, ε=Base.rtoldefault(Float64))
+    @testset "Testing quadratic interface with $solver" begin
+        @testset "QP1 - Quadratic objective" begin
+            # simple quadratic objective
+            # Min x^2 + xy + y^2 + yz + z^2
+            # st  x + 2y + 3z >= 4 (c1)
+            #     x +  y      >= 1 (c2)
+            #     x,y \in R
+
+            @test MOI.supportsproblem(solver, MOI.ScalarQuadraticFunction, [(MOI.ScalarAffineFunction{Float64},MOI.GreaterThan)])
```
mlubin: Missing type parameter on ``MOI.ScalarQuadraticFunction``

test/contquadratic.jl
mlubin: This should now be updated to ``aff \approx ...``

test/contquadratic.jl
mlubin: This should now be replaced with an `isapprox` call (and we also need an implementation of Base.isapprox for two scalar quadratic functions)

test/contquadratic.jl
```diff
@@ -0,0 +1,331 @@
...
+        @testset "QP1" begin
+            # same as QP0 but with duplicate terms
+            # then change the objective and sense
+            # simple quadratic objective
+            # Min x^2 + xy + y^2 + yz + z^2
+            # st  x + 2y + 3z >= 4 (c1)
+            #     x +  y      >= 1 (c2)
+            #     x,y \in R
+
+            @test MOI.supportsproblem(solver, MOI.ScalarQuadraticFunction, [(MOI.ScalarAffineFunction{Float64},MOI.GreaterThan)])
```
mlubin: Missing type parameter on MOI.ScalarQuadraticFunction

test/contquadratic.jl
```diff
@@ -0,0 +1,331 @@
...
+
+            c2 = MOI.addconstraint!(m, MOI.ScalarAffineFunction([v[1],v[2]], [1.0,1.0], 0.0), MOI.GreaterThan(1.0))
+            @test MOI.getattribute(m, MOI.ConstraintCount()) == 2
+
+            obj = MOI.ScalarQuadraticFunction(v, [0.0,0.0,0.0],[v[1], v[1], v[1], v[2], v[2], v[3], v[3]], [v[1], v[2], v[2], v[2], v[3], v[3], v[3]], [2.0, 0.5, 0.5, 2.0, 1.0, 1.0, 1.0], 0.0)
+            MOI.setobjective!(m, MOI.MinSense, obj)
+            @test MOI.getattribute(m, MOI.Sense()) == MOI.MinSense
+
+            if MOI.cangetattribute(m, MOI.ObjectiveFunction())
+                @test obj ≈ MPI.getattribute(m, MOI.ObjectiveFunction())
```
mlubin: This isn't defined yet

test/contquadratic.jl
joaquimg: This type is missing in linear tests also!

test/contquadratic.jl
joaquimg: Should it be defined? Or do I remove from the test?

test/contquadratic.jl
mlubin: My bad

test/contquadratic.jl
mlubin: It should be defined (you should define it)

joaquimg: ok!

test/contquadratic.jl
```diff
@@ -0,0 +1,331 @@
...
+            # st  x + 2y + 3z >= 4 (c1)
+            #     x +  y      >= 1 (c2)
+            #     x,y \in R
+
+            @test MOI.supportsproblem(solver, MOI.ScalarQuadraticFunction{Float64}, [(MOI.ScalarAffineFunction{Float64},MOI.GreaterThan)])
+
+            m = MOI.SolverInstance(solver)
+
+            v = MOI.addvariables!(m, 3)
+            @test MOI.getattribute(m, MOI.VariableCount()) == 3
```
mlubin: ``VariableCount`` was renamed to ``NumberOfVariables``

test/contquadratic.jl
```diff
@@ -0,0 +1,331 @@
...
+
+            c1 = MOI.addconstraint!(m, MOI.ScalarAffineFunction(v, [1.0,2.0,3.0], 0.0), MOI.GreaterThan(4.0))
+            @test MOI.getattribute(m, MOI.ConstraintCount()) == 1
```
mlubin: ``ConstraintCount`` was renamed to ``NumberOfConstraints{F,S}``, now specific to the type of constraint

mlubin: See the linear tests

mlubin:
Looks good! There are just two places now at the beginning of the tests which still have the old code for comparisons instead of \approx.

PR #26, opened 2017-07-13
fix comparisons of variablewise and affine functions

mlubin:
@joaquimg, what do you think of this? For comparisons we canonicalize using a dictionary, so no order is required. Quadratic would look very similar.


joaquimg:
Looks good!

PR #32, opened 2017-07-14
add discussion of interpreting statuses and results

mlubin:
I believe this discussion encapsulates the conclusions of https://github.com/JuliaOpt/MathProgBase.jl/issues/164 and discussions at the meetup.


mlubin:
CC @ccoffrin @blegat

blegat:
We might add for the primal-dual class that when a infeasibility ray is found, if one desire to know whether the problem is unbounded he often needs to set the objective to zero and resolve. Only the combination of a feasible point and a infeasibility ray makes an unbounded ray, certificate of unboundedness.

PR #35, opened 2017-07-15
Port SDP1 test from MPB #23

blegat:


test/contconic.jl
```diff
@@ -503,5 +503,106 @@ function contconictest(solver::MOI.AbstractSolver, ε=Base.rtoldefault(Float64))
         end
     end
 
+    if MOI.supportsproblem(solver, MOI.ScalarAffineFunction, [(MOI.ScalarAffineFunction{Int}, EqualTo{Int}), (MOI.ScalarAffineFunction{Int}, EqualTo{Rational{Int}}), (MOI.VectorVariablewiseFunction, MOI.PositiveSemidefiniteConeTriangle), (MOI.VectorVariablewiseFunction, MOI.SecondOrderCone)])
```
mlubin: I don't think we should expect solvers to implement ``MOI.ScalarAffineFunction{Int}`` and ``EqualTo{Int}`` in addition to ``MOI.ScalarAffineFunction{Float64}`` and ``EqualTo{Float64}`` unless they have a special reason to do so. It imposes a lot of extra work on the interface because they have to be maintained as a separate class of constraints (the constraint references have a different type).
None of the SDP solvers we have connected support rational arithmetic, so I wouldn't expect them to support `EqualTo{Rational{Int}}` either.

blegat: There is SDPA-GMP

mlubin: Ok, but do you expect Mosek to be able to pass this test?

blegat: Yes, a solver usually a type that it supports like Float64 and it can add one method that converts addconstraint of any other type to this type. It does not need one for every pair of function/method.

mlubin: The conversion is not the issue. It also needs to account for the constraint correctly in ``NumberOfConstraints`` and keep track of different constraint reference objects.

mlubin: It's not ok to return a `ConstraintReference{ScalarAffineFunction{Float64},LessThan{Float64}}` if the user provided a `(ScalarAffineFunction{Int}, LessThan{Int})` constraint.

blegat: I see the issue, that would require mosek to return an integer solution.

test/contconic.jl
```diff
@@ -503,5 +503,105 @@ function contconictest(solver::MOI.AbstractSolver, ε=Base.rtoldefault(Float64))
         end
     end
 
+    if MOI.supportsproblem(solver, MOI.ScalarAffineFunction, [(MOI.ScalarAffineFunction{Float64}, EqualTo{Float64}), (MOI.VectorVariablewiseFunction, MOI.PositiveSemidefiniteConeTriangle), (MOI.VectorVariablewiseFunction, MOI.SecondOrderCone)])
```
mlubin: Second argument should be `MOI.ScalarAffineFunction{Float64}`

test/contconic.jl
```diff
@@ -503,5 +503,105 @@ function contconictest(solver::MOI.AbstractSolver, ε=Base.rtoldefault(Float64))
...
+
+            @test MOI.cangetattribute(m, MOI.ObjectiveValue())
+            @test MOI.getattribute(m, MOI.ObjectiveValue()) ≈ 0.705710509 atol=ε
+
+            @test MOI.cangetattribute(m, MOI.VariablePrimal(), X)
+            Xv = MOI.getattribute(m, MOI.VariablePrimal(), X)
+            @test MOI.cangetattribute(m, MOI.VariablePrimal(), x)
+            xv = MOI.getattribute(m, MOI.VariablePrimal(), x)
+
+            y = MathProgBase.getdual(m)
```
mlubin: What's this MathProgBase thing?

test/contconic.jl
blegat: This should also be corrected for the other tests then

mlubin: Indeed

test/contconic.jl
```diff
@@ -503,5 +503,105 @@ function contconictest(solver::MOI.AbstractSolver, ε=Base.rtoldefault(Float64))
...
+            A = [  1   0   0   1   0   1   1   0   0;
+                   1   2   2   1   2   1   0   1   1]
+            b = [1, 1/2]
+            # Check primal objective
+            comp_pobj = dot(c, [Xv; xv])
+            # Check dual objective
+            comp_dobj = -dot(y, b)
+            @test comp_pobj ≈ comp_dobj atol=ε
+
+            var = c + A' * y
```
mlubin: This might be a bit confusing for readers since it follows the old variable duals concepts which don't exist in MOI. Could we write this to be consistent with http://www.juliaopt.org/MathOptInterface.jl/latest/apimanual.html#Duals-1?

mlubin:
Looks good, could we also have a version of the test for `PositiveSemidefinteConeScaled`?

blegat:
Yes, but it can be postponed to another PR

PR #38, opened 2017-07-15
completed the test basic interface in contlin

IssamT:
basic interface test ported from MBP linproginterface
(see issue #23)

### DIFF
--- a/test/contlinear.jl
+++ b/test/contlinear.jl
@@ -192,7 +192,7 @@ function contlineartest(solver::MOI.AbstractSolver, ε=Base.rtoldefault(Float64)
         MOI.modifyconstraint!(m, vc1, MOI.GreaterThan(0.0))
         MOI.delete!(m, vc3)
         vc3 = MOI.addconstraint!(m, MOI.ScalarVariablewiseFunction(v[3]), MOI.EqualTo(0.0))
-        @test MOI.getattribute(m, MOI.NumberOfConstraints{MOI.ScalarVariablewiseFunction,MOI.GreaterThan}()) == 3                
+        @test MOI.getattribute(m, MOI.NumberOfConstraints{MOI.ScalarVariablewiseFunction,MOI.GreaterThan}()) == 3
 
         MOI.optimize!(m)
 
@@ -210,7 +210,6 @@ function contlineartest(solver::MOI.AbstractSolver, ε=Base.rtoldefault(Float64)
         # s.t. x + y + z == 2
         # x,y >= 0, z = 0
 
-        MOI.modifyconstraint!(m, c, MOI.EqualTo(2.0))
         MOI.delete!(m, c)
         cf = MOI.ScalarAffineFunction(v, [1.0,1.0,1.0], 0.0)
         c = MOI.addconstraint!(m, cf, MOI.EqualTo(2.0))
--- a/test/contlinear.jl
+++ b/test/contlinear.jl
@@ -116,6 +116,178 @@ function contlineartest(solver::MOI.AbstractSolver, ε=Base.rtoldefault(Float64)
             @test MOI.getattribute(m, MOI.ConstraintDual(), vc2) ≈ 1 atol=ε
         end
 
+        # add new variable to get :
+        # max x + 2z
+        # s.t. x + y + z <= 1
+        # x,y,z >= 0
+
+        z = MOI.addvariable!(m)
+        push!(v, z)
+        @test v[3] == z
+
+        vc3 = MOI.addconstraint!(m, MOI.ScalarVariablewiseFunction(v[2]), MOI.GreaterThan(0.0))
+        @test MOI.getattribute(m, MOI.NumberOfConstraints{MOI.ScalarAffineFunction{Float64},MOI.GreaterThan}()) == 3
+
+        cf2 = MOI.ScalarAffineFunction(v, [1.0,1.0,1.0], 0.0)
+        MOI.modifyconstraint!(m, c, cf2)
+
+        objf = MOI.ScalarAffineFunction(v, [1.0,0.0,2.0], 0.0)
+        MOI.setobjective!(m, MOI.MaxSense, objf)
+
+        @test MOI.getattribute(m, MOI.NumberOfConstraints{MOI.ScalarAffineFunction{Float64},MOI.LessThan}()) == 1
+        @test MOI.getattribute(m, MOI.NumberOfConstraints{MOI.ScalarAffineFunction{Float64},MOI.GreaterThan}()) == 3
+
+        MOI.optimize!(m)
+
+        @test MOI.cangetattribute(m, MOI.TerminationStatus())
+        @test MOI.getattribute(m, MOI.TerminationStatus()) == MOI.Success
+
+        @test MOI.cangetattribute(m, MOI.PrimalStatus())
+        @test MOI.getattribute(m, MOI.PrimalStatus()) == MOI.FeasiblePoint
+
+        @test MOI.cangetattribute(m, MOI.ObjectiveValue())
+        @test MOI.getattribute(m, MOI.ObjectiveValue()) ≈ 2 atol=ε
+
+        @test MOI.cangetattribute(m, MOI.VariablePrimal(), v)
+        @test MOI.getattribute(m, MOI.VariablePrimal(), v) ≈ [0, 0, 1] atol=ε
+
+        @test MOI.cangetattribute(m, MOI.ConstraintPrimal(), c)
+        @test MOI.getattribute(m, MOI.ConstraintPrimal(), c) ≈ 1 atol=ε
+
+        if MOI.getattribute(solver, MOI.SupportsDuals())
+            @test MOI.cangetattribute(m, MOI.DualStatus())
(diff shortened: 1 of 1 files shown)


test/contlinear.jl
```diff
@@ -116,6 +116,178 @@ function contlineartest(solver::MOI.AbstractSolver, ε=Base.rtoldefault(Float64)
...
+        # x,y,z >= 0
+
+        z = MOI.addvariable!(m)
+        push!(v, z)
+        @test v[3] == z
+
+        vc3 = MOI.addconstraint!(m, MOI.ScalarVariablewiseFunction(v[2]), MOI.GreaterThan(0.0))
+        @test MOI.getattribute(m, MOI.NumberOfConstraints{MOI.ScalarAffineFunction{Float64},MOI.GreaterThan}()) == 3
+
+        cf2 = MOI.ScalarAffineFunction(v, [1.0,1.0,1.0], 0.0)
```
mlubin: This test is meant to test column generation, so I'd say it's more appropriate to use ``ScalarCoeffficientChange`` instead of changing the entire function. This test is correct though. We should keep this version and add a second version with `ScalarCoefficientChange`.

test/contlinear.jl
```diff
@@ -116,6 +116,178 @@ function contlineartest(solver::MOI.AbstractSolver, ε=Base.rtoldefault(Float64)
...
+        MOI.modifyconstraint!(m, c, cf2)
+
+        objf = MOI.ScalarAffineFunction(v, [1.0,0.0,2.0], 0.0)
```
mlubin: You could also use `ScalarCoefficientChange` to modify the objective.

test/contlinear.jl
```diff
@@ -116,6 +116,178 @@ function contlineartest(solver::MOI.AbstractSolver, ε=Base.rtoldefault(Float64)
...
+        @test MOI.getattribute(m, MOI.ConstraintPrimal(), c) ≈ 1 atol=ε
+
+        if MOI.getattribute(solver, MOI.SupportsDuals())
+            @test MOI.cangetattribute(m, MOI.DualStatus())
+            @test MOI.getattribute(m, MOI.DualStatus()) == MOI.FeasiblePoint
+            @test MOI.cangetattribute(m, MOI.ConstraintDual(), c)
+            @test MOI.getattribute(m, MOI.ConstraintDual(), c) ≈ 2 atol=ε
+
+            @test MOI.cangetattribute(m, MOI.ConstraintDual(), vc1)
+            @test MOI.getattribute(m, MOI.ConstraintDual(), vc1) ≈ -1 atol=ε
```
mlubin: `GreaterThan` constraints should have nonnegative feasible duals. This convention is described briefly at http://www.juliaopt.org/MathOptInterface.jl/latest/apimanual.html#Duals-1, and I'm planning on adding more soon.

test/contlinear.jl
```diff
@@ -116,6 +116,178 @@ function contlineartest(solver::MOI.AbstractSolver, ε=Base.rtoldefault(Float64)
...
+            @test MOI.cangetattribute(m, MOI.ConstraintDual(), c)
+            @test MOI.getattribute(m, MOI.ConstraintDual(), c) ≈ 1.5 atol=ε
+            @test MOI.getattribute(m, MOI.ConstraintDual(), c) ≈ -0.5 atol=ε
+
+            @test MOI.cangetattribute(m, MOI.ConstraintDual(), vc1)
+            @test MOI.getattribute(m, MOI.ConstraintDual(), vc1) ≈ 0 atol=ε
+            @test MOI.cangetattribute(m, MOI.ConstraintDual(), vc2)
+            @test MOI.getattribute(m, MOI.ConstraintDual(), vc2) ≈ 0 atol=ε
+            @test MOI.cangetattribute(m, MOI.ConstraintDual(), vc3)
+            @test MOI.getattribute(m, MOI.ConstraintDual(), vc3) ≈ -1.5 atol=ε
```
mlubin: Same comment about the sign of the dual

mlubin:
Looks very good, thanks! I made a couple of comments that should be easy to address.

test/contlinear.jl
```diff
@@ -116,6 +116,178 @@ function contlineartest(solver::MOI.AbstractSolver, ε=Base.rtoldefault(Float64)
...
+        # add new variable to get :
+        # max x + 2z
+        # s.t. x + y + z <= 1
+        # x,y,z >= 0
+
+        z = MOI.addvariable!(m)
+        push!(v, z)
+        @test v[3] == z
+
+        vc3 = MOI.addconstraint!(m, MOI.ScalarVariablewiseFunction(v[2]), MOI.GreaterThan(0.0))
```
mlubin: You mean `v[3]` instead of `v[2]`?

test/contlinear.jl
```diff
@@ -116,6 +116,178 @@ function contlineartest(solver::MOI.AbstractSolver, ε=Base.rtoldefault(Float64)
...
+        @test MOI.getattribute(m, MOI.NumberOfConstraints{MOI.ScalarAffineFunction{Float64},MOI.GreaterThan}()) == 3
```
mlubin: ``MOI.NumberOfConstraints{MOI.ScalarVariablewiseFunction,MOI.GreaterThan}()``

test/contlinear.jl
```diff
@@ -116,6 +116,178 @@ function contlineartest(solver::MOI.AbstractSolver, ε=Base.rtoldefault(Float64)
...
+
+        cf2 = MOI.ScalarAffineFunction(v, [1.0,1.0,1.0], 0.0)
+        MOI.modifyconstraint!(m, c, cf2)
+
+        objf = MOI.ScalarAffineFunction(v, [1.0,0.0,2.0], 0.0)
+        MOI.setobjective!(m, MOI.MaxSense, objf)
+
+        @test MOI.getattribute(m, MOI.NumberOfConstraints{MOI.ScalarAffineFunction{Float64},MOI.LessThan}()) == 1
+        @test MOI.getattribute(m, MOI.NumberOfConstraints{MOI.ScalarAffineFunction{Float64},MOI.GreaterThan}()) == 3
```
mlubin: ``MOI.NumberOfConstraints{MOI.ScalarVariablewiseFunction,MOI.GreaterThan}()``

test/contlinear.jl
IssamT: yes

IssamT: You want to use v[2]

test/contlinear.jl
```diff
@@ -116,6 +116,178 @@ function contlineartest(solver::MOI.AbstractSolver, ε=Base.rtoldefault(Float64)
...
+        @test MOI.cangetattribute(m, MOI.ObjectiveValue())
+        @test MOI.getattribute(m, MOI.ObjectiveValue()) ≈ 3 atol=ε
+
+        # put lb of x back to 0 and fix z to zero to get :
+        # max x + 2z
+        # s.t. x + y + z <= 1
+        # x, y >= 0, z = 0
+
+        MOI.modifyconstraint!(m, vc1, MOI.GreaterThan(0.0))
+        MOI.modifyconstraint!(m, vc3, MOI.EqualTo(0.0))
```
mlubin: ``vc3`` is a ``ConstraintReference{ScalarVariablewiseFunction,GreaterThan}``. You're allowed to replace the set with a different ``GreaterThan`` set, but not ``EqualTo``. The correct thing to do would be to add an upper bound constraint.

test/contlinear.jl
mlubin: I should be `v[3]`

IssamT: sorry i was reading the wrong line :)

test/contlinear.jl
```diff
@@ -116,6 +116,178 @@ function contlineartest(solver::MOI.AbstractSolver, ε=Base.rtoldefault(Float64)
...
+
+        @test MOI.cangetattribute(m, MOI.ObjectiveValue())
+        @test MOI.getattribute(m, MOI.ObjectiveValue()) ≈ 1 atol=ε
+
+        # modify affine linear constraint set to be == 2 to get :
+        # max x + 2z
+        # s.t. x + y + z == 2
+        # x,y >= 0, z = 0
+
+        MOI.modifyconstraint!(m, c, MOI.EqualTo(2.0))
```
mlubin: Similarly you can't modify the sense of a linear constraint. You can add a ``LessThan`` constraint to imply the equality.

test/contlinear.jl
IssamT: You're right. 
- Do we really want to have this restriction?
- In case we keep the restriction Why not delete this constraint and add and EqualTo constraint instead?

mlubin: > Do we really want to have this restriction?

If we allow changing the set, we have to allow changing the function as well. Then `ConstraintReference` can't have any type parameters, which will force solver interfaces to manage their own mappings of constraint references to classes of constraints. Having the constraint references properly typed will make it much easier to implement solver wrappers.

> In case we keep the restriction Why not delete this constraint and add and EqualTo constraint instead?

That's also ok.

mlubin:
Looks good to me. @joaquimg, want to double check?

joaquimg:
Looks good!

PR #39, opened 2017-07-15
expand on duality

mlubin:
- Define duality for maximization problems
- Explicitly write out the special case of LP duality

I'd like to think about how to define duality with quadratic terms after this.

@blegat @chriscoey @IssamT @joaquimg

### DIFF
--- a/docs/src/apimanual.md
+++ b/docs/src/apimanual.md
@@ -458,7 +458,7 @@ and the dual is:
 \end{align}
 ```
 
-where each ``\mathcal{C}_i`` is a closed convex cone and ``\mathcal{C}_i`` is its dual cone.
+where each ``\mathcal{C}_i`` is a closed convex cone and ``\mathcal{C}_i^*`` is its dual cone.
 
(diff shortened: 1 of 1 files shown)


docs/src/apimanual.md
```diff
@@ -455,18 +452,94 @@ and the dual is:
...
 & \max_{y_1, \ldots, y_m} & -\sum_{i=1}^m b_i^T y_i + b_0
 \\
-& \;\;\text{s.t.} & a_0 - \sum_{i=1}^m A_i^T y_i & \in {0}^n
+& \;\;\text{s.t.} & a_0 - \sum_{i=1}^m A_i^T y_i & = 0
 \\
 & & y_i & \in \mathcal{C}_i^* & i = 1 \ldots m
 \end{align}
 ```
 
 where each ``\mathcal{C}_i`` is a closed convex cone and ``\mathcal{C}_i`` is its dual cone.
```
joaquimg: missing `^*` here

mlubin: got it

PR #44, opened 2017-07-16
Add AbstractScalarFunction and AbstractVectorFunction

blegat:


mlubin:
Why is this extra level of abstraction needed?

blegat:
I wouldn't say it is needed but it might be helpful to the solver wrappers. For example they might want to do:
```julia
addconstraint!(m::..., f::AbstractScalarFunction, s) = addconstraint!(m, MOI.vectorize(f), MOI.vectorize(s))
```
so that they do not need to implement the scalar versions.

mlubin:
That's not quite correct because the constraint references will be of the wrong type.

mlubin:
I feel like this is better defined as `Union{ScalarVariablewiseFuncton,ScalarAffineFunction,...}` directly in the solver interface, rather than adding an extra level to the type hierarchy in MOI.

mlubin:
I'm going to close this for now.

Issue #40, opened 2017-07-15
ScalarLinearFunction and redundancy with scalar constraints


joaquimg:
We have the following table in the manual:

#### Linear constraints

| Mathematical Constraint       | MOI Function                 | MOI Set       |
|-------------------------------|------------------------------|---------------|
| ``a^Tx \le u``                | `ScalarAffineFunction`       | `LessThan`    |
| ``a^Tx \ge l``                | `ScalarAffineFunction`       | `GreaterThan` |
| ``a^Tx = b``                  | `ScalarAffineFunction`       | `EqualTo`     |
| ``l \le a^Tx \le u``          | `ScalarAffineFunction`       | `Interval`    |
|...                |... | ...    |

In the first four lines we use `a^Tx` in the first column and we say `ScalarAffineFunction`, should we say/use `ScalarLinearFunction`? This could avoid multiple representations of the same constraint. Since `a^Tx - b = 0` is equivalent to `a^Tx = b`

mlubin:
@joaquimg, that would require introducing a new category of function. We can't get rid of `ScalarAffineFunction` because it's needed for conic constraints. And then we'd have the question of if solvers should support both `ScalarAffineFunction` and `ScalarLinearFunction` in `LessThan` or just `ScalarLinearFunction` in `LessThan`. 
I'm okay with the potential for redundancy here.

mlubin:
Well, to be more precise, `VectorAffineFunction` is needed for conic constraints. It seems a bit inconsistent though to have `ScalarLinearFunction` and `VectorAffineFunction`.

mlubin:
`ScalarAffineFunction` is useful for handling constants in the objective function

joaquimg:
Sure, I don´t think we should get rid of `ScalarAffineFunction`. However, since scalar constraints have their own sets (`Interval`, `LessThan` ...) they could have their own function and we disallow `ScalarAffineFunction` only for them...

mlubin:
What's the difference between introducing ``ScalarLinearFunction`` and letting solvers choose to reject ``ScalarAffineFunction``s when the constant is not zero?

joaquimg:
Querying and modifying these scalar constraints might be a bit "degenerate" and would add more complexity on the solver api side...

joaquimg:
I am fine with rejecting `ScalaAffineFunction` when the constant is not zero

mlubin:
It doesn't bother me if solvers choose to not support this case.

[...]

PR #45, opened 2017-07-16
discussion of alternative formulations and what solvers are expected …

### LINKED ISSUE #19: force Nonnegative, Nonpostive, and Zeros to be vector sets
In order to avoid some level of redundancy and extra work for the solver wrappers, I'm thinking that it would be a good idea to disallow (by convention) constraints of the form: ``ScalarAffineFunction`` in  ``Nonnegative(1)``, ``Nonpositve(1)``, or ``Zeros(1)``. These are already covered by the one-dimensional sets ``GreaterThan``, ``LessThan``, and ``EqualTo``.

### LINKED ISSUE #41: Possible redundant problem formulations
Requires a documentation update.

see issues : #40 and JuliaOpt/MathProgBase.jl#186

mlubin:
…to support

Closes #41 
Closes #19 

CC @IssamT @blegat @joaquimg

### DIFF
--- a/docs/src/apimanual.md
+++ b/docs/src/apimanual.md
@@ -168,6 +168,7 @@ as function-set pairs in MOI. In the notation below, ``x`` is a vector of decisi
 | ``Ax + b \in \mathbb{R}_-^n`` | `VectorAffineFunction`       | `Nonpositive` |
 | ``Ax + b = 0``                | `VectorAffineFunction`       | `Zero`        |
 
+By convention, solvers are not expected to support nonzero constant terms in the `ScalarAffineFunction`s the first four rows above, because they are redundant with the parameters of the sets. For example, ``2x + 1 \le 2`` should be encoded as ``2x \le 1``.
 
 [Define ``\mathbb{R}_+, \mathbb{R}_-``]
 
--- a/docs/src/apimanual.md
+++ b/docs/src/apimanual.md
@@ -551,7 +551,11 @@ Currently, a convention for duals is not defined for problems with non-conic set
 
 [Avoid storing extra copies of the problem when possible.]
 
-[It would be nice if solvers supported the multiple different ways to write the same constraint, e.g., `GreaterThan` and `Nonnegative`, second-order cone and rotated second-order cone.]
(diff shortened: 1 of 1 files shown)


blegat:
Why is it so hard to support the nonzero constant ? It is just a matter of a substraction.
For the AffineFunction in Zeros..., I would go further and disallow it. Otherwise we also need to consider ScalarFunction in PSD(1) for example.
I suggest we define AbstractScalarFunction, AbstractVectorFunction, AbstractScalarSet, AbstractVectorSet and define
```julia
addconstraint!(si, ::AbstractVectorFunction, ::AbstractScalarSet) = throw(...)
addconstraint!(si, ::AbstractScalarFunction, ::AbstractVectorSet) = throw(...)
```

mlubin:
> Why is it so hard to support the nonzero constant ? 

There's more complexity involved when you start talking about modifying the set or the constant.

> For the AffineFunction in Zeros..., I would go further and disallow it.

We can't *prevent* a solver from supporting a certain type of constraint. The most we can do is say that by convention there's no need to support it.

blegat:
Yes we cannot prevent it to support. What I suggested was to make it clear
to the user that it is not supported because it is not part of the
interface, not because the solver is missing a method.

IssamT:
When talking about what the solvers need to implement which keep in mind that the user might be annoyed with error messages like this : "Sorry Feature X is not supported by solver Y you have to write it the other way."

Possible use case: 
A user was using a given solver and wrote a lot of models using a some feature and he wanted to test them with another solver. now he is required to patch all models and write them the other way...

This may hurt the power of JuMP being a **unified** modeling language.

I am now not so sure about the conclusion we draw in #40.

IssamT:
Making transformation is though not easy, as @mlubin have often mentioned, and it would be nice to not deal with it at this stage. 

Maybe we can limit a bit the alternative formulations on the interface side and say in the documentation that an MOI solver which supports a type of math programs is required to implement all the related methods?

blegat:
> A user was using a given solver and wrote a lot of models using some feature and he wanted to test them with another solver. now he is required to patch all models and write them the other way...
>
> This may hurt the power of JuMP being a unified modeling language.

We test in the JuMP's tests that all solvers support the features that JuMP needs so that shouldn't happen for JuMP

mlubin:
> When talking about what the solvers need to implement we should keep in mind that the user might be annoyed with error messages like this : "Sorry Feature X is not supported by solver Y you have to write it the other way."

It's obviously inconvenient, but we don't have unlimited resources to implement all possible transformations. We're in the open source world, so I'm hoping that if a feature is missing, users will be motivated to implement it themselves.

mlubin:
I added a discussion higher up about the constant terms.

PR #53, opened 2017-07-17
Port of MPB test 1d from linproginterface

IssamT:


mlubin:
In my mind, it is reasonable for solvers to reject multiple `ScalarVariablewiseFunction`-`GreaterThan` constraints applying to the same variable, if it's convenient to do so. I think it would be a lot of unnecessary work to keep track of multiple variable bounds when this is usually not needed. If we agree on this point, then we shouldn't be testing this functionality.

mlubin:
Note that the MPB test used one linear constraint and one variable bound, I think that makes more sense to test.

IssamT:
I agree with that.

test/contlinear.jl
```diff
@@ -356,6 +356,68 @@ function contlineartest(solver::MOI.AbstractSolver, ε=Base.rtoldefault(Float64)
...
+
+        MOI.addconstraint!(m, MOI.ScalarVariablewiseFunction(x), MOI.GreaterThan(0.0))
+        cf = MOI.ScalarAffineFunction([x], [1.0], 0.0)
+        MOI.addconstraint!(m, cf, MOI.GreaterThan(3.0))
+
+        @test MOI.getattribute(m, MOI.NumberOfConstraints{MOI.ScalarVariablewiseFunction,MOI.GreaterThan{Float64}}()) == 1
+        @test MOI.getattribute(m, MOI.NumberOfConstraints{MOI.ScalarAffineFunction{Float64},MOI.GreaterThan{Float64}}()) == 1
+
+        objf = MOI.ScalarVariablewiseFunction(x)
+        MOI.setobjective!(m, MOI.MinSense, objf)
```
mlubin: I also feel that it's unnecessary to test ``ScalarVariablewiseFunction`` as an objective function. `ScalarAffineFunction` is sufficient in terms of expressiveness in the context of LPs.

Issue #57, opened 2017-07-17
add rotated second-order cone


mlubin:

joaquimg:
I can do it

joaquimg:
the rotated SOC is always a variablewise constraint?

Or should it be in this table somehow?

| Mathematical Constraint                                       | MOI Function                 | MOI Set                            |
|---------------------------------------------------------------|------------------------------|------------------------------------|
| ``\lVert Ax + b\rVert_2 \le c^Tx + d``                        | `VectorAffineFunction`       | `SecondOrderCone`                  |

mlubin:
No, it's not always variablewise.

mlubin:
Closed by #59

PR #59, opened 2017-07-17
Add rotated second order cone

joaquimg:
#57

### DIFF
--- a/docs/src/apimanual.md
+++ b/docs/src/apimanual.md
@@ -178,7 +178,7 @@ By convention, solvers are not expected to support nonzero constant terms in the
 | Mathematical Constraint                                       | MOI Function                 | MOI Set                            |
 |---------------------------------------------------------------|------------------------------|------------------------------------|
 | ``\lVert Ax + b\rVert_2 \le c^Tx + d``                        | `VectorAffineFunction`       | `SecondOrderCone`                  |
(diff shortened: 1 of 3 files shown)


docs/src/apimanual.md
```diff
@@ -178,9 +178,13 @@ By convention, solvers are not expected to support nonzero constant terms in the
 | Mathematical Constraint                                       | MOI Function                 | MOI Set                            |
 |---------------------------------------------------------------|------------------------------|------------------------------------|
 | ``\lVert Ax + b\rVert_2 \le c^Tx + d``                        | `VectorAffineFunction`       | `SecondOrderCone`                  |
+| ``\lVert x[2:end] \rVert_2 \le x[1]``                         | `ScalarVariablewiseFunction` | `SecondOrderCone`                  |
```
mlubin: I'd rather use latex notation than `[2:end]` here. I think it's fine to say ``y \ge \lVert x \rVert` and let users refer to the documentation for the variable ordering. In the line above we don't make a point of the order of the affine functions.

docs/src/apimanual.md
```diff
@@ -178,9 +178,13 @@ By convention, solvers are not expected to support nonzero constant terms in the
...
+| ``\lVert x[3:end] \rVert_2^2 \le 2x[1]x[2]``                  | `ScalarVariablewiseFunction` | `RotatedSecondOrderCone`           |
```
mlubin: Same comment here

PR #64, opened 2017-07-19
segregate scalar and vector sets and functions

mlubin:
Trying this out to see if it's a useful concept. Is it the case that we would only want to allow constraints of the form "vector-valued function" in "vector-input set" and "scalar-valued function" in "scalar-input set"?

This is a revival and extension of @blegat's https://github.com/JuliaOpt/MathOptInterface.jl/pull/44.

### DIFF
--- a/src/functions.jl
+++ b/src/functions.jl
@@ -8,12 +8,27 @@ Abstract supertype for function objects.
 abstract type AbstractFunction end
 
 """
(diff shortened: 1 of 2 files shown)


blegat:
We could also do it for function modification.

mlubin:
@blegat, makes sense. I'll leave that for a separate PR.

Issue #47, opened 2017-07-17
Helper functions


joaquimg:
Some functions like converting Matrices to MOI VectorAffineFunction format might be very useful for users, for instance this sketch:
```julia
function triplets(A::SparseMatrixCSC, cols::Vector{MOI.VariableReference}, rows::Vector{I})
    @assert A.m == length(rows)
    @assert A.n == length(cols)
    colval = colval(A.colptr, length(A.rowval))
    return rows[A.rowval], cols[colval], A.nzvals    
end
function colval(colptr::Vector{I}, n::Integer)
    colval = zeros(I, n)
    c = 1
    for i in 1:length(colptr)-1
        for j in colptr[i]:(colptr[i+1]-1)
            colval[c] = I(i)
            c+=1
        end
    end
    return colval
end
```
Would it be good to have a MOIExtras.jl?
Maybe solver APIs have similar functions that could also go into MOIExtras.jl ...

mlubin:
Aren't these two functions covered by `findnz` and `sparse`?

mlubin:
Oh, I guess `SparseMatrixCSC` won't let you do variables as column indices.

joaquimg:
yep, thats the problem

blegat:
@joaquimg You are welcome to add it to https://github.com/JuliaOpt/MathOptInterfaceUtilities.jl :)

joaquimg:
Can I close this? Creating https://github.com/JuliaOpt/MathOptInterfaceUtilities.jl is enough, right?

blegat:
Yes, you can close it, we can reopen it in MOIU if we add to add these functions there.

PR #65, opened 2017-07-19
Add function modification

blegat:
Add implementation of function modification.

* It creates a new function since the type is immutable.
* `change.rows` and `outputindex` are not assumed to be sorted
* Copy are done only for the fields that are modified (should we copy the other ones too ?).
* Should it belong to a MOIExtra package or should it belong here ? It will be used by [SemidefiniteOptInterface](https://github.com/JuliaOpt/SemidefiniteOptInterface.jl).

### DIFF
--- a/src/functions.jl
+++ b/src/functions.jl
@@ -155,3 +155,99 @@ struct MultirowChange{T} <: AbstractFunctionModification
     rows::Vector{Int}
     new_coefficients::Vector{T}
 end
+
+"""
+    modifyfunction(f::AbstractFunction, change::AbstractFunctionModification)
+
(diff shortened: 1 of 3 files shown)


mlubin:
I think this should go in an MOI extras package (https://github.com/JuliaOpt/MathOptInterface.jl/issues/47). The function types were designed for passing data together and not really as a final storage format that all wrappers should use.

blegat:
This might also be using by the package #9

blegat:
It has been added in https://github.com/JuliaOpt/MathOptInterfaceUtilities.jl, feel free to comment it there

PR #66, opened 2017-07-19
Fix conic duals

blegat:
The duals needs to be flipped according to the convention written in the docs (which makes sense since conic dual do not match LPQP duals in MPB and they match in MOI).
I haven't checked in the SOC and SDP tests yet but I will do that in a later PR.


mlubin:
Makes sense. The duals on equalities are hard to spot check.

Issue #63, opened 2017-07-18
performance hints


mlubin:
@joaquimg proposed a performance hint to tell the MOI wrapper to load the problem into the solver. I expect that we may decide that we want other performance hints as we start implementing things, so opening this for general discussion.

joaquimg:
I posted [this comment](https://github.com/JuliaOpt/CPLEX.jl/pull/139#issuecomment-317437852) in JuliaOpt/CPLEX.jl#139

For Xpress there is a 10x speed and memory difference between adding variables 1 by 1 and adding them in a batch.
https://gist.github.com/joaquimg/5dbaf83e7555b94d79f4d01f8ed53119

mlubin:
I expect we'll have to put a bit of work into efficiently moving the data from the JuMP `Instance` to the `SolverInstance`, but we can already create all of the variables at once with the current interface. Adding constraints one-by-one is potentially more of a performance issue.

joaquimg:
Very similar result for constraints:
https://gist.github.com/joaquimg/68c67de76eea7adea0b9ab880c667dfc

joaquimg:
What are you planning to do with instance?

Send everything to the solver as a `VectorAffineFuntion ` in `Nonpositives/Zeros`?

Otherwise, I can only see the solver having to figure out when to add everything, and then `load` is very important. Moreover, `load` and `SolverInstance` become a bit complex, since they will have to do some bookkeeping.

Any ideas?

mlubin:
> Send everything to the solver as a `VectorAffineFuntion` in `Nonpositives/Zeros`?

No, we'll be providing the same constraints to the solver as the user provided to us.

We can try:
```julia
addconstraints!(::SolverInstance,::Vector{F},::Vector{S})
```
The solver can build a sparse matrix based on this give all the constraints to the solver at once. We'd normally have to call this function three times, once for each of `LessThan`, `GreaterThan`, and `EqualTo`. That's better than 100,000 times for a problem with 100,000 constraints.

joaquimg:
3 (actually once per Set + Function combination) times should be fine.
I like this proposal.

mlubin:
This can also be used to pass in all bounds, integrality constraints, etc.

joaquimg:
yep!

mlubin:
Actually can we just overload the broadcast syntax to intercept `addconstraint!.(...)`?

mlubin:
[...]

PR #67, opened 2017-07-19
Copy vector of variable before giving it to ScalarAffineFunction

blegat:
We do `push!(v, z)` later so if we consider that `v` is owned by the function (since it does not copy), we need to copy before giving it

### DIFF
--- a/test/contlinear.jl
+++ b/test/contlinear.jl
@@ -125,7 +125,7 @@ function contlineartest(solver::MOI.AbstractSolver, ε=Base.rtoldefault(Float64)
         # x,y,z >= 0
 
         z = MOI.addvariable!(m)
-        push!(v, z)
+        v = [v; z]
         @test v[3] == z
 
         vc3 = MOI.addconstraint!(m, MOI.SingleVariable(v[3]), MOI.GreaterThan(0.0))


mlubin:
Mmm. This needs a discussion on whether solvers can modify input (https://github.com/JuliaOpt/MathProgBase.jl/issues/43). I'd tend to say that we should ask solvers not to modify the input.

odow:
Solvers should never modify input. Otherwise you might be in the position where Gurobi does, but Clp doesn't. That could lead to a change in behaviour just by changing the solver...

blegat:
Since MOI is low-level we can ask more cautiousness to the user to avoid unnecessary copy.
I would be tempted to let the solver modify it. It might do that if we call modifyconstraint later.
However it seems more reasonable to say that no one can modify the input: not the solver and not the user.

mlubin:
@blegat, I think it's a corner case that solvers would want to keep the original vectors and modify them directly. Usually I would expect MOI wrappers to copy the input into a different data structure (e.g., loading it into the solver's API). I agree with @odow here.

mlubin:
Another point is that when loading a lot of linear constraints, users might want to re-use the same input vectors to avoid allocating new ones for each constraint.

blegat:
So that means that every time JuMP will add a constraint to `Instance`, the instance will copy it. That is very inefficient.

IssamT:
Use case:
- create vector : v
- create function : f
- add constraint : f in s1
- modify vector v
- add constraint : f in s2

With the current implementation f has changed no? if yes then there isn't only a problem on the solver side but also on the user side if the function is pointing on the same vector that the user has provided.

I agree with @odow that solvers should not modify input data. but if the user is planning to modify v after giving it to a function he should provide a copy to the function. In the case the user will not modify the vector then he can provide it without copying it. In C++ I would put the argument of the function as a const reference to the vector.

blegat:
I have update the PR into a less controversial change : I do not change `v` the vector of variables in the user side.

IssamT:
How about modifying functions to use Wrappers of Tuples instead of vectors? If I am not wrong, this way a modification will be possible but only through modifying the whole Tuple.

blegat:
The conversion from Vector to tuple is not type stable so it might be inefficient.

IssamT:
But why would you use Vector if the functions are using Tuple?
The user will have:
`v = (x,y)`
create a function using that v
`v = (v... , y)`
create another function using the new v
This way the `push!` is denied by design.

blegat:
This would require JuMP to store his variables as an `NTuple`. If we want `JuMP.Model` to have concrete types, it will need to have `N` as a type parameter which means that `@variable` would need to change the type of the model.
In general, it seems that requiring every user of optimization to use tuple is rather restrictive.

mlubin:
Tuples are not an appropriate replacement for arbitrary-length vectors. We could easily have an affine function with 1,000,000 coefficients:

```
julia> tuple(ones(1000000)...)

signal (11): Segmentation fault
```

mlubin:
> So that means that every time JuMP will add a constraint to Instance, the instance will copy it. That is very inefficient.

We can make some optimizations between JuMP and its internal instance if needed. I don't think that should guide the interface design.

I would propose that we view the vectors passed though MOI as temporary and belonging to the users, not to solvers. Solvers should not assume that the contents of the vectors will not change in the future. Users *may* assume that solvers will not modify the vectors.

I believe we agree that solvers should not modify the input vectors because that creates a whole class of potential bugs. If we say that users cannot modify the input vectors after providing them to the solver, then we're removing the possibility of reusing the vectors later. I think the following is a pretty reasonable pattern for the user to do:
```
variables = Vector{VariableReference}(0)
coefficients = Vector{Float64}(0)
for each constraint we want to add
    resize variables and coefficients
    fill variables and coefficients with values
    addconstraint!(m, ScalarAffineFunction(variables, coefficients, constant), ...)
end
```

If we disallow the user to modify vectors afterwords, we have to create new ``Vector{Float64}`` and ``Vector{VariableReference}`` objects for each constraint, which could incur GC cost.

blegat:
The issue is that SOI's solvers, I store everything in MOIU's `Instance` type and only load it into the solver when `optimize!` is called. The reason is that these solvers need to first know the size of the problem before you can fill the coefficients of the constraints, ...
If I copy each function inputted before giving it to the instance, it's very inefficient.

Why not having a performance hint (see #63) saying that the input vectors will not be modified by the user ? Most solvers would not care but since JuMP will set this hint, SOI's solvers will not do extra copies when called through JuMP.

mlubin:
Is it really that inefficient? Could you quantify this?

blegat:
I will make me sleep bad when I think about it and I will not enjoy using these solvers for sum of squares thinking about the copies that are being done :D

mlubin:
I mean quantify the actual performance penalty, not the psychological damage :)

mlubin:
I'd like to see an example where this copying has a performance hit of greater than 5% of the solve time.

blegat:
If the constraints are given in a huge vector constraint then at the moment
of the copy there is twice more memory used than required and this might
cause failure.

blegat:
By the way, JuMP may also use this performance hint with Instance. This
shows that this performance hint may be useful for users of Instance

mlubin:
> If the constraints are given in a huge vector constraint then at the moment
of the copy there is twice more memory used than required and this might
cause failure.

There will also be a moment when the same amount of memory is used while the wrapper is copying the problem into the C interface, so wouldn't you have a failure at that point also?

blegat:
Closing since this will not be merged. The discussion for the performance hint remains open, e.g. https://github.com/JuliaOpt/MathOptInterfaceUtilities.jl/issues/2

PR #72, opened 2017-07-21
Add missing R in docstring

blegat:


mlubin:
We should probably move away from this syntax because it's deprecated in 0.7.

blegat:
What is the replacement in v0.7 ?

Issue #21, opened 2017-07-08
better mechanism for choosing which tests to run


mlubin:
Right now we have a few different ways of deciding whether to run a particular test or subtest:
[Example 1](https://github.com/JuliaOpt/MathOptInterface.jl/blob/ad237aef1fadfc7fd3402555987904c405661969/test/contconic.jl#L145) [Example 2](https://github.com/JuliaOpt/MathOptInterface.jl/blob/ad237aef1fadfc7fd3402555987904c405661969/test/contlinear.jl#L73).

I wanted to get rid of the ``duals`` keyword and replace it with the assumption that the solver wants us to test duals if the ``SupportsDuals`` attribute returns true, but that doesn't quite work since some solvers (ahem) do duals for linear and quadratic but not cones.

We should think up a nicer way to handle this.

blegat:
Should `SupportDuals` returns false in that case if the instance have conic constraints ?

mlubin:
``SupportsDuals`` is currently a property of a solver, not of a solver instance. I don't know if we should complicate MOI too much just because gurobi has weird behavior for conic duals. It would be better to give gurobi a way to skip the corresponding tests.

ccoffrin:
what about asking about an attribute pair, something like `(SupportsDuals, SupportsConstType)`?

blegat:
We could have one function for each test and one function that runs all `contlinear` tests, one function that runs all `contconic` tests, ...

mlubin:
That works

joaquimg:
More modularization in tests looks very good

PR #75, opened 2017-07-23
Check whether solver supports problem with only variable constraint

blegat:
CSDP does not support problems where there is only variablewise constraint but it passes the rest of the tests.


mlubin:
We should do this check consistently for all tests. What's missing is a way for a solver to assert which tests should pass (https://github.com/JuliaOpt/MathOptInterface.jl/issues/21).

PR #77, opened 2017-07-23
Modularize tests #21

blegat:


blegat:
@chriscoey What do you suggest ? Calling linear tests with linear1tset, linear2test, ... ?

blegat:
@chriscoey done

chriscoey:
OK looks good. The naming is a little redundant but we can revisit it later when all the different classes of tests are set up.

PR #78, opened 2017-07-23
Fix SemiContinuous for consistency

joaquimg:


mlubin:
`SemiInteger` is a special case because of the repeated vowel. You're forced to write "semi-integer". All of the other sets like `Semicontinuous`, `Nonnegatives`, `Nonpositives`, can be written like "semicontinuous", "nonnegatives", and "nonpositives". Changing `Semicontinous` to `SemiContinous` would not be consistent with these other sets.

odow:
I'm not a fan of special cases... `Semiinteger` isn't that bad?

mlubin:
It's a special case for the English language, so we're going have an inconsistency either in the name of the set or how we refer to it in the documentation.

joaquimg:
I don´t like `SemiInteger` to be different, looks weird

mlubin:
I won't object to `Semiinteger` if it doesn't bother people

joaquimg:
I like it better

mlubin:
`Semïnteger` anyone?

mlubin:
The diff right now is not quite as intended, I think.

joaquimg:
Would you like `const Semïnteger = Semiinteger` ?

mlubin:
> Would you like ``const Semïnteger = Semiinteger`` ?

I don't think that will be well received, unfortunately.

chriscoey:
semiinteger is fine IMO

mlubin:
We also need a line updating the name in `apireference.md`.

mlubin:
Bump. Needs docs update.

PR #80, opened 2017-07-24
Accept more solution for linear5

blegat:
Any solution such that x+y = 2, x >= 0, y >= 0 (i.e. the simplex scaled by 2) is optimal. For instance, CSDP returns [1.3095908090667123, 0.6904091909332876] which is optimal but wasn't passing the test :(

### DIFF
--- a/test/contlinear.jl
+++ b/test/contlinear.jl
@@ -547,7 +547,7 @@ function linear5test(solver::MOI.AbstractSolver, ε=Base.rtoldefault(Float64))
         #        1 x + 2 y <= 4
         #        x >= 0, y >= 0
         #
-        #   solution: x = 0, y = 2, objv = 2
+        #   solution: x = 2, y = 0, objv = 2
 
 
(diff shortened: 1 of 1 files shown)


mlubin:
It would be preferable to change the test so that there's a unique optimal solution.

blegat:
Indeed, I have update the coefficient change, there is now a unique sol

test/contlinear.jl
```diff
@@ -543,14 +543,14 @@ function linear5test(solver::MOI.AbstractSolver, ε=Base.rtoldefault(Float64))
         # change coeff
         #   maximize x + y
         #
-        #   s.t. 2 x + 2 y <= 4
+        #   s.t. 2 x + 3 y <= 4
         #        1 x + 2 y <= 4
         #        x >= 0, y >= 0
         #
         #   solution: x = 0, y = 2, objv = 2
```
joaquimg: needs fix here also

Issue #71, opened 2017-07-21
Questions implementing a wrapper


odow:
I'm going to use this issue as a collection place for a range of questions I have implementing the CPLEX solver wrapper:

**On the purpose of low-level interface**
- [x] should we expose the full (potentially ad-hoc, complicated) API, or a simplified one to satisfy MOI requirements. **Don't remove existing functionality**

**On variable bounds**
- [x] is `addconstraint!(m, v::VectorOfVariables, set::LessThan{Float64})` a valid constraint? **oO**
- [x] should `addconstraint!(m, v::VectorOfVariables, set::Vector{Union{LessThan{Float64}, GreaterThan{Float64}, EqualsTo{Float64}, Interval{Float64}}})` be a valid constraint so users can set bounds all at once? **No. <s>Overload the broadcast operator https://github.com/JuliaOpt/MathOptInterface.jl/issues/63#issuecomment-317545139</s> https://github.com/JuliaOpt/MathOptInterface.jl/issues/63#issuecomment-317526848**

joaquimg:
Maybe with plural: `addconstraints!(m, v::VectorOfVariables, set::Vector{Union{LessThan{Float64}, GreaterThan{Float64}, EqualsTo{Float64}, Interval{Float64}}})` this could return vector of references...

mlubin:
>  is ``addconstraint!(m, v::VectorOfVariables, set::LessThan{Float64})`` a valid constraint?

We can't prevent a solver from recognizing the constraint, but no (https://github.com/JuliaOpt/MathOptInterface.jl/pull/64). `VectorOfVariables` is a vector-valued function and `LessThan` is a scalar-input set.

> should ``addconstraint!(m, v::VectorOfVariables, set::Vector{Union{LessThan{Float64}, GreaterThan{Float64}, EqualsTo{Float64}, Interval{Float64}}})`` be a valid constraint so users can set bounds all at once?

It's one function per set in MOI. `VectorOfVariables` is one function, so it doesn't make sense to provide multiple sets there. We could define ``addconstraints!(m, ::Vector{F}, ::Vector{S})`` to return a vector of `ConstraintReference{F,S}` for concrete `F` and `S`. I don't see a reasonable (type-stable) way to do that when `S` is a union type .

> should we expose the full (potentially ad-hoc, complicated) API, or a simplified one to satisfy MOI requirements.

[...]

PR #82, opened 2017-07-24
define addconstraints! (#71, #63)

mlubin:


mlubin:
Should we let solvers define a fallback or should we define a global fallback?
Also needs tests.

joehuchette:
+1 for a global fallback

mlubin:
Added a fallback

PR #90, opened 2017-07-27
attributes to access list of references present in an instance

mlubin:
This is particularly useful if you only have an instance object and want to extract data from it (e.g., in JuMP). Right now you're forced to keep track of the variable and constraint references separately.


odow:
Can this be merged?

Issue #79, opened 2017-07-24
move test/function_utilities.jl into src or MathOptInterfaceUtilities


chriscoey:
https://github.com/JuliaOpt/MathOptInterface.jl/blob/master/test/function_utilities.jl

is there anything else we want to redefine `Base.isapprox` for? may want to rename the file.

any preference where it goes?

blegat:
It is already in MOIU: https://github.com/JuliaOpt/MathOptInterfaceUtilities.jl/blob/f82dc2a0746efc20621db35118acde5fad395ef1/src/functions.jl#L28-L72

chriscoey:
is it OK to use MOIU and remove the file from test?

blegat:
That is also what I would expect to happen

mlubin:
Fine with me

PR #93, opened 2017-07-27
Fixes needed to get tests to run

odow:
Not sure how other solvers passed tests without these?

Note: not a complete set of changes. I may discover more

### DIFF
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -86,10 +86,15 @@ end
 
 Return a `Bool` indicating whether it is possible to query attribute `attr` from the solver `s`.
 
-    cangetattribute(m::AbstractSolverInstance, attr::AbstractVariableAttribute, R::Type{VariableReference})::Bool
-    cangetattribute(m::AbstractSolverInstance, attr::AbstractConstraintAttribute, R::Type{ConstraintReference{F,S})::Bool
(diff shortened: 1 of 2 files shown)


blegat:
I think I was passing the tests because MOIU was loaded and it defines `Base.isapprox`, see #79

src/attributes.jl
```diff
@@ -86,21 +86,31 @@ end
...
 cangetattribute(m, ObjectiveValue())
 cangetattribute(m, VariablePrimalStart(), varref)
 cangetattribute(m, ConstraintPrimal(), conref)
+cangetattribute(m, VariablePrimal(), [ref1, ref2])
 ```
 """
 function cangetattribute end
 cangetattribute(m::AbstractSolverInstance, attr::AnyAttribute) = false
+cangetattribute(m::AbstractSolverInstance, attr::AnyAttribute, ref::AnyReference) = false
+function cangetattribute(m::AbstractSolverInstance, attr::AnyAttribute, refs::Vector{T}) where T <: AnyReference
```
blegat: You can write `Vector{<: AnyReference}` directly without needing to create the variable `T`. I think it is new in Julia v0.6 and it is quite handy :-P

PR #96, opened 2017-07-28
Add solver list

blegat:
This should add coverage to linear and conic tests and make sure they keep passing.
Depends on #94


mlubin:
Isn't it early for this? We will still be making some changes in both MOI and the solver interfaces. Also I don't think we should replicate JuMP's bad design here. The solver interfaces should decide which MOI tests they should pass, not the other way around

blegat:
What do you suggest to do to run the tests in MOI ? Do like in MPB and hardcode for each test a solver that passes them ?

mlubin:
I think the most correct approach would be to include a reference solver implementation in MOI and test that. Whether we want to do that is another question.

blegat:
In a separate package or in MOI/test folder itself ? That would be a lot of code to have a solver that supports everything.

mlubin:
In MOI/src I'd guess. Having a reference implementation tied to MOI makes much more sense than using external solvers to test MOI's tests.

The reference implementation could do a grid search on continuous variables and enumeration for discrete variables. It would be fine if it only works for 3 dimensions.

PR #101, opened 2017-08-01
rename Sense attribute to ObjectiveSense

### LINKED ISSUE #91: rename sense attribute?
The attribute name ``Sense`` seems inconsistently short for MOI. I keep thinking it should be ``ObjectiveSense``. Opinions?

mlubin:
Closes #91

### DIFF
--- a/docs/src/apireference.md
+++ b/docs/src/apireference.md
@@ -61,7 +61,7 @@ List of solver instance attributes
 
 ```@docs
 RawSolver
-Sense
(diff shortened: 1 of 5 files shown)


mlubin:
It was probably a bad idea to depend on MOIU in the MOI tests. That's what's failing.

mlubin:
No objections so I'll just merge so the breakage happens sooner rather than later.

Issue #99, opened 2017-08-01
More info about TerminationStatus/PrimalStatus etc


joaquimg:
Should we have a minimally standardized way to return more info about the status?

Some output info of [Xpress and CPLEX](https://github.com/JuliaOpt/Xpress.jl/pull/10#discussion_r130509494) do not match any existing status (and probably should not match at all) but we could have a default way to return a string/symbol/error/warning with that extra info.

blegat:
Having an attribute that returns a string would be nice indeed.

mlubin:
``SpecificTerminationStatus`` (or another name) is fine as an attribute where solvers could return a solver-dependent string. Could we discuss the specific cases that don't map well? We put a lot of work into designing the status codes...

odow:
I think the statuses are great. It's more a case of how much effort we should put in.

CPLEX solves an LP, returns `CPX_STAT_INForUNBD`. It also returns `CPX_NO_SOLN` so I don't have any information on the primal or dual. Should I stop there or try to find a certificate? https://github.com/JuliaOpt/CPLEX.jl/pull/139/files#diff-5b9fa9efaaa021b1607ca6ccdfc75248R40 

And if I do, is the termination status now `Success`? Because what about the time limit?

mlubin:
``CPX_STAT_INForUNBD`` maps directly to ``InfeasibleOrUnbounded`` and I presume no solution is available so there is no primal status code to return.

``CPX_NO_SOLN`` is seems like a solution status, not a termination status. If there is no solution available you should simply say that ``cangetattribute(...,PrimalStatus()) == false``.

odow:
Should there be a resultstatuscode for `NoResult`?

mlubin:
No, primal status is tied to a result number (http://www.juliaopt.org/MathOptInterface.jl/latest/apireference.html#MathOptInterface.PrimalStatus). If no number is provided, it defaults to 1. If there are zero results, then the primal status of result number 1 simply cannot be accessed.

odow:
Ah I missed this. I'll add some tests

odow:
Should there be a `PrimalResultCount` and `DualResultCount`?

joaquimg:
Makes sense for me

mlubin:
The original concept was to have a single result count since I'd consider the primal-dual pair to be tied together. If either the primal or dual part of a result is missing, then that information is available though ``cangetattribute``.

IssamT:
[...]

PR #103, opened 2017-08-01
WIP: More tests

odow:
Port some of the MPB tests from linprog (1 c,d,e from https://github.com/JuliaOpt/MathOptInterface.jl/issues/23) and add tests for the result count attribute https://github.com/JuliaOpt/MathOptInterface.jl/issues/99#issuecomment-319501666

Closes https://github.com/JuliaOpt/MathOptInterface.jl/issues/105

### DIFF
--- a/test/contlinear.jl
+++ b/test/contlinear.jl
@@ -14,6 +14,10 @@ function linear1test(solver::MOI.AbstractSolver, ε=Base.rtoldefault(Float64))
 
         @test MOI.supportsproblem(solver, MOI.ScalarAffineFunction{Float64}, [(MOI.ScalarAffineFunction{Float64},MOI.LessThan{Float64}),(MOI.SingleVariable,MOI.GreaterThan{Float64})])
 
+        @test MOI.getattribute(solver, MOI.SupportsAddConstraintAfterSolve())
+        @test MOI.getattribute(solver, MOI.SupportsAddVariableAfterSolve())
+        @test MOI.getattribute(solver, MOI.SupportsDeleteConstraint())
+
         m = MOI.SolverInstance(solver)
 
         v = MOI.addvariables!(m, 2)
@@ -509,7 +513,7 @@ end
 
 function linear5test(solver::MOI.AbstractSolver, ε=Base.rtoldefault(Float64))
     @testset "Change coeffs, del constr, del var" begin
-
+        @test MOI.getattribute(solver, MOI.SupportsDeleteVariable())
         #####################################
         # Start from simple LP
         # Solve it
--- a/test/contlinear.jl
+++ b/test/contlinear.jl
@@ -933,7 +933,7 @@ function linear10test(solver::MOI.AbstractSolver, ε=Base.rtoldefault(Float64))
             [MOI.GreaterThan(0.0), MOI.GreaterThan(0.0)]
         )
 
-        MOI.addconstraint!(m, MOI.ScalarAffineFunction([x,y], [1.0, 1.0], 0.0), MOI.Interval(5.0, 10.0))
+        c = MOI.addconstraint!(m, MOI.ScalarAffineFunction([x,y], [1.0, 1.0], 0.0), MOI.Interval(5.0, 10.0))
 
         MOI.setobjective!(m, MOI.MaxSense,
             MOI.ScalarAffineFunction([x, y], [1.0, 1.0], 0.0)
@@ -945,6 +945,13 @@ function linear10test(solver::MOI.AbstractSolver, ε=Base.rtoldefault(Float64))
         @test MOI.getattribute(m, MOI.PrimalStatus()) == MOI.FeasiblePoint
         @test MOI.getattribute(m, MOI.ObjectiveValue()) ≈ 10.0 atol=ε
 
+        if MOI.getattribute(solver, MOI.SupportsDuals())
+            @test MOI.cangetattribute(m, MOI.DualStatus())
+            @test MOI.getattribute(m, MOI.DualStatus()) == MOI.FeasiblePoint
+            @test MOI.cangetattribute(m, MOI.ConstraintDual(), c)
+            @test MOI.getattribute(m, MOI.ConstraintDual(), c) ≈ -1 atol=ε
+        end
+
         MOI.setobjective!(m, MOI.MinSense,
             MOI.ScalarAffineFunction([x, y], [1.0, 1.0], 0.0)
         )
@@ -954,6 +961,13 @@ function linear10test(solver::MOI.AbstractSolver, ε=Base.rtoldefault(Float64))
         @test MOI.getattribute(m, MOI.TerminationStatus()) == MOI.Success
         @test MOI.getattribute(m, MOI.PrimalStatus()) == MOI.FeasiblePoint
         @test MOI.getattribute(m, MOI.ObjectiveValue()) ≈ 5.0 atol=ε
+
+        if MOI.getattribute(solver, MOI.SupportsDuals())
+            @test MOI.cangetattribute(m, MOI.DualStatus())
+            @test MOI.getattribute(m, MOI.DualStatus()) == MOI.FeasiblePoint
+            @test MOI.cangetattribute(m, MOI.ConstraintDual(), c)
+            @test MOI.getattribute(m, MOI.ConstraintDual(), c) ≈ 1 atol=ε
+        end
     end
 
 end
--- a/test/contlinear.jl
+++ b/test/contlinear.jl
@@ -5,7 +5,7 @@ using MathOptInterfaceUtilities # Defines isapprox for ScalarAffineFunction
 
 # Continuous linear problems
 
-function linear1test(solver::MOI.AbstractSolver, ε=Base.rtoldefault(Float64))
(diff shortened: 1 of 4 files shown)


test/contlinear.jl
```diff
@@ -144,8 +144,11 @@ function linear1test(solver::MOI.AbstractSolver, ε=Base.rtoldefault(Float64))
         @test MOI.cangetattribute(m, MOI.TerminationStatus())
         @test MOI.getattribute(m, MOI.TerminationStatus()) == MOI.Success
 
-        @test MOI.cangetattribute(m, MOI.PrimalStatus())
-        @test MOI.getattribute(m, MOI.PrimalStatus()) == MOI.FeasiblePoint
+        @test MOI.cangetattribute(m, MOI.ResultCount())
+        @test MOI.getattribute(m, MOI.ResultCount()) == 1
```
joaquimg: I am not sure, but maybe we could test for `MOI.getattribute(m, MOI.ResultCount()) >= 1` instead of `== 1`.

test/intlinear.jl
```diff
@@ -1,6 +1,80 @@
...
+        @test MOI.getattribute(m, MOI.ResultCount()) >= 1
+
+        @test MOI.cangetattribute(m, MOI.PrimalStatus())
+        @test MOI.getattribute(m, MOI.PrimalStatus()) == MOI.FeasiblePoint
+
+        @test MOI.cangetattribute(m, MOI.ObjectiveValue())
+        @test MOI.getattribute(m, MOI.ObjectiveValue()) ≈ 19 atol=ε
+
+        @test MOI.cangetattribute(m, MOI.VariablePrimal(), v)
+        @test MOI.getattribute(m, MOI.VariablePrimal(), v) ≈ [4,5,1] atol=ε
```
joaquimg: The problem seems to be degenerate, `x = 0, y = 7, z = 1`, maybe we could exchange the bounds between `x` and `y`

odow: Yeah. I just copied the CPLEX problem

odow:
@joaquimg do you have SOS working in Xpress? I can't get CPLEX to pass the `SOSII` test with MOI...

It's like I'm adding in the wrong constraint somewhere...

odow:
This should probably get merged soon as it contains a lot of new tests and various fixes to current ones that are necessary to get passing.

test/contlinear.jl
```diff
@@ -726,6 +750,100 @@ function linear7test(solver::MOI.AbstractSolver, ε=Base.rtoldefault(Float64))
...
+        MOI.setobjective!(m, MOI.MinSense, MOI.ScalarAffineFunction([x], [1.0], 0.0))
+        MOI.optimize!(m)
+
+        @test MOI.cangetattribute(m, MOI.ResultCount())
+        if MOI.getattribute(m, MOI.ResultCount()) == 1
+            # solver returned an infeasibility ray
+            @test MOI.getattribute(m, MOI.TerminationStatus()) == MOI.Success
+            @test MOI.getattribute(m, MOI.DualStatus()) == MOI.InfeasibilityCertificate
+            @test MOI.cangetattribute(m, MOI.ConstraintDual(), c)
+            @test MOI.getattribute(m, MOI.ConstraintDual(), c) ≈ -1 atol=ε
```
mlubin: What should the dual values be on the bound constraints? These should be tested also.

test/contlinear.jl
```diff
@@ -726,6 +750,100 @@ function linear7test(solver::MOI.AbstractSolver, ε=Base.rtoldefault(Float64))
...
+        MOI.addconstraint!(m, MOI.SingleVariable(x), MOI.GreaterThan(0.0))
+        MOI.addconstraint!(m, MOI.SingleVariable(y), MOI.GreaterThan(0.0))
+        MOI.setobjective!(m, MOI.MinSense, MOI.ScalarAffineFunction([x, y], [-1.0, -1.0], 0.0))
+        MOI.optimize!(m)
+
+        @test MOI.cangetattribute(m, MOI.ResultCount())
+        if MOI.getattribute(m, MOI.ResultCount()) == 1
+            # solver returned an unbounded ray
+            @test MOI.getattribute(m, MOI.TerminationStatus()) == MOI.Success
+            @test MOI.getattribute(m, MOI.PrimalStatus()) == MOI.ReductionCertificate
```
mlubin: ``ReductionCertificate`` is incorrect here. Looks like we forgot to include a ray status.

odow: What exactly is a reduction certificate then? I just assumed it was some fancy conic word for an unbounded ray

mlubin: Actually the ray status is there and is called `InfeasibilityCertificate` because a primal unbounded ray is an infeasibility certificate of the dual. Facial reduction certificates are mentioned briefly [in the manual](http://www.juliaopt.org/MathOptInterface.jl/latest/apimanual.html#Primal-dual-convex-solver-1); they're for cases of duality failures or unattained solutions which never happen for LP.

joaquimg:
@odow I am planning to add SOS support this week

test/intlinear.jl
```diff
@@ -26,6 +298,10 @@ function knapsacktest(solver::MOI.AbstractSolver, eps=Base.rtoldefault(Float64))
 
         MOI.setobjective!(m, MOI.MaxSense, MOI.ScalarAffineFunction(v, [5.0, 3.0, 2.0, 7.0, 4.0], 0.0))
 
+        if MOI.cansetattribute(m, MOI.VariablePrimalStart(), v)
+            MOI.setattribute!(m, MOI.VariablePrimalStart(), v, [0.0, 0.0, 0.0, 0.0, 0.0])
```
joaquimg: Any ideas on how do we check this method is working?

joaquimg:
I remember a bug when modifying quadratic objectives: the quadratic terms were not being re-set to zero. Are we testing that?

odow:
Okay. Before I make this PR any longer, lets get this merged. @blegat or @joaquimg looks like you have write access?

test/intlinear.jl
odow: Nope

joaquimg:
Great job! Contlinear and intlinear are all passing in CPLEX? I agree with merging today. @blegat do you have any problem with this?

test/contlinear.jl
```diff
@@ -726,6 +754,241 @@ function linear7test(solver::MOI.AbstractSolver, ε=Base.rtoldefault(Float64))
...
+        )
+
+        MOI.optimize!(m)
+
+        @test MOI.getattribute(m, MOI.TerminationStatus()) == MOI.Success
+        @test MOI.getattribute(m, MOI.PrimalStatus()) == MOI.FeasiblePoint
+
+        @test MOI.getattribute(m, MOI.ObjectiveValue()) ≈ 79e4/11 atol=ε
+        @test MOI.getattribute(m, MOI.VariablePrimal(), x) ≈ 650/11 atol=ε
+        @test MOI.getattribute(m, MOI.VariablePrimal(), y) ≈ 400/11 atol=ε
```
blegat: I would use `rtol` instead of `atol` here. `isapprox` is implemented as
```julia
abs(x-y) <= atol + rtol*max(abs(x), abs(y)))
```
where `rtol` is `1.4901161193847656e-8` for `Float64` by default.
For the objective `rtol * max(abs(x), abs(y))` is approximately 1e-3 so if `atol` is like `1e-5` it will be neglibeable.
CSDP fails the test for `VariablePrimal`/`x` because epsilon is 1e-7 which means that `atol` is 1e-7` but `rtol * max(abs(x), abs(y))` is about `1e-7` too. CSDP has a relative tolerance of `1e-7` so it should pass the test but here it is tested for an *absolute* tolerance of 1e-7.

In fact, it makes more sense to only use `atol` when one of the two numbers being compared is 0 or very close to 0. Why we use `atol` everywhere in MOI's tests ? Usually it does not matter much but here since the numbers are large, it does matter.

odow:
> Contlinear and intlinear are all passing in CPLEX?

Contlinear, intlinear, contquadratic, and most of contconic are passing.

test/contlinear.jl
odow: Solvers are able to specify different tolerances... for example https://github.com/JuliaOpt/CPLEX.jl/pull/139/files#diff-fce720c43af3c52c862fd7451c7374b8R30

blegat: Yes but the tolerance should be relative, not absolute right ?

blegat: It might be cumbersome since usually the solver just tells its accuracy and we should expect to pass the solvers with epsilon = this accuracy.
I would suggest using atol when what we compare to zero and rtol when we compare to nonzero. `rtol` is pretty bad close to zero:
```julia
isapprox(1e-9, 1e-10, rtol=1e-2)
```
So we do
```julia
@test x ≈ 0 atol=ε
@test x ≈ 1 rtol=ε
@test x ≈ y atol=ε rtol=ε # x and y might be close to zero or far from zero, we don't know. I don't know if this case happens in the tests though
```

odow: Have you seen 83f919e?

blegat: Yes, we can use this solution if everyone agree on it.

Issue #69, opened 2017-07-20
add shorthand for SingleVariable and VectorOfVariables


mlubin:
https://github.com/JuliaOpt/MathOptInterface.jl/issues/60#issuecomment-316474508

PR #107, opened 2017-08-09
Add shorthand fallbacks for constraints

odow:
Closes https://github.com/JuliaOpt/MathOptInterface.jl/issues/69


odow:
@mlubin you're meant to be on holiday...

joaquimg:
I also liked! I would add to docstrings also

Issue #89, opened 2017-07-27
Change constraint "sense"


joaquimg:
Many solvers allow us to change the constraint from equality to inequality. The first idea was not to support it https://github.com/JuliaOpt/MathOptInterface.jl/pull/38/commits/f245a6ac10a18da2463c612c7859017e00e76414 , one had to delete a constraint and add a new one, however some solver might be able to modify the sign and not be able to delete. Moreover modifying looks simpler.

The idea is to have something along the lines:
```julia
c_new = modifyconstraint(m::SolverInstance, ::MOI.ConstraintSet, 
    c::MOI.ConstraintReference{F, MOI.LessThan{T}}, 
    newset::MOI.EquaTo{T})::MOI.ConstraintReference{F, MOI.EqualTo{T}}
```

mlubin:
I think we should have separate names for methods that modify constraints without invalidating the constraint reference and those which modify constraints and return a new reference. Maybe ``transformconstraint!`` instead of ``modifyconstraint!``?

joaquimg:
It could `transformconstraint!` to make it very clear it does something a bit different.

blegat:
I would remove the `MOI.ConstraintSet` argument in order to be consistent with `modifyconstraint!` ?

PR #108, opened 2017-08-09
Transform constraint

odow:
Closes https://github.com/JuliaOpt/MathOptInterface.jl/issues/89

### DIFF
--- a/src/constraints.jl
+++ b/src/constraints.jl
@@ -73,3 +73,31 @@ modifyconstraint!(m, c, ScalarConstantChange(10.0))
 ```
 """
 function modifyconstraint! end
+
+
+"""
+## Transform Constraint Set
+
+    transformconstraint!(m::AbstractSolverInstance, c::ConstraintReference{F,S1}, newset::S2)::ConstraintReference{F,S2}
+
+Replace the set in constraint `c` with `newset`. The constraint reference `c`
+will no longer be valid, and the function returns a new constraint reference.
+
+Solvers may only support a subset of constraint transforms that they perform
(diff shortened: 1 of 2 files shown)


blegat:
Why don't we do
```julia
function transformconstraint!(m::AbstractSolverInstance, c::ConstraintReference{F,S1}, newset::S2)
    f = getattribute(m, ConstraintFunction(), c)
    delete!(m, c)
    addconstraint!(m, f, newset)
end
```
That way, if the solver have something faster to do, it will be called instead.
The user knows that by calling `transformconstraint!` he will always do the most efficient thing.

joaquimg:
That is nice. But then `transformconstraint!` always return a new reference right?

blegat:
How could it return the same reference since it has a different type ?

joaquimg:
Indeed, I got confused with the `modify`. 
A second problem, the solver might not support `delete`, so there must e a `candelete` or `supporttransform`

blegat:
Yes, we could have a
```julia
function cantransformconstraint!(m::AbstractSolverInstance, c::ConstraintReference{F,S1}, newset::S2)
    candelete(m, c)
end
```

odow:
I'm away for a few days. I think we need canmodifyconstraint and cantransformconstraint with the same arguments. Or a better way of throwing errors

joaquimg:
better:
```julia
function cantransformconstraint!(m::AbstractSolverInstance, c::ConstraintReference{F,S1}, newset::S2)
    cangetattribute(m, ConstraintFunction(), c)
    candelete(m, c)
end
```

blegat:
Just need a rebase :)

odow:
Rebased

PR #109, opened 2017-08-11
Fix test args, test for SOS-constraint suppoort

ulfworsoe:
Disable SOS tests for non-SOS solvers. Absense of SOS support should not produce a test fail.
I have added `NearlyFeasible` as an acceptable result for some tests.

### DIFF
--- a/test/intlinear.jl
+++ b/test/intlinear.jl
@@ -49,7 +49,7 @@ function int1test(solver::MOI.AbstractSolver; atol=Base.rtoldefault(Float64), rt
         @test MOI.getattribute(m, MOI.ObjectiveSense()) == MOI.MaxSense
 
         MOI.optimize!(m)
-
+        
         @test MOI.cangetattribute(m, MOI.TerminationStatus())
         @test MOI.getattribute(m, MOI.TerminationStatus()) == MOI.Success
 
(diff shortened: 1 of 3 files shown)


test/intlinear.jl
```diff
@@ -49,15 +49,15 @@ function int1test(solver::MOI.AbstractSolver; atol=Base.rtoldefault(Float64), rt
         @test MOI.getattribute(m, MOI.ObjectiveSense()) == MOI.MaxSense
 
         MOI.optimize!(m)
-
+
```
blegat: Some blank space added here

odow:
I think this is another indication we need a better way for solvers to choose which tests to run

blegat:
What we could do is defining:
* the problems/operations that a solver should support if it expects to pass intlinear/contlinear/contconic tests.
* If the problem is part of these then we do `@test supportsproblem ...`, otherwise we do `if supportsproblem ...`
* if a solver does not want to support one of these problems/operations then it can skip the test by calling each tests individually without this one.

For example it is too restrictive to ask for every contconic solvers to support SDP.
However we can probably expect for every contlinear solver to support interval constraint since it is easy to transform it to a `LessThan` and a `GreaterThan` constraints in the wrapper.
@ulfworsoe do you think that it would be possible to make the transformation of SOS constraints into classical constraints ?

PR #111, opened 2017-08-12
add canmodifyconstraint

### LINKED ISSUE #85: Check if constraints can be modified

joaquimg:
closes #85

### DIFF
--- a/src/constraints.jl
+++ b/src/constraints.jl
@@ -19,7 +19,51 @@ function addconstraints! end
 # default fallback
 addconstraints!(m::AbstractSolverInstance, funcs, sets) = addconstraint!.(m, funcs, sets)
 
-# TODO: method to query if solver supports this type of modification
+"""
+## Modify Function
+
+    canmodifyconstraint(m::AbstractSolverInstance, c::ConstraintReference, func::F)::Bool
+
+Return a `Bool` indicating whether it is possible to replace the function in constraint `c` with `func`. `F` must match the original function type used to define the constraint.
+
+### Examples
+
+If `c` is a `ConstraintReference{ScalarAffineFunction,S}` and `v1` and `v2` are `VariableReference` objects,
+
+```julia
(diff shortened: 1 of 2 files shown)


joaquimg:
Should we have a fallback like in https://github.com/JuliaOpt/MathOptInterface.jl/pull/108#issuecomment-321199454

blegat:
We can have a fallback for AbstractFunctionModification but it needs to be in MOIU since function modification is implemented there

src/constraints.jl
```diff
@@ -19,7 +19,51 @@ function addconstraints! end
 # default fallback
 addconstraints!(m::AbstractSolverInstance, funcs, sets) = addconstraint!.(m, funcs, sets)
 
-# TODO: method to query if solver supports this type of modification
+"""
+## Modify Function
+
+    canmodifyconstraint(m::AbstractSolverInstance, c::ConstraintReference, func::F)::Bool
```
odow: `c::ConstraintReference{F,S}, func::F)::Bool`

src/constraints.jl
```diff
@@ -19,7 +19,51 @@ function addconstraints! end
...
+If `c` is a `ConstraintReference{ScalarAffineFunction,S}` and `v1` and `v2` are `VariableReference` objects,
+
+```julia
+canmodifyconstraint(m, c, ScalarAffineFunction([v1,v2],[1.0,2.0],5.0))
+canmodifyconstraint(m, c, SingleVariable(v1)) # false
+```
+
+## Modify Set
+
+    canmodifyconstraint(m::AbstractSolverInstance, c::ConstraintReference, S::S)::Bool
```
odow: `c::ConstraintReference{F,S}, set::S)::Bool`

odow:
> Should we have a fallback like in #108 (comment) ?

No. I think these methods should stay as simple as possible. Modification means in-place.

PR #115, opened 2017-08-12
Include test scripts

joaquimg:
At least we see if there are no weird sintax mistakes

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,8 @@ git:
 ## uncomment the following lines to override the default test script
 script:
   - julia -e 'Pkg.clone(pwd()); Pkg.build("MathOptInterface"); Pkg.test("MathOptInterface"; coverage=true)'
+    Pkg.clone("https://github.com/JuliaOpt/MathOptInterfaceUtilities.jl.git");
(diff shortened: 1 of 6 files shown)


joaquimg:
How we add MathOptInterfaceUtilities?

blegat:
I am not personally against it but it is going against #104 this change will probably last the time of @mlubin 's vacations :D

joaquimg:
we can wait for him, I just wanted to include the files to check sintax errors

joaquimg:
@mlubin are you against this one after all?

mlubin:
I think we should try as hard as possible to make MOI's runtests not depend on any downstream packages.

joaquimg:
I agree with you. But the idea here is just to load MOI (and nothing else) in runtests so that we avoid some syntax mistakes.

mlubin:
Well we can move the definition of ``isapprox`` back into MOI or duplicate it in the tests folder. Having standalone tests is a sufficient motivation.

blegat:
That is not the only thing being used. IIRC there is also getindex and VectorOfVariables -> VectorAffineFunction conversion

mlubin:
No longer relevant given MOIT

PR #122, opened 2017-08-15
Add fallback for transformconstraint!

blegat:


odow:
I am not in favour of this.

blegat:
@odow I don't understand what you mean by your second sentence, which route are you talking about ?

odow:
If it is not supported natively, a solver like CPLEX could choose to implement
```julia
function transformconstraint!(m::CplexSolverInstance, c::ConstraintReference{ScalarAffineFunction{Float64}, LessThan{Float64}}, newset::GreaterThan{Float64})
    f = getattribute(m, ConstraintFunction(), c)
    delete!(m, c)
    addconstraint!(m, f, newset)
end
```
but it should not be a default fallback. Many solvers will not support querying constraint functions, or deleting constraints, or even adding constraints of `(f, newset)`.

blegat:
Yes that is why we will do
```julia
function cantransformconstraint(m, c, newset)
    cangetattribute(m, ConstraintFunction(), c) && candelete(m, c) && canaddconstraint(m, f, newset)
end
```

If a user calls `transformconstraint!` which `cantransformconstraint!` returns false, can he complain to get a method error on `getattribute`, `delete!` or `addconstraint!` instead of a MethodError on `cantransformconstraint` ?

odow:
I still think we should hold off adding too many layers like this until there is demand. MOI is meant to be a thin wrapper for native solver abilities

blegat:
What we are doing here is exactly what we did for `addconstraints!`. `addconstraints!` and `transformconstraint!` do not enable the user of MOI to do things he wasn't able to do before, however the user knows that if he wants to do exactly one of these 2 tasks, he'd better call the corresponding function since in the case the solver has something faster to do, it will do it.

If we do not provide a fallback, we allow solvers not to implement `transformconstraint!` or `addconstraints!` so it sends a confusing message to the user. "If you use one of this function, for some solver, it will be faster, for some it will be the same as doing the equivalent using `addconstraint!`, ... and for some solver, it will just throw a MethodError so please use `cantransformconstraint` before and in case it does not support it do the equivalent too". Every single user will have to do this `if`, `else` and the copy paste of the fallback in their code. So the fallback will be copy pasted both in every user codes and every wrapper code that decide to implement it.

It seems to me that we need to distinguish two types of functions: some that have a fallback and the others. The functions that have a fallback are functions that the user is advised to use since it will always be at least as fast as not using it and in case the solver can do something faster then it will be strictly faster.

mlubin:
We should be consistent with the fallbacks at least. When they're simple like here or ``addconstraints!`` then we should either always have a fallback or never have a fallback.

blegat:
What do you mean by "always" ?

mlubin:
As in this PR

PR #124, opened 2017-08-16
Add tests for candelete and canmodify

odow:
So far only added to contlinear and intlinear. There should be more modification tests in conic and quadratic...

I also reverted https://github.com/JuliaOpt/MathOptInterface.jl/pull/121 to default to the boolean `false` rather than throw an error which seems like the better thing to do.

### DIFF
--- a/docs/src/apireference.md
+++ b/docs/src/apireference.md
@@ -118,6 +118,7 @@ ConstraintReference
 candelete
 isvalid
 delete!(::AbstractSolverInstance,::AnyReference)
+candelete
 ```
 
 ### Variables
@@ -148,6 +149,8 @@ addconstraint!
 addconstraints!
 modifyconstraint!
 canmodifyconstraint
+transformconstraint!
+cantransformconstraint
 ```
 
 List of attributes associated with constraints. [category AbstractConstraintAttribute]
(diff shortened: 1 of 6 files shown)


blegat:
Returning false instead of throwing a method error might makes bug harder to spot. In which case would it be useful ? If we ask `isvalid` or `candelete` to a solver with some type that it does not know, wouldn't it be better to throw a MethodError ?

odow:
Why? By default, things can't be modified so `canmodify() = false`. Solvers can explicitly state that they can.

```julia
m = SolverInstance(CplexSolver())

v = addvariable(m)
candelete(v) # true
delete!(m, v)

f = ConstraintReference{Int, Int}(1)
candelete(f) # false
delete!(m, f) # MethodError: no method matching delete!(::CplexSolverInstance, ::ConstraintReference{Int, Int})
```

blegat:
The question really is : what should a solver say if we ask `isvalid(m, v)` with a ConstraintReference that doesn't make sense.
Does `isvalid` means "has already been deleted" or more generally "is it a valid reference". Currently in the docs it is closer to the second one so it goes in your direction.

odow:
`isvalid=true` means that it points to a valid constraint that exists in the solver instance and can be used to query/modify etc.

PR #127, opened 2017-09-08
Improves infeasibility ray check of lineartest8

blegat:
The dual of the constraint does not have to be `-1`, it can be scaled by any positive constant since it is a ray.

### DIFF
--- a/test/contlinear.jl
+++ b/test/contlinear.jl
@@ -804,13 +804,12 @@ function linear8test(solver::MOI.AbstractSolver; atol=Base.rtoldefault(Float64),
             @test MOI.canget(instance, MOI.ConstraintDual(), c)
             cd = MOI.get(instance, MOI.ConstraintDual(), c)
             @test cd < -atol
-            # TODO: farkas dual on bounds - see #127
-            # xd = MOI.get(instance, MOI.ConstraintDual(), bndx)
-            # yd = MOI.get(instance, MOI.ConstraintDual(), bndy)
-            # @test xd > atol
-            # @test yd > atol
-            # @test yd ≈ -cd atol=atol rtol=rtol
-            # @test xd ≈ -2cd atol=atol rtol=rtol
+            xd = MOI.get(instance, MOI.ConstraintDual(), bndx)
+            @test xd > atol
+            yd = MOI.get(instance, MOI.ConstraintDual(), bndy)
+            @test yd > atol
+            @test yd ≈ -cd atol=atol rtol=rtol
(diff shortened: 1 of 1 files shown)


joaquimg:
Actually I cant approve :( Xpress does not return rays for variable bounds.
However I need the first part because the constraint ray should be just negative as you fixed

mlubin:
> Xpress does not return rays for variable bounds.

Are you sure? That seems strange.

joaquimg:
Yep, pretty sure. At least for Xpress up to 8.0 version (the last is 8.3).

They have a `get[primal/dual]ray` API that is a bit limited.

blegat:
Can't you determine the value of the ray for the variable bounds ?

joaquimg:
How?

blegat:
In the test of this PR for example, the problem is
```
min x
s.t. 2x+y <= -1
x,y >= 0
```
so the dual is
```
max -z
s.t.
1 - 2z >= 0 dual of x >= 0
-z >= 0 dual of y >= 0
z <= 0 dual variable of constraint
```
when you look for a ray, it becomes
```
max -z
s.t.
-2z >= 0
-z >= 0
z <= 0
```
The value of the variable bounds are the slack of the first and second constraint of the dual hence the checks `yd ≈ -cd` and `xd ≈ -2cd` of this PR (`z` is `cd`).

joaquimg:
Oh, I see, you mean having the interface to compute these

blegat:
@joaquimg Is Xpress ready to pass the test with the changes of this PR now ?

joaquimg:
Nope. It does not returns an infeasibility certificate with the bounds. I also think Gurobi and CPLEX won’t do it.

mlubin:
Ok to close now given #159?

blegat:
I still think this test should be finished.

blegat:
@joaquimg I just checked, [Gurobi](https://github.com/JuliaOpt/MathOptInterfaceGurobi.jl) passes linear12 test and linear8 test with the modification of this PR.

joaquimg:
That is interesting, I am pretty sure I do not set Gurobi's `"FarkasDual"` for bounds. Are we just having some luck?

blegat:
What do you return as dual of the variable bounds ?

joaquimg:
I should be writing ´NaN´. My Gurobi licence expired, I will update it tomorrow.

blegat:
Replaced by https://github.com/JuliaOpt/MathOptInterfaceTests.jl/pull/34

PR #132, opened 2017-09-24
Update abstraction to make room for standalone instances

### LINKED ISSUE #9: an Instance type

mlubin:
Following up on https://github.com/JuliaOpt/MathOptInterface.jl/pull/131, this PR updates the type hierarchy to allow for a standalone instance that's not attached to a solver.
- `AbstractSolverInstance` and `AbstractStandaloneInstance` inherit from `AbstractInstance`.
- Attributes are split into `AbstractSolverInstanceAttribute` and `AbstractInstanceAttribute`. There are no attributes specific to a standalone instance.
- No change to variable and constraint attributes, standalone instances can just implement `cangetattribute` appropriately.

I'm heading in the direction of making a simple reader/writer for MOI instances that we can use in JuMP's unit tests to check if JuMP generates the problem correctly. After this PR we can just call these objects standalone instances, as a first-class MOI concept.

Closes #9

### DIFF
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -262,7 +262,7 @@ struct ListOfVariableReferences <: AbstractInstanceAttribute end
 A `Vector{ConstraintReferences{F,S}}` with references to all constraints of
 type `F`-in`S` in the instance (i.e., of length equal to the value of `NumberOfConstraints{F,S}()`).
 """
-struct ListOfConstraintReferences{F,S} <: AbstractSolverInstanceAttribute end
+struct ListOfConstraintReferences{F,S} <: AbstractInstanceAttribute end
 
 """
     NumberOfConstraints{F,S}()
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -8,23 +8,31 @@ Abstract supertype for attribute objects that can be used to set or get attribut
 abstract type AbstractSolverAttribute end
 
 """
+    AbstractInstanceAttribute
+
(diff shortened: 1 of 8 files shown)


src/attributes.jl
```diff
@@ -218,6 +230,64 @@ A `Bool` indicating if the solver interprets certain quadratic constraints as se
...
+"""
+struct ListOfVariableReferences <: AbstractInstanceAttribute end
+
+"""
+    ListOfConstraintReferences{F,S}()
+
+A `Vector{ConstraintReferences{F,S}}` with references to all constraints of
+type `F`-in`S` in the instance (i.e., of length equal to the value of `NumberOfConstraints{F,S}()`).
+"""
+struct ListOfConstraintReferences{F,S} <: AbstractSolverInstanceAttribute end
```
blegat: Shouldn't this be a `AbstractInstanceAttribute` ? You have added `Solver`

mlubin: Good catch

mlubin:
I'll merge this tomorrow barring objections.

PR #142, opened 2017-10-16
Change MultirowChange to VectorCoefficientChange

odow:
For better consistency.

### DIFF
--- a/docs/src/apireference.md
+++ b/docs/src/apireference.md
@@ -197,7 +197,7 @@ List of function modifications.
 ScalarConstantChange
 VectorConstantChange
 ScalarCoefficientChange
-MultirowChange
+VectorCoefficientChange
 ```
 
 ## Sets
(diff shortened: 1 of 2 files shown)


mlubin:
This name is a bit confusing because the "coefficient" of a vector-valued constraint isn't well defined. If we had a general change that allowed you to modify coefficients for multiple variables in multiple rows, what would it be called?

odow:
What about making it
```julia
struct VectorCoefficientChange{T}
    variables::Vector{VariableReference}
    rows::Vector{Int}
    new_coefficients::Vector{T}
end
```
Solvers can complain if `variables` contains different variable references and they don't support that.

mlubin:
The intention for the single-column change is that it could have a special meaning in the context of column generation, i.e., adding a new variable and then doing a `MultirowChange` could be implemented efficiently. Don't really want to get rid of that.

odow:
Can this be closed?

mlubin:
Yes, if you don't want to bikeshed further.

PR #144, opened 2017-10-16
Proposal: tolerances for constraints

odow:
The other day @mlubin and I discussed the idea that tolerances and other options should be a component of any model instance with floating point arithmetic. (The discussion was based around what features to include in https://github.com/odow/MathOptFormat.jl)

For example, integrality tolerances
```julia
struct ConstraintPrimalTolerance <: AbstractConstraintAttribute end
c = addconstraint!(m, x, Integer())
setattribute!(m, ConstraintPrimalTolerance(), c, 1e-2)
# would permit
#     x = 1.01
# as feasible
```
and primal feasibility
```julia
c = addconstraint!(m, ScalarAffineFunction([x], [1.0], 1.0), LessThan(1.0))
setattribute!(m, ConstraintPrimalTolerance(), c, 1e-2)
# would permit
#     x =0.01
# as feasible: x + 1 <= 1 ... => ... 1.01 <= 1
```

For things like SOS
```julia
c = addconstraint!(m, VectorOfVariables([x,y,z]), SOS1([1.0, 2.0, 3.0]))
setattribute!(m, ConstraintPrimalTolerance(), c, 1e-2)
# would permit
#     (x, y, z) = (1.0, 0.0, 0.01)
# as feasible
```

### DIFF
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -476,7 +476,7 @@ Return the `AbstractSet` object used to define the constraint.
 struct ConstraintSet <: AbstractConstraintAttribute end
 
 """
-    PrimalFeasibilityTolerance()
+    ConstraintPrimalTolerance()
 
 For scalar functions `f ∈ S`, this tolerance corresponds to the amount by which
 primal constraint solution can lie outside the set and still be considered
@@ -491,7 +491,7 @@ feasible.
 For sets `SOS1` and `SOS2`, this corresponds to the maximum value that a zero
 element can take and still be considered feasible.
 """
-struct PrimalFeasibilityTolerance  <: AbstractConstraintAttribute end
+struct ConstraintPrimalTolerance  <: AbstractConstraintAttribute end
 
 ## Termination status
 """
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -475,6 +475,24 @@ Return the `AbstractSet` object used to define the constraint.
 """
 struct ConstraintSet <: AbstractConstraintAttribute end
 
+"""
+    PrimalFeasibilityTolerance()
+
+For scalar functions `f ∈ S`, this tolerance corresponds to the amount by which
+primal constraint solution can lie outside the set and still be considered
+feasible.For vector functions, this is applied row-wise.
+
+#### Special Cases
+
+For sets `ZeroOne`, `Integer`, and `Semiinteger`: specify the amount by which an
+integer variable can differ from an integer value and still be considered
+feasible.
+
+For sets `SOS1` and `SOS2`, this corresponds to the maximum value that a zero
+element can take and still be considered feasible.
+"""
+struct PrimalFeasibilityTolerance  <: AbstractConstraintAttribute end
+
 ## Termination status
 """
     TerminationStatus()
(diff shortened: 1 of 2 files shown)


src/attributes.jl
```diff
@@ -475,6 +475,24 @@ Return the `AbstractSet` object used to define the constraint.
 """
 struct ConstraintSet <: AbstractConstraintAttribute end
 
+"""
+    ConstraintPrimalTolerance()
+
+For scalar functions `f ∈ S`, this tolerance corresponds to the amount by which
```
mlubin: This is not a proper definition. What's the distance metric? Row-wise doesn't make sense for PSD constraints, for example.

odow: Yes it's not a proper definition. Mainly put this PR up so we can discuss. [Gurobi](http://www.gurobi.com/documentation/7.5/refman/feasibilitytol.html) have an explicit parameter but don't specify the distance metric. [CPLEX](https://www-01.ibm.com/support/docview.wss?uid=swg21400045) have parameters for variable bound violations (again, no metric given), and some parameters to control the barrier method, but no explicit parameter for constraint tolerances.

I would expect this means something different for each (Function, Set) combination so maybe that's how it should be approached?

mlubin: Usually it would just depend on the set and not the function. We could define canonical violation metrics for each set. It would be another question to what extent solvers would support the metric.

odow: Something along the lines of 

```
EqualTo: [value-δ, value+δ]
LessThan: (-∞, upper+δ]
GreaterThan: [lower-δ, ∞)
Interval: [lower-δ, upper+δ]

ZeroOne: [0-δ, 0+δ] ∪ [1-δ, 1+δ]
Integers: U_{i∈Z} [i-δ, i+δ]

Reals: x∈ R^n
Zeros: {x ∈ R^n: -δ ≤ x ≤ δ}
Nonpositives: {x ∈ R^n: x ≤ δ}
Nonnegatives: {x ∈ R^n: x ≥ -δ}
Semicontinuous: [0-δ, 0+δ] ∪ [lower-δ, upper+δ]
Semiinteger: [0-δ, 0+δ] ∪ {∪_{i∈lower, lower+1, ..., upper} [i-δ, i+δ]}

SOS1: at most one variable outside the range [-δ, δ]
SOS2: at most two variables outside the range [-δ, δ], and if there are two, 
    they must be adjacent
```

mlubin:
Actually functions need tolerances as well, to limit the gap between `ConstraintPrimal()` and the true value. Then a separate tolerance for how far `ConstraintPrimal()` is from the set. Not sure how much sense this makes algorithmically though.

src/attributes.jl
mlubin: Do you know if this definition for SOS feasibility tolerance is used anywhere?

odow: No. It just seemed like a reasonable thing to do

mlubin:
@odow, I'd like to write a little testing framework for MOI models (for testing both solvers and JuMP's model generation) and we'll need to deal with tolerances there. Could we experiment with constraint-wise tolerances there before baking them into MOI? I feel that it's premature to put this into MOI since the definitions feel a bit arbitrary and we don't know how solvers will behave.

odow:
That seems very reasonable.

IssamT:
Interesting discussion. The suggestion is more generic than mine in #165 with global tolerances. Although I admit I also don't have the answer to the metric question.

odow:
Closing this as stale. We might re-visit this in the future, but not in the near-term.

PR #146, opened 2017-10-18
Add missing MOI prefix in contconic

blegat:


mlubin:
Oops. This is an issue with all tests, right?

mlubin:
That's strange, don't know what I was thinking.

PR #148, opened 2017-10-18
[RFC] Add coefficienttype function

daviddelaat:
This adds a coefficient type function that defaults to `Float64`. (This is used in the pull request https://github.com/JuliaOpt/SemidefiniteOptInterface.jl/pull/3). I hope this is an idiomatic way to add this function. Comments welcome.

### DIFF
--- a/src/MathOptInterface.jl
+++ b/src/MathOptInterface.jl
@@ -113,6 +113,13 @@ should be `true` for a second-order cone solver `s`.
 """
 function supportsproblem end
 
+"""
+    coefficienttype(m::AbstractInstance)
+
+Returns the coefficient type used by the solver instance. Defaults to Float64.
+"""
+function coefficienttype(s::AbstractInstance); Float64; end
+
 include("references.jl")
 include("attributes.jl")
 include("functions.jl")
--- a/src/MathOptInterface.jl
+++ b/src/MathOptInterface.jl
@@ -114,11 +114,19 @@ should be `true` for a second-order cone solver `s`.
 function supportsproblem end
 
 """
+    coefficienttype(m::AbstractSolver)
+
+Returns the coefficient type used by the solver. Defaults to Float64.
+"""
+function coefficienttype(::AbstractSolver); Float64; end
+
+"""
     coefficienttype(m::AbstractInstance)
 
-Returns the coefficient type used by the solver instance. Defaults to Float64.
+Returns the coefficient type used by the instance. Defaults to Float64.
 """
-function coefficienttype(s::AbstractInstance); Float64; end
+function coefficienttype(::AbstractInstance); Float64; end
+
 
 include("references.jl")
 include("attributes.jl")


odow:
Where does this actually get used?

mlubin:
1. Properties like this should be attributes, not new methods.
2. I'm still confused about the use case and why this isn't redundant and/or in direct conflict with `supportsproblem`. If you want to check if a solver/instance supports certain classes of functions (e.g., with specific coefficient types), `supportsproblem` answers that question for you.
3. I could see a settable solver attribute relating to the internal precision used when solving the problem, but this isn't a very general feature so it seems more appropriate to be solver specific.

blegat:
An attribute like `CoefficientType`, `InternalCoefficientType` or `PreferredCoefficientType` would make sense to me.
Suppose you want to solve a problem A. You would write a function like
```julia
function solveA(g::AbstractGraph, solver::MOI.AbstractSolver)
    T = MOI.getattribute(solver, MOI.CoefficientType())
    # Generate MOI functions data using type T and feed them to an instance.
end
```
For instance, if `T` is `BigFloat`, the whole problem data will be generated in `BigFloat` and so if you want a coefficient to be `1/3`, it will be more accurate.
You would like this function to be solver independent so having a solver specific attribute does not help here.

odow:
Can't you do this with supportsproblem though? 

Because what about solvers that support mixed constraint types? Like Float64 and bigfloat constraint matrices but only float64 objectives (for example). 

Better to check exicitly the constraints you want to add. You could loop through different number types

blegat:
Looping is not really an option. The type wanted could be a type that is defined out of base and that is not even defined when `MathOptInterface` is loaded.
If the solver supports being given everything by some type, it gives this type. If there are several ones, it can give its "preferred" one.
If this does not make sense for this solver, then it is free no not define `getattribute` for this attribute;

mlubin:
@blegat, what are the possible values for `T` besides `Float64`, `Rational`, and `BigFloat`? Can't you write some helpers around `supportsproblem` that return the same information?

mlubin:
> You would like this function to be solver independent so having a solver specific attribute does not help here.

> If this does not make sense for this solver, then it is free no not define getattribute for this attribute;

These two statements are a bit contradictory. The solver-independent code won't work if the solver doesn't define this attribute?

mlubin:
What about an attribute that returns a list of *possible* coefficient types that are supported by the solver, which you can then use to check more specifically with `supportsproblem`? Most solvers would return `[Float64]` but others might return `[Float64,BigFloat]` if `BigFloat` is allowed as a coefficient type somewhere.

blegat:
@mlubin A list would be nice, maybe a tuple would be better in term of type stability. Would we give a meaning to the order of the types in the list ? Like the first one is somehow prefered/closer to internal representation.

mlubin:
> A list would be nice, maybe a tuple would be better in term of type stability.

Yep, tuple is okay also.

> Would we give a meaning to the order of the types in the list ? Like the first one is somehow prefered/closer to internal representation.

Solvers can choose to do this, but it's unclear to me if this concept is really well defined. I'd prefer to leave that as an undocumented convention.

daviddelaat:
The main reason for this pull request is that SemidefiniteOptInterface.jl currently only supports Float64. To generalize this so that it also supports other types (in particular high precision floats) we need to paramameterize the type SOItoMOIBridge with a coefficient type. One reason for this is that it has fields like `varmap::Vector{Vector{Tuple{Int, Int, Int, T, T}}}`, where it is good to know T for memory efficiency. Another reason is that uses factors like 1/sqrt(2) without relation to a specific constraint, where we need to know T to compute this in the right precision. See the pull request https://github.com/JuliaOpt/SemidefiniteOptInterface.jl/pull/3 mentioned above.

This means that when constructing a SOItoMOIBridge from an AbstractSolver type, we need to know the type that this abstract solver wants to use. The solvers I typically use (SDP solvers) all support only one type. (The SDPA family supports several types, including BigFloat, but this is a different binary per type.) 

Iterating over coefficient types does not seem to be an option here (as mentioned), but I think an attribute for a tuple of coefficient types would be a good solution indeed.

mlubin:
@daviddelaat, I understand the need to parameterize in the context of SemidefiniteOptInterface.jl. However, you can accomplish this cleanly without any changes to MOI.

From what I can tell from your PR (JuliaOpt/SemidefiniteOptInterface.jl#3), you need the ability to get the coefficient type from an `AbstractSDSolver`, so why not just define ``coefficienttype(::AbstractSDSolver)``? A unique `coefficienttype` concept is just not consistent with the abstractions we've defined at the level of MOI, but you're free to use this idea within SemidefiniteOptInterface.

blegat:
@daviddelaat Why can't we just do `first(MOI.get(solver, MOI.CoefficientType()))` ? For the SDP solvers, we can just return a tuple with one element, e.g. `(BigFloat,)`.

daviddelaat:
@blegat Yes, that's what I intended to do. 

@mlubin You're right, this is only needed for AbstractSDSolver. I thought it would be more clean to implement this at the level of AbstractSolver, but perhaps it is better to do this at the level of AbstractSDSolver. Then we could indeed use a function `coefficienttype(::AbstractSDSolver)` that returns just a single type. What do you thing @blegat?

blegat:
It seems to make more sense to have `CoefficientType` at the MOI level and return a tuple. A SDSolver could also support several types and return a tuple type.

mlubin:
There's no such thing as a solver anymore, so I'll close this PR.

Issue #151, opened 2017-10-19
lookup by name


mlubin:
Now that we have variable and constraint names, it makes sense to be able to look up the references by names. This seems especially relevant when transferring an instance across files/solvers. Strings are a reasonable key to match up the pieces of the model; they're generally simpler to handle than e.g., what [JuMP does](https://github.com/JuliaOpt/JuMP.jl/blob/9466fe9bc15caae3684f980b34d52a545a6c1d12/src/solverinterface.jl#L21) to map references between the solver instance and the standalone instance.

To start the discussion I would propose:
```julia
MOI.has(m::AbstractInstance, VariableReference, "foo") # returns true or false
MOI.lookup(m::AbstractInstance, VariableReference, "foo") # returns a VariableReference if it exists, `KeyError` otherwise

MOI.has(m::AbstractInstance, ConstraintReference{F,S}, "foo") # returns true or false
MOI.lookup(m::AbstractInstance, ConstraintReference{F,S}, "foo") # returns a ConstraintReference{F,S} if it exists, `KeyError` otherwise

# lazy, type unstable interface
MOI.has(m::AbstractInstance, ConstraintReference, "foo") # returns true or false
MOI.lookup(m::AbstractInstance, ConstraintReference, "foo") # returns a ConstraintReference{F,S} if it exists for some (F,S), `KeyError` otherwise
```

I would say that every reference needs to have either a unique name or an empty string for a name. Immediate error when creating a duplicate name.

odow:
To minimize the number of different MOI functions I propose
```julia
MOI.isvalid(m, VariableReference,  "foo")
MOI.isvalid(m, ConstraintReference{F,S}, "foo")
MOI.isvalid(m, ConstraintReference, "foo")
# or
MOI.cangetattribute(m, VariableReference,  "foo")
MOI.cangetattribute(m, ConstraintReference{F,S}, "foo")
MOI.cangetattribute(m, ConstraintReference, "foo")
```
instead of `MOI.has` and
```julia
MOI.getattribute(m, VariableReference,  "foo")
MOI.getattribute(m, ConstraintReference{F,S},  "foo")
MOI.getattribute(m, ConstraintReference,  "foo")
```
 instead of `MOI.lookup`.

> I would say that every reference needs to have either a unique name or an empty string for a name. Immediate error when creating a duplicate name.

+100

mlubin:
[...]

PR #152, opened 2017-10-20
rename getattribute to get and setattribute! to set!

mlubin:
The tests and examples seem to look pretty clean now. This should be fine as along as we never plan on exporting these names.

Will allow us to use `MOI.get` to look up variables and constraints by name (#151). I'll merge this automatically if the following conditions hold:
- 24 hours have passed
- No objections
- At least three thumbs up

### DIFF
--- a/docs/src/apimanual.md
+++ b/docs/src/apimanual.md
@@ -30,7 +30,7 @@ MOI does not export functions, but for brevity we often omit qualifying names wi
 using MathOptInterface
 const MOI = MathOptInterface
 ```
-and prefix all MOI methods with `MOI.` in user code. If a name is also available in base Julia, we always use explicitly use the module prefix, for example, with `MOI.get`.
+and prefix all MOI methods with `MOI.` in user code. If a name is also available in base Julia, we always explicitly use the module prefix, for example, with `MOI.get`.
 
 ## Standard form problem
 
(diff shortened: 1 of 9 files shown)


docs/src/apimanual.md
```diff
@@ -25,6 +25,13 @@ MOI is designed to:
 This manual introduces the concepts needed to understand MOI and give a high-level picture of how all of the pieces fit together. The primary focus is on MOI from the perspective of a user of the interface. At the end of the manual we have a section on [Implementing a solver interface](@ref).
 The [API Reference](@ref) page lists the complete API.
 
+MOI does not export functions, but for brevity we often omit qualifying names with the MOI module. Best practice is to have
+```julia
+using MathOptInterface
+const MOI = MathOptInterface
+```
+and prefix all MOI methods with `MOI.` in user code. If a name is also available in base Julia, we always use explicitly use the module prefix, for example, with `MOI.get`.
```
blegat: There is twice "use"

mlubin:
Waiting for one more thumbs up. @joaquimg @chriscoey @joehuchette

joehuchette:
Looks good to me. The ``MOI.`` prefix is used inconsistently, though, e.g. https://github.com/JuliaOpt/MathOptInterface.jl/pull/152/files#diff-d8be55aa74364fe9aa7535b696fe9920R329

mlubin:
@joehuchette we'll have to decide if we want to use the MOI prefix *everywhere* in the documentation since it feels a bit excessive, but I'd agree that the full examples should be copy-pasteable. There's a discussion about the MOI prefix at the top of the manual.

PR #153, opened 2017-10-22
define lookup by name 

mlubin:
Overload `get` and `canget` to look up a variable or constraint reference by name as proposed in #151.


joaquimg:
Maybe we could me more explicit on the case of vector functions and make it clear that there is only one name for `VectorAffineFunction` in `Zeros`.

mlubin:
@joaquimg, that's a concept that applies throughout MOI. How would you suggest we word it here?

joaquimg:
I will open an issue and then propose a PR. This should not stop this PR.

PR #154, opened 2017-10-28
Add test for variable name and constraint with deletion

blegat:


mlubin:
Could you move this into a separate function, or only run if the instance claims to support deletion?

blegat:
Good idea

PR #157, opened 2017-11-01
Make functions generic with respect to variable type

mlubin:
But keep simple names for the functions using VariableReference

https://github.com/JuliaOpt/MathOptInterfaceUtilities.jl/pull/22#discussion_r147913099

PR #158, opened 2017-11-02
Revert "Make functions generic with respect to variable type"

mlubin:
This reverts commit 971a87de4455a5488df3272a31c234d5aee45112. See https://github.com/JuliaOpt/MathOptInterface.jl/pull/157#issuecomment-341285653.

I think this is a case where we're shooting ourselves in the foot to save a few lines of tedious redefinitions that only need to be written once and are easy to verify, and as a result we're inflicting unnecessary pain on anyone using MOI. I don't like the fact that the `Generic` names will leak through all sorts of error messages. The shortcut definitions don't let you hide the extra cognitive load of the generic types.

### DIFF
--- a/src/functions.jl
+++ b/src/functions.jl
@@ -1,70 +1,26 @@
 # Functions
 
 """
-    AbstractFunction{V}
+    AbstractFunction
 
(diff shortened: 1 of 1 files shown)


joaquimg:
Its good that we tried that, but good error messages are way more important that reusing a few lines...

PR #159, opened 2017-11-02
added a new test of infeas. prob. with 2 constraints

IssamT:
replaces #156

### DIFF
--- a/test/contlinear.jl
+++ b/test/contlinear.jl
@@ -1063,6 +1063,45 @@ function linear11test(solver::MOI.AbstractSolver; atol=Base.rtoldefault(Float64)
     end
 end
 
+function linear12test(solver::MOI.AbstractSolver; atol=Base.rtoldefault(Float64), rtol=Base.rtoldefault(Float64))
+    @testset "test infeasible problem with 2 linear constraints" begin
+        # min x
+        # s.t. 2x-3y <= -7
+        #      y <= 2
+        # x,y >= 0
+        @test MOI.supportsproblem(solver, MOI.ScalarAffineFunction{Float64}, [(MOI.ScalarAffineFunction{Float64},MOI.GreaterThan{Float64}),(MOI.SingleVariable,MOI.GreaterThan{Float64})])
+
+        m = MOI.SolverInstance(solver)
+        x = MOI.addvariable!(m)
+        y = MOI.addvariable!(m)
+        c1 = MOI.addconstraint!(m, MOI.ScalarAffineFunction([x,y], [2.0,-3.0], 0.0), MOI.LessThan(-7.0))
+        c2 = MOI.addconstraint!(m, MOI.ScalarAffineFunction([y], [1.0], 0.0), MOI.LessThan(2.0))
+        bndx = MOI.addconstraint!(m, MOI.SingleVariable(x), MOI.GreaterThan(0.0))
+        bndy = MOI.addconstraint!(m, MOI.SingleVariable(y), MOI.GreaterThan(0.0))
+        MOI.set!(m, MOI.ObjectiveFunction(), MOI.ScalarAffineFunction([x], [1.0], 0.0))
+        MOI.set!(m, MOI.ObjectiveSense(), MOI.MinSense)
+        MOI.optimize!(m)
+
+        @test MOI.canget(m, MOI.ResultCount())
+        if MOI.get(m, MOI.ResultCount()) >= 1
+            # solver returned an infeasibility ray
+            @test MOI.get(m, MOI.TerminationStatus()) == MOI.Success
+            @test MOI.get(m, MOI.DualStatus()) == MOI.InfeasibilityCertificate
+            @test MOI.canget(m, MOI.ConstraintDual(), c1)
+            cd1 = MOI.get(m, MOI.ConstraintDual(), c1)
+            cd2 = MOI.get(m, MOI.ConstraintDual(), c2)
+            bndyd = MOI.get(m, MOI.ConstraintDual(), bndy)
+            @test (cd2-bndyd)/cd1 ≈ 3 atol=atol rtol=rtol
+        else
+            # solver returned nothing
+            @test MOI.get(m, MOI.ResultCount()) == 0
+            @test MOI.canget(m, MOI.PrimalStatus(1)) == false
+            @test MOI.get(m, MOI.TerminationStatus()) == MOI.InfeasibleNoResult ||
+                MOI.get(m, MOI.TerminationStatus()) == MOI.InfeasibleOrUnbounded
+        end
+    end
+end
+
 function contlineartest(solver::MOI.AbstractSolver; atol=Base.rtoldefault(Float64), rtol=Base.rtoldefault(Float64))
     linear1test(solver, atol=atol, rtol=rtol)
     linear2test(solver, atol=atol, rtol=rtol)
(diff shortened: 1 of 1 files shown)


test/contlinear.jl
```diff
@@ -1063,6 +1063,45 @@ function linear11test(solver::MOI.AbstractSolver; atol=Base.rtoldefault(Float64)
...
+        @test MOI.canget(m, MOI.ResultCount())
+        if MOI.get(m, MOI.ResultCount()) >= 1
+            # solver returned an infeasibility ray
+            @test MOI.get(m, MOI.TerminationStatus()) == MOI.Success
+            @test MOI.get(m, MOI.DualStatus()) == MOI.InfeasibilityCertificate
+            @test MOI.canget(m, MOI.ConstraintDual(), c1)
+            cd1 = MOI.get(m, MOI.ConstraintDual(), c1)
+            cd2 = MOI.get(m, MOI.ConstraintDual(), c2)
+            bndyd = MOI.get(m, MOI.ConstraintDual(), bndy)
+            @test (cd2-bndyd)/cd1 ≈ 3 atol=atol rtol=rtol
```
blegat: Could you add tests for the signs of the dual values ?
I would also rather write `- 3 cd1 + cd2 ≈ bndyd` and `2 cd1 ≈ bndxd` to make it clearer why the test should be true by being closer to the dual constraint.

blegat: We could also add a test for the objective `-7a + 2b > 0`

test/contlinear.jl
```diff
@@ -1063,6 +1063,49 @@ function linear11test(solver::MOI.AbstractSolver; atol=Base.rtoldefault(Float64)
...
+            @test MOI.get(m, MOI.TerminationStatus()) == MOI.Success
+            @test MOI.get(m, MOI.DualStatus()) == MOI.InfeasibilityCertificate
+            @test MOI.canget(m, MOI.ConstraintDual(), c1)
+            cd1 = MOI.get(m, MOI.ConstraintDual(), c1)
+            cd2 = MOI.get(m, MOI.ConstraintDual(), c2)
+            bndyd = MOI.get(m, MOI.ConstraintDual(), bndy)
+            @test cd1 < atol
+            @test cd2 < atol
+            @test - 3 * cd1 + cd2 ≈ bndyd atol=atol rtol=rtol
+            @test 2 * cd1 ≈ bndxd atol=atol rtol=rtol
```
blegat: it should be `-bndyd` and `-bndxd`

test/contlinear.jl
```diff
@@ -1063,6 +1063,49 @@ function linear11test(solver::MOI.AbstractSolver; atol=Base.rtoldefault(Float64)
...
+
+        @test MOI.canget(m, MOI.ResultCount())
+        if MOI.get(m, MOI.ResultCount()) >= 1
+            # solver returned an infeasibility ray
+            @test MOI.get(m, MOI.TerminationStatus()) == MOI.Success
+            @test MOI.get(m, MOI.DualStatus()) == MOI.InfeasibilityCertificate
+            @test MOI.canget(m, MOI.ConstraintDual(), c1)
+            cd1 = MOI.get(m, MOI.ConstraintDual(), c1)
+            cd2 = MOI.get(m, MOI.ConstraintDual(), c2)
+            bndyd = MOI.get(m, MOI.ConstraintDual(), bndy)
```
blegat: You should define `bndxd` here

test/contlinear.jl
```diff
@@ -1063,6 +1063,49 @@ function linear11test(solver::MOI.AbstractSolver; atol=Base.rtoldefault(Float64)
...
+            @test MOI.canget(m, MOI.PrimalStatus(1)) == false
+            @test MOI.get(m, MOI.TerminationStatus()) == MOI.InfeasibleNoResult ||
+                MOI.get(m, MOI.TerminationStatus()) == MOI.InfeasibleOrUnbounded
+        end
+    end
+end
+
 function contlineartest(solver::MOI.AbstractSolver; atol=Base.rtoldefault(Float64), rtol=Base.rtoldefault(Float64))
     linear1test(solver, atol=atol, rtol=rtol)
     linear2test(solver, atol=atol, rtol=rtol)
```
blegat: You should add linear12test in this function

test/contlinear.jl
```diff
@@ -1063,6 +1063,50 @@ function linear11test(solver::MOI.AbstractSolver; atol=Base.rtoldefault(Float64)
...
+            @test MOI.canget(m, MOI.ConstraintDual(), c1)
+            cd1 = MOI.get(m, MOI.ConstraintDual(), c1)
+            cd2 = MOI.get(m, MOI.ConstraintDual(), c2)
+            bndxd = MOI.get(m, MOI.ConstraintDual(), bndx)
+            bndyd = MOI.get(m, MOI.ConstraintDual(), bndy)
+            @test cd1 < atol
+            @test cd2 < atol
+            @test - 3 * cd1 + cd2 ≈ -bndyd atol=atol rtol=rtol
+            @test 2 * cd1 ≈ -bndxd atol=atol rtol=rtol
+            @test -7 * cd1 + 2 * cd2 > -atol
```
mlubin: Shouldn't this be `> atol`? The objective needs to be strictly positive.

IssamT: I don't know. For me, having a tolerance towards a solver meaning that we accept to have an error but only within that tolerance.

mlubin: Here we have no actual value to test, we're checking for strict positivity. If `-7 * cd1 + 2 * cd2 == 0` then we do not have a valid infeasibility proof.

IssamT: Here is my reasoning: 

The test would be correct with `0` tolerance if there exist `epsilon > 0` such that `-7 * cd1 + 2 * cd2 = epsilon`. 

Therefore the test would be correct with `absoluteTolerance`  if there exist `epsilon > 0` such that `-7 * cd1 + 2 * cd2 = actualvalue` and `abs(epsilon - actualValue) <= absoluteTolerance`

mlubin: We can't let epsilon get arbitrarily small (smaller than `absoluteTolerance`). The current tests accepts ``cd1, cd2, bndxd, bndyd = 0.0, 0.0 , 0.0 , 0.0`` which is incorrect, and I'd say an important case to catch.

IssamT: I understand your concern but I don't think using `> atol` is the correct solution because it is not consistant with the general definition of tolerance ("accepting incorrect answers as long as they are not too far from the correct answer"). It is better to define precisely what `atol` means for us. Solvers accept many tolerances such as feasibility/optimality tolerance... Since we use only `atol` everywhere I assume it s the largest one. We can enforce that infeasibility/unbounded ray are correct as well with `atol` tolerance, when scaled such as the biggest component in the vector has absolute value 1. In this case we need to first scale the ray using infinite norm, and then test the exact values using `atol`.

blegat: I agree with @mlubin it should be `> atol`

IssamT: Ok, since you both agree on it.

mlubin: I agree that this is a nominally improper use of `atol`. Rescaling the ray and then comparing with a value would also be okay. We are going to have to revisit tolerances in the future anyway: https://github.com/JuliaOpt/MathOptInterface.jl/pull/144.

PR #160, opened 2017-11-03
Add function readinstance

odow:
This PR renames `writeinstance` to `write` and adds two new methods:

`MathOptInstance.read!(instance::AbstractInstance, filename::String)`

and

`MathOptInstance.copy!(dest::AbstractInstance, src::AbstractInstance)`

We use `MathOptInstance.copy!` over `Base.copy!` for consistency with other MathOptInterface methods that have identical names (i.e. `get`) but do not overload the Base version.

For example:
```julia
grb = MOI.SolverInstance(GurobiSolver())
# ... problem creation etc ...
MOI.write(grb, "test.mps")

cpx = MOI.SolverInstance(CPLEXSolver())
MOI.read!(cpx, "test.mps")

xpr = MOI.SolverInstance(XpressSolver())
MOI.copy!(xpr, cpx)
```
<s>Another option is

```julia
"""
    SolverInstance(solver::AbstractSolver)

Create a solver instance from the given solver.

    SolverInstance(solver::AbstractSolver, filename::String)

Create a new solver instance from the file `filename` using the solver `solver`.
Supported file types depend on the solver.
"""
function SolverInstance end
```</s>

### DIFF
--- a/src/MathOptInterface.jl
+++ b/src/MathOptInterface.jl
@@ -70,6 +70,14 @@ Supported file types depend on the solver or standalone instance type.
 function writeinstance end
 
 """
+    readinstance(solver::AbstractSolver, filename::String)
+
+Create a new solver instance from the file `filename` using the solver `solver`.
+Supported file types depend on the solver.
+"""
+function readinstance end
+
+"""
     supportsproblem(s::AbstractSolver, objective_type::F, constraint_types::Vector)::Bool
 
 Return `true` if the solver supports optimizing a problem with objective type `F` and constraints of the types specified by `constraint_types` which is a list of tuples `(F,S)` for `F`-in-`S` constraints. Return false if the solver does not support this problem class.
--- a/src/MathOptInterface.jl
+++ b/src/MathOptInterface.jl
@@ -82,7 +82,7 @@ Supported file types depend on the instance type.
 Read the model from the instance `src` into the instance `m`. If `m` is
 non-empty, this may throw an error.
 """
-function readinstance end
+function readinstance! end
 
 """
     supportsproblem(s::AbstractSolver, objective_type::F, constraint_types::Vector)::Bool
--- a/src/MathOptInterface.jl
+++ b/src/MathOptInterface.jl
@@ -70,10 +70,17 @@ Supported file types depend on the solver or standalone instance type.
 function writeinstance end
 
 """
-    readinstance(solver::AbstractSolver, filename::String)
+    readinstance!(m::AbstractInstance, filename::String)
 
-Create a new solver instance from the file `filename` using the solver `solver`.
-Supported file types depend on the solver.
+Read the file `filename` into the instance `m`. If `m` is non-empty, this may
+throw an error.
+
+Supported file types depend on the instance type.
+
+    readinstance!(m::AbstractInstance, src::AbstractInstance)
+
+Read the model from the instance `src` into the instance `m`. If `m` is
+non-empty, this may throw an error.
 """
 function readinstance end
 
--- a/src/MathOptInterface.jl
+++ b/src/MathOptInterface.jl
@@ -62,27 +62,30 @@ Users must discard the solver instance object after this method is invoked.
 function free! end
 
 """
-    writeinstance(m::AbstractInstance, filename::String)
+    write(m::AbstractInstance, filename::String)
 
 Writes the current instance data to the given file.
 Supported file types depend on the solver or standalone instance type.
 """
-function writeinstance end
+function write end
 
 """
-    readinstance!(m::AbstractInstance, filename::String)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by mlubin** [comment]
This is a bit confusing, why not make this a method of ``AbstractInstance`` that, if successful, replaces whatever is in the current instance with the instance loaded from the file?

I think a more commonly used method would be one that loads an `AbstractSolverInstance` from an `AbstactInstance`, and users can use our standard file readers to create the `AbstractInstance` from a file.

**Comment by odow** [comment]
> This is a bit confusing, why not make this a method of AbstractInstance that, if successful, replaces whatever is in the current instance with the instance loaded from the file?

This seems like a really bad idea? What if solvers don't support deleting constraints and stuff. Isn't it just going to have to create a new model anyway? Loading from a file feels like it's a fresh start.

> I think a more commonly used method would be one that loads an AbstractSolverInstance from an AbstactInstance, and users can use our standard file readers to create the AbstractInstance from a file.

I'm confused with all of these AbstractInstances and AbstractSolverInstances. Are you saying
```julia
# given a file "lpfile.lp"
using LPFiles, MathOptInterface, Gurobi

lp = LPFiles.read("lpfile.lp")::MathOptInterface.AbstractStandaloneInstance

m = MathOptInterface.SolverInstance(GurobiSolver(), lp)
# or
m = MathOptInterface.SolverInstance(GurobiSolver())
MathOptInterface.readinstance!(m, lp)
```
because, while nice, shouldn't we leverage the solvers' built-in readers?

mlubin:
What if `MyFancyStandaloneInstance` has a native file reader? The proposal only allows for solver instances to have native file readers, because you need a solver object to start with. If the instance doesn't support reading a file with a nonempty model, just throw an error.

> Are you saying ...
```julia
# given a file "lpfile.lp"
using LPFiles, MathOptInterface, Gurobi

lp = LPFiles.read("lpfile.lp")::MathOptInterface.AbstractStandaloneInstance

m = MathOptInterface.SolverInstance(GurobiSolver())
MathOptInterface.readinstance!(m, lp)
```

Yes. Both options should be available, but for things like MOF I imagine there won't be native readers.

mlubin:
For example, I would replace
https://github.com/odow/MathOptFormat.jl/blob/9856bf267e47fbad70fcbdf6540ded76f29d9b88/src/reader.jl#L39
with
```julia
instance = MOI.SolverInstance(solver)
MOI.readinstance!(instance, MOFInstance(filename))
```

src/MathOptInterface.jl
```diff
@@ -70,6 +70,21 @@ Supported file types depend on the solver or standalone instance type.
...
+throw an error.
+
+Supported file types depend on the instance type.
+
+    readinstance!(m::AbstractInstance, src::AbstractInstance)
+
+Read the model from the instance `src` into the instance `m`. If `m` is
+non-empty, this may throw an error.
+"""
+function readinstance end
```
mlubin: Missing `!`

src/MathOptInterface.jl
```diff
@@ -70,6 +70,21 @@ Supported file types depend on the solver or standalone instance type.
...
 
 """
+    readinstance!(m::AbstractInstance, filename::String)
+
+Read the file `filename` into the instance `m`. If `m` is non-empty, this may
+throw an error.
+
+Supported file types depend on the instance type.
+
+    readinstance!(m::AbstractInstance, src::AbstractInstance)
```
mlubin: Maybe we can use `Base.copy!` for this version? It's exactly a destructive copy.

odow: I'm not a fan of using the Base version if everything else in MOI is unexported (i.e. get).
What about
```julia
MathOptInterface.write(m::AbstractInstance, filename::String)
MathOptInterface.read!(m::AbstractInstance, filename::String)
MathOptInterface.copy!(m::AbstractInstance, src::AbstractInstance)
```

mlubin: Someone's going to ask why we're not consistent with Base, but yes, that's okay with me.

odow: Imagine if this worked...
```julia
mof = MOFInstance("test.mof.json")

grb = MOI.SolverInstance(GurobiSolver())
MOI.copy!(grb, mof)
MOI.write(grb, "test.mps")

cpx = MOI.SolverInstance(CPLEXSolver())
MOI.read!(cpx, "test.mps")

mof2 = MOFInstance()
MOI.copy!(mof2, cpx)

mof2 == mof
```

mlubin: Replace `==` with my `test_instances_equal` (github.com/JuliaOpt/MathOptInterfaceUtilities.jl/pull/22/files#diff-d96547c1b2a274f333396d4d54349d4aR263) and it should work.

Issue #164, opened 2017-11-06
A "solver" for standalone instances


odow:
As raised [here](https://github.com/odow/MathOptFormat.jl/pull/14#issuecomment-342039829), what is the parallel of a solver for a standalone instance?

Maybe `solver` is a misnomer and they should be called
 - InstanceCreator
 - Instantiator
 - InstanceGenerator
 - InstanceController

```julia
"""
    AbtractInstantiator

Abstract supertype for "solver" objects.
An instantiator is a lightweight object used for selecting solvers and parameters.
It does not store any solver instance data.
"""
abstract type AbtractInstantiator end

"""
    StandaloneInstance(creator::AbtractInstantiator)::AbstractStandaloneInstance

Create a standalone instance from the given instantiator.
"""
function StandaloneInstance end

"""
    SolverInstance(creator::AbtractInstantiator)::AbstractSolverInstance 

Create a solver instance from the given instantiator.
"""
function SolverInstance end
```

Then you could have something like
```julia
using Gurobi, MathOptInterface, LPFiles

grb = MOI.SolverInstance(GurobiSolver())
MOI.StandaloneInstance(GurobiSolver()) # errors

MOI.SolverInstance(LPFile()) # errors
lp = MOI.StandaloneInstance(LPFile())
```

mlubin:
It's good to discuss this, since the solver concept is something we imported from MPB without much thought. The point of the solver is to hold the arguments for the constructor of an instance in a generic way. Maybe we don't need this anymore?

Here's what it could look like without solver objects:
```julia
m = Model()
.....
JuMP.solve(m) # error: no solver instance attached
JuMP.attach(m, GurobiInstance(TimeLimit=..., env=...))
JuMP.solve(m)
```
This clarifies what JuMP will do internally anyway, so I'm in favor.

odow:
Maybe `SolverInstance` should be called a `WrapperInstance` as it wraps an underlying model without storing the model.

```julia
m = Model() # creates a standalone instance
@variable(m, x >= 0)
@objective(m, Min, x)
c = @constraint(m, 2x >= 2)
JuMP.attach(m, CbcModel())
JuMP.solve(m)
# x = 1
JuMP.setcoefficient!(m, c, x, 4.0) # error: solver is currently attached.
JuMP.detach(m)
JuMP.setcoefficient!(m, c, x, 4.0) # modifies coefficent in standalone instance
JuMP.attach(m, CbcModel())
[...]

PR #165, opened 2017-11-10
WIP: restructure tests

### LINKED ISSUE #21: better mechanism for choosing which tests to run

mlubin:
Solvers choose tolerances and which tests to run with a `TestConfig` object and an exclude list, e.g.,
```julia
contlineartest(GurobiSolver(), TestConfig(atol=...,rtol=...), exclude = ["linear9"])
```
The current definition of `TestConfig` is:
```julia
struct TestConfig
atol::Float64 # absolute tolerance for ...
rtol::Float64 # relative tolerance for ...
query::Bool # can get objective function, and constraint functions, and constraint sets
duals::Bool # test dual solutions
infeas_certificates::Bool # check for infeasibility certificates when appropriate
end
```
which will easily grow. Probably makes sense to provide keyword constructor so it's easy to add things in the future.

This change removes the dependence on solver attributes to prepare for removing solver objects as discussed in https://github.com/JuliaOpt/MathOptInterface.jl/issues/164.

Only done for contlinear test so far, will proceed if this seems reasonable. I also plan on addressing https://github.com/JuliaOpt/MathOptInterface.jl/pull/115 by removing the dependency on MOIU.

@IssamT @odow @blegat 

Closes #21

### DIFF
--- a/test/config.jl
+++ b/test/config.jl
@@ -1,0 +1,7 @@
+struct TestConfig
+atol::Float64 # absolute tolerance for ...
+rtol::Float64 # relative tolerance for ...
+query::Bool # can get objective function, and constraint functions, and constraint sets
+duals::Bool # test dual solutions
+infeas_certificates::Bool # check for infeasibility certificates when appropriate
+end
--- a/test/contlinear.jl
+++ b/test/contlinear.jl
@@ -5,1119 +5,1145 @@ using MathOptInterfaceUtilities # Defines isapprox for ScalarAffineFunction
 
 # Continuous linear problems
 
-function linear1test(solver::MOI.AbstractSolver; atol=Base.rtoldefault(Float64), rtol=Base.rtoldefault(Float64))
-    @testset "Basic solve, query, resolve" begin
-        # simple 2 variable, 1 constraint problem
-        # min -x
-        # st   x + y <= 1   (x + y - 1 ∈ Nonpositives)
-        #       x, y >= 0   (x, y ∈ Nonnegatives)
+# Basic solver, query, resolve
+function linear1test(solver::MOI.AbstractSolver, config::TestConfig)
+    atol = config.atol
+    rtol = config.rtol
+    # simple 2 variable, 1 constraint problem
+    # min -x
+    # st   x + y <= 1   (x + y - 1 ∈ Nonpositives)
+    #       x, y >= 0   (x, y ∈ Nonnegatives)
 
-        @test MOI.supportsproblem(solver, MOI.ScalarAffineFunction{Float64}, [(MOI.ScalarAffineFunction{Float64},MOI.LessThan{Float64}),(MOI.SingleVariable,MOI.GreaterThan{Float64})])
+    @test MOI.supportsproblem(solver, MOI.ScalarAffineFunction{Float64}, [(MOI.ScalarAffineFunction{Float64},MOI.LessThan{Float64}),(MOI.SingleVariable,MOI.GreaterThan{Float64})])
 
-        @test MOI.get(solver, MOI.SupportsAddConstraintAfterSolve())
-        @test MOI.get(solver, MOI.SupportsAddVariableAfterSolve())
-        @test MOI.get(solver, MOI.SupportsDeleteConstraint())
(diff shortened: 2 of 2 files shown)


IssamT:
Good idea, but can we include in the structure more than just one `atol` and `rtol`? Solvers may still use the same value.

Cplex for instance defines

- feasibility tolerance
- optimality tolerance
- integrity tolerance

The definitions given by Cplex are abstract enough to be solver independent.

blegat:
@IssamT You can give the maximum of these tolerance as `atol` and `rtol`. The tests are not meant to provide an in dept testing of the solver but rather a test of the solver wrapper. If the wrapper does something wrong, you will notice it even if you use a tolerance of 1e-3

@mlubin What would be the replacement of the solvers in the tests ? A function returning an instance ?

mlubin:
@IssamT, adding more tolerances is reasonable but out of scope for this PR (https://github.com/JuliaOpt/MathOptInterface.jl/pull/144#issuecomment-337448990). I would want to restructure the tests to use these tolerances at the same time, not just add them to the config structure.

@blegat, yes. When we remove solvers (in a separate PR), then I think having a function that returns a new instance is reasonable here.

IssamT:
@blegat, I am fine with your suggestion, as long as we agree that it is exactly the maximum of those.

Assuming the bound of a variable x is 3

testing `x <= 3 + atol` is correct.
but testing `100000*x <= 300000*3 +atol` is not correct.

having `atol` as the maximum of some defined tolerances is fine, but one shouldn't add `atol` in any tested line where he believes the theoretical result is correct without `atol`.

mlubin:
@IssamT, I'd invite you to follow up on this PR with one that clarifies and precisely defines the tolerances we use.

joaquimg:
About the keyword constructor.
There is this nice macro: `Base.@kwdef` that creates a constructor with default values. Then one can simply set fields.

mlubin:
The solver tests should really be in `src/` instead of `test/` because they are used by other packages. They're not tests for MOI itself. That leaves a few options:
1. ~Submodule `MathOptInterface.Test` includes tests and needed definitions (`isapprox`) from MOIU.~ Not feasible because submodules can't refer to outer modules?
2. Tests and needed definitions move directly into MOI. This seems a bit messy.
3. Solver tests move to MOIU.
3. Solver tests move into separate package (MOITests?).

I'd vote for option 3. Thoughts?

joaquimg:
3 looks good, but a solver might not need MOIU at all, in that case 4 seems nicer.
I agree that 2 looks messy.

mlubin:
MOIU could be a test dependency if the solver doesn't need it otherwise.

joaquimg:
Yep.
Do you see the tests growing a lot (in size)? Maybe using a few external files to hold data? In this case they would make MOIU large with no need.

blegat:
4 seems best. 3 night make MOIU grow too much

mlubin:
I'm ok with 4.

mlubin:
I'm traveling for the next week. If someone would like to create MathOptInterfaceTests and move over my PR to there, would speed things up. After this and #164 go through, I'll be able to get back to working on JuMP.

joaquimg:
Initialized: https://github.com/JuliaOpt/MathOptInterfaceTests.jl

This modification will be very useful for testing GLPK MIP solver, which is almost ready (https://github.com/joaquimg/MathOptInterfaceGLPK.jl)

mlubin:
Replaced by MOIT

PR #166, opened 2017-11-11
Change packed storage format for PSD cone

blegat:
Comparisons between the storage formats can be found [here](http://www.netlib.org/lapack/lug/node123.html) or [here](https://software.intel.com/en-us/mkl-developer-reference-c-matrix-storage-schemes-for-lapack-routines).
The mapping is a lot easier with the `'U'` storage format and it does not depend on the dimension of the matrices. For this reason, I think we should pick this one.

### DIFF
--- a/src/sets.jl
+++ b/src/sets.jl
@@ -161,8 +161,8 @@ dimension(s::Union{ExponentialCone, DualExponentialCone, PowerCone, DualPowerCon
     PositiveSemidefiniteConeTriangle(dimension)
 
 The (vectorized) cone of symmetric positive semidefinite matrices, with off-diagonals unscaled.
-The entries of the upper triangular part of the matrix are given row by row (or equivalently, the entries of the lower triangular part are given column by column).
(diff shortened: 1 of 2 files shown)


mlubin:
I don't have a strong preference for one or the other. What do most solvers do?
@ulfworsoe

ulfworsoe:
Internally, mosek stores PSD entries in 'L' format, i.e. the lower triangular part in column-major format. As @blegat says, it is slightly nicer to convert to linear index for the `'U'` format.

blegat:
@ulfworsoe Do you think it might induce an overhead to translate from `'U'` format to Mosek's `'L'` format ?

ulfworsoe:
@mlubin No overhead I would deem relevant. I use `U` format internally in Fusion.

mlubin:
CC @joehuchette @daviddelaat

blegat:
Note that for the Julia function `Symmetric`, the default is `uplo=:U` and the matrix layout in Julia is column major (the elements are stored column by column). So the change of this PR seems consistent with Julia Base :)

mlubin:
Ok ok :)

PR #169, opened 2017-11-18
Create AbstractSolver[Constraint|Variable] attribute

blegat:
This eases https://github.com/JuliaOpt/MathOptInterfaceUtilities.jl/pull/31

### DIFF
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -30,13 +30,27 @@ Abstract supertype for attribute objects that can be used to set or get attribut
 abstract type AbstractVariableAttribute end
 
 """
+    AbstractSolverVariableAttribute
+
+Abstract supertype for attribute objects that can be used to set or get attributes (properties) of variables in the solver instance.
+"""
+abstract type AbstractSolverVariableAttribute <: AbstractVariableAttribute end
+
+"""
     AbstractConstraintAttribute
 
 Abstract supertype for attribute objects that can be used to set or get attributes (properties) of constraints in the instance.
 """
 abstract type AbstractConstraintAttribute end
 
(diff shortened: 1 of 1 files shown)


mlubin:
What is a solver variable and why can't these attributes apply to standalone instances? Starting point, for example could be considered part of the specification of the instance for nonconvex problems.

blegat:
I was hesitant for starting point. We can move it

mlubin:
The bridges should also transform the starting point, however. It's still not too clear why the distinction between solver variable attribute and variable attribute is needed. Why not list the attributes that the bridge intercepts and just pass the other ones through?

blegat:
It is not whether the attribute is intercepted or not. For a constraint of
the set type Interval, we know it is bridge. However, for
ConstraintFunction, this is not specific to the solver hence we can just
ask the MOIU instance.

An AbstractSolverConstraintAttribute is an attribute that is only available
once optimize! has been called hence it is not implemented for MOIU
instances or files of the MOI format

Le dim. 19 nov. 2017 à 4:05, Miles Lubin <notifications@github.com> a
écrit :

blegat:
We could also call it AbstractResultConstraintAttribute or
AbstractConstraintResultAttribute

Le dim. 19 nov. 2017 à 12:05, Benoît Legat <benoit.legat@gmail.com> a
écrit :

mlubin:
> An AbstractSolverConstraintAttribute is an attribute that is only available
once optimize! has been called hence it is not implemented for MOIU
instances or files of the MOI format

Shouldn't we allow a result to exist without being attached to a particular solver? You could also store a result (or multiple results) in a file together with the MOI instance. It seems a bit arbitrary to say that results can only come from solvers.

I'd be more comfortable with a separate list of attributes (i.e., an explicit `Union` type) instead of changing the type hierarchy.

blegat:
I will simply use an `Union` in MOIT.

PR #171, opened 2017-11-18
Remove tests; relocated to MOIT

mlubin:
Closes https://github.com/JuliaOpt/MathOptInterface.jl/issues/21.
We should also move over all of the testing-related issues to MOIT.

@blegat @joaquimg, all of the tests have been copied over, right?


mlubin:
I've disabled AppVeyor for MOI since there's nothing to tests and the failing status is unhelpful.

odow:
Moving the issues but I don't have access rights to close the MOI ones.

joaquimg:
Yes, I copied everything.

PR #172, opened 2017-11-19
Fix AnyAttribute

blegat:
I guess this was a typo

### DIFF
--- a/docs/src/apireference.md
+++ b/docs/src/apireference.md
@@ -12,7 +12,6 @@ List of attribute categories.
 
 ```@docs
 AbstractInstanceAttribute
-AbstractSolverInstanceAttribute
 AbstractVariableAttribute
 AbstractConstraintAttribute
 ```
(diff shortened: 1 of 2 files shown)


mlubin:
We do have an `AbstractSolverInstanceAttribute`, but maybe we should remove this given the recent discussion in #169.

blegat:
Yes, we should discuss this on an issue. The reason of this PR is that since `AbstractSolverInstanceAttribute <: AbstractInstanceAttribute`, using only `AbstractSolverInstanceAttribute` in `AnyAttribute` does not include the attributes that are `AbstractInstanceAttribute` but not `AbstractSolverInstanceAttribute`.
This causes `AnyAttribute` to contain not all attributes.

mlubin:
I'd propose to remove `AbstractSolverInstanceAttribute` completely. And of course replace it with `AbstractInstanceAttribute` in `AnyAttribute`.

blegat:
I am ok with this, done in 407ce2e.

mlubin:
Looks good. Needs a rebase and doc updates.

PR #175, opened 2017-11-21
Add PSDSquare + LogDet + RootDet + GeometricMean cones

blegat:
I have also defined `dimension` for PSD cones since `s.dimension` is not the dimension of the cone.

### DIFF
--- a/docs/src/apireference.md
+++ b/docs/src/apireference.md
@@ -207,6 +207,13 @@ PowerCone
 DualPowerCone
 PositiveSemidefiniteConeTriangle
 PositiveSemidefiniteConeScaled
+PositiveSemidefiniteConeSquare
+LogDetConeTriangle
+LogDetConeScaled
+LogDetConeSquare
+RootDetConeTriangle
+RootDetConeScaled
+RootDetConeSquare
 Integer
 ZeroOne
 Semicontinuous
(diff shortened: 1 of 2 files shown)


mlubin:
I was going to suggest to get rid of the Scaled versions of the cones completely since it doesn't look like we're putting any effort into supporting or testing them.

mlubin:
CC @chriscoey

chriscoey:
we are heading in the direction of replacing Convex.jl it seems (which may be a good thing). why not also have lambda_min and lambda_max cones, and various other useful function cones which could, like the above, be reformulated with PSD cones plus possibly other cones (linear, exponential, geomean etc)?

blegat:
I have added the geomean cone :)
It seems SCS is the only one using the scaled representation natively. In fact [SDOI](https://github.com/JuliaOpt/SemidefiniteOptInterface.jl) supports it, MOIMosek has already some code to support it and there are more tests for the scaled cone than for the unscaled one (sdp2 is scaled only). I am not against removing it although it does not hurt and I can move the bridge Scaled -> triangle from SDOI to MOIU to make it easy for Mosek to support it without needed to complicate its code. However, it would simplify things to remove it and the little gain in SCS is maybe not worth the complication so I am undecided.

blegat:
Its faith is going to be decided in #176 (looks like Gladiator :D)

blegat:
I'll merge this shortly if there is no objection and then I will proceed with removing Scaled and clarifying dual for Triangle.

PR #177, opened 2017-11-22
Remove scaled and add duality note for triangle

### LINKED ISSUE #176: Scaled symmetric matrices
As explained [here](https://github.com/cvxgrp/scs), scaled symmetric matrix format has the advantage that the scalar product of the vector correspond to the scalar product of the matrices. However in modelling, we always prefer to talk about the upper triangular part of the matrix without scaling (otherwise there is complicates cases to check with the dual and so on).
As we are created more cones storing a matrix, for consistency we need to create one for scaled symmetric matrix format each time.
Is it worth keeping this PositiveSemidefiniteConeScaled or should we get rid of it ?

* Keep it :+1: 
* Get rid of it :-1: 

cc @ulfworsoe

blegat:
Closes #176

### DIFF
--- a/docs/src/apimanual.md
+++ b/docs/src/apimanual.md
@@ -201,9 +201,9 @@ Constraints with `SingleVariable` in `LessThan`, `GreaterThan`, `EqualTo`, or `I
 | ``2yz \ge \lVert x \rVert_2^2, y,z \ge 0``                    | `VectorOfVariables`          | `RotatedSecondOrderCone`           |
 | ``(a_1^Tx + b_1,a_2^Tx + b_2,a_3^Tx + b_3) \in \mathcal{E}``  | `VectorAffineFunction`       | `ExponentialCone`                  |
 | ``A(x) \in \mathcal{S}_+``                                    | `VectorAffineFunction`       | `PositiveSemidefiniteConeTriangle` |
-| ``A(x) \in \mathcal{S}'_+``                                   | `VectorAffineFunction`       | `PositiveSemidefiniteConeScaled`   |
+| ``A(x) \in \mathcal{S}'_+``                                   | `VectorAffineFunction`       | `PositiveSemidefiniteConeSquare`   |
 | ``x \in \mathcal{S}_+``                                       | `VectorOfVariables`          | `PositiveSemidefiniteConeTriangle` |
-| ``x \in \mathcal{S}'_+``                                      | `VectorOfVariables`          | `PositiveSemidefiniteConeScaled`   |
+| ``x \in \mathcal{S}'_+``                                      | `VectorOfVariables`          | `PositiveSemidefiniteConeSquare`   |
 
 
 [Define ``\mathcal{E}`` (exponential cone), ``\mathcal{S}_+`` (smat), ``\mathcal{S}'_+`` (svec). ``A(x)`` is an affine function of ``x`` that outputs a matrix.]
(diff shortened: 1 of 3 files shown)


mlubin:
The duality discussion needs to be worked into http://www.juliaopt.org/MathOptInterface.jl/latest/apimanual.html#Duals-1 as well.

docs/src/apimanual.md
```diff
@@ -446,6 +446,12 @@ An important note for the LP case is that the signs of the feasible duals depend
 
 Currently, a convention for duals is not defined for problems with non-conic sets ``\mathcal{S}_i`` or quadratic functions ``f_0, f_i``.
 
+#### Duality and scalar product
+
+The nonzero entries of the matrix ``A_i`` of the section [Duals](@ref) may be different from the `coefficients` field in the `VectorAffineFunction`.
```
mlubin: "may be" -> "are"

docs/src/apimanual.md
```diff
@@ -446,6 +446,12 @@ An important note for the LP case is that the signs of the feasible duals depend
...
+The row of ``A_i`` should the vector such that its scalar product with the vector ``x`` is equal to the function defined in the `VectorAffineFunction`.
```
mlubin: "should" -> "should be"?

docs/src/apimanual.md
mlubin: Never mind, I see. You should start this subsection by listing sets which this discussion applies to.

src/sets.jl
```diff
@@ -196,36 +195,54 @@ The advantage of the first format is the mapping between the `(i, j)` matrix ind
...
-The (vectorized) cone of symmetric positive semidefinite matrices, with off-diagonals scaled.
-The entries of the upper triangular part of the matrix are given column by column (or equivalently, the entries of the lower triangular part are given row by row).
-An ``n \\times n`` matrix has ``n(n+1)/2`` lower-triangular elements, so for the vectorized cone of dimension ``d``, the corresponding symmetric matrix has side dimension ``\\sqrt{1/4 + 2 d} - 1/2`` elements.
-The off-diagonal entries of the matrices of both the cone and its dual are scaled by ``\\sqrt{2}`` and the scalar product is simply the sum of the pairwise product of the entries.
-
-### Examples
-
-The matrix
+### Duality note
+The scalar product for the symmetric matrix in its vectorized form is the sum of the pairwise product of the diagonal entries plus twice the sum of the pairwise product of the upper diagonal entries.
```
mlubin: I would add a reference to slide 6 of http://docs.mosek.com/slides/ismp2012/sdo.pdf, or even better to a relevant paper.

docs/src/apimanual.md
```diff
@@ -446,6 +446,12 @@ An important note for the LP case is that the signs of the feasible duals depend
 
 Currently, a convention for duals is not defined for problems with non-conic sets ``\mathcal{S}_i`` or quadratic functions ``f_0, f_i``.
 
+#### Duality and scalar product
+
+The scalar product is different from the canonical one for the sets [PositiveSemidefiniteConeTriangle](@ref), [LogDetConeTriangle](@ref), [RootDetConeTriangle](@ref).
+For these cones, the nonzero entries of the matrix ``A_i`` of the section [Duals](@ref) may be different from the `coefficients` field in the `VectorAffineFunction`.
```
mlubin: "may be different" is ambiguous. Say exactly which are different and which are not.

blegat:
I have updated the explanation, it should be clearer now. I have also update the docstring for PSDCTriangle so that it matches the primal form used in MOI docs.

mlubin:
CC @chriscoey for review

chriscoey:
:+1:  looks good to me

PR #178, opened 2017-11-30
Add instance id to references. Closes #163

### LINKED ISSUE #163: Reference ownership

```julia
src = SolverInstance(solver)
x = addvariable!(src)

copy!(dest, src)

# what is the result of
isvalid(dest, x)
```

odow:
As proposed in https://github.com/JuliaOpt/MathOptInterface.jl/issues/163#issuecomment-348080360

Probably needs some more discussion / thinking about how this would work.

### DIFF
--- a/src/MathOptInterface.jl
+++ b/src/MathOptInterface.jl
@@ -67,6 +67,9 @@ function read! end
 
 Copy the model from the instance `src` into the instance `dest`. If `dest` is
 non-empty, this may throw an error.
+
+All data is copied rather than passed by reference, and all valid variable and
+constraint references in `src` must become valid instances in `dest`.
 """
 function copy! end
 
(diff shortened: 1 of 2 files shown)


mlubin:
I think we should treat the `instance` field as a hash. Equal `instance` values do no imply the references belong to the same instance. Mismatched `instance` values indicate a user error. So I would remove the change to the description of `copy`, and we consider this PR separate from that issue.

blegat:
Is it really worth it to double the size of a variable reference just for error handling ? It is a low level interface, we should emphasize lightness more than user friendliness.

mlubin:
Let's keep the discussion at https://github.com/JuliaOpt/MathOptInterface.jl/issues/163.

PR #179, opened 2017-12-02
copy! returns a reference map.

### LINKED ISSUE #163: Reference ownership

mlubin:
Discussion on validity of references across instances.

Closes #163 
Closes #178


src/MathOptInterface.jl
```diff
@@ -65,8 +65,23 @@ function read! end
...
+# Given `AbstractInstance`s `src` and `dest`.
+# `dest` must be empty.
+
+x = addvariable!(src)
+isvalid(src, x)   # true
+isvalid(dest, x)  # false (`dest` has no variables)
+
+reference_map = copy!(dest, src)
+
+isvalid(dest, x) # false (unless x == reference_map[x])
```
IssamT: If I am not wrong, `isvalid(dest, x)` can also be `true` if `x == reference_map[y]` for some `y` in `src`. if you add the assumption that `src` is also initially empty, then the "`(unless x == reference_map[x])`" will be correct but not exhaustive.

mlubin: That's correct, good catch.

mlubin:
The names `ConstraintReference` and `VariableReference` don't really match the semantics of the implementation. If we called them `ConstraintKey` or `ConstraintIndex` then it would be clearer that it's possible that the object could be used to "look up" constraints in other instances.

I'd vote for `ConstraintKey` as `ConstraintIndex` would tend to imply that the `value` field would contain linearly increasing integers.

mlubin:
Now I'm leaning towards `ConstraintIndex`. Wikipedia defines it as "a key in an associative array".

src/MathOptInterface.jl
```diff
@@ -65,8 +65,23 @@ function read! end
 """
     copy!(dest::AbstractInstance, src::AbstractInstance)
 
-Copy the model from the instance `src` into the instance `dest`. If `dest` is
-non-empty, this may throw an error.
+Copy the model from the instance `src` into the instance `dest`. The target instance `dest` must be empty. Returns a dictionary-like object that translates variable and constraint references from the `src` instance to the corresponding references in the `dest` instance.
```
blegat: Why don't we say "is emptied" ?

docs/src/apimanual.md
```diff
@@ -481,3 +481,5 @@ All data passed to the solver should be copied immediately to internal data stru
...
 
 MOI solver interfaces may be in the same package as the solver itself (either the C wrapper if the solver is accessible through C, or the Julia code if the solver is written in Julia, for example). In some cases it may be more appropriate to host the MOI wrapper in its own package; in this case it is recommended that the MOI wrapper package be named `MathOptInterfaceXXX` where `XXX` is the solver name.
+
+Recall the discussion on references in [`isvalid`](@ref MathOptInterface.isvalid). A simple way assign integer values to references to avoid unintentional conflicts with other instances is to use `xor`, an invertible operation, with an arbitrary integer value like `object_id` as the key. `xor` can be used again to covert back to the internal numbering scheme. This is not required but is a small trick that will most likely quickly catch a dangerous class of bugs.
```
blegat: covert -> convert ?

IssamT:
In general I would just use `MOI.Variable`/`MOI.Constraint` and let the documentation clarifies what they really are. But if we follow the coding style of `MOI` they should be clear names for the developers.

How about `VariableHandle`/`ConstraintHandle` ?

Well, if we want be clear enough to avoid the confusion of #163 we should use something like `InstanceVariableHandle`/`InstanceConstraintHandle` but that would be too much. no?

mlubin:
"Handle" is about the same as "Reference". They both imply that the object is something like a pointer to the original variable or constraint, which is not how we've implemented it. `VariableIndex` and `ConstraintIndex`, on the other hand, imply that these are indices into a collection of variables and constraints, and that it's possible for a single index to map to different variables if you use them as keys in different collections (different instances). That seems like what we're going for.

IssamT:
> that it's possible for a single index to map to different variables if you use them as keys in different collections (different instances)

Then I think I misunderstood. I thought this would be possible because of the implementation but isn't part of the specification and shouldn't be done. So you want to be able to use the same `VariableReference` to access variables in different instances. Right?

mlubin:
>  So you want to be able to use the same VariableReference to access variables in different instances. Right?

Definitely not. But I want users to know that they should think of the `VariableIndex` objects as integers and not be surprised if, by accident, they refer to other variables in a different instance.

IssamT:
In that case I still prefer `Handle` since it s abstract and more flexible, while talking about `Index` is more implementation dependent, especially when there is no corresponding array. The first thing a new developer may do when reading VariableIndex is to look for the array ...

From https://en.wikipedia.org/wiki/Handle_(computing)

"A resource handle can be an opaque identifier, in which case it is often an integer number (often an array index in an array or "table" that is used to manage that type of resource)"

IssamT:
Yet, I agree that only switching Reference for Handle wouldn't help much when it comes to misusing a VariableReference outside of its instance.

IssamT:
How about having the handle explained only in the doc and showing explicitly the link to the instance in the name ?

`InstanceVariable/InstanceConstraint`

mlubin:
I'm not sure opaqueness is really helpful here. Why don't we just say exactly what it is?

```julia
"""
    VariableIndex

A type-safe wrapper for `Int64` for use in referencing variables in an instance.
To allow for deletion, indices need not be consecutive.
"""
struct VariableIndex
    value::Int64
end
```

IssamT:
You are right. Anyway it is something easy to understand for new developers as soon as they read it its 2 line definition.

mlubin:
I'll put together a separate PR for the renaming.

blegat:
It is correct that references is confusing. Having `x == y` would mean that they are the same with that name. With index, it would just mean they represent the same index but probably corresponding to different collections. I am voting for the `Index` suffix.

odow:
I'm not sure I like `copy!` returning a dictionary. Is that going to be type stable with `ConstraintIndex`? Why not just force the `dest` to support the same Indexes:
```julia
src = SolverInstance(solver)
x = addvariable!(src)

copy!(dest, src)

# force 
isvalid(dest, x) == true
```
Either `dest` is able to add VariableIndexes with the right `value`, or it stores the mapping internally.

mlubin:
@odow, https://github.com/JuliaOpt/MathOptInterface.jl/issues/163#issuecomment-348488637

> The return type doesn't need to be a `Dict`. It can be a dict-like object that supports variable and constraint references efficiently.


> Why not just force the dest to support the same Indexes

I can imagine that causing pain for solver wrappers. They'll need to internally remap the indices and decide if they're old indices from a copied model or newly added indices.

odow:
Okay well the follow up question is really, who implements `copy!`? Is it `src`? Or `dest`? Or `MOIU`?

mlubin:
`dest`, because some instances may choose not to implement `addconstraint!`, so `copy!` would be the only way to load in the data.

mlubin:
New PR at #181

PR #180, opened 2017-12-03
rename *Reference to *Index

mlubin:
As discussed in https://github.com/JuliaOpt/MathOptInterface.jl/pull/179

@IssamT @blegat @odow


odow:
I go offline for one weekend and return to find the world a different place...

mlubin:
Note the change from `UInt64` to `Int64`. We don't commonly use unsigned integers for indices, so I don't see a need to use them here.

PR #181, opened 2017-12-03
copy! returns an index map

### LINKED ISSUE #163: Reference ownership

mlubin:
Closes #179
Closes #178
Closes #163


mlubin:
Related question, should we have a `cancopy` to check if the `dest` instance can support the copy given the types of constraints/objective in the `src`?

PR #187, opened 2017-12-11
Add isempty and empty!

blegat:
* `isempty` is needed in https://github.com/JuliaOpt/MathOptInterfaceUtilities.jl/issues/32#issuecomment-350852899
* `empty!` is needed in https://github.com/JuliaOpt/MathOptInterfaceUtilities.jl/pull/47

### DIFF
--- a/src/MathOptInterface.jl
+++ b/src/MathOptInterface.jl
@@ -63,6 +63,20 @@ Supported file types depend on the instance type.
 function read! end
 
 """
+    isempty(instance::AbstractInstance)
+
+Returns whether the `instance` has an objective set or has any variables or constraints its model.
+"""
+function isempty end
+
+"""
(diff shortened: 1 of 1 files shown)


src/MathOptInterface.jl
```diff
@@ -62,6 +62,20 @@ Supported file types depend on the instance type.
 """
 function read! end
 
+"""
+    isempty(instance::AbstractInstance)
+
+Returns whether the `instance` has an objective set or has any variables or constraints its model.
```
mlubin: "Returns `false` if ..."

What about attributes? Is the instance still empty if you've set `ObjectiveSense()`?

blegat: No, it is empty if copying into an empty instance is a no-op.

PR #189, opened 2018-01-05
Add list of attributes set

blegat:
It seems to me that *copyable* attribute is vague but it corresponds precisely to non-solver specific attributes that were set.

### DIFF
--- a/src/MathOptInterface.jl
+++ b/src/MathOptInterface.jl
@@ -113,7 +113,11 @@ end
 
 Copy the model from the instance `src` into the instance `dest`. The target instance `dest` is emptied, and all previous indices to variables or constraints in `dest` are invalidated. Returns a `CopyResult` object. If the copy is successfully, the `CopyResult` contains a dictionary-like object that translates variable and constraint indices from the `src` instance to the corresponding indices in the `dest` instance.
 
-If an attribute `attr` cannot be copied from `src` to `dest` then if `MOI.mustcopy(attr)` is `true` then an error is thrown and otherwise, if `warnattributes` is `true`, a warning is displayed.
+If an attribute `attr` cannot be copied from `src` to `dest` then:
+
+* If `mustcopy(attr)` is `true` then an error is thrown, otherwise,
+* If `warnattributes` is `true`, a warning is displayed, otherwise,
+* The attribute is silently ignored.
 
 ### Example
 
--- a/src/MathOptInterface.jl
+++ b/src/MathOptInterface.jl
@@ -109,9 +109,11 @@ struct CopyResult{T}
 end
 
 """
-    copy!(dest::AbstractInstance, src::AbstractInstance)::CopyResult
+    copy!(dest::AbstractInstance, src::AbstractInstance, warnattributes=true)::CopyResult
 
 Copy the model from the instance `src` into the instance `dest`. The target instance `dest` is emptied, and all previous indices to variables or constraints in `dest` are invalidated. Returns a `CopyResult` object. If the copy is successfully, the `CopyResult` contains a dictionary-like object that translates variable and constraint indices from the `src` instance to the corresponding indices in the `dest` instance.
+
+If an attribute `attr` cannot be copied from `src` to `dest` then if `MOI.mustcopy(attr)` is `true` then an error is thrown and otherwise, if `warnattributes` is `true`, a warning is displayed.
 
 ### Example
 
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -185,6 +185,13 @@ end
 ## Instance attributes
 
 """
+    ListOfInstanceAttributesSet()
+
+A `Vector{AbstractInstanceAttribute}` of all instance attributes that were set to the instance and are not instance-specific.
+"""
+struct ListOfInstanceAttributesSet <: AbstractInstanceAttribute end
+
+"""
     Name()
 
 A string identifying the instance.
@@ -319,6 +326,13 @@ struct ResultCount <: AbstractInstanceAttribute end
 ## Variable attributes
 
 """
+    ListOfVariableAttributesSet()
+
+A `Vector{AbstractVariableAttribute}` of all variable attributes that were set to the variable and are not instance-specific.
+"""
+struct ListOfVariableAttributesSet <: AbstractVariableAttribute end
+
+"""
     VariableName()
 
 A string identifying the variable. It is invalid for two variables to have the same name.
@@ -366,6 +380,13 @@ Possible values are:
 @enum BasisStatusCode Basic Nonbasic NonbasicAtLower NonbasicAtUpper SuperBasic
 
 ## Constraint attributes
+
+"""
+    ListOfConstraintAttributesSet()
+
+A `Vector{AbstractConstraintAttribute}` of all constraint attributes that were set to the constraint and are not instance-specific.
+"""
+struct ListOfConstraintAttributesSet <: AbstractVariableAttribute end
 
 """
     ConstraintName()
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -24,6 +24,13 @@ abstract type AbstractConstraintAttribute end
 const AnyAttribute = Union{AbstractInstanceAttribute, AbstractVariableAttribute, AbstractConstraintAttribute}
 
 """
+    mustcopy(attr::Union{Type{AbstractInstanceAttribute}, Type{AbstractVariableAttribute}, Type{AbstractConstraintAttribute}})
+
+Returns whether the attribute affects the instance is thus mandatory to copy in `MOI.copy!` or whether it only affects how the instance is solved.
+"""
+function mustcopy end
+
+"""
     get(instance::AbstractInstance, attr::AbstractInstanceAttribute)
 
 Return an attribute `attr` of the instance `instance`.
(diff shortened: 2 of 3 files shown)


src/attributes.jl
```diff
@@ -183,6 +183,13 @@ end
 
 ## Instance attributes
 
+"""
+    ListOfInstanceAttributesSet()
+
+A `Vector{AbstractInstanceAttribute}` of all instance attributes that were set to the instance and are not instance-specific.
```
mlubin: What does `instance-specific` mean?

mlubin:
I find "non-solver specific" also vague. There could be attributes supported by a subset of solvers but not all. How do we decide what to copy? We should start out by deciding which attributes are necessary for the mathematical description of the problem (`ObjectiveFunction`) and which are not (`GurobiLogLevel`). Does `LogLevel` get copied if some solvers support it and some don't?

src/attributes.jl
blegat: `solver-specific` only makes sense for solver instance so I used `instance-specific` which should make sense for any `AbstractInstance`

mlubin: What is the definition of instance-specific? Anything that is not defined directly in MOI?

mlubin: If so, this definition does not work. An instance may have an attribute that's essential to the definition of the problem that's not defined in MOI. Imagine an MOI extension for multiobjective problems.

blegat:
Yes, that's tricky. Maybe we should copy every attributes that were set but then we need to know to react a `MOI.canset` returning false. For some we might just ignore it. For others we might give a warning. For others we need to error.

We could split attributes in 2 category:
* The attributes that affect how the instance is solved (log-level, simplex-or-interior_point, ...)
* The attributes that affect the instance (e.g. ObjectiveSense, ObjectiveFunction, Name, VariableName, ConstraintName)

For the second category, we error. For the first we either do a warning or we don't say anything depending on a keyword argument of `MOI.copy!`.

We would distinguish them with a function
```julia
MOI.mandatorytocopy(att)
```

What do you think ?

mlubin:
I agree with the proposal. `MOI.mandatorytocopy` is a bit long, how about `MOI.mustcopy`?

For reference, Gurobi calls your first category ["parameters"](http://www.gurobi.com/documentation/7.5/refman/parameters.html) and the second category ["attributes"](http://www.gurobi.com/documentation/7.5/refman/attributes.html).

mlubin:
I'll merge #190 so you can make a PR that touches the description of copy.

blegat:
@mlubin Any objection ?

src/attributes.jl
```diff
@@ -317,6 +331,13 @@ struct ResultCount <: AbstractInstanceAttribute end
 
 ## Variable attributes
 
+"""
+    ListOfVariableAttributesSet()
+
+A `Vector{AbstractVariableAttribute}` of all variable attributes that were set to the variable.
+"""
+struct ListOfVariableAttributesSet <: AbstractVariableAttribute end
```
mlubin: This should be an instance attribute instead of a variable attribute. It could be expensive to query this for every variable. The definition of `canget` says:
    
    Return a Bool indicating whether the instance instance currently has a value for the attributed specified by attribute type attr applied to every variable indexed by v or constraint indexed by c.

Any attribute that's available for one variable has to be available for all. This also applies to constraints, but we may need to revise this definition to restrict it to all constraints of the same `F`-in-`S` type.

blegat: Indeed, good point.
Once we have the list, we still need to do `MOI.get` for every attribute and every attribute but I guess this is unavoidable.

mlubin: You mean every attribute and every variable? We have the vector version of `get`.

blegat: Ah yes right. 
There is also the `canget` version for single variable. If an attribute is only set to a few variables, calling `MOI.canget(instance, attr, MOI.get(instance, MOI.ListOfVariableIndices())`, will return false because it is not applied to all variables so they will be ignored by `MOI.copy!`.

mlubin: I don't think we should allow attributes to be defined only for some variables.

blegat: I have updated the PR so that the list of attributes does not depend on the variable. As far as this PR is concerned this solve the issue. It remains to decide what we do once we have the list of attributes but this is not related to this PR so I have moved the discussion to https://github.com/JuliaOpt/MathOptInterface.jl/issues/191

src/attributes.jl
```diff
@@ -366,6 +387,13 @@ Possible values are:
 
 ## Constraint attributes
 
+"""
+    ListOfConstraintAttributesSet()
+
+A `Vector{AbstractConstraintAttribute}` of all constraint attributes that were set to the instance.
```
mlubin: This may need to be redefined to include `F` and `S` as parameters depending on the outcome of #191.

src/attributes.jl
```diff
@@ -317,6 +331,13 @@ struct ResultCount <: AbstractInstanceAttribute end
 
 ## Variable attributes
 
+"""
+    ListOfVariableAttributesSet()
+
+A `Vector{AbstractVariableAttribute}` of all variable attributes that were set to the instance.
+"""
+struct ListOfVariableAttributesSet <: AbstractVariableAttribute end
```
mlubin: `AbstractInstanceAttribute`

src/attributes.jl
```diff
@@ -366,6 +387,13 @@ Possible values are:
 
 ## Constraint attributes
 
+"""
+    ListOfConstraintAttributesSet()
+
+A `Vector{AbstractConstraintAttribute}` of all constraint attributes that were set to the instance.
+"""
+struct ListOfConstraintAttributesSet <: AbstractVariableAttribute end
```
mlubin: `AbstractInstanceAttribute`

src/attributes.jl
```diff
@@ -23,6 +23,13 @@ abstract type AbstractConstraintAttribute end
 
 const AnyAttribute = Union{AbstractInstanceAttribute, AbstractVariableAttribute, AbstractConstraintAttribute}
 
+"""
+    mustcopy(attr::Union{Type{AbstractInstanceAttribute}, Type{AbstractVariableAttribute}, Type{AbstractConstraintAttribute}})
+
+Returns whether the attribute affects the instance is thus mandatory to copy in `MOI.copy!` or whether it only affects how the instance is solved.
```
mlubin: "affects the instance" is unclear. Just say "is mandatory to copy" and give examples like `ObjectiveSense` compared with `GurobiLogLevel` (hypothetical).

src/attributes.jl
```diff
@@ -184,6 +191,13 @@ end
 
 ## Instance attributes
 
+"""
+    ListOfInstanceAttributesSet()
+
+A `Vector{AbstractInstanceAttribute}` of all instance attributes that were set to the instance.
```
mlubin: How is a solver supposed to decide if an attribute was set, does it need to store a bool for every attribute it supports?

blegat: I do not see any way around that. most solver specific attributes have default values but they should return these attributes. For instance if we have an MOI attributes for loglevel, if we do not set the log level of `src`, it should not impose its log level of `dest`.

mlubin: Ok, fair enough.

src/attributes.jl
```diff
@@ -23,6 +23,17 @@ abstract type AbstractConstraintAttribute end
 
 const AnyAttribute = Union{AbstractInstanceAttribute, AbstractVariableAttribute, AbstractConstraintAttribute}
 
+"""
+    mustcopy(attr::Union{Type{AbstractInstanceAttribute}, Type{AbstractVariableAttribute}, Type{AbstractConstraintAttribute}})
+
+Returns is mandatory to copy in `MOI.copy!` or whether it only affects how the instance is solved.
```
mlubin: Returns `true` if it is mandatory to copy the attribute in `MOI.copy!` and `false` if the attribute only affects how the instance is solved.

src/attributes.jl
```diff
@@ -23,6 +23,17 @@ abstract type AbstractConstraintAttribute end
...
+
+### Examples
+
+The attributes `ObjectiveFunction` and `ObjectiveSense` are mandatory but an hypothetical attribute such as `GurobiLogLevel` is not mandatory.
```
mlubin: "a hypothetical"

src/MathOptInterface.jl
```diff
@@ -109,10 +109,12 @@ struct CopyResult{T}
 end
 
 """
-    copy!(dest::AbstractInstance, src::AbstractInstance)::CopyResult
+    copy!(dest::AbstractInstance, src::AbstractInstance, warnattributes=true)::CopyResult
 
 Copy the model from the instance `src` into the instance `dest`. The target instance `dest` is emptied, and all previous indices to variables or constraints in `dest` are invalidated. Returns a `CopyResult` object. If the copy is successfully, the `CopyResult` contains a dictionary-like object that translates variable and constraint indices from the `src` instance to the corresponding indices in the `dest` instance.
 
+If an attribute `attr` cannot be copied from `src` to `dest` then if `MOI.mustcopy(attr)` is `true` then an error is thrown and otherwise, if `warnattributes` is `true`, a warning is displayed.
```
mlubin: Sorry to keep tweaking the wording but this isn't super clear. Can we make this into a list?

If an attribute `attr` cannot be copied from `src` to `dest` then:

- If `mustcopy(attr)` is `true` then an error is thrown, otherwise,
- If `warnattributes` is `true`, a warning is displayed, otherwise,
- The attribute is silently ignored.

PR #190, opened 2018-01-06
Implement CopyResult and remove cancopy

### LINKED ISSUE #188: Interpreting the output of cancopy
As mentionned in this PR https://github.com/JuliaOpt/MathOptInterfaceUtilities.jl/pull/50 (it would also be useful for an automatic bridger (described here: https://github.com/JuliaOpt/JuMP.jl/issues/1152#issuecomment-354373234)), it would be nice to have a way to detect why `cancopy` return `false`.

For instance we could have
```julia
cancopyconstraint(instance::AbstractInstance, f::AbstractFunction, s::AbstractSet)
```
It returns false if `cancopy(instance, src)` returns false for any possible instance `src` that contains the constraint `f`-in-`s`.
That is, if the constraint is supported but cannot be mixed with specific other constraints, then it should return true.

This is useful for automatic bridger as it shouls that this constraint definitely need to be bridged.

What do you think ?

mlubin:
Fixes #188

@blegat @odow

### DIFF
--- a/src/MathOptInterface.jl
+++ b/src/MathOptInterface.jl
@@ -77,9 +77,38 @@ Empty the instance, that is, remove the objective and all variables and constrai
 function empty! end
 
 """
-    copy!(dest::AbstractInstance, src::AbstractInstance)
+    CopyStatusCode
 
-Copy the model from the instance `src` into the instance `dest`. The target instance `dest` is emptied, and all previous indices to variables or constraints in `dest` are invalidated. Returns a dictionary-like object that translates variable and constraint indices from the `src` instance to the corresponding indices in the `dest` instance.
+An Enum of possible statuses returned by a `copy!` operation through the `CopyResult` struct.
+
+* `CopySuccess`: The copy was successful.
+* `CopyUnsupportedAttribute`: The copy failed because the destination does not support an attribute present in the source.
+* `CopyUnsupportedConstraint`: The copy failed because the destination does not support a constraint present in the source.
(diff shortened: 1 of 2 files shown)


src/MathOptInterface.jl
```diff
@@ -77,9 +77,38 @@ Empty the instance, that is, remove the objective and all variables and constrai
...
+struct CopyResult{DictLike}
+    status::CopyStatusCode
+    message::String    # Human-friendly explanation why the copy failed
+    indexmap::DictLike # Only valid if status is CopySuccess
+end
+
+"""
+    copy!(dest::AbstractInstance, src::AbstractInstance)::CopyResult
+
+Copy the model from the instance `src` into the instance `dest`. The target instance `dest` is emptied, and all previous indices to variables or constraints in `dest` are invalidated. Returns a `CopyResult` object. If the copy is successfuly, the `CopyResult` contains a dictionary-like object that translates variable and constraint indices from the `src` instance to the corresponding indices in the `dest` instance.
```
blegat: successfuly -> successfully

blegat: If the copy failed. What is the state of `dest` ? Does it need to be as it was before (without even having been emptied), empty, or anything ? I suggest, we do not impose any state, the user have just triggered a copy into `dest` so he is prepared to loose its content.

mlubin: Agreed that the state of dest should be undefined in this case. Updated the documentation.

PR #192, opened 2018-01-09
canget on typeof of index instead of index

### LINKED ISSUE #191: Setting attributes for a subset of variables only
Currently we have
```julia
    canget(instance::AbstractInstance, attr::AbstractVariableAttribute, v::VariableIndex)::Bool
    canget(instance::AbstractInstance, attr::AbstractConstraintAttribute, c::ConstraintIndex{F,S})::Bool

Return a `Bool` indicating whether the instance `instance` currently has a value for the attributed specified by attribute type `attr` applied to the variable indexed by `v` or constraint indexed by `c`.

    canget(instance::AbstractInstance, attr::AbstractVariableAttribute, v::Vector{VariableIndex})::Bool
    canget(instance::AbstractInstance, attr::AbstractConstraintAttribute, c::Vector{ConstraintIndex{F,S}})::Bool

Return a `Bool` indicating whether the instance `instance` currently has a value for the attributed specified by attribute type `attr` applied to *every* variable indexed by `v` or constraint indexed by `c`.
```
This leaves the possibility that for some attributes can be got from some variables but not others. During a copy, it would be inefficient to check for every variable attributes if canget is false or not.
One possible solution would be to have `canget(instance, attr)` without the indices arguments and disallow attributes to be set only for some variables but not all.
One other solution would be to do something like
```julia
# inside copy!(dest, src)
vars = MOI.get(src, MOI.ListOfVariableIndices())
for attr in MOI.get(src, MOI.ListOfVariablesAttributesSet())
    # We know that attr has been set to at least one variable
    if MOI.canget(src, attr, vars)
        ...
    else
        for var in vars
             if MOI.canget(src, attr, var)
                ...
            end
        end
    end
end
```
With this code, `MOI.canget` will be called on every variable **only if** `src` gives a variable attribute, so it means that it has been set for at least one variable, and then return false for `MOI.canget(src, attr, vars)` which means that is has been set for only some variables. So instead of disallowing the case of variable attribute only set for some variables, we just make the inefficiency only happen in this case.

blegat:
Closes #191

### DIFF
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -86,36 +86,38 @@ function get!(output, instance::AbstractInstance, attr::AnyAttribute, args...)
 end
 
 """
-    canget(instance::AbstractInstance, attr::AbstractVariableAttribute, v::VariableIndex)::Bool
-    canget(instance::AbstractInstance, attr::AbstractConstraintAttribute, c::ConstraintIndex{F,S})::Bool
+    canget(instance::AbstractInstance, attr::AbstractInstanceAttribute)::Bool
 
-Return a `Bool` indicating whether the instance `instance` currently has a value for the attributed specified by attribute type `attr` applied to the variable indexed by `v` or constraint indexed by `c`.
+Return a `Bool` indicating whether `instance` currently has a value for the attribute specified by attribute type `attr`.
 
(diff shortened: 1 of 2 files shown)


mlubin:
Using `VariableIndex` in the API is bit unnecessary and weird, but I can see the analogy with `get`.

blegat:
It was in fact to be consistent with `canset` which does exactly this.

PR #194, opened 2018-01-11
Implements mustcopy for MOI attributes

blegat:
Should we set `mustcopy` to `true` or `false` by default or no no default ?
If we set it to `true`, it may mean more work from the wrapper since most of their attributes will need to return `false` but it may be safer.

### DIFF
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -415,7 +415,6 @@ mustcopy(::ConstraintName) = false
 An initial assignment of the constraint primal values that the solver may use to warm-start the solve.
 """
 struct ConstraintPrimalStart <: AbstractConstraintAttribute end
-mustcopy(::ConstraintPrimalStart) = false
 
 
 """
@@ -424,7 +423,6 @@ mustcopy(::ConstraintPrimalStart) = false
 An initial assignment of the constraint duals that the solver may use to warm-start the solve.
 """
 struct ConstraintDualStart <: AbstractConstraintAttribute end
-mustcopy(::ConstraintDualStart) = false
 
 """
     ConstraintPrimal(N)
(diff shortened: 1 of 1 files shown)


mlubin:
We need to pick a safe default, which is either `true` or no default at all.

blegat:
I have picked `true` at the moment but no default may be better because it needs to be well though of and forcing explicit choice with a `MethodError` reminder seems better. What do you think ?

src/attributes.jl
```diff
@@ -356,6 +356,7 @@ struct VariableName <: AbstractVariableAttribute end
 An initial assignment of the variables that the solver may use to warm-start the solve.
 """
 struct VariablePrimalStart <: AbstractVariableAttribute end
+mustcopy(::VariablePrimalStart) = false
```
mlubin: Now this is tricky. For nonconvex problems, I consider the starting point an essential part of the problem definition, since the combination of the starting point and algorithm define the solution. Let's not set `mustcopy` to `false` here until it becomes an issue. We may eventually need to define two different attributes to capture the two different meanings of starting points.

mlubin:
I'm okay with letting `true` be the default for now, but we may revise this decision when we see some of its implications.

src/attributes.jl
blegat: Done. Does is also old for ConstraintPrimalStart and ConstraintDualStart ?

mlubin:
Yes, same holds there.

src/attributes.jl
```diff
@@ -349,6 +350,7 @@ struct ListOfVariableAttributesSet <: AbstractInstanceAttribute end
 A string identifying the variable. It is invalid for two variables to have the same name.
 """
 struct VariableName <: AbstractVariableAttribute end
+mustcopy(::VariableName) = false
```
mlubin: Eh, why is this a special case? Solvers should generally support names.

blegat: I just thought that instance used through an instance manager do not have to support it. But it's easy to implement in the wrapper so we might as well do it.

PR #195, opened 2018-01-12
Update to match doc

blegat:


src/attributes.jl
```diff
@@ -162,7 +162,7 @@ canset(instance, ConstraintPrimal(), ConstraintIndex{VectorAffineFunction{Float6
 """
 function canset end
 canset(instance::AbstractInstance, attr::AnyAttribute) = false
-canset(instance::AbstractInstance, attr::AnyAttribute, ref::Index) = false
+canset(instance::AbstractInstance, attr::AnyAttribute, ref::Type{<:Index}) = false
 canset(instance::AbstractInstance, attr::AnyAttribute, refs::Vector{<:Index}) = false
```
mlubin: Should this vector version still exist?

blegat: I'd vote to remove it

mlubin: Is it even documented?

mlubin: Yes, it is. But anyway it's redundant. Let's remove it.

PR #170, opened 2017-11-18
remove the "solver" concept

mlubin:
Closes https://github.com/JuliaOpt/MathOptInterface.jl/issues/164

PR #196, opened 2018-01-15
Add supportconstraint and supportattribute

blegat:
Follow up of #188, alternative of `supportsproblem` that was axed in #170

### DIFF
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -37,7 +37,7 @@ Return a `Bool` indicating whether `instance` supports the variable attribute `a
 Return a `Bool` indicating whether `instance` supports the constraint attribute `attr`.
 
 In other words, it should return `true` if `copy!(instance, src)` does not return `CopyUnsupportedAttribute` when the attribute `attr` is set to `src`.
-If the attribute is not supported in specific circumstances, it should still return `true`.
+If the attribute is only not supported in specific circumstances, it should still return `true`.
 """
 function supportsattribute end
 supportsattribute(::AbstractInstance, ::AnyAttribute, ::Type{<:Any}) = false
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -34,7 +34,7 @@ Return a `Bool` indicating whether `instance` supports the variable attribute `a
 
     supports(instance::AbstractInstance, attr::AbstractConstraintAttribute, ::Type{ConstraintIndex{F,S}})::Bool where {F,S}
 
-Return a `Bool` indicating whether `instance` supports the constraint attribute `attr`.
+Return a `Bool` indicating whether `instance` supports the constraint attribute `attr` applied to an `F`-in-`S` constraint.
 
 In other words, it should return `true` if `copy!(instance, src)` does not return `CopyUnsupportedAttribute` when the attribute `attr` is set to `src`.
 If the attribute is only not supported in specific circumstances, it should still return `true`.
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -24,6 +24,23 @@ abstract type AbstractConstraintAttribute end
 const AnyAttribute = Union{AbstractInstanceAttribute, AbstractVariableAttribute, AbstractConstraintAttribute}
 
 """
+    supportsattribute(instance::AbstractInstance, attr::AbstractInstanceAttribute, ::Type{V}) where V
+
+Return a `Bool` indicating whether there exists an instance `src` that contains the attribute `attr` and `get(src, attr)` is of type `V` such that `copy!(instance, src)` returns a `CopyResult` object with a status `CopySuccess`.
+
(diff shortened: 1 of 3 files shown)


src/attributes.jl
```diff
@@ -23,6 +23,23 @@ abstract type AbstractConstraintAttribute end
 
 const AnyAttribute = Union{AbstractInstanceAttribute, AbstractVariableAttribute, AbstractConstraintAttribute}
 
+"""
+    supportsattribute(instance::AbstractInstance, attr::AbstractInstanceAttribute, ::Type{V}) where V
+
+Return a `Bool` indicating whether there exists an instance `src` that contains the attribute `attr` and `get(src, attr)` is of type `V` such that `copy!(instance, src)` returns a `CopyResult` object with a status `CopySuccess`.
```
mlubin: s/`src`/`instance`/

mlubin: "There exists" is confusing. What does this mean?

blegat: I think it should be `src`, it is different to the `instance` that is passed as first argument

blegat: This is an attempt of rigorous version of "`instance` support the attribute/constraint except eventually on some special case such as constraints that cannot be together in the same instance`. With the "There exists" we only require that there is at least one situation in which the constraint is supported. Therefore, if this return false, it means that the presence if such constraint/attribute in `src` will make it impossible to copy it into `instance` whatever is the rest of `src`.

mlubin:
What is the intended use case for this?

mlubin:
Sorry, I see that it's the automatic bridger. Let's be clear about this instead of having obscure definitions. Something like "a `Bool` indicating if the attribute is supported in the context of the automatic bridger".

blegat:
It could be more general than that. For instance, if you want to add a SOC constraints but you want to support both Gurobi/CPLEX and Mosek/SCS (and you don't want to use bridges), you could do
```julia
if supportsconstraint(... Quadratic ...)
 ...
elseif supportsconstraint(... SOC cone...)
 ...
else
 error(...)
end
```
Since the AutomaticBridger is outside of MOI, it might be cleaner to try to have a definition independent of it.

mlubin:
Okay, then the definition needs some work.

mlubin:
For example, describe the intention of methods first and then add a note about what should happen in corner cases.

blegat:
It should be better now :)

src/attributes.jl
```diff
@@ -23,6 +23,26 @@ abstract type AbstractConstraintAttribute end
...
+Return a `Bool` indicating whether `instance` supports the variable attribute `attr`.
+
+    supportsattribute(instance::AbstractInstance, attr::AbstractConstraintAttribute, ::Type{ConstraintIndex{F,S}})::Bool where {F,S}
+
+Return a `Bool` indicating whether `instance` supports the constraint attribute `attr`.
+
+In other words, it should return `true` if `copy!(instance, src)` does not return `CopyUnsupportedAttribute` when the attribute `attr` is set to `src`.
+If the attribute is only not supported in specific circumstances, it should still return `true`.
+"""
+function supportsattribute end
```
mlubin: Would it be consistent with `set` and `get!` to name this `supports`?

blegat: Yes, it would

src/attributes.jl
```diff
@@ -23,6 +23,26 @@ abstract type AbstractConstraintAttribute end
...
+
+Return a `Bool` indicating whether `instance` supports the instance attribute `attr`.
+
+    supports(instance::AbstractInstance, attr::AbstractVariableAttribute, ::Type{VariableIndex})::Bool
+
+Return a `Bool` indicating whether `instance` supports the variable attribute `attr`.
+
+    supports(instance::AbstractInstance, attr::AbstractConstraintAttribute, ::Type{ConstraintIndex{F,S}})::Bool where {F,S}
+
+Return a `Bool` indicating whether `instance` supports the constraint attribute `attr`.
```
mlubin: ... applied to an `F`-in`S` constraint

blegat: Indeed, fixed :)

src/attributes.jl
```diff
@@ -23,6 +23,26 @@ abstract type AbstractConstraintAttribute end
...
+
+    supports(instance::AbstractInstance, attr::AbstractConstraintAttribute, ::Type{ConstraintIndex{F,S}})::Bool where {F,S}
+
+Return a `Bool` indicating whether `instance` supports the constraint attribute `attr` applied to an `F`-in-`S` constraint.
+
+In other words, it should return `true` if `copy!(instance, src)` does not return `CopyUnsupportedAttribute` when the attribute `attr` is set to `src`.
+If the attribute is only not supported in specific circumstances, it should still return `true`.
+"""
+function supports end
+supports(::AbstractInstance, ::AnyAttribute, ::Type{<:Any}) = false
```
mlubin: This doesn't match the documentation. Is the last argument supposed to be there?

blegat: No it doesn't

Issue #31, opened 2017-07-14
Return type of getattribute functions


IssamT:
Is a detailed documentation enough or should we have something like the following?

```julia
abstract type AbstractSolverInstanceAttribute{T} end
...
```

and

```julia
function getattribute(m, attr::AnyAttribute{T}, args...)::T
```

We can have shortcuts like `NumberOfVariables <: AbstractSolverInstanceAttribute{Int}`

mlubin:
I can see it being useful to be able to know programmatically which type an attribute corresponds to.

- For `RawSolver` we say `<: AbstractSolverInstanceAttribute{Any}`?
- `ListOfConstraints` is `<: AbstractSolverInstanceAttribute{Vector{Any}}` or `<: AbstractSolverInstanceAttribute{Any}` or something more specific?

Note that most solver instances won't implement
```julia
function getattribute(m, attr::AnyAttribute{T}, args...)::T
```
They will implement separate `getattribute` methods, e.g.,
```julia
function getattribute(m::GurobiSolverInstance, attr::NumberOfVariables)
...
```
so it's not that relevant to the implementation.

IssamT:
You are right. My suggestion will not put any restriction on the method implementation of the solver. And I don't think it is possible to do it in Julia as long as the method implemented by the solver is the final one to be called by the user. 

What about adding a check like this. But I don't know it s very useful... It only helps find errors that are on the solver implementation.

```julia
function getattribute{T}(m, attr::AnyAttribute{T})
    ret = solvergetattribute(m, attr)
    if isa(ret, T)
        ret
    else
        error("attribute $(typeof(attr)) should have a return of type $T")
    end
end 
```

mlubin:
I don't see a reason to impose that extra level of indirection, except for debugging purposes.
MOI is an interface and it's up to the solvers to implement it properly. There are lots of other places where we have to trust that the implementation is correct.

IssamT:
can julia inline this?

getattribute{T}(m, attr::AnyAttribute{T})::T = solvergetattribute(m, attr)

P.S. This would only check that the return type is convertible to T.  So not that strong...

I think we are not the only ones facing such issue with interfaces.

https://github.com/JuliaLang/julia/issues/6975

by the way, why do you provide the arguments? do they restrict something , or they could as well have been written just in the documentation ?

mlubin:
[...]

PR #198, opened 2018-01-19
Parametrize ObjectiveFunction

### LINKED ISSUE #197: get ObjectiveFunction is not type stable
We can query the function type...

blegat:
Let's break the MOI world again \o/
Closes #197

### DIFF
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -252,12 +252,15 @@ has value greater than zero.
 struct ListOfConstraints <: AbstractInstanceAttribute end
 
 """
-    ObjectiveFunction()
+    ObjectiveFunction{F<:AbstractFunction}()
 
-An `AbstractFunction` instance which represents the objective function.
+An `F` instance which represents the objective function.
 It is guaranteed to be equivalent but not necessarily identical to the function provided by the user.
+Throws an `InexactError` if the objective function cannot be converted to `F`,
+e.g. the objective function is quadratic and `F` is `ScalarAffineFunction{Float64}` or
+it has non-integer coefficient and `F` is `ScalarAffineFunction{Int}`.
 """
-struct ObjectiveFunction <: AbstractInstanceAttribute end
+struct ObjectiveFunction{F<:AbstractScalarFunction} <: AbstractInstanceAttribute end
 
 ## Solver instance attributes
 
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -252,7 +252,7 @@ has value greater than zero.
 struct ListOfConstraints <: AbstractInstanceAttribute end
 
 """
-    ObjectiveFunction{F<:AbstractFunction}()
+    ObjectiveFunction{F<:AbstractScalarFunction}()
 
 An `F` instance which represents the objective function.
 It is guaranteed to be equivalent but not necessarily identical to the function provided by the user.
--- a/docs/src/apimanual.md
+++ b/docs/src/apimanual.md
@@ -118,7 +118,7 @@ The [`ObjectiveSense`](@ref MathOptInterface.ObjectiveSense) attribute is used f
 For example,
 ```julia
 x = addvariables!(instance, 2)
-set!(instance, ObjectiveFunction(), ScalarAffineFunction([x[1],x[2]],[5.0,-2.3],1.0))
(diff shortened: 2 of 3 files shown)


mlubin:
This is a usability loss for MOI. It would be nice if you could still use `ObjectiveFunction()` in a `set!` call, because the type information is already encoded in the object being set. Similarly if you don't care about the type, you might want to use `ObjectiveFunction()` in `get`.

Instead of making a special case for `ObjectiveFunction`, what about trying to address https://github.com/JuliaOpt/MathOptInterface.jl/issues/31?

blegat:
We could have a default constructor of ObjectiveFunction:
```julia
ObjectiveFunction() = ObjectiveFunction{AbstractScalarFunction}()
```
and define this so that solver instance do not need to worry about it
```julia
set!(instance::AbstractInstance, ::ObjectiveFunction{AbstractScalarFunction}, f::F) = set!(instance, ObjectiveFunction{F}, f)
```

blegat:
For all attributes (except `RawSolver` and `ObjectiveFunction`), the return type can be determined from the documentation. For `ListOfConstraints`, it is `Vector{Tuple{AbstractFunction, AbstractSet}}` for instance. This PR allows `ObjectiveFunction` not to be an exception.

mlubin:
> For all attributes (except RawSolver and ObjectiveFunction), the return type can be determined from the documentation.

Not quite, we're ambiguous about the types of numbers, for example. But this is a practical fix for the current issues, so I'll support it.

blegat:
> Not quite, we're ambiguous about the types of numbers, for example. But this is a practical fix for the current issues, so I'll support it.

Yes, in most cases, solver support just one number type as input and output the same type but we want to remain flexible :-P

src/attributes.jl
```diff
@@ -252,12 +252,15 @@ has value greater than zero.
 struct ListOfConstraints <: AbstractInstanceAttribute end
 
 """
-    ObjectiveFunction()
+    ObjectiveFunction{F<:AbstractFunction}()
```
mlubin: `AbstractFunction` or `AbstractScalarFunction`?

mlubin:
We should experiment with https://github.com/JuliaOpt/MathOptInterface.jl/pull/198#issuecomment-358971403 before merging this PR.

blegat:
It turns out
```julia
set!(instance::AbstractInstance, ::ObjectiveFunction{AbstractScalarFunction}, f::F) = set!(instance, ObjectiveFunction{F}(), f)
```
creates an ambiguity with `set(::InstanceType, ::AbstractInstanceAttribute, value)` that most intermediate layer (such as InstanceManager) define.

blegat:
I think it is better not to do this trick and force the user to do
```julia
MOI.set!(instance, MOI.ObjectiveFunction{MOI.ScalarAffineFunction{Float64}}, ...)
```
It uses a few additional keystrokes but it is clean and MOI will most often be used in package so more keystroke is no issue. For quick scripts and command line, people use JuMP where keystroke matters.

mlubin:
What if we have attributes called `TypedObjectiveFunction{F}()` and `ObjectiveFunction()`? `ObjectiveFunction()` would be a shortcut for `TypedObjectiveFunction` and would not show up in `ListOfInstanceAttributesSet`.

blegat:
`set!(instance, ObjectiveFunction(), ...)` will be `set!(instance, ::Type{TypedObjectiveFunction})` ?
We could also add `setobjective!` as a syntactic sugar for MOI user (implementations of MOI shouldn't care about it
```julia
function setobjective!(instance::AbstractInstance, sense::OptimizationSense, f::AbstractScalarFunction)
    MOI.set!(instance, MOI.ObjectiveSense(), sense)
    MOI.set!(instance, MOI.ObjectiveFunction{typeof(f)}, f)
end
```
Setting these two attributes separately from the user feels weird anyway.

mlubin:
Ok, we can merge the PR as is or with `setobjective!`, up to you. **But** first I want to see PRs queued up on MOIU, MOIT, and JuMP with the corresponding fixes so that the MOI ecosystem will only be in a broken state for a short period of time.

blegat:
All PRs are queued up :) I have chosen not to add `setobjective!` for now. It would have maybe made things a bit simpler in MOIU and JuMP (it can be added later smoothly without breaking anything) but in MOIT, I think we shouldn't use `setobjective!` as it would allow solver to pass tests by only implementing this function and not the `set!` equivalent.

PR #199, opened 2018-01-21
Remove mustcopy and add AbstractSolverParameter

blegat:
See discussion following https://github.com/JuliaOpt/MathOptInterfaceTests.jl/pull/44#issuecomment-359189967

### DIFF
--- a/docs/src/apireference.md
+++ b/docs/src/apireference.md
@@ -11,6 +11,7 @@ CurrentModule = MathOptInterface
 List of attribute categories.
 
 ```@docs
+AbstractSolverParameter
 AbstractInstanceAttribute
 AbstractVariableAttribute
 AbstractConstraintAttribute
@@ -43,7 +44,6 @@ Copying
 copy!
 CopyResult
 CopyStatusCode
-mustcopy
 ```
 
 List of instance attributes
--- a/docs/src/apireference.md
+++ b/docs/src/apireference.md
@@ -6,18 +6,22 @@ CurrentModule = MathOptInterface
 
 [Some introduction to API. List basic standalone methods.]
 
-## Attributes
+## Parameters and Attributes
+
+Parameter abstract type.
+```@docs
+AbstractSolverParameter
+```
 
 List of attribute categories.
 
 ```@docs
-AbstractSolverParameter
 AbstractInstanceAttribute
 AbstractVariableAttribute
 AbstractConstraintAttribute
 ```
 
-Functions for getting and setting attributes.
+Functions for getting and setting parameters and attributes.
 
 ```@docs
 canget
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -1,6 +1,18 @@
 # Attributes
 
 """
+    AbstractSolverParameter
+
+Abstract supertype for parameter objects that can be used to set or get parameters of the solver.
+
+### Note
+
+The difference between `AbstractSolverParameter` and `AbstractInstanceAttribute` lies in the behavior of `isempty`, `empty!` and `copy!`.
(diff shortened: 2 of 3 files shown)


blegat:
Here, all variables and constraints attribute are obviously removed with the variables. Does it mean that all variable and constraint attributes should have `mustcopy` equal to `true` ?
Then, should we move `AbstractInstanceAttribute` that have `mustcopy` equal to `false` simply be `AbstractSolverAttribute` and get rid of `mustcopy` ?

mlubin:
> Then, should we move AbstractInstanceAttribute that have mustcopy equal to false simply be AbstractSolverAttribute and get rid of mustcopy ?

@blegat I'm having trouble coming up with a reason not to do this. It matches up with Gurobi's concept of a parameter, so we might as well call it `AbstractSolverParameter`. Should the MOIU instance support setting arbitrary parameters?

@joaquimg @odow

joaquimg:
We have to be careful if one set a solver1 parameter and then calls solver2. If parameters are integers MOIU might not break and silently set a completely different parameter. Solvers should have ther AbstractGurobiSolverParameter or we make these concrete GurobiSolverParameter(parameter,value)...

blegat:
Each parameter will have its own type, subtype of `AbstractSolverParameter`, when parameters will be passed from a solver to another solver through a `copy!`, all solver specific parameters will be dropped since the other solver will return `false` to `canset` since it does not know the parameter.

blegat:
@mlubin @joaquimg I have redirected the PR to introduce AbstractSolverParameter and remove mustcopy. What do you think of it ?

Note that `ListOfSolverParametersSet` is a `AbstractSolverParameter` since it is not affected by `empty!`.

src/attributes.jl
```diff
@@ -21,9 +33,13 @@ Abstract supertype for attribute objects that can be used to set or get attribut
 """
 abstract type AbstractConstraintAttribute end
 
-const AnyAttribute = Union{AbstractInstanceAttribute, AbstractVariableAttribute, AbstractConstraintAttribute}
+const AnyAttribute = Union{AbstractSolverParameter, AbstractInstanceAttribute, AbstractVariableAttribute, AbstractConstraintAttribute}
```
mlubin: Is a parameter a special kind of attribute or is it a different thing? We should be precise about this.

blegat: I think it is better to split but still have a common supertype

blegat:
For default implementations of `get`, `set!`, ... I have included the parameter inside an existing method since otherwise that would unallow an instance to do e.g. `Union{AbstractSolverParameter, AbstractInstanceAttribute}` since it would give an ambiguity.

src/attributes.jl
```diff
@@ -201,11 +229,18 @@ Assign a value to the attribute `attr` of constraint `c` in instance `instance`.
...
+    throw(ArgumentError("AbstractInstance of type $(typeof(instance)) does not support setting the $(_name(attr))"))
 end
 
-# TODO: solver-independent parameters
+## Solver parameters
+
+"""
+    ListOfSolverParametersSet()
+
+A `Vector{AbstractSolverParameters}` of all solver parameters that were set to the instance.
```
joaquimg: Should be sigular, right?

blegat: Good catch !

docs/src/apireference.md
```diff
@@ -11,6 +11,7 @@ CurrentModule = MathOptInterface
 List of attribute categories.
 
 ```@docs
+AbstractSolverParameter
```
mlubin: The documentation needs restructuring (parameters aren't attributes).

blegat:
I have restructured the doc :)

PR #201, opened 2018-01-25
Add SolverName

blegat:
This is needed to solve https://github.com/JuliaOpt/JuMP.jl/issues/1092

### DIFF
--- a/docs/src/apireference.md
+++ b/docs/src/apireference.md
@@ -70,6 +70,12 @@ optimize!
 free!
 ```
 
+List of attributes optimizers attributes
+
+```@docs
+SolverName
+```
+
 List of attributes useful for optimizers
 
 
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -243,6 +243,13 @@ A `Vector{AbstractOptimizerAttribute}` of all optimizer attributes that were set
 """
 struct ListOfOptimizerAttributesSet <: AbstractOptimizerAttribute end
 
+"""
+    SolverName()
+
+A string identifying the solver.
+"""
+struct SolverName <: AbstractOptimizerAttribute end
+
 ## Model attributes
 
 """


mlubin:
I don't know if this really fits in the attribute or parameter framework. It's get-only and we'll never want to copy it. Maybe a separate `solvername` method is more appropriate?

Also it's unfortunate but I think we need to do another big renaming of the MOI abstractions. "Solver" isn't currently a clear MOI term, but we're converging to it in MOIU and JuMP, e.g.,
https://github.com/JuliaOpt/MathOptInterfaceUtilities.jl/blob/d9d1c4f20b1667e440b2a54831fefe8a117e433a/src/instancemanager.jl#L5

blegat:
The advantage of it being a solver parameter is that the InstanceManager can do
```julia
function MOI.get(m::InstanceManager, param::MOI.AbstractSolverParameter)
    MOI.get(m.solver, param)
end
```
so it does not need to do a particular case for `solvername`.
This shows that `AbstractSolverParameter` can also have a meaning for get-only parameter.

mlubin:
This doesn't really fit with my understanding of a "parameter" though, it's not something that affects the solution procedure. "Attribute" vs "parameter" might have been the wrong naming choice. Using "property" as the union of the two doesn't make much sense either.
https://github.com/JuliaOpt/MathOptInterface.jl/blob/831b87ba2dcfd9b8cf50ff73e32c856cb3488f82/src/attributes.jl#L37

What about "InstanceAttribute" and "SolverAttribute"? This again pushes on us to rename `AbstractSolverInstance` to `AbstractSolver` and `AbstractStandaloneInstance` to `AbstractInstance`.

blegat:
That seems better, but I like `AbstractInstance` for the supertype of both.
What about `AbstractSolver <: AbstractInstance` and `AbstractStandaloneInstance <: AbstractInstance` ? Or do you have a better idea for the super type ?

mlubin:
I have trouble calling a solver a kind of instance, but on the other hand it doesn't bother me to say that a solver implements the instance interface. This is awkward to deal with in Julia. Usually you'd duck-type it and not even have a super type, right?

blegat:
We could duck-type but it seems unfortunate to loose the abstract type for types implementing the MOI interface. Most of the time, we need to use `AbstractInstance`. The distinction between `AbstractStandaloneInstance` and `AbstractSolverInstance` and the separation is sometimes blurry (e.g. instance storing results) so the only thing that really separates them is `optimize!`. Even `AbstractSolverAttribute` could also be supported by `AbstractStandaloneInstance`, see https://github.com/JuliaOpt/MathOptInterfaceUtilities.jl/issues/61.
I support renaming `SolverParameter` to `SolverAttribute` but I am not convinced that we should get rid of `AbstractInstance`.

mlubin:
What about `AbstractSolver` replaces `AbstractSolverInstance` and `AbstractInstance` replaces `AbstractStandaloneInstance`, with `InstanceLike = Union{AbstractSolver,AbstractInstance}`. Solvers and instances are fundamentally different things that just overlap in their API in some places for convenience. It doesn't make sense to call a copy-only solver an "instance," but "instance-like" is less objectionable. I'm open to bikeshedding on `InstanceLike`.

blegat:
That would work, I like `InstanceLike` ! We are also using the `Like` suffix in [MultivariatePolynomials](https://github.com/JuliaAlgebra/MultivariatePolynomials.jl).

blegat:
I have update the PR, do you prefer `SolverName` or `OptimizerName` ?

mlubin:
I'm not bothered by `SolverName` but some might see it as inconsistent.

mlubin:
Then again I'm also responsible for `defVar`.

blegat:
What do you mean by defVar ?

odow:
https://github.com/JuliaOpt/JuMP.jl/blob/v0.1.1/examples/basic.jl#L19

blegat:
Ah ok got it :D @odow @joaquimg ok with SolverName ?

joaquimg:
Ok!

odow:
Not bothered.

PR #206, opened 2018-02-04
Implements Base.isapprox

blegat:
Closes https://github.com/JuliaOpt/MathOptInterfaceTests.jl/issues/33.
MOIU used to implements indexing, copy and comparison of MOI types. Thanks to `eachscalar`, it does not directly define Base methods on MOI types. With https://github.com/JuliaOpt/MathOptInterfaceUtilities.jl/pull/72, it does not implement copy anymore.
The last remaining type piracy is with `isapprox`.
Adding a method define externally with arguments having only external types should be avoided in Julia. Here `isapprox` is defined in Base so MOIU shouldn't define it for MOI types.
If MOIT defines it, that automatically makes it impossible to use by another package without adding MOIT in its requirement. If it doesn't know about it, it might give a weird failure, the first time sometime try to load the package with MOIT at the same time.
It seems to me that the only clean way to deal with this is to define `isapprox` in MOI directly.

What do you think ?

### DIFF
--- a/src/functions.jl
+++ b/src/functions.jl
@@ -170,3 +170,39 @@ struct MultirowChange{T} <: AbstractFunctionModification
     rows::Vector{Int}
     new_coefficients::Vector{T}
 end
+
+# Implementation of comparison for MOI functions
+import Base: ==
+
+==(f::VectorOfVariables, g::VectorOfVariables) = f.variables == g.variables
+
+# For affine and quadratic functions, terms are compressed in a dictionary using `_dicts` and then the dictionaries are compared with `dict_isapprox`
+function dict_isapprox(d1::Dict, d2::Dict{<:Any, T}; kwargs...) where T
+    all(kv -> isapprox(kv.second, Base.get(d2, kv.first, zero(T)); kwargs...), d1)
+end
+
+# Build a dictionary where the duplicate keys are summed
+function sum_dict(kvs::Vector{Pair{K, V}}) where {K, V}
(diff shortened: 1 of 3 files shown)


src/functions.jl
```diff
@@ -170,3 +170,39 @@ struct MultirowChange{T} <: AbstractFunctionModification
     rows::Vector{Int}
     new_coefficients::Vector{T}
 end
+
+# Implementation of comparison for MOI functions
+import Base: ==
+
+==(f::VectorOfVariables, g::VectorOfVariables; kwargs...) = f.variables == g.variables
```
mlubin: Why take `kwargs` here?

blegat: Oops

src/functions.jl
```diff
@@ -170,3 +170,39 @@ struct MultirowChange{T} <: AbstractFunctionModification
...
+        d[key] = kv.second + Base.get(d, key, zero(V))
+    end
+    d
+end
+
+_dicts(f::ScalarAffineFunction) = sum_dict(Pair.(f.variables, f.coefficients))
+_dicts(f::VectorAffineFunction) = sum_dict(Pair.(tuple.(f.outputindex, f.variables), f.coefficients))
+
+# For quadratic terms, x*y == y*x
+_sort(v1::VariableIndex, v2::VariableIndex) = VariableIndex.(extrema((v1.value, v2.value)))
```
mlubin: "canonicalize" instead of "sort"?

Issue #209, opened 2018-02-09
Implementation Questions


odow:
_Some feedback from @tkoolen from gitter. Bolded statements are replies to his questions. People are welcome to edit my replies._

First, some observations about MOI{SCS, ECOS}. These are just my notes from digging into the code. Please correct me if anything is wrong. MOISCS and MOIECOS each define an `AbstractSolverInstance`, which stores a solver-specific version of an optimization problem. The MOI documentation says that the data should be loaded in the solver’s API, but for SCS and ECOS that’s not really the case; it’s just put in a format that’s easily loadable into the solver’s API. The actual loading (as I would call it) happens in `MOI.optimize!`. So really the {SCS, ECOS}SolverInstances are very limited in their capabilities; you can basically only copy an `AbstractInstance` into them and then solve. They currently don’t support problem modification (the `MOIU.loadconstraint!` function is only used during the copying process).

Now for OSQP. OSQP gets a lot of its strength from its support for problem data modification. To support that, the `modify{constraint!, objective!}` methods for OSQPSolverInstance would ideally directly call the `OSQP.update_*` methods, directly modifying the problem data loaded into OSQP itself, as opposed to maintaining a list of changes to be made upon the next call to `optimize!`. That would probably entail loading the data into OSQP in `copy!`, similar to what @mlubin proposed for SCS.
**`@mlubin` said: "I'm also in favor of loading the data into OSQP on copy!, that's the intention of the API"**

Additionally, methods for `MOI.canmodify{constraint, objective}` are needed. There are two problems here:

 1. for constraints, I don’t think I can currently query that information from OSQP. OSQP only allows modification of constraints if the sparsity structure is maintained, and the only way to know the sparsity structure given the current OSQP interface is to try modification and see if you get an error code.
2. for the objective function, I believe `MOI.canmodifyobjective` cannot capture the constraints that OSQP places on objective modification (again, the sparsity structure must be maintained).
**It's okay to return true from canmodifyXXX and then an error on modifyXX! if the data doesn't meet the necessary conditions.**

Here's what I have at the moment: https://github.com/tkoolen/OSQP.jl/compare/tk/mathprogbase...tk/mathoptinterface. Untested. 

[...]

PR #210, opened 2018-02-09
Clarity get/set! when canget/set returns false

odow:
As raised in https://github.com/JuliaOpt/MathOptInterface.jl/issues/209

### DIFF
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -100,7 +100,7 @@ If *any* constraint with name `name` exists in the instance `instance`, return t
 ### Note
 
 It is the user's responsibility to check that `canget` returns `true` prior to calling `get`. 
-Calling `get` when `canget` returns `false` may throw an error or return an incorrect result.
+Calling `get` when `canget` returns `false` will throw an error.
 
 ### Examples
 
@@ -236,7 +236,7 @@ Assign a value respectively to the attribute `attr` of each constraint in the co
 ### Note
 
 It is the user's responsibility to check that `canset` returns `true` prior to calling `set!`. 
-Calling `set!` when `canset` returns `false` may throw an error or do nothing.
+Calling `set!` when `canset` returns `false` will throw an error.
 """
 function set! end
 function set!(instance::AbstractInstance, attr::AnyProperty, args...)
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -97,6 +97,11 @@ If an `F`-in-`S` constraint with name `name` exists in the instance `instance`, 
 
 If *any* constraint with name `name` exists in the instance `instance`, return the corresponding index, otherwise throw a `KeyError`. This version is available for convenience but may incur a performance penalty because it is not type stable.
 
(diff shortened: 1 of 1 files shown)


mlubin:
Hmm. I don't think we should allow `get` to return incorrect results or say that `set!` values can be ignored. These should be errors.

tkoolen:
I think that ideally:

* It should always be possible to get any attribute in a 'safe' way, e.g. by having a `checkedget`/`safeget` (or maybe *this* should be called `get`) that first checks `canget` and then returns the result. This should probably be the default behavior.
* There should be an `unsafeget` version (currently called `get`) for at least some properties that foregoes the `canget` check for efficiency reasons, (e.g when you want to iterate over a bunch of `VariableIndex`es) and you've already checked `canget` for `::Type{VariableIndex}`.

This is starting to sound like the `@inbounds` machinery. Unfortunately, that's hard to replicate, but maybe some kind of simple `@unsafe` macro solution to redirect to the `unsafeget` version would be nice.

mlubin:
@tkoolen is there really a need for an unsafe version? If you're iterating over a bunch of `VariableIndex`es then you can use the ``::Vector{VariableIndex}`` version of `get` instead.

blegat:
What would be done in `unsafeget` if `canget` is `false` ? The solvers returns an incorrect value and the error is thrown in `safeget` which is implemented in MOI like this ?
```julia
function MOI.safeget(...)
    if MOI.canget(...)
        MOI.get(...)
    else
        error(...)
    end
end
```

tkoolen:
@mlubin, I thought that there was a need for that given your earlier comment, https://github.com/JuliaOpt/MathOptInterfaceUtilities.jl/issues/87#issuecomment-364540567. If that's not the case, what I proposed is probably unnecessary.

@blegat, yep.

odow:
This has been updated to reflect that the call should throw an error.

src/attributes.jl
```diff
@@ -96,6 +96,11 @@ If an `F`-in-`S` constraint with name `name` exists in the model `model`, return
 
 If *any* constraint with name `name` exists in the model `model`, return the corresponding index, otherwise throw a `KeyError`. This version is available for convenience but may incur a performance penalty because it is not type stable.
 
+### Note
+
+It is the user's responsibility to check that `canget` returns `true` prior to calling `get`. 
+Calling `get` when `canget` returns `false` will throw an error.
```
mlubin: This is now a weird statement. Do we need to add this note for `addconstraint!` `modifyconstraint!`, etc (anything that as a corresponding `canXXX`)?
I would phrase it more of a note for implementations. These functions *must* throw errors if the corresponding `canXXX` returned false.

odow: Should this just be a general note in the manual then?

mlubin: Yeah, could be in the section on notes for solver wrappers.

odow:
Closing in favour of #228

PR #217, opened 2018-02-11
Last big renaming

### LINKED ISSUE #203: Last round of renaming
Now that we've mostly settled on the MOI/MOIU/JuMP API structure, it's time to rename the different components to match what they do. Last time we do this before MOI is released.

Let's list the things we have:
- `supports-MOI-interface` describes a type that implements the MOI interface for specifying problems, i.e., `addvariable!`, `addconstraint!`, `copy!`. There are different levels of support, i.e., copy-only, 2-pass load, and general `addconstraint!`.
- `data-with-optimize!` is a `supports-MOI-interface` object that lets you call `optimize!`. The data are usually stored and owned directly by the solver.
- `data-without-optimize!` is a `supports-MOI-interface` object that does not support `optimize!`. It just stores the problem data.
- `data-with-optimize!-plus-caching` is a `supports-MOI-interface` object that lets you attach and synchronize with a `data-with-optimize!`. It stores an extra copy of the data and deals with the case of the `data-with-optimize!` not supporting the full MOI interface by re-copying the data to the solver when needed. In "automatic" mode, a `data-with-optimize!-plus-caching` can act exactly like a `data-with-optimize!` (https://github.com/JuliaOpt/MathOptInterfaceUtilities.jl/pull/65). In "manual" mode it gives more control over when copies happen, and you can't naively use it as a drop-in for `data-with-optimize!`.
- `optimize!-attributes` are gettable/settable attributes related to how `optimize!` works like stopping limits, solution methods, and the name of the solver (https://github.com/JuliaOpt/MathOptInterface.jl/pull/201). These are dropped on `copy!`
- `data-attributes` are essential parts of the data specification (names, `VariablePrimalStart`, `NLPBlock`) that cannot be dropped on `copy!`.
- `result-attributes` are properties of the result computed by `optimize!`. I don't think we've discussed if they can be dropped on `copy` or not.

JuMP provides a frontend to any `supports-MOI-instance` object depending on what its `moibackend` field is set to. The default use case is for JuMP to be a frontend for`data-with-optimize!-plus-caching`.

Currently:
- `supports-MOI-interface` is `AbstractInstance`
- `data-with-optimize!` is `AbstractSolverInstance`
- `data-without-optimize!` is `AbstractStandaloneInstance`
- `data-with-optimize!-plus-caching` is `InstanceManager`
- Attributes are a mix of `AbstractSolverParameter` and `AnyAttribute` (https://github.com/JuliaOpt/MathOptInterface.jl/pull/201).

I'm going to throw out a couple questions to start the discussion. I don't have a full proposal yet, but hopefully this summary above will get everyone on the same page on the layout of the abstractions. I'd like to brainstorm and make a decision on this quickly so that we can make this change and move on!

**Question 1:** Looking at this now, the definition of `data-without-optimize` is a bit strange. Usually we don't define objects by what they don't support. Should we just remove this concept in favor of `supports-MOI-interface`? If we need to dispatch on `data-with-optimize!` and `data-without-optimize!` we can dispatch instead between `data-with-optimize!` and `supports-MOI-interface`.

**Question 2:** Do we want to call `data-with-optimize!` `AbstractSolver`? A documentation-level definition of a solver could be:
>  "A solver implements the MOI interface for loading problem data and supports `optimize!`".

This has a downside of being completely different from MPB's definition of a solver, but that's not a blocker. In the `data-with-optimize!-plus-caching` I found it convenient to talk about the "solver" and the "instance":

https://github.com/JuliaOpt/MathOptInterfaceUtilities.jl/blob/d9d1c4f20b1667e440b2a54831fefe8a117e433a/src/instancemanager.jl#L12

Note that any use of "solver" in MOI should be consistent with the use of `solver=` in JuMP.

**Question 3:** Can we make JuMP's `Model` somehow consistent with the MOI naming convention? JuMP is now essentially a nice interface to any `supports-MOI-interface` object. That would suggest picking a name for `supports-MOI-interface` that's similar to JuMP's `Model`, or renaming JuMP's `Model`.

@blegat @odow @joaquimg @chriscoey @joehuchette @ccoffrin @IssamT

mlubin:
`AbstractInstance` -> `ModelLike`
`AbstractStandaloneInstance` -> gone
`AbstractSolverInstance` -> `AbstractOptimizer`
`AbstractInstanceAttribute` -> `AbstractModelAttribute`
`AbstractSolverParameter` -> `AbstractOptimizerAttribute`

Updated the discussion in the manual, let me know if it reads well. If there's consensus on this renaming, I'll queue up PRs on the downstream packages and merge all at once.

Closes #203

### DIFF
--- a/docs/src/apimanual.md
+++ b/docs/src/apimanual.md
@@ -93,7 +93,7 @@ v2 = addvariables!(model, 3)
 set!(model, VariablePrimalStart(), v2, [1.3,6.8,-4.6])
 ```
 
-A variable can be deleted from a model with [`delete!(::ModelLike, ::VariableIndex)`](@ref MathOptInterface.delete!(::MathOptInterface.ModelLike, ::MathOptInterface.Index)), if this functionality is supported.
+A variable can be deleted from a model with [`delete!(::ModelLike, ::VariableIndex)`](@ref MathOptInterface.delete!(::MathOptInterface.ModelLike, ::MathOptInterface.Index)). Not all models support deleting variables; the [`candelete`](@ref MathOptInterface.candelete) method is used to check if this is supported.
 
 ## Functions
 
@@ -348,7 +348,7 @@ primal_variable_result = MOI.get(optimizer, MOI.VariablePrimal(), x)
 ### Duals
 
 
-Conic duality is the starting point for MOI's duality conventions. When all functions are affine (or coordinate projections), and all constraint sets are closed convex cone, the model may be called a conic optimization problem.
+Conic duality is the starting point for MOI's duality conventions. When all functions are affine (or coordinate projections), and all constraint sets are closed convex cones, the model may be called a conic optimization problem.
 For conic-form minimization problems, the primal is:
 
 ```math
--- a/docs/src/apimanual.md
+++ b/docs/src/apimanual.md
@@ -78,7 +78,7 @@ provide methods to solve the model.
 Through the rest of the manual, `model` is used as a generic `ModelLike`, and
 `optimizer` is used as a generic `AbstractOptimizer`.
 
-[Discuss how models are constructed, optimizer parameters.]
+[Discuss how models are constructed, optimizer attributes.]
 
 ## Variables
 
--- a/docs/src/apimanual.md
+++ b/docs/src/apimanual.md
@@ -3,7 +3,7 @@
 
 ## Purpose
 
-Each mathematical optimization solver API has its own concepts and data structures for representing optimization instances and obtaining results.
+Each mathematical optimization solver API has its own concepts and data structures for representing optimization models and obtaining results.
 However, it is often desirable to represent an instance of an optimization problem at a higher level so that it is easy to try using different solvers.
 MathOptInterface (MOI) is an abstraction layer designed to provide a unified interface to mathematical optimization solvers so that users do not need to understand multiple solver-specific APIs.
 MOI can be used directly, or through a higher-level modeling interface like [JuMP](https://github.com/JuliaOpt/JuMP.jl).
@@ -19,7 +19,7 @@ MOI is designed to:
 - Enable a solver to more precisely specify which problem classes it supports
 - Enable both primal and dual warm starts
 - Enable adding and removing both variables and constraints by indices that are not required to be consecutive
-- Enable any modification that the solver supports to an existing instance
+- Enable any modification that the solver supports to an existing model
 - Avoid requiring the solver wrapper to store an additional copy of the problem data
 
 This manual introduces the concepts needed to understand MOI and give a high-level picture of how all of the pieces fit together. The primary focus is on MOI from the perspective of a user of the interface. At the end of the manual we have a section on [Implementing a solver interface](@ref).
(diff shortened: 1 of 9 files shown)


docs/src/apimanual.md
```diff
@@ -56,39 +56,44 @@ The current function types are:
...
-v2 = addvariables!(instance, 3)
-set!(instance, VariablePrimalStart(), v2, [1.3,6.8,-4.6])
+v = addvariable!(model)
+set!(model, VariablePrimalStart(), v, 10.5)
+v2 = addvariables!(model, 3)
+set!(model, VariablePrimalStart(), v2, [1.3,6.8,-4.6])
 ```
 
-A variable can be deleted from an instance with [`delete!(::AbstractInstance, ::VariableIndex)`](@ref MathOptInterface.delete!(::MathOptInterface.AbstractInstance, ::MathOptInterface.Index)), if this functionality is supported.
+A variable can be deleted from a model with [`delete!(::ModelLike, ::VariableIndex)`](@ref MathOptInterface.delete!(::MathOptInterface.ModelLike, ::MathOptInterface.Index)), if this functionality is supported.
```
odow: Mention `candelete`?

docs/src/apimanual.md
```diff
@@ -343,7 +348,7 @@ primal_variable_result = MOI.get(instance, MOI.VariablePrimal(), x)
 ### Duals
 
 
-Conic duality is the starting point for MOI's duality conventions. When all functions are affine (or coordinate projections), and all constraint sets are closed convex cone, the instance may be called a conic optimization problem.
+Conic duality is the starting point for MOI's duality conventions. When all functions are affine (or coordinate projections), and all constraint sets are closed convex cone, the model may be called a conic optimization problem.
```
odow: s/cone/cones

blegat:
Shouldn't `AbstractOptimizerParameter` be `AbstractOptimizerAttribute` ? See https://github.com/JuliaOpt/MathOptInterface.jl/pull/201#issuecomment-360614461

docs/src/apimanual.md
```diff
@@ -56,39 +56,44 @@ The current function types are:
...
 MOI defines some commonly used sets, but the interface is extensible to other sets recognized by the solver.
 [Describe currently supported sets.]
 
-## Instances
+## The `ModelLike` and `AbstractOptimizer` APIs
 
-An **Instance** ([`AbstractInstance`](@ref MathOptInterface.AbstractInstance)) is a representation of a concrete instance of an optimization problem, i.e., with all data specified.  Instances are either **standalone instances** or **solver instances**:
+The most significant part of MOI is the definition of the **model API** that is
+used to specify an instance of an optimization problem (e.g., by adding
+variables and constraints). Objects that implement the model API should inherit
```
blegat: "Objects that implement the model API" -> "Objects that implement **parts of** the model API" ? Or we consider that if an object does not implement something and say `can...` -> `false` then it is a complete implementation ?

mlubin: We should definitely add a discussion in the documentation of the different ways to implement the model API, but for the purposes of this sentence, I consider copy-only a valid implementation of the model API.

blegat: Yes, I think that if `can...` return false, it should be considered a valid implementation

PR #219, opened 2018-02-12
Add convenience constructors for Interval.

tkoolen:
I found these to be useful while writing the OSQP interface.

### DIFF
--- a/src/sets.jl
+++ b/src/sets.jl
@@ -102,9 +102,9 @@ struct Interval{T <: Real} <: AbstractScalarSet
     lower::T
     upper::T
 end
-Interval(s::GreaterThan) = Interval(s.lower, typemax(s.lower))
-Interval(s::LessThan) = Interval(typemin(s.upper), s.upper)
-Interval(s::EqualTo) = Interval(s.value, s.value)
+Interval(s::GreaterThan{<:AbstractFloat}) = Interval(s.lower, typemax(s.lower))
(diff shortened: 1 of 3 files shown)


odow:
Is this more suited to MOIU? I think we've tried to avoid convienence constructors like this in MOI.

tkoolen:
I would consider that to be type piracy.

src/sets.jl
```diff
@@ -102,6 +102,9 @@ struct Interval{T <: Real} <: AbstractScalarSet
     lower::T
     upper::T
 end
+Interval(s::GreaterThan) = Interval(s.lower, typemax(s.lower))
```
blegat: What do we do for `GreaterThan{Bool}` or `GreaterThan{Int}` ? In MathProgBase, we [throw a warning in that case](https://github.com/JuliaOpt/MathProgBase.jl/blob/7c6f0a8414fdc018cf7ba90eb3c5f29c2e68a738/src/HighLevelInterface/HighLevelInterface.jl#L6-L10).
We could also display a warning for non-`AbstractFloat` type or simply restrict this definition to `AbstractFloat` (I would prefer this second approach).

tkoolen: Good point. I also prefer the second approach.

mlubin:
These need to be documented in some form. Either in the docstring for `Interval` or somewhere else.

tkoolen:
Docs added.

PR #221, opened 2018-02-12
Test travis integration with gitter

mlubin:


mlubin:
I didn't find an easy way to only notify for the master branch, but I think this is still useful.

PR #223, opened 2018-02-13
Reword manual section on implementation

### LINKED ISSUE #213: Discuss implementing all constraint types
Raised by @tkoolen: 

> I can't use the MOIT tests because I only advertise support for Interval constraint sets (which I think is the right thing to do). I thought that the proper way to add support for LessThan etc. would be to have a CombineIntoIntervalBridge in MOIU, similar to SplitIntervalBridge.

> I’m still not sure whether I should advertise support for vector-valued constraints in OSQP. Perhaps document that better.

I added the following to the docs, however when I attempted to use the web editor I clicked `merge into master` rather than `make a PR`. Commit is https://github.com/JuliaOpt/MathOptInterface.jl/commit/51c8f7243e884cda860672e34aea34727aa1eb62. I can revert if people want changes.

With this in mind, developers should attempt to support as many constraint types as possible if it makes sense on the solver side. For example, if the solver only supports constraints of the form `l <= f(x) <= u`, where `l` or `u` can be `inf` or `-inf`, the wrapper should support `ScalarAffineFunction`-in-`LessThan`/`GreaterThan`/`EqualTo` instead of just `Interval`.

odow:
Closes #213


docs/src/apimanual.md
```diff
@@ -470,7 +470,7 @@ See [`PositiveSemidefiniteConeTriangle`](@ref MathOptInterface.PositiveSemidefin
...
...feature, not a bug. MOI is designed to make it possible to experiment with alternative representations of an optimization problem at both the solving and modeling level. When implementing an interface, it is important to keep in mind that the constraints which a solver supports via MOI will have a near 1-to-1 correspondence with how users can express problems in JuMP, because JuMP does not perform automatic transformations. (Alternative systems like Convex.jl do.) For example, if a solver wrapper does not support `ScalarAffineFunction`-in-`LessThan` constraints, then users will not be able to write: `@constraint(m, 2x + y <= 10)` in JuMP. With this in mind, developers should support all the constraint types that they want to be usable from JuMP. For example, if the solver only supports constraints of the form `l <= f(x) <= u`, where `l` or `u` can be `inf` or `-inf`, the wrapper should probably support `ScalarAffineFunction`-in-`LessThan`/`GreaterThan`/`EqualTo` in addition to just `Interval`. The alternative will force users to write: `@constraint(m, -Inf <= 2x + y <= 10)`. That said, from the perspective of JuMP, solvers can safely choose to not support the following constraints:
```
mlubin: This should be expanded into a list of bullet points on how JuMP transforms typical constraints.

blegat: We might as well also talk about bridges here or give a link to the bridge doc (which is currently not there)

docs/src/apimanual.md
```diff
@@ -470,7 +470,14 @@ See [`PositiveSemidefiniteConeTriangle`](@ref MathOptInterface.PositiveSemidefin
...
+MOI defines a very general interface, with multiple possible ways to describe the same constraint. This is considered a feature, not a bug. MOI is designed to make it possible to experiment with alternative representations of an optimization problem at both the solving and modeling level. When implementing an interface, it is important to keep in mind that the constraints which a solver supports via MOI will have a near 1-to-1 correspondence with how users can express problems in JuMP, because JuMP does not perform automatic transformations. (Alternative systems like Convex.jl do.) The following bullet points show examples of how JuMP expressions are translated into MOI sets:
+ - `@constraint(m, 2x + y <= 10)` becomes `ScalarAffineFunction`-in-`LessThan`;
```
mlubin: Looks good. Variable bounds are also important to mention. We translate lower and upper bounds into separate constraints even if the user writes `@variable(m, 0 <= x <= 1)`.

PR #225, opened 2018-02-14
More instance->model

odow:
Things missed in the Last Big Renaming (#217).


src/MathOptInterface.jl
```diff
@@ -40,17 +40,17 @@ function free! end
     write(model::ModelLike, filename::String)
 
 Writes the current model data to the given file.
-Supported file types depend on the solver or standalone instance type.
+Supported file types depend on the model.
```
mlubin: "model" here and "model type" for `read!`. Either is ok but should be consistent.

PR #226, opened 2018-02-14
Clarify how read! works

### LINKED ISSUE #185: how does read! work?
How does the user get the indices of the variables and constraints loaded using `read!`?

odow:
Closes #185

### DIFF
--- a/src/MathOptInterface.jl
+++ b/src/MathOptInterface.jl
@@ -54,10 +54,11 @@ Supported file types depend on the instance type.
 
 ### Note
 
-Once the model is loaded into the model, users can query the variables via
-`get(model, ListOfVariableIndices())`. However, this could return the variables
-in any order. Therefore, users should look up up each variable index by name:
-`get(model, VariableIndex, "name")`.
+Once the contents of the file are loaded into the model, users can query the variables via
+`get(model, ListOfVariableIndices())`. However, some filetypes, such as LP files, do not
+maintain an explicit ordering of the variables. Therefore, the returned list may be in an
+arbitrary order. To guarantee the ordering of the returned indices, users should look up
+up each variable index by name: `get(model, VariableIndex, "name")`.
 """
 function read! end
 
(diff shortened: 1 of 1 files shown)


src/MathOptInterface.jl
```diff
@@ -51,6 +51,13 @@ Read the file `filename` into the instance `instance`. If `m` is non-empty, this
 throw an error.
 
 Supported file types depend on the instance type.
+
+### Note
+
+Once the model is loaded into the model, users can query the variables via
+`get(model, ListOfVariableIndices())`. However, this could return the variables
+in any order. Therefore, users should look up up each variable index by name:
```
mlubin: MPS files have a well-defined order on the variables. Would we still expect `ListOfVariableIndices()` to return them back in an arbitrary order in that case?

odow: Presumably. But that seems a little flakey.

odow: This is another manual doc then? Recommending `get.(model, VariableIndex, ["x", "y", "z"])` without mentioning `ListOfVariableIndices`?

src/MathOptInterface.jl
```diff
@@ -51,6 +51,13 @@ Read the file `filename` into the instance `instance`. If `m` is non-empty, this
 throw an error.
 
 Supported file types depend on the instance type.
+
+### Note
+
+Once the model is loaded into the model, users can query the variables via
```
blegat: "Once the model is loaded into the model" is a bit difficult to grasp :-P Maybe "Once `filename`'s model is loaded in to `model`" reads better

mlubin: "Once the contents of the file are loaded into the model" maybe

odow:
Updated to reflect comments.

src/MathOptInterface.jl
```diff
@@ -51,6 +51,14 @@ Read the file `filename` into the model `model`. If `m` is non-empty, this may
 throw an error.
 
 Supported file types depend on the model type.
+
+### Note
+
+Once the contents of the file are loaded into the model, users can query the variables via
+`get(model, ListOfVariableIndices())`. However, some filetypes, such as LP files, do not
+maintain an explicit ordering of the variables. Therefore, the returned list may be in an
+arbitrary order. To guarantee the ordering of the returned indices, users should look up
```
mlubin: "To avoid depending on the order of the indices, users should look up ..."

PR #232, opened 2018-02-16
Add MOIU as a submodule

blegat:
I haven't included the bridges yet as they can be decoupled from the rest of MOIU. Bridges requires MOIU but none of MOIU requires bridges. Moreover, bridges are not necessary to test MOI so we have 3 choices:
1. Leave bridges in MOIU
2. Add bridges in a submodule MOI.Bridges
3. Put bridges in a separate package MOIBridges.jl

What do you prefer ?

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,1 +1,2 @@
 julia 0.6
+Compat 0.51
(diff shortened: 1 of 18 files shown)


joaquimg:
I have a slight preference for MOIBridges.jl but all 3 seem ok

joaquimg:
Compat is being used to keep compatibility between 0.6 and 0.7?

mlubin:
MathOptInterfaceBridges sounds good to me.

PR #233, opened 2018-02-17
Fix SOS set parameter

joaquimg:


blegat:
@joaquimg You are asking to merge it into bl/canaddvariables instead of master

joaquimg:
Opss my bad it was really late heheheh

PR #237, opened 2018-02-17
Add sizehint! in MOIU canonical

joaquimg:


blegat:
What is the performance speedup for functions without duplicates ?

joaquimg:
Well this fix is aimed for all functions not only the ones with no duplicate. Is there any harm in that sizehint?

blegat:
If there are lots of duplicates, this might allocate more than needed. So in case we have a performance win when there is only a few duplicates then the increased memory used when there are lots of duplicates is no big deal but otherwise we might reconsider.

odow:
Closing this as stale. It can be re-addressed if, and when, benchmarking shows it as a bottleneck.

PR #264, opened 2018-02-28
Test ordered ListofXXXIndices

### LINKED ISSUE #230: Decision on ListOfVariableIndices
The documentation now is pretty sparse:

> A Vector{VariableIndex} indexing all variables present in the model (i.e., of length equal to the value of NumberOfVariables()).

We should either 1) make a big point here about the variables in the list appearing in an arbitrary order or 2) define an ordering on the variables. I would propose to enforce that `ListOfVariableIndices` returns variables ordered by creation time, because:
- Most wrappers do this anyway.
- It matches user expectations.
- In the worst case the model needs to maintain an extra ordered set.

Ref https://github.com/JuliaOpt/MathOptInterface.jl/pull/226

odow:
I'm a little behind on the whole testing infrastructure so I hope this is the correct way to test it.

Closes #230

### DIFF
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -278,16 +278,17 @@ struct NumberOfVariables <: AbstractModelAttribute end
 """
     ListOfVariableIndices()
 
-A `Vector{VariableIndex}` indexing all variables present
-in the model (i.e., of length equal to the value of `NumberOfVariables()`).
+A `Vector{VariableIndex}` containing all variables present in the model (i.e.,
+of length equal to the value of `NumberOfVariables()`) ordered by creation time.
 """
 struct ListOfVariableIndices <: AbstractModelAttribute end
 
 """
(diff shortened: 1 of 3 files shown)


src/attributes.jl
```diff
@@ -278,16 +278,18 @@ struct NumberOfVariables <: AbstractModelAttribute end
 """
     ListOfVariableIndices()
 
-A `Vector{VariableIndex}` indexing all variables present
-in the model (i.e., of length equal to the value of `NumberOfVariables()`).
+A `Vector{VariableIndex}` containing all variable indices present in the model
+(i.e., of length equal to the value of `NumberOfVariables()`), ordered by
+creation time.
```
mlubin: "in the order in which they were added" (or something similar)
"creation time" almost suggests that the solver should keep a timestamp

PR #265, opened 2018-02-28
Add badges

blegat:


mlubin:
The table doesn't display correctly in the preview, are you sure it's correct?

blegat:
Indeed, it's fixed now :)

mlubin:
There's no parallelism in the meaning of the rows (stable vs. latest/travis vs. codecov?), it's a bit confusing.

blegat:
No, it is in two rows for aesthetic reason, we can do it in one row as in 1b9ec7f1a8af6b5001536166338c6df392973f01 to make it less confusing.

PR #267, opened 2018-02-28
Test primal InfeasiblePoint case

blegat:


src/Utilities/mockoptimizer.jl
```diff
@@ -300,6 +300,11 @@ function mock_primal!(mock::MockOptimizer, conduals::Pair...)
     mock.hasprimal = false
     mock_dual!(mock, MOI.InfeasibilityCertificate, conduals...)
 end
+function mock_primal!(mock::MockOptimizer, primstatus::MOI.ResultStatusCode, conduals::Pair...)
```
mlubin: Where is this method called from?

blegat: From test/contconic.jl lines 26, 32 and 58 (see the changes of this commit)

mlubin: It's not obvious to track how the call to `mock_optimize` gets routed here. Do there really need to 4 versions of `mock_primal!` with overlapping signatures?

blegat: Is it better with c00a217 ?

src/Utilities/mockoptimizer.jl
```diff
@@ -286,21 +286,39 @@ end
...
-    MOI.set!(mock, MOI.VariablePrimal(), MOI.get(mock, MOI.ListOfVariableIndices()), varprim)
-    mock_dual!(mock, dual...)
-end
-function mock_primal!(mock::MockOptimizer, varprim::Vector, dual...)
-    # Feasible primal solution
-    mock_primal!(mock, MOI.FeasiblePoint, varprim, dual...)
+    mock_varprimal!(mock, varprimdual...)
 end
+# Shortcut for feasible primal solution
+mock_primal!(mock::MockOptimizer, varprim::Vector, dual...) = mock_primal!(mock, MOI.FeasiblePoint, varprim, dual...)
```
mlubin: Why is there a `dual` argument to `mock_primal!`?

src/Utilities/mockoptimizer.jl
```diff
@@ -286,21 +286,39 @@ end
...
 function mock_primal!(mock::MockOptimizer, conduals::Pair...)
     # No primal solution
     mock.hasprimal = false
+    mock_varprimal!(mock, conduals...)
+end
+
+# Sets variable primal to varprim
+function mock_varprimal!(mock::MockOptimizer, varprim::Vector, dual...)
+    MOI.set!(mock, MOI.VariablePrimal(), MOI.get(mock, MOI.ListOfVariableIndices()), varprim)
+    mock_dual!(mock, dual...)
```
mlubin: Why does `mock_varprimal!` call `mock_dual!` which calls `mock_condual!?

mlubin: The reader shouldn't need to go through 6 levels of recursion to understand all of the properties that are being set.

blegat:
I have split primal and dual in 78e464b, I am merging but we can continue the discussion on the PR for contlinear tests.

mlubin:
It's much better now that `mock_primal!` doesn't call `mock_dual!`.

PR #273, opened 2018-03-02
Bad arguments shouldn't be got since canset(dest, ...) returns false

blegat:


src/Test/modellike.jl
```diff
@@ -142,19 +142,16 @@ MOI.get(src::BadModel, ::MOI.ConstraintSet, ::MOI.ConstraintIndex{MOI.SingleVari
 struct BadModelAttribute <: MOI.AbstractModelAttribute end
 struct BadModelAttributeModel <: BadModel end
 MOI.canget(src::BadModelAttributeModel, ::BadModelAttribute) = true
-MOI.get(src::BadModelAttributeModel, ::BadModelAttribute) = 0
```
mlubin: If `canget` returns true, why remove the implementation of `get`?

blegat: During `copy(dest, src::BadModelAttributeModel)`, `canget(src, ...)` will return `true` but `canset(dest, ...)` will return `false`. In this case, a correct implementation of `copy` shouldn't call `get(src, ...)` since the result will not be used as it won't do `set!(dest, ...)`.
If `get` is defined here, a bad implementation of `copy` would pass the test. If it is not defined, the bad implementation will get `UndefinedMethod`

mlubin: Could you add a comment with this explanation of why get is not defined?

blegat: Done in 69e43a1 :)

PR #275, opened 2018-03-02
Test supports and supportsconstraint

blegat:


mlubin:
This is moving in the opposite direction from what we discussed in https://github.com/JuliaOpt/MathOptInterface.jl/issues/224, but I guess it's ok for now.

blegat:
If we create a test just for `supports` and `supportsconstraint`, we risk that some solvers forget to use this test and it is difficult to test all cases just in one test.
I agree with the conclusion of #224 but we should still have basic checks in `contlinear`, `contconic`, ... tests like checking `canXXX` everytime before `XXX`, and `support...` before loading a model.

PR #279, opened 2018-03-11
add vector valued function and set example

joaquimg:


docs/src/apimanual.md
```diff
@@ -174,7 +174,26 @@ addconstraint!(model, SingleVariable(x[1]), GreaterThan(0.0))
 addconstraint!(model, SingleVariable(x[2]), GreaterThan(-1.0))
 ```
 
-[Example with vector-valued set.]
+Besides scalar valued functions in scalar valued sets it possible to use vector valued functions and sets.
```
mlubin: scalar-valued and vector-valued

docs/src/apimanual.md
```diff
@@ -174,7 +174,26 @@ addconstraint!(model, SingleVariable(x[1]), GreaterThan(0.0))
...
+
+The code example below encodes the linear optimization problem:
```
mlubin: It may be confusing to call this a linear optimization problem because it has a nonlinear (convex) constraint.

PR #281, opened 2018-03-11
Rename Section

joaquimg:
This Section even starts describing constraints.
Should we add and intermediary section about sets?


docs/src/apimanual.md
```diff
@@ -140,7 +140,7 @@ sets the objective to the function just discussed in the minimization sense.
 
 See [Functions and function modifications](@ref) for the complete list of functions.
 
-## Sets
+## Constraints
```
mlubin: Maybe Sets and Constraints?

PR #290, opened 2018-03-20
Check that optimizer is empty in constructor

blegat:
The optimizer and model need to be empty in order to be in sync at the start.
This has the additional advantage that if solvers use this constructor for testing, they will be tested with `isempty` so that they will also work when they will be used with `resetsolver!` which calls `isempty` (issue raised by @joaquimg on Gitter).

### DIFF
--- a/test/cachingoptimizer.jl
+++ b/test/cachingoptimizer.jl
@@ -163,22 +163,25 @@ end
 end
 
 @testset "CachingOptimizer constructor with optimizer" begin
-    s = MOIU.MockOptimizer(ModelForMock{Float64}())
-    model = ModelForCachingOptimizer{Float64}()
     @testset "Empty model and optimizer" begin
+        s = MOIU.MockOptimizer(ModelForMock{Float64}())
+        model = ModelForCachingOptimizer{Float64}()
         m = MOIU.CachingOptimizer(model, s)
         @test MOI.isempty(m)
         @test MOIU.state(m) == MOIU.AttachedOptimizer
         @test MOIU.mode(m) == MOIU.Automatic
     end
     @testset "Non-empty optimizer" begin
+        s = MOIU.MockOptimizer(ModelForMock{Float64}())
         MOI.addvariable!(s)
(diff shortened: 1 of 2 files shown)


test/cachingoptimizer.jl
```diff
@@ -164,10 +164,20 @@ end
 
 @testset "CachingOptimizer constructor with optimizer" begin
     s = MOIU.MockOptimizer(ModelForMock{Float64}())
-    m = MOIU.CachingOptimizer(ModelForCachingOptimizer{Float64}(), s)
+    model = ModelForCachingOptimizer{Float64}()
+    m = MOIU.CachingOptimizer(model, s)
     @test MOI.isempty(m)
     @test MOIU.state(m) == MOIU.AttachedOptimizer
     @test MOIU.mode(m) == MOIU.Automatic
+    MOI.addvariable!(s)
```
mlubin: Please split into separate tests. We should try to structure tests as:
1. Setup
2. Do something
3. `@test` that something did what we expected

This makes it easier to understand what the test is checking for and how to address it if there's a failure.

blegat: Done :)

test/cachingoptimizer.jl
```diff
@@ -164,10 +164,26 @@ end
...
-    @test MOIU.mode(m) == MOIU.Automatic
+    model = ModelForCachingOptimizer{Float64}()
+    @testset "Empty model and optimizer" begin
+        m = MOIU.CachingOptimizer(model, s)
+        @test MOI.isempty(m)
+        @test MOIU.state(m) == MOIU.AttachedOptimizer
+        @test MOIU.mode(m) == MOIU.Automatic
+    end
+    @testset "Non-empty optimizer" begin
+        MOI.addvariable!(s)
```
mlubin: Tests shouldn't really share state like this, it defeats the readability point. It's better to regenerate `s` and `model`.

blegat: Done :)

Issue #211, opened 2018-02-09
Use `Pair{VariableIndex, T}` instead of separate variable and coefficient lists


odow:
Proposed by @tkoolen: In ScalarAffineFunction, why not use a Vector{Pair{VariableIndex, T}}, so as to ensure that you don’t have more coeffs than vars? Same thing for some of the other functions.

Ref #209

joaquimg:
I don’t like Pair. I think it might be interesting having those separate vectors for inplace modifications easily. We can enforce size check on the constructor if thats a big thing

tkoolen:
But then other methods can still push different numbers of things to each of the vectors.

mlubin:
I agree that this is a good idea from the software engineering perspective. I would use a struct instead of a Pair (Pair doesn't generalize well to the quadratic case.)

One concern is that MOI functions usually go through [`mapvariables`](https://github.com/JuliaOpt/MathOptInterfaceUtilities.jl/blob/1ea446c08f5d2f32d52b10974992beeb2670b12e/src/functions.jl#L5) at some point in their life. This change would force us to 1) make a more expensive copy that includes the coefficients, or 2) modify the functions in place instead.

joaquimg:
The current design is very easy to modify variable indices or the coefficients. There are cases like cut generation routines that always use the same variable indices and just changes the coefficients. Pair would make this a pain. I don´t see a big gain with that, this error is really easy to catch and solve.

tkoolen:
I agree that a `struct` other than `Pair` would be better.

> There are cases like cut generation routines that always use the same variable indices and just changes the coefficients. Pair would make this a pain.

Is that something a couple of utility functions could fix? Something like
```julia
struct LinearTerm{T}
    variable::VariableIndex
    coefficient::T
end

setvariable(term::LinearTerm, var::VariableIndex) = LinearTerm(variable, term.coefficient)
setcoefficient(term::LinearTerm, coefficient) = LinearTerm(term.variable, coefficient)
```
> this error is really easy to catch and solve.

I disagree. It's really tempting to write code like

```julia
fun = ScalarAffineFunction([VariableIndex(1), VariableIndex(2)], [5.0, -2.3, 3.0], 1.0) # oops

for i in eachindex(fun.variables)
    fun.variables[i] = VariableIndex(3 * i)
    fun.coefficients[i] += 1
end # last coefficient wasn't updated

# and then in some completely different solver package:
for coeff in fun.coefficients
[...]

PR #293, opened 2018-04-01
switch to arrays of structs in the NLP interface

mlubin:
I'd like to make this change throughout MOI (#211). NLP is the easiest since it doesn't break much yet. I have some preliminary NLP work that I'll push after this.

@odow @ccoffrin @tkoolen @blegat

### DIFF
--- a/docs/src/apireference.md
+++ b/docs/src/apireference.md
@@ -261,6 +261,7 @@ canmodifyobjective
 
 ```@docs
 NLPBlock
+NLPBoundsPair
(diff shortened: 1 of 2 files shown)


mlubin:
Good point, replaced `IndexPair` with `Tuple{Int64, Int64}`.

blegat:
Why `Int64` and not simply `Int` ?

mlubin:
We use `Int64` for `VariableIndex` as well.

PR #297, opened 2018-04-05
Add copynames argument to copy

### LINKED ISSUE #291: Add option to copy! to silently ignore failing copy of names
When a solver is used through a caching optimizer, the support of names in the optimizer is not mandatory as names are already handled by the model. Hence when copying the model to the optimizer, we should allow the optimizer not to support names.

blegat:
Closes #291

### DIFF
--- a/src/Utilities/cachingoptimizer.jl
+++ b/src/Utilities/cachingoptimizer.jl
@@ -117,7 +117,7 @@ from the model cache to the optimizer failed.
 function attachoptimizer!(m::CachingOptimizer)
     @assert m.state == EmptyOptimizer
     # We do not need to copy names because name-related operations are handled by `m.model_cache`
-    copy_result = MOI.copy!(m.optimizer, m.model_cache; copynames=false)
+    copy_result = MOI.copy!(m.optimizer, m.model_cache, copynames=false)
     if copy_result.status != MOI.CopySuccess
         return copy_result
     end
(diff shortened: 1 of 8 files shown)


src/Utilities/cachingoptimizer.jl
```diff
@@ -116,7 +116,8 @@ from the model cache to the optimizer failed.
 """
 function attachoptimizer!(m::CachingOptimizer)
     @assert m.state == EmptyOptimizer
-    copy_result = MOI.copy!(m.optimizer, m.model_cache)
+    # We do not need to copy names because name-related operations are handled by `m.model_cache`
+    copy_result = MOI.copy!(m.optimizer, m.model_cache; copynames=false)
```
mlubin: Semicolon can be comma here also

src/Utilities/copy.jl
```diff
@@ -87,12 +87,16 @@ function copyconstraints!(dest::MOI.ModelLike, src::MOI.ModelLike, idxmap::Index
...
     end
 
-    return passattributes!(dest, src, idxmap, cis_src)
+    return passattributes!(dest, src, copynames, idxmap, cis_src)
 end
 
 attribute_value_map(idxmap, f::MOI.AbstractFunction) = mapvariables(idxmap, f)
 attribute_value_map(idxmap, attribute_value) = attribute_value
 function defaultcopy!(dest::MOI.ModelLike, src::MOI.ModelLike)
+    warn("defaultcopy!(dest, src) is deprecated, use defaultcopy!(dest, src, false) instead or defaultcopy!(dest, src, true) if you do not want to copy names.")
```
mlubin: Are `false` and `true` swapped here? `true` is the default, and you use `false` if you do not want to copy names

blegat: Oups, indeed

PR #304, opened 2018-04-09
Implements ListOfAttributesSet for UniversalFallback

### LINKED ISSUE #303: UniversalFallback should implement ListOfAttributesSet
... otherwise all attributes are silently dropped when copying. I ran into this with `NLPBlock`.

blegat:
Closes #303

### DIFF
--- a/src/Utilities/universalfallback.jl
+++ b/src/Utilities/universalfallback.jl
@@ -76,6 +76,24 @@ function MOI.get(uf::UniversalFallback, attr::Union{MOI.AbstractVariableAttribut
     end
 end
 
+_dict(uf, attr::MOI.ListOfOptimizerAttributesSet)  = uf.optattr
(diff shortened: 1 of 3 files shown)


src/Utilities/universalfallback.jl
```diff
@@ -76,6 +76,24 @@ function MOI.get(uf::UniversalFallback, attr::Union{MOI.AbstractVariableAttribut
     end
 end
 
+_dict(uf, attr::MOI.ListOfOptimizerAttributesSet)  = uf.optattr
```
mlubin: I don't find the `_dict`/`_args` indirection very helpful. Someone is going to be unnecessarily confused when reading this code. (E.g.Why are we splatting empty tuples?) I would just a standalone implementation for each of the ListOfXXXSet attributes.

PR #319, opened 2018-04-18
Add Bridges submodule

### LINKED ISSUE #313: Move MOIBridges.jl to this package
The vote of https://github.com/JuliaOpt/MathOptInterface.jl/issues/173 was against moving MOIBridges to MOI but things have changed now as MOITest and MOIUtilities have been moved to MOI.
What do you prefer ?
* Move it to MOI under a Bridges submodule :+1: 
* Move it to MOI under a Utilities.Bridges submodule :tada: 
* Leave it separate (and maybe rename it to MathOptBridges) :-1: 

It opens the question on what we should include in this package as submodule instead of separate packages
1) Modules needed to test MOI (Test, Utilities)
2) Modules defining new MOI layers needed by JuMP (Bridges)
3) Modules defining new MOI layers (...)

For now we only include the first category and Bridges has been kept in separate packages. Including Bridges in this package means we also want to include the second category. I think it is important to maintain the third category out of MOI to promote a decentralized ecosystem. We don't want to send the message that MOI layers defined as MOI submodules are "better" that MOI layers defined outside.

blegat:
Closes #313

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -12,6 +12,7 @@ v0.2.0 (prerelease)
 - Addition of NLP tests.
 - Introduction of `UniversalFallback`.
 - `copynames` keyword argument to `MOI.copy!`.
+- Add Bridges submodule.
 
 
 v0.1.0 (February 28, 2018)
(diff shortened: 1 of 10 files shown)


mlubin:
These are all the bridges, right?

blegat:
Yes, they are all there :)

PR #320, opened 2018-04-18
Add guideline in doc

blegat:


docs/src/apimanual.md
```diff
@@ -565,4 +565,39 @@ If `canaddconstraint` returns `false`, then calling `addconstraint!` must throw
...
+The skeleton below can be used for the wrapper test file of a solver name `FooBar`:
+```julia
+using MathOptInterface
+const MOI = MathOptInterface
+const MOIT = MOI.Test
+const MOIU = MOI.Utilities
+const MOIB = MOI.Bridges
+
+# Include here the functions/sets supported by the solver wrapper (not those that are supported through bridges)
+MOIU.@model FooBarModelData () (EqualTo, GreaterThan, LessThan) (Zeros, Nonnegatives, Nonpositives) () (SingleVariable,) (ScalarAffineFunction,) (VectorOfVariables,) (VectorAffineFunction,)
```
mlubin: Could the UniversalFallback work here? There's no performance concerns for testing.

docs/src/apimanual.md
```diff
@@ -565,4 +565,39 @@ If `canaddconstraint` returns `false`, then calling `addconstraint!` must throw
...
+
+MOIB.@bridge SplitInterval MOIB.SplitIntervalBridge () (Interval,) () () () (ScalarAffineFunction,) () ()
```
mlubin: Why do users need to see this macro call? Can't MOIB just export a SplitInterval type? The macros calls in this code snippet are enough to scare potential solver authors away. (Only partially kidding.)

docs/src/apimanual.md
blegat: Universal Fallback still requires an internal model so we need to create a model anyway. Doing `MOIU.@model FooBarModelData () () () () () () () ()` and then using UniversalFallback would be a bit obscure. Besides, UniversalFallback only do attribute fallback for now :)

docs/src/apimanual.md
blegat: We could define all bridges in MOIB by default yes, I initially thought that it wasn't necessary and would increase the loading time of MOIB but it could be reconsidered

docs/src/apimanual.md
```diff
@@ -565,4 +565,39 @@ If `canaddconstraint` returns `false`, then calling `addconstraint!` must throw
...
+using MathOptInterface
+const MOI = MathOptInterface
+const MOIT = MOI.Test
+const MOIB = MOI.Bridges
+
+const optimizer = FooBarOptimizer()
+const config = MOIT.TestConfig(atol=1e-6, rtol=1e-6)
+
+@testset "MOI Continuous Linear" begin
+    MOIT.contlineartest(MOIB.SplitInterval{Float64}(optimizer), config)
```
mlubin: This would be a good place to explain why `SplitInterval` is used.

PR #324, opened 2018-04-20
Add test_delete_bridge function

blegat:


test/bridge.jl
```diff
@@ -76,6 +76,26 @@ MOIB.@bridge RSOCtoPSD MOIB.RSOCtoPSDCBridge () () (RotatedSecondOrderCone,) ()
...
+        test_noc(m, noc...)
+    end
+    @test MOI.isvalid(m, ci)
+    @test MOI.candelete(m, ci)
+    MOI.delete!(m, ci)
+    @test !MOI.isvalid(m, ci)
+    @test isempty(m.bridges)
+    test_noc(m, F, S, 0)
+    # As the bridge has been removed, if the constraints it has created where not removed, it wouldn't be there to decrease this counter anymore
+    @test MOI.get(m, MOI.NumberOfVariables()) == nvars
```
mlubin: Do bridges change the value of `MOI.NumberOfVariables()` that's visible to users?

blegat: If it adds variables as part of the reformulation, it needs to hide it from the user

mlubin: Good, just double checking!

PR #326, opened 2018-04-21
Add RSOC bridge

blegat:
Closes https://github.com/JuliaOpt/MathOptInterfaceBridges.jl/issues/38

### DIFF
--- a/src/Bridges/rsocbridge.jl
+++ b/src/Bridges/rsocbridge.jl
@@ -1,0 +1,62 @@
+"""
+    RSOCBridge{T}
+
+The `RotatedSecondOrderCone` is `SecondOrderCone` representable; see [1, p. 104].
+Indeed, we have ``2tu = (t/√2 + u/√2)^2 - (t/√2 - u/√2)^2 = (t + u/2)^2 - (t - u/2)^2`` hence
+```math
+2tu \\ge || x ||_2^2
+```
+is equivalent to
+```math
+(t + u/2)^2 \\ge || x ||_2^2 + (t - u/2)^2.
+```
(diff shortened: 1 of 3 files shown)


src/Bridges/rsocbridge.jl
```diff
@@ -0,0 +1,62 @@
...
+Indeed, we have ``2tu = (t/√2 + u/√2)^2 - (t/√2 - u/√2)^2 = (t + u/2)^2 - (t - u/2)^2`` hence
+```math
+2tu \\ge || x ||_2^2
+```
+is equivalent to
+```math
+(t + u/2)^2 \\ge || x ||_2^2 + (t - u/2)^2.
+```
+We can therefore use the transformation ``(t, u, x) \\mapsto (t+u/2, t-u/2, x)``.
+The transformation ``(t, u, x) \\mapsto (t/\\sqrt{2}+u/\\sqrt{2}, t/\\sqrt{2}-u/\\sqrt{2}, x)``
```
mlubin: Hmm. Why is preserving rationality of the input important? For preserving the user's intentions regarding feasibility tolerances, it would be nice to use the form of the transformation that preserves the norm of the vectors.
@chriscoey

blegat: This allows solver with rational input to use this bridge. We could also have both bridge or have a parameter for choosing the transformation or have the `√2` bridge and wait for a solver with rational input to exist before adding this one.

mlubin: Given that we don't have any solvers with rational input yet, let's choose the version that's more appropriate for existing solvers. I'm also questioning the motivation for a rational SOCP solver given that the optimal solution could be irrational even with rational input data.

PR #329, opened 2018-04-27
RFC: unit tests

odow:
This PR adds some atomic tests that aim to minimally test each expected feature in addition to the full end-to-end tests in `contlinear.jl` etc.

It has already found a few bugs in LinQuadOptInterface that weren't following the API to the letter.

### Bug fixes to existing tests
- If we delete the variable as it was, we can't add the constraints.

### Questions
<s>- Should we enforce a default objective sense?</s> Raised as separate issue #335 

### Other

Closes #244


src/Test/atomic_tests.jl
```diff
@@ -0,0 +1,244 @@
+"""
+    This function tests adding a single variable.
+"""
+function add_variable(model::MOI.ModelLike, config::TestConfig)
```
blegat: It would be more consistent with the rest of the tests to call it `addvariabletest`

blegat:
>  Should we enforce a default objective sense?

This is related to the question : Should a solver error if `optimize!` is called and `ObjectiveSense` has not been set or assume it is `FeasibilitySense` ?

odow:
I guess we can just leave each solver to have a default. Otherwise despite the backend having a default sense, every MOIWrapper will have to store a flag indicating whether the sense has been set.

odow:
I had to exclude a few of the atomic tests since MockOptimizer doesn't implement names.

blegat:
> I had to exclude a few of the atomic tests since MockOptimizer doesn't implement names.

Should be fixed in https://github.com/JuliaOpt/MathOptInterface.jl/pull/336 :)

src/Test/AtomicTests/constraints.jl
```diff
@@ -0,0 +1,24 @@
+"""
+    Test getting constraints by name.
+"""
+function getconstraint(model::MOI.ModelLike, config::TestConfig)
+    MOI.empty!(model)
```
mlubin: It's weird that we have to call `MOI.empty!` at the beginning of every test. This should happen somewhere else.

src/Test/AtomicTests/objectives.jl
```diff
@@ -0,0 +1,105 @@
...
+
+"""
+    Test constant in objective.
+"""
+function constant_obj(model::MOI.ModelLike, config::TestConfig)
+    atol, rtol = config.atol, config.rtol
+    MOI.empty!(model)
+    @test MOI.isempty(model)
+    MOIU.loadfromstring!(model,"""
+        variables: x
```
mlubin: This is a great use of `loadfromstring!`

src/Test/AtomicTests/objectives.jl
```diff
@@ -0,0 +1,105 @@
...
+    MOI.set!(model, MOI.ObjectiveSense(), MOI.FeasibilitySense)
+    @test MOI.canget(model, MOI.ObjectiveSense())
+    @test MOI.get(model, MOI.ObjectiveSense()) == MOI.FeasibilitySense
+end
+atomictests["feasibility_sense"] = feasibility_sense
+
+"""
+    Test constant in objective.
+"""
+function constant_obj(model::MOI.ModelLike, config::TestConfig)
```
mlubin: If the test calls `optimize!`, I'd propose to have "solve" somewhere in the name, like `solve_constant_obj`. You could easily think that `constant_obj` would just set and get the objective attribute without solving.

src/Test/AtomicTests/variables.jl
```diff
@@ -0,0 +1,178 @@
...
+    MOI.set!(model, MOI.VariableName(), v, "y")
+    @test MOI.get(model, MOI.VariableName(), v) == "y"
+    x = MOI.addvariable!(model)
+    MOI.set!(model, MOI.VariableName(), x, "x")
+    @test MOI.get(model, MOI.VariableName(), x) == "x"
+end
+atomictests["variablenames"]    = variablenames
+
+"""
+    Test the setting of an upper bound
```
mlubin: Another example where "solve" in the name would be informative, e.g., `solve_with_upperbound`.

src/Test/AtomicTests/constraints.jl
odow: I agree. Maybe it should go here?
https://github.com/JuliaOpt/MathOptInterface.jl/blob/84c5b19fa7cea292b715f6e02658bf977db08134/src/Test/config.jl#L42-L43

That can be a job for a different PR.

test/Test/unit.jl
```diff
@@ -0,0 +1,14 @@
...
+    mock = MOIU.MockOptimizer(Model{Float64}())
+    config = MOIT.TestConfig()
+    MOIT.unittest(mock, config, [
+        "solve_blank_obj",
+        "solve_constant_obj",
+        "solve_singlevariable_obj",
+        "solve_with_lowerbound",
+        "solve_with_upperbound",
+        "getconstraint",
+        "getvariable",
```
blegat: `getconstraint`, `getvariable` and `variablenames` should work now that https://github.com/JuliaOpt/MathOptInterface.jl/pull/336 is merged.
You can test the `solve` tests

odow: I'm not sure that the mock optimizer implements the names correctly. I got some errors. I'm going to merge this as is, the outstanding tests can be addressed in later PR's.

PR #337, opened 2018-04-29
Allow MOIWrapper folder

blegat:
Some wrapper may loose readability if everything is stuffed into a large file, what do you think ?


mlubin:
It's better to have some convention than none, but I expect we'll revisit this when Julia eventually adopts some conventions on file and directory naming.

PR #341, opened 2018-05-01
WIP: add some more unit tests

odow:


src/Test/UnitTests/constraints.jl
```diff
@@ -22,3 +22,143 @@ function getconstraint(model::MOI.ModelLike, config::TestConfig)
...
+            variable_primal   = [(x, 0.5)],
+            constraint_primal = [(c, 1.0)],
+            constraint_dual   = [(c, 0.5)]
+        )
+    end
+end
+unittests["solve_affine_equalto"] = solve_affine_equalto
+
+"""
+    Add an ScalarAffineFunction-in-Interval constraint.
```
blegat: Why is it indented ? Why don't you have `solve_affine_interval(...)` indented and then the docstring unindented ?

src/Test/UnitTests/constraints.jl
```diff
@@ -22,3 +22,143 @@ function getconstraint(model::MOI.ModelLike, config::TestConfig)
     @test MOI.isvalid(model, c2)
 end
 unittests["getconstraint"]    = getconstraint
+
+function testsolution(model, config;
```
mlubin: `test_model_solution`?

src/Test/UnitTests/constraints.jl
```diff
@@ -22,3 +22,143 @@ function getconstraint(model::MOI.ModelLike, config::TestConfig)
...
+        end
+    end
+    if constraint_primal != nothing
+        for (index, solution_value) in constraint_primal
+            @test MOI.canget(model, MOI.ConstraintPrimal(), typeof(index))
+            @test MOI.get(model, MOI.ConstraintPrimal(), index) ≈ solution_value atol=atol rtol=rtol
+        end
+    end
+    if constraint_dual != nothing
+        if config.duals
```
mlubin: It's worth documenting that the constraint dual isn't tested unless `config.duals == true`, because that's surprising.

src/Test/UnitTests/constraints.jl
mlubin: This method probably deserves a full docstring given how it's used everywhere.

test/Test/unit.jl
```diff
@@ -1,11 +1,118 @@
 @testset "Unit Tests" begin
     mock = MOIU.MockOptimizer(Model{Float64}())
+    mock.evalobjective = true
```
odow: @blegat I had to set this flag. Is it the correct way to do this? Should it be enabled by default???

blegat: I am in favor of this being the default. Currently, JuMP uses `evalobjective=false` so using `evalobjective=true` by default will force JuMP to either do `mock.evalobjective=false` or rely on automatic evalutation of the objective.
@mlubin Which one do you prefer ?

mlubin: We can make `mock.evalobjective == true` the default. Parameters like this should be passed to the constructor, no manually set though.

blegat: > Can't we just pass the objective value as an argument, otherwise it attempts to evaluate it?

In the constructor or in `mock_optimize!` ?

blegat: > We can make `mock.evalobjective == true` the default. Parameters like this should be passed to the constructor, no manually set though.

Yes, we could have keyword arguments

PR #342, opened 2018-05-01
Define dimension for AbstractScalarSet, add some tests.

tkoolen:


src/sets.jl
```diff
@@ -7,26 +7,30 @@ Abstract supertype for set objects used to encode constraints.
 """
 abstract type AbstractSet end
 
+"""
+    dimension(s::AbstractSet)
+
+Return the underlying dimension (number of vector components) in the set `s`, i.e.,
```
mlubin: "number of vector components" doesn't quite make sense anymore. Please clarify the definition.

tkoolen: Hmm, yeah, what should it be?

Typically, mathematical definitions of dimension refer to an intrinsic property of a space (e.g. a vector space). Dimension isn't defined for arbitrary sets. In addition, for `AbstractSet` there are cases like `PositiveSemidefiniteConeTriangle` and `PositiveSemidefiniteConeSquare` where the number we want to return is not the dimension of the vector space to which a given member of the set belongs: same mathematical object (a member of `$S^n_+$`), two different dimensions.

So I guess concrete `AbstractSet` subtypes implicitly define some particular vector space that is used to parameterize members of the set and the `AbstractSet`'s `dimension` is the dimension of that vector space. (Note by the way that the real line is in fact a vector space, and as such it has a well-defined dimension.)

So how about:

> The dimension of the vector space used to parameterize members of `s`.

blegat: Alternatively, it is equal to the dimension that an `AbstractFunction` should have to be used with the set. Then the dimension of an `AbstractFunction` is 1 for scalar function and the number of rows for a vector function.

src/sets.jl
```diff
@@ -184,7 +188,7 @@ dimension(s::Union{ExponentialCone, DualExponentialCone, PowerCone, DualPowerCon
 
 The (vectorized) cone of symmetric positive semidefinite matrices, with side length `dimension` and with off-diagonals unscaled.
 The entries of the upper triangular part of the matrix are given column by column (or equivalently, the entries of the lower triangular part are given row by row).
-An ``dimension \\times dimension`` matrix has ``dimension(dimension+1)/2`` lower-triangular elements, so for the vectorized cone of dimension ``n``, the corresponding symmetric matrix has side dimension ``\\sqrt{1/4 + 2 n} - 1/2`` elements.
+An ``dimension \\times dimension`` matrix has ``dimension * (dimension+1)/2`` lower-triangular elements, so for the vectorized cone of dimension ``n``, the corresponding symmetric matrix has side dimension ``\\sqrt{1/4 + 2 n} - 1/2`` elements.
```
blegat: As we use double backquote, we are in LaTeX mode. In LaTeX, I prefer `\cdot` or `\times` or nothing over `*`. Here `\times` would be ambiguous with matrix dimensions so `\cdot` seems more appropriate

src/sets.jl
tkoolen: Whichever you want, just let me know and I'll change it.

blegat: I think the link with the function is simpler, the vector space might be ambiguous for cases like PSDCTriangle

PR #343, opened 2018-05-02
RFC: make ScalarAffineFunction and ScalarQuadraticFunction mutable.

tkoolen:
`ScalarAffineFunction` and `ScalarQuadraticFunction` are not `isbits` types due to their `Vector` fields, which means that construction necessarily allocates, and as such is not an option if zero allocation is a requirement.

The alternative, pre-allocating `ScalarAffineFunction` and `ScalarQuadraticFunction` objects and then modifying them, is currently almost a solution, save for the `constant` fields, which cannot be modified. Making these types mutable fully enables this second workflow.

An alternative could be to store the constants in `Base.RefValue{T}` fields, but for little gain, and simply making these types mutable seems like the least amount of hassle.

### DIFF
--- a/src/functions.jl
+++ b/src/functions.jl
@@ -56,6 +56,10 @@ struct ScalarAffineTerm{T}
     variable_index::VariableIndex
 end
 
+# Note: ScalarAffineFunction is mutable because its `constant` field is likely of an immutable
+# type, while its `terms` field is of a mutable type, meaning that creating a `ScalarAffineFunction`
+# allocates, and it is desirable to provide a zero-allocation option for working with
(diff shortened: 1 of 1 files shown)


mlubin:
Makes sense but we need to be consistent. Should all non-`isbits` types be mutable?

tkoolen:
That's one option. Alternatively, a rule like 

> An MOI type should be mutable if and only if it has both `isbits`  and non-`isbits` fields when all type parameters are `isbits` types.

could work. I guess `MultiRowChange` should be changed regardless.

mlubin:
Could you just add a comment above the docstring for `ScalarAffineFunction` explaining why it's mutable and pointing to this discussion?

tkoolen:
Updated.

PR #345, opened 2018-05-02
Fix Model names using unit tests

blegat:
Unit tests allowed to find a few bugs in the names implementation of `MOIU.@model`


odow:
Hurrah for better tests

PR #348, opened 2018-05-04
Disallow non-optimizer in bridge

blegat:
Since `AbstractBridgeOptimizer <: MOI.AbstractOptimizer`, the inner model should be an optimizer. Otherwise, `optimize!` will not work. It may seem restrictive to prevent models that are not optimizer to use bridges but they can always use a mockoptimizer to turn into an optimizer.

### DIFF
--- a/src/Bridges/bridge.jl
+++ b/src/Bridges/bridge.jl
@@ -3,9 +3,9 @@ export @bridge
 """
     AbstractBridge
 
-A bridge represents a bridged constraint in an `AbstractBridgeOptimizer`. It contains the indices of the constraints that it has created in the model.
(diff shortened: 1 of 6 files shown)


mlubin:
This seems like a lot of churn and forcing extra complexity without a particularly good reason. There's nothing about `MOI.AbstractOptimizer` that says that `optimize!` can't fail with an error under some conditions that depend on the optimizer's internal state.

blegat:
I am ok to leave it as is if you don't feel that it violates the MOI definition :)

mlubin:
It doesn't violate the definition. `optimize!` is allowed to return an error.

PR #351, opened 2018-05-05
redefine MOI functions to use arrays of structs instead of structs of arrays

### LINKED ISSUE #211: Use `Pair{VariableIndex, T}` instead of separate variable and coefficient lists

mlubin:
Closes #211 

Please comment and double check that this is what we want. Overall I think the new definitions:
- are much clearer.
- reflect better coding practices (no more need to verify that `length(indices) == length(coefficients)`).
- could lead to faster code (by better cache performance when looping through terms).

After an agreement on the new definitions, I'll need help updating *everything* (by PRs to this branch first, then in the solver wrappers). We can tag this as MOI 0.3.

@tkoolen @odow @blegat @joaquimg @joehuchette @ccoffrin


ccoffrin:
Without any context, inclined to agree.  This pattern frequently pops up.

tkoolen:
Just wondering: why did you decide to use `Tuple`s instead of separate `struct`s? Ref: https://github.com/JuliaOpt/MathOptInterface.jl/issues/211#issuecomment-364451858; I understand you didn't want to use both `Pair`s and `Tuple`s, but are `struct`s with explicit field names not just easier to use/more self-documenting than `Tuple`s?

mlubin:
I convinced myself that structs make sense for `ScalarAffineTerm` and `ScalarQuadraticTerm` for the self-documenting feature and because they're used in multiple places. For `MultirowChange`, it seems like overkill to define a struct for the `(output_index, coefficient)` tuple that's only used in one context.

src/functions.jl
```diff
@@ -43,77 +43,140 @@ struct VectorOfVariables <: AbstractVectorFunction
...
+    variable_index::VariableIndex
+end
+
+"""
+    ScalarAffineFunction{T}(terms, constant)
 
 The scalar-valued affine function ``a^T x + b``, where:
-* ``a`` is a sparse vector specified in tuple form by `variables::Vector{VariableIndex}` and `coefficients::Vector{T}`
+* ``a`` is a sparse vector specified by a list of
+  [`ScalarAffineTerm`](@ref ScalarAffineTerm) structs.
```
blegat: Isn't
```julia
[`ScalarAffineTerm`](@ref)
```
sufficient ?

src/functions.jl
```diff
@@ -43,77 +43,140 @@ struct VectorOfVariables <: AbstractVectorFunction
...
     constant::T
 end
 
+"""
+    struct VectorQuadraticTerm{T}
+        output_index::Int64
+        scalar_term::ScalarQuadraticTerm{T}
+    end
 
+A `ScalarQuadraticTerm` plus its index of the output component of a
```
blegat: Could add a `@ref` for ScalarQuadraticTerm

src/functions.jl
mlubin: It's a bit awkward to say
```
[`ScalarAffineTerm`](@ref)s
```
(with the s).

mlubin: I realize that this rule isn't applied consistently though.

blegat: There is currently no "s", is there ?

mlubin: "a list of ScalarAffineTerm" is not really grammatical. It should be "a list of ScalarAffineTerms" or "a list of ScalarAffineTerm structs".

mlubin: Oops, my bad for misunderstanding. Yes, I'll change this.

blegat:
What remains to be done

- [ ] Uncomment and fix parser tests
- [ ] Uncomment and fix unit tests
- [ ] Uncomment and fix bridge tests
- [ ] https://github.com/JuliaOpt/MathOptInterface.jl/pull/363#discussion_r189255529
- [ ] https://github.com/JuliaOpt/MathOptInterface.jl/pull/363#discussion_r189269518

blegat:
I have rebased it and fixed basic constraints tests in 2bd77d4

mlubin:
Thanks! Let's wait until we have a PR on the JuMP side and merge both at the same time, given that https://github.com/JuliaOpt/JuMP.jl/pull/1305 is blocked.

PR #352, opened 2018-05-07
[array_of_tuples] Add isbits test

blegat:


mlubin:
What is the motivation for testing `isbits`? This follows from the definition of the structs, and it doesn't add any extra coverage.

blegat:
> This follows from the definition of the structs

If someone (e.g. external contributor) want to modify the structs (e.g. by making it mutable), it might not be `isbits` anymore and the tests will indicate that these should be isbits

mlubin:
> If someone (e.g. external contributor) want to modify the structs (e.g. by making it mutable), it might not be isbits anymore and the tests will indicate that these should be isbits

I'm not worried about accidentally merging a PR that changes `struct` to `mutable struct`. If it's *really* that important, this would be better addressed by adding a comment in the source explaining why we use `struct` instead of `mutable struct`. (I'm not convinced this is needed.)

mlubin:
Another way to convince me would be to add an explanation to the `isbits` test explaining why we test `isbits` for only these structs and not the others defined in MOI. The explanation could include a TODO.

PR #353, opened 2018-05-08
Split bridge file in two

blegat:
It is cleaner to separate `AbstractBridge` from `AbstractBridgeOptimizer`


mlubin:
Note the travis failure.

PR #354, opened 2018-05-11
RFC: Comprehensive testing of basic constraint functionality

odow:
### Background

At the moment it is pretty hard from the solver's point-of-view to check that they are implementing the full MOI API correctly (e.g., LQOI). This PR attempts to test this.

For every constraint type supported by the solver, it tests the basic usage of the MOI API. It does not test corner cases. That can come later. In most cases, checking that the solver implemented something correctly will involve forming the problem, modifying, solving, and then checking the solution. This PR just tries to ensure that the methods are implemented.

It has exposed a number of bugs an un-implemented stuff in LQOI (ref https://github.com/JuliaOpt/LinQuadOptInterface.jl/pull/10).

See [the tests in Gurobi.jl](https://github.com/JuliaOpt/Gurobi.jl/blob/2033629e3a67d72cbfb99ac68fbc417f9c7d2b8e/test/MOIWrapper.jl#L10-L26) for an example of how it can be called.

### What this does

N.B. some unchecked things still todo.

In particular, it adds tests for basic usage of:
- [x] supportsconstraint
- [x] canget/get `NumberOfConstraints{F,S}()`
- [x] canaddconstraint
- [x] addconstraint!
- [x] addconstraints!
- [x] canget/get `ListOfConstraintIndices{F,S}()`
- [x] isvalid
- [x] candelete/delete!
- [x] get `ConstraintFunction`
- [x] get `ConstraintSet`
- [ ] canmodify/modify
  - [ ] modify set
  - [ ] modify function
  - [ ] [scalarconstant](http://www.juliaopt.org/MathOptInterface.jl/latest/apireference.html#MathOptInterface.ScalarConstantChange)
  - [ ] [vectorconstant](http://www.juliaopt.org/MathOptInterface.jl/latest/apireference.html#MathOptInterface.VectorConstantChange)
  - [ ] [scalarcoefficient](http://www.juliaopt.org/MathOptInterface.jl/latest/apireference.html#MathOptInterface.ScalarCoefficientChange)
  - [ ] [multirow](http://www.juliaopt.org/MathOptInterface.jl/latest/apireference.html#MathOptInterface.MultirowChange)
- [ ] cantransform/transform
- [x] `ConstraintName`
- [ ] `ConstraintPrimalStart`
- [ ] `ConstraintDualStart`
- [ ] `ConstraintBasisStatus`
- [ ] `ConstraintDual`
- [ ] `ConstraintPrimal`

for the function-set combinations
- [x] `ScalarAffineFunction{Float64}`-in-{`LessThan, GreaterThan, EqualTo, Interval`}
- [x] `ScalarQuadraticFunction{Float64}`-in-{`LessThan, GreaterThan, EqualTo, Interval`}
- [x] `VectorAffineFunction{Float64}`-in-{`Zeros,Nonnegatives,Nonpositives,Reals`}
- [x] `VectorQuadraticFunction{Float64}`-in-{`Zeros,Nonnegatives,Nonpositives,Reals`}
- [x] `SingleVariable`-in-{`LessThan, GreaterThan, EqualTo, Interval`}
- [x] `SingleVariable`-in-{`ZeroOne, Integer, Semicontinuous, Semiinteger`}
- [x] `VectorOfVariables`-in-{`SOS1, SOS2, Zeros,Nonnegatives,Nonpositives,Reals`}
- [x] conic stuff
  - [x] `VectorOfVariables`-in-`SecondOrderCone`
  - [x] `VectorOfVariables`-in-`RotatedSecondOrderCone`
  - [x] `VectorOfVariables`-in-`GeometricMeanCone`
  - [x] `VectorOfVariables`-in-`ExponentialCone`
  - [x] `VectorOfVariables`-in-`DualExponentialCone`
  - [x] PowerCone
  - [x] DualPowerCone
  - [x] PositiveSemidefiniteConeTriangle
  - [x] PositiveSemidefiniteConeSquare
  - [x] LogDetConeTriangle
  - [x] LogDetConeSquare
  - [x] RootDetConeTriangle
  - [x] RootDetConeSquare

<s>Considering doing something like
```julia
const BasicConstraintTests = Dict(
    (MOI.SingleVariable, MOI.ZeroOne)        => test_singlevariable_in_zeroone
)

function basic_constraint_tests(model::MOI.ModelLike, config::TestConfig; delete::Bool=true)
    for ( (F,S), func ) in BasicConstraintTests
        if MOI.supportsconstraint(model, F, S)
            @testset "$(string(func))" begin
                func(model, config; delete=delete)
            end
        end
    end
end
unittests["basic_constraint_tests"] = basic_constraint_tests
```
But, we probably need a way for solvers to specify if they don't support deleting or querying a subset of constraint types.</s> I did something like this.

### DIFF
--- a/src/Test/UnitTests/basic_constraint_tests.jl
+++ b/src/Test/UnitTests/basic_constraint_tests.jl
@@ -1,5 +1,5 @@
 """
-    basic_constraint_tests(f::Function, model::MOI.ModelLike, config::TestConfig, set::MOI.AbstractSet, N::Int; delete::Bool=true)
+    test_basic_constraint_functionality(f::Function, model::MOI.ModelLike, config::TestConfig, set::MOI.AbstractSet, N::Int; delete::Bool=true)
 
 Test some basic constraint tests.
 
@@ -11,11 +11,11 @@ If `config.query=true`, test getting `ConstraintFunction` and `ConstraintSet`.
 
 ### Example
 
-    basic_constraint_tests(model, config, MOI.LessThan(1.0), 1; delete=false) do x
+    test_basic_constraint_functionality(model, config, MOI.LessThan(1.0), 1; delete=false) do x
         MOI.ScalarAffineFunction(model, [x], [1.0], 0.0)
     end
 """
-function basic_constraint_tests(f::Function, model::MOI.ModelLike, config::TestConfig, set::MOI.AbstractSet, N::Int=1; delete::Bool=true)
+function test_basic_constraint_functionality(f::Function, model::MOI.ModelLike, config::TestConfig, set::MOI.AbstractSet, N::Int=1; delete::Bool=true)
     MOI.empty!(model)
     x = MOI.addvariables!(model, N)
     constraint_function = f(x)
@@ -100,7 +100,7 @@ Basic tests for constraints of the form `ScalarAffineFunction{Float64}`-in-`Less
 If `delete=true`, test `MOI.candelete` and `MOI.delete!`.
 """
 function test_scalaraffine_in_lessthan(model::MOI.ModelLike, config::TestConfig; delete::Bool=true)
-    basic_constraint_tests(model, config, MOI.LessThan(1.0); delete=delete) do x
+    test_basic_constraint_functionality(model, config, MOI.LessThan(1.0); delete=delete) do x
         MOI.ScalarAffineFunction(x, [1.0], 0.0)
     end
 end
@@ -115,7 +115,7 @@ Basic tests for constraints of the form `ScalarAffineFunction{Float64}`-in-`Grea
 If `delete=true`, test `MOI.candelete` and `MOI.delete!`.
 """
 function test_scalaraffine_in_greaterthan(model::MOI.ModelLike, config::TestConfig; delete::Bool=true)
-    basic_constraint_tests(model, config, MOI.GreaterThan(1.0); delete=delete) do x
+    test_basic_constraint_functionality(model, config, MOI.GreaterThan(1.0); delete=delete) do x
         MOI.ScalarAffineFunction(x, [1.0], 0.0)
     end
 end
@@ -129,7 +129,7 @@ Basic tests for constraints of the form `ScalarAffineFunction{Float64}`-in-`Equa
 If `delete=true`, test `MOI.candelete` and `MOI.delete!`.
 """
 function test_scalaraffine_in_equalto(model::MOI.ModelLike, config::TestConfig; delete::Bool=true)
-    basic_constraint_tests(model, config, MOI.EqualTo(1.0); delete=delete) do x
+    test_basic_constraint_functionality(model, config, MOI.EqualTo(1.0); delete=delete) do x
         MOI.ScalarAffineFunction(x, [1.0], 0.0)
     end
 end
@@ -143,7 +143,7 @@ Basic tests for constraints of the form `ScalarAffineFunction{Float64}`-in-`Inte
 If `delete=true`, test `MOI.candelete` and `MOI.delete!`.
 """
 function test_scalaraffine_in_interval(model::MOI.ModelLike, config::TestConfig; delete::Bool=true)
-    basic_constraint_tests(model, config, MOI.Interval(0.0, 1.0); delete=delete) do x
+    test_basic_constraint_functionality(model, config, MOI.Interval(0.0, 1.0); delete=delete) do x
         MOI.ScalarAffineFunction(x, [1.0], 0.0)
     end
 end
@@ -157,7 +157,7 @@ Basic tests for constraints of the form `ScalarQuadraticFunction{Float64}`-in-`L
 If `delete=true`, test `MOI.candelete` and `MOI.delete!`.
 """
 function test_scalarquadratic_in_lessthan(model::MOI.ModelLike, config::TestConfig; delete::Bool=true)
-    basic_constraint_tests(model, config, MOI.LessThan(1.0); delete=delete) do x
+    test_basic_constraint_functionality(model, config, MOI.LessThan(1.0); delete=delete) do x
         MOI.ScalarQuadraticFunction(x, [1.0], x, x, [1.0], 0.0)
     end
 end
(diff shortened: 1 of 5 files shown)


src/Test/UnitTests/basic_constraint_tests.jl
```diff
@@ -0,0 +1,233 @@
...
+    (MOI.VectorOfVariables, MOI.DualExponentialCone)    => ( dummy_vectorofvariables, 3, MOI.DualExponentialCone() ),
+    (MOI.VectorOfVariables, MOI.PowerCone)              => ( dummy_vectorofvariables, 3, MOI.PowerCone(2.0) ),
+    (MOI.VectorOfVariables, MOI.DualPowerCone)          => ( dummy_vectorofvariables, 3, MOI.DualPowerCone(2.0) ),
+
+    (MOI.VectorOfVariables, MOI.PositiveSemidefiniteConeTriangle) => ( dummy_vectorofvariables, 6, MOI.PositiveSemidefiniteConeTriangle(3) ),
+    (MOI.VectorOfVariables, MOI.PositiveSemidefiniteConeSquare)   => ( dummy_vectorofvariables, 9, MOI.PositiveSemidefiniteConeSquare(3) ),
+    (MOI.VectorOfVariables, MOI.LogDetConeTriangle)               => ( dummy_vectorofvariables, 6, MOI.LogDetConeTriangle(3) ),
+    (MOI.VectorOfVariables, MOI.LogDetConeSquare)                 => ( dummy_vectorofvariables, 9, MOI.LogDetConeSquare(3) ),
+    (MOI.VectorOfVariables, MOI.RootDetConeTriangle)              => ( dummy_vectorofvariables, 6, MOI.RootDetConeTriangle(3) ),
+    (MOI.VectorOfVariables, MOI.RootDetConeSquare)                => ( dummy_vectorofvariables, 9, MOI.RootDetConeSquare(3) ),
```
odow: @blegat are these right? I'm not sure if I understood the docs correctly.

odow:
This is now RFC. It doesn't attempt to test each method thoroughly. It just tries to check that solvers have implemented the full API for all the `F`-in-`S` combinations they claim to support.

This helps avoid situations where you think you've implemented `delete!` for all constraints, only to find that you missed `ScalarAffine`-in-`Interval`. This was helpful for LQOI.

There are a few unchecked items that can probably be addressed in a new PR.

src/Test/UnitTests/basic_constraint_tests.jl
blegat: Almost, you should do +1 (6->7 and 9->10) for the Det-cones to account for the `t`

blegat: You could check that with the `dimension` function

odow: Maybe this should be clarified in the docs? It currently says:
"The argument `dimension` is the side dimension of the matrix `X`, i.e., its number of rows or columns."
I took that to mean excluding the `t`.

blegat: Yes, the `dimension` field excludes the `t` but the `dimension` function returns the size  that a vector set should have to belong to that set so it includes `t` and all element of the matrices hence it is `1 + d^2` for LogDetSquare where `d` is the field

odow: Oh right. I got confused. The docs are clear now that I read them properly.

src/Test/UnitTests/basic_constraint_tests.jl
```diff
@@ -0,0 +1,233 @@
...
+    delete                  = true,
+    get_constraint_function = true,
+    get_constraint_set      = true,
+    include                 = Any[],
+    exclude                 = Any[]
+    )
+    if length(include) > 0 && length(exclude) > 0
+        error("Don't use `include` and `exclude` together.")
+    end
+    test_keys = length(include) > 0 ? include : Iterators.filter(x->!(x in exclude), keys(BasicConstraintTests))
```
blegat: Why don't you do `test_keys = keys(BasicConstraintTests)` in the case where `include` and `exclude` are both empty. Since you do `if MOI.supportsconstraint` later, this would test all supported constraints and remove the need for a `ListOfSupportedConstraintTypes` discussed in https://github.com/JuliaOpt/MathOptInterface.jl/pull/357#issuecomment-388686469

src/Test/UnitTests/basic_constraint_tests.jl
```diff
@@ -0,0 +1,233 @@
...
+        end
+    end
+end
+
+# x
+const dummy_single_variable   = (x::Vector{MOI.VariableIndex}) -> MOI.SingleVariable(x[1])
+# x₁, x₂
+const dummy_vectorofvariables = (x::Vector{MOI.VariableIndex}) -> MOI.VectorOfVariables(x)
+# 1.0 * x
+const dummy_scalar_affine     = (x::Vector{MOI.VariableIndex}) -> MOI.ScalarAffineFunction(x, [1.0], 0.0)
```
blegat: [1.0] should be replaced by `ones(Float64, length(x))`

src/Test/UnitTests/basic_constraint_tests.jl
blegat: > This does test keys(BasicConstraintTests) if include and exclude are both empty.

Indeed ^^

> If we had ListOfSupportedConstraintTypes, we could throw a warning that we didn't test SingleVariable-in-LessThan{Int} for example.

The warning would be called when `include` is non-empty and it does not contain all supported sets ? This can already be done now by iterating through the keys of `BasicConstraintTests`

mlubin:
The naming of `test_basic_constraint_functionality` and `basic_constraint_tests` is a bit confusing. Are they both public functions that users should care about? They should have parallel names in this case. Is `test_basic_constraint_functionality` a helper for `basic_constraint_tests`? `do_x` is a confusing name for a helper for `x_do`.

odow:
Merged this as is. We can address the modification tests after #360 is addressed.

Issue #302, opened 2018-04-08
UniversalFallback should throw key error rather than return canget == false if attribute partially set


mlubin:
https://github.com/JuliaOpt/MathOptInterface.jl/blob/7164fdc573f3f92acb93de1bf18a3d3bd5e315c7/src/Utilities/universalfallback.jl#L54-L57

The check on the length of the dictionary means that:
```julia
m = Model()
@variable(m, x, start = 1)
@variable(m, y)
JuMP.startvalue(x)
```
gives
```
 Attribute MathOptInterface.VariablePrimalStart() not accessible
  Stacktrace:
   [1] get(::MathOptInterface.Utilities.CachingOptimizer, ::MathOptInterface.VariablePrimalStart, ::MathOptInterface.VariableIndex) at /home/mlubin/.julia/v0.6/MathOptInterface/src/Utilities/cachingoptimizer.jl:380
   [2] get(::JuMP.Model, ::MathOptInterface.VariablePrimalStart, ::JuMP.Variable) at /home/mlubin/.julia/v0.6/JuMP/src/JuMP.jl:379
   [3] startvalue(::JuMP.Variable) at /home/mlubin/.julia/v0.6/JuMP/src/variables.jl:379
```
The attribute values should be available immediately after you set them, and there should be a key error in the case that you ask for a value that's not defined.

blegat:
In the doc, we have say that `canget` returns `true` if it is

> applied to *every* variable

https://github.com/JuliaOpt/MathOptInterface.jl/blob/9f75cea0bc305d0628827b29427a329809b0f33d/src/attributes.jl#L141-L143

mlubin:
It's evident to me that this is a bad design. `Attribute MathOptInterface.VariablePrimalStart() not accessible` is an incorrect and misleading error message in this situation. If an attribute is partially set, the model is still responsible for that attribute, and `get` should be called. The model may choose to return an error if it can't access the attribute, and that's ok. We might need to brainstorm a bit on how to resolve this situation and #295 now that we have a bit more experience working with MOI.

blegat:
When `MOI.get` is called on CachingOptimizer, and `canget` is false for both the model and the optimizer, it could try to still call `get` on one of them as you suggest. It should try `get` on the one that supports this even it is not set for all variables. `supports` is probably the solution then.

[...]

Issue #295, opened 2018-04-02
supports vs canset


mlubin:
Why do we have both
```julia
MOI.canset(model, MOI.NLPBlock())
```
and
```julia
MOI.supports(model, MOI.NLPBlock())
```
?

@blegat

mlubin:
This is potentially more confusing than helpful for users and implementers of solver interfaces. If the optimizer `supports` an attribute but `canset` would be false, why not just throw an error explaining why the attribute is not settable right now?

blegat:
`supports` is the replacement of `cancopy`. It is useful for `copy`-only solvers as we can check programatically if a `copy` might fail. For solvers that support `addconstraint!` and `set!`, it might seem cumbersome to also implement `supports` and `supportsconstraint`. We might change the fallback to
```julia
supports(m::ModelLike, attr...) = canset(m, attr)
```
so that solvers for which canset is equivalent to supports only need to implement `canset`.

mlubin:
Could you provide an example where we need to distinguish between `supports` and `canset` programmatically? Default fallbacks help but don't really remove the additional complexity of having these two concepts.

blegat:
If a user adds a rotated second order cone constraints, we need to know whether the optimizer supports it to know whether we need to transform it into a second order cone constraint (e.g. using a bridge) or to leave the constraint as it is.

mlubin:
Why not use `supportsconstraint` instead of `canaddconstraint` in that context?

blegat:
That's exactly my point, we need `supportsconstraint` in this situation

blegat:
We need `canaddconstraint` for the CachingOptimizer to see whether the optimizer supports adding a constraint or it needs to go in `EmptyOptimizer` state
https://github.com/JuliaOpt/MathOptInterface.jl/blob/7164fdc573f3f92acb93de1bf18a3d3bd5e315c7/src/Utilities/cachingoptimizer.jl#L213-L215

mlubin:
I'm also proposing to replace that usage of `canaddconstraint` with `supportsconstraint`.

blegat:
But `supportsconstraint` does not help distinguishing whether the solver supports adding constraints or needs to be emptied and needs to be passed the whole problem using `copy!`.

mlubin:
What about `MOI.canaddconstraint(m.optimizer) && MOI.supportsconstraint(m.optimizer, typeof(func), typeof(set))` in this case?

[...]

PR #357, opened 2018-05-13
ListOfConstraints -> ListOfConstraintTypes

### LINKED ISSUE #214: rename ListOfConstraints?
https://github.com/JuliaOpt/MathOptInterface.jl/issues/203#issuecomment-364592475

@tkoolen

odow:
Closes #214 

I also propose changing
```
 A list of tuples of the form `(F,S)`, where `F` is a function type
 and `S` is a set type indicating that the attribute `NumberOfConstraints{F,S}()`
 has value greater than zero.
```
to 
```
 A list of tuples of the form `(F,S)`, where `F` is a function type
 and `S` is a set type indicating that `supportsconstraint(model, F, S)` is `true`.
```

Updated: changed `cansupport` to `supportsconstraint`.

### DIFF
--- a/docs/src/apireference.md
+++ b/docs/src/apireference.md
@@ -53,7 +53,7 @@ Name
 ObjectiveSense
 NumberOfVariables
 ListOfVariableIndices
-ListOfConstraints
+ListOfConstraintTypes
 NumberOfConstraints
 ListOfConstraintIndices
 ListOfModelAttributesSet
--- a/src/Bridges/bridgeoptimizer.jl
+++ b/src/Bridges/bridgeoptimizer.jl
@@ -42,9 +42,9 @@ function MOI.get(b::AbstractBridgeOptimizer, attr::Union{MOI.NumberOfConstraints
     end
     s
 end
-MOI.canget(b::AbstractBridgeOptimizer, attr::MOI.ListOfConstraints) = MOI.canget(b.model, attr) && MOI.canget(b.bridged, attr)
+MOI.canget(b::AbstractBridgeOptimizer, attr::MOI.ListOfConstraintTypes) = MOI.canget(b.model, attr) && MOI.canget(b.bridged, attr)
 _noc(b, fs) = MOI.get(b, MOI.NumberOfConstraints{fs...}())
-function MOI.get(b::AbstractBridgeOptimizer, attr::MOI.ListOfConstraints)
+function MOI.get(b::AbstractBridgeOptimizer, attr::MOI.ListOfConstraintTypes)
     loc = MOI.get(b.model, attr)
     rm = find(_noc.(b, loc) .== 0)
     deleteat!(loc, rm)
--- a/src/Test/contconic.jl
+++ b/src/Test/contconic.jl
@@ -38,7 +38,7 @@ function _lin1test(model::MOI.ModelLike, config::TestConfig, vecofvars::Bool)
     c = MOI.addconstraint!(model, MOI.VectorAffineFunction([1,1,1,2,2], [v;v[2];v[3]], ones(5), [-3.0,-2.0]), MOI.Zeros(2))
     @test MOI.get(model, MOI.NumberOfConstraints{vecofvars ? MOI.VectorOfVariables : MOI.VectorAffineFunction{Float64},MOI.Nonnegatives}()) == 1
     @test MOI.get(model, MOI.NumberOfConstraints{MOI.VectorAffineFunction{Float64},MOI.Zeros}()) == 1
-    loc = MOI.get(model, MOI.ListOfConstraints())
+    loc = MOI.get(model, MOI.ListOfConstraintTypes())
     @test length(loc) == 2
     @test (vecofvars ? MOI.VectorOfVariables : MOI.VectorAffineFunction{Float64},MOI.Nonnegatives) in loc
     @test (MOI.VectorAffineFunction{Float64},MOI.Zeros) in loc
@@ -357,7 +357,7 @@ function _soc1test(model::MOI.ModelLike, config::TestConfig, vecofvars::Bool)
 
     @test MOI.get(model, MOI.NumberOfConstraints{MOI.VectorAffineFunction{Float64},MOI.Zeros}()) == 1
     @test MOI.get(model, MOI.NumberOfConstraints{vecofvars ? MOI.VectorOfVariables : MOI.VectorAffineFunction{Float64},MOI.SecondOrderCone}()) == 1
-    loc = MOI.get(model, MOI.ListOfConstraints())
+    loc = MOI.get(model, MOI.ListOfConstraintTypes())
     @test length(loc) == 2
     @test (MOI.VectorAffineFunction{Float64},MOI.Zeros) in loc
     @test (vecofvars ? MOI.VectorOfVariables : MOI.VectorAffineFunction{Float64},MOI.SecondOrderCone) in loc
--- a/src/Test/intconic.jl
+++ b/src/Test/intconic.jl
@@ -33,7 +33,7 @@ function intsoc1test(model::MOI.ModelLike, config::TestConfig)
 
     @test MOI.get(model, MOI.NumberOfConstraints{MOI.VectorAffineFunction{Float64},MOI.Zeros}()) == 1
     @test MOI.get(model, MOI.NumberOfConstraints{MOI.VectorOfVariables,MOI.SecondOrderCone}()) == 1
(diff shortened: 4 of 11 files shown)


mlubin:
The change looks good to me.
By `cansupport` do you mean `canaddconstraint` or `supportsconstraint`? This is really a mess (https://github.com/JuliaOpt/MathOptInterface.jl/issues/295).

mlubin:
Agreed that would be an improvement from what we do now.

blegat:
`(F,S) in get(m, ListOfSupportedConstraintTypes())` would be inefficient if `get` returns a `Vector`. Why would we change it to the list of constraint types supported ? This is not possible with the UniversalFallback since it would support an infinite number of constraint types :-P

odow:
https://github.com/JuliaOpt/MathOptInterface.jl/blob/f238c74fa323e3e1f29d45bf11fc4c63de169e77/src/Utilities/universalfallback.jl#L152-L153

Surely `supportsconstraint` is only true if `uf.model` supports it?

> would be inefficient if get returns a Vector

The type stable way is to check `canaddconstraint`. `supportsconstraint` could/should be used by extensions etc to check if the solver supports a constraint. For example, in SDDP.jl, I add a `ScalarAffineFunction`-in-`EqualTo` constraint without telling the user. I might want to check that the solver supports such a constraint at the start before I start building the models.

mlubin:
There is a point there that having an explicit list of all possible constraint types supported could be problematic and even infinite for `UniversalFallback`.

odow:
How about renaming `supportsconstraint` to `isconstrainttypesupported`.

Long and verbose (and has an ugly `tt`), but at least it is explicit enough to clear up the confusion.

Other options
- `is_constraint_type_supported`
- `get(m, IsSupportedConstraintType(), F, S)`

blegat:
> Surely supportsconstraint is only true if uf.model supports it?

No it supports everything, if what is not supported by `uf.model`. Currently `UniversalFallback` only implements attribute fallback so looking at the source code will not help :-P

> The type stable way is to check `canaddconstraint`

`canaddconstraint` should be fast since it is used by `CachingOptimizer` and `supportsconstraint` should also be fast since it is used by `LazyBridgeOptimizer` (which is coming in the next few days).

> How about renaming `supportsconstraint` to `isconstrainttypesupported`.

Why would it be less confusing ? The distinction that should be made is:

* `canaddconstraint` : The constraint **can be added** now *in the current state* of the model
* `supportsconstraint` : The model **supports** the constraint *independently of it state*

It seems to me that the `canadd` and `supports` wording convey the right distinction.

mlubin:
> `canaddconstraint`: The constraint **can be added** now *in the current state* of the model

I'm going to re-suggest (https://github.com/JuliaOpt/MathOptInterface.jl/issues/295#issuecomment-378418100) that we remove this concept completely. If a model claims to support a constraint independently of its state, but for some reason the constraint cannot be added right now, I would like to see an error message on `addconstraint` rather than a `false` without an explanation.

blegat:
Copy-only solvers will return `false` to `canaddconstraint` and `true` to `supportsconstraint` and the `CachingOptimizer` uses this fact to determine that the solver needs to go to the `EmptyState`. Having a `try` `catch` to handle this would be less efficient and less elegant I'd say

mlubin:
> Copy-only solvers will return false to canaddconstraint and true to supportsconstraint

https://github.com/JuliaOpt/MathOptInterface.jl/issues/295#issuecomment-378592148

blegat:
> https://github.com/JuliaOpt/MathOptInterface.jl/issues/295#issuecomment-378602220

https://github.com/JuliaOpt/MathOptInterface.jl/issues/295#issuecomment-378602220 :D

mlubin:
> Some solvers might, after optimize! is called, support adding some kind of constraints but not other kind of constraints, even if they support these constraints.

That's fine, they can return errors in that case, and users can work around them. The extra complexity in the API is not worth it, IMO.

blegat:
> The extra complexity in the API is not worth it, IMO.

We may keep the current `MOI.canaddconstraint` and add `MOI.canaddconstraint(::MOI.ModelLike)` which returns a `Bool` indicating whether **any** *supported* constraint can be added.
A direct consequence of this definition gives the following valid default implementation
```julia
MOI.canaddconstraint!(m::ModelLike, F::Type{<:AbstractFunction}, S::Type{<:AbstractSet}) = MOI.canaddconstraint!(m) && MOI.supports
```
Most solvers will only define `MOI.canaddconstraint!(::ModelLike)` but for a solver for which this depends on the constraint types, it can overwrite this default implementation so that the CachingOptimizer takes its particularity into account.

mlubin:
> it can overwrite this default implementation so that the CachingOptimizer takes its particularity into account.

Could you walk through a realisitc example of why this is important and what would go wrong if `CachingOptimizer` just uses `MOI.canaddconstraint!(m) && MOI.supports`?

blegat:
Suppose that once `optimize!` has been called, a mixed integer solver supports adding linear cuts after `optimize!` (i.e. `ScalarAffineFunction`-in-`LessThan{Float64}`) but does not support changing the integrality of the variables (i.e. `SingleVariable`-in-`Integer` and `SingleVariable`-in-`Reals`).
When the user change the integrality through JuMP, the CachingOptimizer should empty the solver but then a linear constraint is added, the CachingOptimizer should just add the constraint (if it empties the solver, there will be a big performance loss since the optimization needs to be restarted from scratch).
The solver will do
```julia
MOI.supportsconstraint!(m::MySolver, ::Type{<:ScalarAffineFunction}, S::Type{<:LessThan}) = true
MOI.supportsconstraint!(m::MySolver, ::Type{<:ScalarAffineFunction}, S::Type{<:LessThan}) = true
MOI.canaddconstraint!(m::MySolver) = !m.optimized
MOI.canaddconstraint!(m::MySolver, ::Type{<:ScalarAffineFunction}, S::Type{<:LessThan}) = true
MOI.canaddconstraint!(m::MySolver, ::Type{<:ScalarAffineFunction}, S::Type{<:LessThan}) = !m.optimized
```

blegat:
The user could be an algorithm and the algorithm could be solver independent where some solvers support the action and some do not. If reset is not called, the algorithm will fail for some solver and if reset is called, the algorithm will unnecessarily reset some solver.

mlubin:
FWIW if you want to rename `canaddconstraint` to `should_caching_optimizer_try_to_add_constraint` then go ahead. That would at least be more honest :)

blegat:
I agree that they are tradeoffs to be made. In the use case I was mentionning can still be solved by the solver caching the integrality changes internally. It is more work but there is also the cost of having `canaddconstraint` that you have just mentionned.

If everyone agrees on replacing `canaddconstraint(::MOI.ModelLike, F, S)` by `canaddconstraint(::MOI.ModelLike)` I won't oppose it.
@joaquimg @odow What do you think ?

mlubin:
> It has a clear concept, and lines up with the rest of the canXXX API.

I'm also proposing to dump the whole `canXXX` API, or at least rename it to `should_caching_optimizer_try_to_do_XXX` because that's what it really is.

blegat:
> * It induces hard-to-explain behavior in CachingOptimizer. ("Why did it just drop the solver?")
> * It obscures error messages that the user might really want to see.

The `Manual` mode of the `CachingOptimizer` is the solution for both of these issues

> I'm also proposing to dump the whole canXXX API, or at least rename it to should_caching_optimizer_try_to_do_XXX because that's what it really is.

The first solution would kill the `Automatic` mode of the `CachingOptimizer` and I don't see how the second solution would make things clearer. `canget` has a self-contained definition independent on the `CachingOptimizer`, mentioning the `CachingOptimizer` in the name wouldn't make it clearer IMO. Adding a note in the definition may help though.

mlubin:
> canget has a self-contained definition independent on the CachingOptimizer

It's a problematic definition however. The issue is that `canXXX` leads `CachingOptimizer` to hide errors that the user might want to see (#302). `Manual` mode is *sometimes* but not always the answer to this (it doesn't resolve #302).

The difference between `canXXX` and `should_caching_optimizer_try_to_do_XXX` is that the latter should return true also in the case that `XXX` is known to return an error that the user should see.

chriscoey:
what's the status of this particular PR?

mlubin:
I don't see a pressing need to change the definition. I support a renaming, maybe to `ListOfConstraintsPresent`.

odow:
Closing this as stale. I'll leave the issue open for now.

PR #361, opened 2018-05-16
[array_of_tuples] Fix MOI tests

blegat:
The MOI testset works.
For building the vector of terms, I alternate between
```julia
MOI.ScalarAffineTerm.([1, 2], [x, y])
```
and
```julia
[MOI.ScalarAffineTerm.(1, x), MOI.ScalarAffineTerm.(2, y)]
```
As a rule, the thumb, the first one is used when there are more terms.
If we want to have a style guideline for the Test module, now is the time :)


test/functions.jl
```diff
@@ -12,7 +12,7 @@
         fvv = MOI.VectorOfVariables([x, z, y])
         @test MOIU.evalvariables(vi -> vals[vi], fvv) ≈ [3, 5, 1]
         @test MOIU.evalvariables(vi -> vals[vi], fvv) ≈ [3, 5, 1]
-        fsa = MOI.ScalarAffineFunction([x, z, y], [1., 3., 2.], 2.)
+        fsa = MOI.ScalarAffineFunction([MOI.ScalarAffineTerm(1., x), MOI.ScalarAffineTerm(3., z), MOI.ScalarAffineTerm(2., y)], 2.)
```
odow: All of the `Float64`s should be `1.0` instead of `1.`

src/functions.jl
```diff
@@ -254,18 +254,24 @@ function sum_dict(kvs::Vector{Pair{K, V}}) where {K, V}
     d
 end
 
-_dicts(f::ScalarAffineFunction) = sum_dict(Pair.(f.variables, f.coefficients))
-_dicts(f::VectorAffineFunction) = sum_dict(Pair.(tuple.(f.outputindex, f.variables), f.coefficients))
-
+function _pair(t::Union{VectorAffineTerm, VectorQuadraticTerm})
```
mlubin: Could you add some comments explaining what these methods do? Underscore isn't a license to provide no comments :)

mlubin:
`MOI.ScalarAffineTerm.([1, 2], [x, y])` is shorter and just as readable, so I'd say go with that.

PR #362, opened 2018-05-16
Add LazyBridgeOptimizer

blegat:


test/runtests.jl
```diff
@@ -9,7 +9,7 @@ using Compat.Test
 
 # Tests for solvers are located in MOI.Test.
 
-# MOI tests not relying on any submodule
+ MOI tests not relying on any submodule
```
odow: Stripped a comment char

blegat: Oups :-P

docs/src/apireference.md
```diff
@@ -304,3 +307,4 @@ Bridges.LogDetBridge
 Bridges.SOCtoPSDCBridge
 Bridges.RSOCtoPSDCBridge
 ```
+For each bridge defined in this package, a corresponding bridge optimizer is available with the same name but "Bridge" suffix, e.g. `SplitInterval` is an `SingleBridgeOptimizer` for the `SplitIntervalBridge`.
```
odow: but _without the_ "Bridge" suffix?

Also, I'm not sure what our policy is on `e.g.`, but it either needs to be "suffix, e.g., `SplitInterval`" or "suffix (e.g. `SplitInterval`"

src/Bridges/bridge.jl
```diff
@@ -30,6 +30,20 @@ type `F`-in`S` created by the bride `b` in the model (i.e., of length equal to t
 """
 MOI.get(b::AbstractBridge, ::MOI.ListOfConstraintIndices{F, S}) where {F, S} = CI{F, S}[]
 
+"""
+    supportedconstrainttypes(BT::Type{<:AbstractBridge})::Bool
+
+Return a list of the types of constraints that can be bridged with a bridge of type `BT`.
+"""
+function supportedconstrainttypes end
```
odow: Is this evidence for why I think this is necessary for all optimizers?

src/Bridges/lazybridgeoptimizer.jl
```diff
@@ -0,0 +1,72 @@
...
+"""
+function addbridge!(b::LazyBridgeOptimizer, BT::Type{<:AbstractBridge})
+    push!(b.bridgetypes, BT)
+    # Bellman-Ford algorithm
+    changed = true # Has b.dist changed in the last iteration ?
+    while changed
+        changed = false
+        i = 0
+        for BT in b.bridgetypes
+            i += 1
```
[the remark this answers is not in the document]
blegat: Oups, it was for debug, good catch

src/Bridges/bridge.jl
blegat: Or the opposite, an evidence that `supportedconstrainttypes` is a bad idea since it does not support bridges that can bridge infinitely many constraints (e.g. bridges that bridge AnyFunction-in-SomeSet)

PR #373, opened 2018-05-24
Add fullbridgeoptimizer function

blegat:
This would allow JuMP to transform
https://github.com/JuliaOpt/JuMP.jl/blob/a59a431419f885a0e11c2838b50bea0e63061647/src/JuMP.jl#L164
into
```julia
m.moibackend = MOIB.fullbridgeoptimizer(MOIU.CachingOptimizer(MOIU.UniversalFallback(JuMPMOIModel{Float64}()), mode == Automatic ? MOIU.Automatic : MOIU.Manual), Float64)
```
and not have to explicitly add each MOI bridge.
It would be nice if this could also be part of MOI v0.3 :)

### DIFF
--- a/src/Bridges/bridgeoptimizer.jl
+++ b/src/Bridges/bridgeoptimizer.jl
@@ -91,42 +91,42 @@ function MOI.canget(b::AbstractBridgeOptimizer, attr::Union{MOI.NumberOfConstrai
 end
 function MOI.get(b::AbstractBridgeOptimizer, loc::MOI.ListOfConstraintIndices{F, S}) where {F, S}
     if isbridged(b, F, S)
-        MOI.get(b.bridged, loc)
+        locr = MOI.get(b.bridged, loc)
     else
         locr = MOI.get(b.model, loc)
-        for bridge in values(b.bridges)
-            for c in MOI.get(bridge, loc)
-                i = findfirst(locr, c)
-                if (VERSION >= v"0.7.0-DEV.3395" && i !== nothing) || (VERSION < v"0.7.0-DEV.3395" && !iszero(i))
-                    MOI.deleteat!(locr, i)
-                end
+    end
+    for bridge in values(b.bridges)
+        for c in MOI.get(bridge, loc)
+            i = findfirst(locr, c)
+            if (VERSION >= v"0.7.0-DEV.3395" && i !== nothing) || (VERSION < v"0.7.0-DEV.3395" && !iszero(i))
+                MOI.deleteat!(locr, i)
             end
         end
-        locr
     end
+    locr
 end
-function _numberof(b::AbstractBridgeOptimizer, attr::Union{MOI.NumberOfConstraints, MOI.NumberOfVariables})
-    s = MOI.get(b.model, attr)
+function _numberof(b::AbstractBridgeOptimizer, model::MOI.ModelLike, attr::Union{MOI.NumberOfConstraints, MOI.NumberOfVariables})
+    s = MOI.get(model, attr)
     for v in values(b.bridges)
         s -= MOI.get(v, attr)
     end
     s
 end
-MOI.get(b::AbstractBridgeOptimizer, attr::MOI.NumberOfVariables) = _numberof(b, attr)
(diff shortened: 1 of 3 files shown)


src/Bridges/bridgeoptimizer.jl
```diff
@@ -91,42 +91,42 @@ function MOI.canget(b::AbstractBridgeOptimizer, attr::Union{MOI.NumberOfConstrai
...
+        _numberof(b, b.model, attr)
     end
 end
 MOI.canget(b::AbstractBridgeOptimizer, attr::MOI.ListOfConstraints) = MOI.canget(b.model, attr) && MOI.canget(b.bridged, attr)
 _noc(b, fs) = MOI.get(b, MOI.NumberOfConstraints{fs...}())
 function MOI.get(b::AbstractBridgeOptimizer, attr::MOI.ListOfConstraints)
-    loc = MOI.get(b.model, attr)
-    rm = find(_noc.(b, loc) .== 0)
+    loc = [MOI.get(b.model, attr); MOI.get(b.bridged, attr)]
+    rm = find(iszero.(_noc.(b, loc)))
```
mlubin: I don't understand what's going on here. The names `loc`, `_noc`, and `rm` don't help with readability. Why do entries need to be deleted?

test/bridge.jl
```diff
@@ -101,28 +123,27 @@ MOIU.@model NoRSOCModel () (EqualTo, GreaterThan, LessThan, Interval) (Zeros, No
         @test MOIB.bridge(bridgedmock, c) isa MOIB.RSOCtoPSDCBridge
     end
 
+    @testset "Combining two briges" begin
+        fullbridgedmock = MOIB.fullbridgeoptimizer(mock, Float64)
+        mock.optimize! = (mock::MOIU.MockOptimizer) -> MOIU.mock_optimize!(mock, [1, 1, 0, 1, 1, 0, 1, √2])
+        config = MOIT.TestConfig()
+        MOIT.rootdett1vtest(fullbridgedmock, config)
+        MOIT.rootdett1ftest(fullbridgedmock, config)
+        @test !MOI.canget(fullbridgedmock, MOI.ConstraintDual(), MOI.ConstraintIndex{MOI.VectorAffineFunction{Float64}, MOI.RootDetConeTriangle})
```
mlubin: Why do you expect to not be able to get the constraint dual? Because it's not implemented in the bridge?

blegat: Yes, it's not yet implemented.

mlubin: Better to add a comment to that effect then

src/Bridges/bridgeoptimizer.jl
```diff
@@ -122,11 +122,11 @@ function MOI.get(b::AbstractBridgeOptimizer, attr::MOI.NumberOfConstraints{F, S}
...
 end
 MOI.canget(b::AbstractBridgeOptimizer, attr::MOI.ListOfConstraints) = MOI.canget(b.model, attr) && MOI.canget(b.bridged, attr)
-_noc(b, fs) = MOI.get(b, MOI.NumberOfConstraints{fs...}())
 function MOI.get(b::AbstractBridgeOptimizer, attr::MOI.ListOfConstraints)
-    loc = [MOI.get(b.model, attr); MOI.get(b.bridged, attr)]
-    rm = find(iszero.(_noc.(b, loc)))
-    deleteat!(loc, rm)
+    list_of_types = [MOI.get(b.model, attr); MOI.get(b.bridged, attr)]
+    types_to_remove = find(iszero.(map(FS -> MOI.get(b, MOI.NumberOfConstraints{FS...}()), list_of_types)))
+    deleteat!(list_of_types, types_to_remove)
```
mlubin: Why do we need to remove some types? Why would the model or the bridge return constraint types where `NumberOfConstraints` is zero?

blegat: Because these are constraints created by a bridge and not by the user. `NumberOfConstraints` takes care of removing them.

blegat: I have added a comment explaining this :)

PR #380, opened 2018-06-06
Fix #377, Likely persistent annoyances with 32 bit support due to explicit Int64s

### LINKED ISSUE #377: Likely persistent annoyances with 32 bit support due to explicit `Int64`s
Explicit `Int64`s, e.g. in these locations:

https://github.com/JuliaOpt/MathOptInterface.jl/blob/0b5afa2b32fa76f3e5bb203f077e6891a9bd1ffb/src/functions.jl#L86-L89

https://github.com/JuliaOpt/MathOptInterface.jl/blob/0b5afa2b32fa76f3e5bb203f077e6891a9bd1ffb/src/functions.jl#L154-L157

are pretty likely to cause headaches with trying to support 32 bit architectures. I ran into this already in https://github.com/oxfordcontrol/OSQP.jl/pull/14.

I personally don't think it would be so bad to have those fields just be `Int`s, as more than two billion rows (`typemax(Int32)`) in one `VectorAffineConstraint` seems like an uncommon scenario, but I think at least there should be constructors for these types that take `Integer`s.

tkoolen:
Fixes #377.

### DIFF
--- a/test/functions.jl
+++ b/test/functions.jl
@@ -3,6 +3,14 @@
     x = MOI.VariableIndex(1)
     y = MOI.VariableIndex(2)
     z = MOI.VariableIndex(3)
+    @testset "VectorAffineTerm/VectorQuadraticTerm construction" begin
+        scalaraffine = MOI.ScalarAffineTerm(2.0, z)
+        @test MOI.VectorAffineTerm(Int32(3), scalaraffine) === MOI.VectorAffineTerm(Int64(3), scalaraffine)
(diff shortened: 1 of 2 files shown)


test/functions.jl
```diff
@@ -3,6 +3,15 @@
     x = MOI.VariableIndex(1)
     y = MOI.VariableIndex(2)
     z = MOI.VariableIndex(3)
+    @testset "VectorAffineTerm/VectorQuadraticTerm construction" begin
+        NonNativeInt = Int == Int32 ? Int64 : Int32
```
mlubin: It's sufficient to just test with `Int32`. Native vs. nonnative is unneeded complexity in this context.

tkoolen: Addressed.

Issue #360, opened 2018-05-14
Proposal: change API for modifications


odow:
To date, I'm not sure if anyone has dug into the modification API yet. I was about to start writing tests (https://github.com/JuliaOpt/MathOptInterface.jl/pull/354) and had some thoughts.

---

In addition to some [function modifications](http://www.juliaopt.org/MathOptInterface.jl/latest/apireference.html#MathOptInterface.ScalarConstantChange), we currently have two [modifyconstraint!](http://www.juliaopt.org/MathOptInterface.jl/latest/apireference.html#MathOptInterface.modifyconstraint!) methods:

`modifyconstraint!(m, c::ConstraintIndex{F,S}, new_set::S)`
`modifyconstraint!(m, c::ConstraintIndex{F,S}, new_function::F)`

Considering the rest of the API, doesn't it make more sense to make these `set!` methods?

```julia
set!(m::ModelLike, ::ConstraintFunction, c::ConstraintIndex{F,S}, new_function::F)
set!(m::ModelLike, ::ConstraintSet, c::ConstraintIndex{F,S}, new_set::S)
```

---

Furthermore, instead of `modifyconstraint!` and `modifyobjective!`, we could have
```julia
canmodify(m, ::ConstraintFunction, c::Type{ConstraintIndex{F,S}}, change::Type{AbstractFunctionModification})
modify!(m, ::ConstraintFunction, c::ConstraintIndex{F,S}, change::AbstractFunctionModification)

canmodify(m, ::ObjectiveFunction, change::Type{AbstractFunctionModification})
modify!(m, ::ObjectiveFunction, change::AbstractFunctionModification)
```

Edit: we could also change the `transformconstraint` to 
```julia
canmodify(m, ::ConstraintSet, c::Type{ConstraintIndex{F,S1}}, ::Type{S2})
c2 = modify!(m, ::ConstraintSet, c::ConstraintIndex{F,S1}, new_set::S2)
```

blegat:
I am in favor of these changes.

For `transform` -> `modify`, I remember we named it transform to make it different than `modify` with a set of the same type but if modifying the set is moved to `set!` then we could add changing the set with a new set type to `modify`. However, `transform` need to return a new constraint index while `modify` does not change the constraint index so it is probably better to keep then separate functions.

mlubin:
[...]

PR #386, opened 2018-06-13
Added member types to *Change struct docstrings

BenChung:
I've been using raw MOI to solve an optimization problem with problem modification, and I regularly need to remind myself what the types of the change struct fields are. This PR adds type information to docstrings that did not previously indicate what the type of the fields are, so that it's easy to tell what they take from Julia's help system.

### DIFF
--- a/src/functions.jl
+++ b/src/functions.jl
@@ -205,7 +205,7 @@ An abstract supertype for structs which specify partial modifications to functio
 abstract type AbstractFunctionModification end
 
 """
-    ScalarConstantChange{T}(new_constant)
+    ScalarConstantChange{T}(new_constant::T)
 
 A struct used to request a change in the constant term of a scalar-valued function.
 Applicable to `ScalarAffineFunction` and `ScalarQuadraticFunction`.
@@ -215,7 +215,7 @@ struct ScalarConstantChange{T} <: AbstractFunctionModification
 end
 
 """
-    VectorConstantChange{T}(new_constant)
+    VectorConstantChange{T}(new_constant::Vector{T})
 
 A struct used to request a change in the constant vector of a vector-valued function.
 Applicable to `VectorAffineFunction` and `VectorQuadraticFunction`.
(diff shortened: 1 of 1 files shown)


odow:
> I've been using raw MOI to solve an optimization problem with problem modification

Really!?! Which solvers? What are your thoughts? Do you have a view on https://github.com/JuliaOpt/MathOptInterface.jl/issues/360?

odow:
> Modulo some easy-to-fix bugs I ran into in their MOI implementation

I would expect that there a number of bugs in all of the solver wrappers.

> The perils of developing code against pre-prerelease libraries!

Your code is probably the most complex usage of MOI yet! We'd love to hear any and all gripes and complaints you have with the API.

> I'd go so far as to suggest that the transformconstraint! equivalent should not handle the case where the old set is the same as the new one at all

Yes, this is what we do [in LinQuadOptInterface](https://github.com/JuliaOpt/LinQuadOptInterface.jl/blob/37bdd17ad71f51d9ba44d43313476fc87583bf06/src/constraints/scalaraffine.jl#L212).

p.s., super late notice, but are you aware of http://www.juliaopt.org/meetings/bordeaux2018/

BenChung:
> We'd love to hear any and all gripes and complaints you have with the API.

The new function definitions in 0.3 fixed my real issue. The old API required a lot of clever indexing for VectorAffineFunctions in particular, but with the explicit structs it's a lot nicer. While I haven't ported my problem forward yet, I'm expecting it to be much easier to read at the very least! With 0.2, I've spent a long time picking apart VectorAffineFunctions to figure out what I intended them to do and how precisely the index acrobatics achieves this. The one thing that I would like though is some way to change the entire VectorAffineFunction matrix in one go, but this is a convenience that I suspect doesn't map down to the solver well.

> p.s., super late notice, but are you aware of http://www.juliaopt.org/meetings/bordeaux2018/

I am! Unfortunately, while I'll be in Europe at that time, I'm busy that week. If it comes back to MIT I'd love to attend though :).

odow:
> The one thing that I would like though is some way to change the entire VectorAffineFunction matrix in one go

If you're changing the sparsity pattern, the correct way to do this is probably to delete the constraint and add a new one. If you're keeping the same sparsity pattern it shouldn't be too hard to write a wrapper that loops through calling `MultirowChange` on the variables.

Issue #218, opened 2018-02-11
candelete is inconsistent with canget


mlubin:
We would probably like to pass in the type of the index instead of the index value itself.

blegat:
There might be cases where some variables can be deleted and some cannot. For instance, a solver could say that a variable cannot be deleted if it still appears in some constraints

mlubin:
That sounds like it could also just be an error on `delete!`. Unclear though...

blegat:
There is the same performance issue in `CachingOptimizer` for deleting constraints so I think we should go for it

odow:
Bump. I just encountered this.

PR #388, opened 2018-06-14
RFC: new modification syntax

odow:
### Overview

This pull request changes the modification syntax in MOI.

The biggest change is to refactor `modifyconstraint!(m, c::ConstraintIndex{F,S}, set::S)` to `set!(m, ::ConstraintSet, c::ConstraintIndex{F,S}, set::S)`. (A similar refactor is done with `ConstraintFunction`.) This plays much nicer with the `get`/`set!` paradigm.

In addition, `modifyconstraint!` methods that implement `AbstractFunctionModification` is refactored to `modify!` and `transformconstraint!` is refactored into `transform!`.

This PR also adds tests and documentation.

### Changes
See #360 for discussion. Proposed changes:
- [x] `set!(m::ModelLike, ::ConstraintFunction, c::ConstraintIndex{F,S}, new_function::F)`
- [x] `set!(m::ModelLike, ::ConstraintSet, c::ConstraintIndex{F,S}, new_set::S)`
<s>@blegat, how do I make the change for `ConstraintSet` in `model.jl`? There is quite a few levels of in-direction and the macro and I don't know what to touch.</s>
- [x] `modify!(m::ModelLike, c::ConstraintIndex{F,S}, change::AbstractFunctionModification)` 
(Note: this is slightly different to that proposed in #360.)
- [x] `modify!(m::ModelLike, ::ObjectiveFunction, change::AbstractFunctionModification)`
- [x] `transformconstraint!` -> `transform!`

### Tests
- [x] ConstraintSet
  - [x] changing variable bound
  - [x] changing rhs in scalaraffine
- [x] ConstraintFunction
- [x] modify! objective
  - [x] scalarconstant
  - [x] scalarcoefficient
- [ ] modify! constraint
  - [ ] scalarconstant
  Given that we say the constant should be part of the set, what F-in-S pair makes sense to test?
  - [x] vectorconstant
  - [x] multirowchange
  - [x] scalarcoefficient
- [ ] transform!
  Not tested. See below

### Blockers / future PR

- [ ] `cantransform` still takes the instance of the constraint index of an argument rather than the type. This is needed because there is a default fallback and `candelete` also takes the instance rather than the type (see #218).
- [ ] `transform!` is not tested as it is not implemented in `MockOptimizer`.
- [ ] `transform!` only applies to constraint sets. Should it apply to constraint functions for consistency? I'm not sure if it makes sense to change the function to a different  type of function without deleting the constraint though...

These points can probably be addressed by a future PR that changes `candelete` and `cantransform` at the same time and adds tests. This PR is already quite large.


blegat:
Do we now disallow a set or function being passed to `modify!` ?

odow:
Yes. `modify` will be reserved for `AbstractFunctionModification`s.

src/Utilities/mockoptimizer.jl
```diff
@@ -210,8 +210,9 @@ function MOI.modifyconstraint!(mock::MockOptimizer, c::CI, change)
     MOI.modifyconstraint!(mock.inner_model, xor_index(c), xor_variables(change))
 end
 
-function MOI.modifyconstraint!(mock::MockOptimizer, c::CI{F,S}, set::S) where {F<:MOI.AbstractFunction, S<:MOI.AbstractSet}
-    MOI.modifyconstraint!(mock.inner_model, xor_index(c), set)
+MOI.canset(mock::MockOptimizer, ::MOI.ConstraintSet, c::CI{F,S}, set::Type{S}) where {F<:MOI.AbstractFunction, S<:MOI.AbstractSet} = true
+function MOI.set!(mock::MockOptimizer, ::MOI.ConstraintSet, c::CI{F,S}, set::S) where {F<:MOI.AbstractFunction, S<:MOI.AbstractSet}
```
blegat: This is already handled by line 94

src/constraints.jl
```diff
@@ -43,18 +43,44 @@ function addconstraints! end
...
+### Examples
+
+If `c` is a `ConstraintIndex{F,Interval}`
+
+```julia
+set!(model, ConstraintSet(), c, Interval(0, 5))
+set!(model, ConstraintSet(), c, NonPositives)    # Error
+```
+"""
+function set!(model::ModelLike, cs::ConstraintSet, c::ConstraintIndex{F,S1}, s::S2) where F where  S1 where S2
```
blegat: The issue of having a specific method like this is that it prevents models to handle it with a more general method as this would lead to ambiguity (this affects mockoptimizer, bridges, ...)
We could instead call something like `seterror` in  line 234.
The default `seterror` method would give the ArgumentError currently at line 234 and we would add this method as a `seterror` method.

blegat:
Why don't we use a different signature for `canset` than the one used for all other attributes ?

src/Bridges/bridgeoptimizer.jl
```diff
@@ -209,19 +209,59 @@ function MOI.addconstraint!(b::AbstractBridgeOptimizer, f::MOI.AbstractFunction,
...
     end
 end
-function MOI.canmodifyconstraint(b::AbstractBridgeOptimizer, ci::CI, change)
-    if isbridged(b, typeof(ci))
-       MOI.canmodifyconstraint(b.bridged, ci, change) && MOI.canmodifyconstraint(b, MOIB.bridge(b, ci), change)
+function MOI.canmodify(b::AbstractBridgeOptimizer, ::Type{C}, ::Type{Chg}) where {C<:CI,Chg<:MOI.AbstractFunctionModification}
+    if isbridged(b, C)
+       MOI.canmodify(b.bridged, C, Chg)
+       # TODO(@blegat): how should the following be implemented?
+       # && MOI.canmodify(b, MOIB.bridge(b, ci), Chg)
```
odow: Okay, @blegat new question. Is this necessary and in what use-case? It makes tests like
https://github.com/JuliaOpt/MathOptInterface.jl/pull/388/files#diff-5d9f4af23e8ec976db542b6b9af239c4R135
break, but everything else passes. There are also things like
https://github.com/JuliaOpt/MathOptInterface.jl/pull/388/files#diff-91d11cfb910f6ae0036ae89cac114d3eR144
Are they necessary?

src/Bridges/bridgeoptimizer.jl
```diff
@@ -209,19 +209,59 @@ function MOI.addconstraint!(b::AbstractBridgeOptimizer, f::MOI.AbstractFunction,
...
+    else
+        MOI.modify!(b.model, ci, change)
+    end
+end
+
+function MOI.canset(b::AbstractBridgeOptimizer, ::MOI.ConstraintSet, ::Type{C}) where C <: CI
+    if isbridged(b, C)
+       MOI.canset(b.bridged, MOI.ConstraintSet(), C)
+        # TODO(@blegat) is this necessary? How do I do it without types
+        # && MOI.canset(b, MOIB.bridge(b, ci), change)
```
odow: Ditto here.

src/Bridges/bridgeoptimizer.jl
```diff
@@ -209,19 +209,59 @@ function MOI.addconstraint!(b::AbstractBridgeOptimizer, f::MOI.AbstractFunction,
...
+    else
+        MOI.set!(b.model, MOI.ConstraintSet(), constraint_index, set)
+    end
+end
+
+function MOI.canset(b::AbstractBridgeOptimizer, ::MOI.ConstraintFunction, ::Type{C}) where C <: CI
+    if isbridged(b, C)
+       MOI.canset(b.bridged, MOI.ConstraintFunction(), C)
+        # TODO(@blegat) is this necessary? How do I do it without types
+        # && MOI.canset(b, MOIB.bridge(b, ci), change)
```
odow: And here.

src/Bridges/bridgeoptimizer.jl
blegat: Yes, it checks that the bridge supports the modification. Currently, only the interval bridge supports modification

odow: I missed the second part to the question. If they are necessary, how do we implement this using only type information...

src/Test/UnitTests/modifications.jl
```diff
@@ -0,0 +1,282 @@
...
+            constraint_dual   = [(c, -1.0)]
+        )
+    end
+    @test MOI.canset(model, MOI.ConstraintFunction(), typeof(c))
+    MOI.set!(model, MOI.ConstraintFunction(), c,
+        MOI.ScalarAffineFunction([MOI.ScalarAffineTerm(2.0, x)], 0.0)
+    )
+    @test MOI.canget(model, MOI.ConstraintFunction(), typeof(c))
+    foo = MOI.get(model, MOI.ConstraintFunction(), c)
+    @test_broken foo == MOI.ScalarAffineFunction([MOI.ScalarAffineTerm(2.0, x)], 0.0)
```
odow: @blegat another mystery. The `VariableIndex` in `foo` is `VariableIndex(12345679)` instead of `VariableIndex(1)`, but I couldn't work out where the missing `xor` is.

src/Bridges/bridgeoptimizer.jl
blegat: You need `bridgetype` :)

blegat: I have pushed a fix

src/Test/UnitTests/modifications.jl
blegat: Fixed by 1ad6f61

odow:
This is now RFC. 

cc: @mlubin, @joaquimg

src/Bridges/bridgeoptimizer.jl
```diff
@@ -209,25 +209,59 @@ function MOI.addconstraint!(b::AbstractBridgeOptimizer, f::MOI.AbstractFunction,
...
+
+function MOI.canset(b::AbstractBridgeOptimizer, ::MOI.ConstraintFunction, ::Type{CI{F, S}}) where {F, S}
+    if isbridged(b, CI{F, S})
+        MOI.canset(b.bridged, MOI.ConstraintFunction(), CI{F, S}) && MOI.canset(b, MOI.ConstraintFunction(), MOIB.bridgetype(b, F, S))
+    else
+        MOI.canset(b.model, MOI.ConstraintFunction(), CI{F, S})
+    end
+end
+function MOI.set!(b::AbstractBridgeOptimizer, ::MOI.ConstraintFunction, constraint_index::CI, func)
+    # Note: we also under-type this function to avoid ambiguity
```
blegat: What do you mean by this comment ? Why don't you just define `set!(..., attr::Union{ConstraintFunction, ConstraintSet}, ...)` ?

src/constraints.jl
```diff
@@ -44,83 +44,81 @@ function addconstraints! end
...
 
 ### Examples
 
 If `c` is a `ConstraintIndex{F,Interval}`
 
 ```julia
-modifyconstraint!(model, c, Interval(0, 5))
-modifyconstraint!(model, c, NonPositives) # Error
+set!(model, ConstraintSet(), c, Interval(0, 5))
+set!(model, ConstraintSet(), c, NonPositives)    # Error
```
mlubin: `Nonpositives` is the type. Maybe use a scalar set instead so it's a bit more clear, e.g., `LessThan(3)`.

src/constraints.jl
```diff
@@ -164,13 +162,13 @@ Return a `Bool` indicating whether the set of type `S1` in constraint `c` can be
 If `c` is a `ConstraintIndex{ScalarAffineFunction{Float64},LessThan{Float64}}`,
 
 ```julia
-cantransformconstraint(model, c, GreaterThan(0.0)) # true
-cantransformconstraint(model, c, ZeroOne())        # false
+cantransform(model, c, GreaterThan(0.0)) # true
+cantransform(model, c, ZeroOne())        # false
```
mlubin: This is false for all current solvers, but MOI allows you to implement a solver where this could work. Maybe clarify that this is expected behavior but not a hard rule.

src/modifications.jl
```diff
@@ -0,0 +1,56 @@
...
+
+### Examples
+
+```julia
+canmodify(model, MOI.ConstraintIndex{MOI.ScalarAffineFunction{Float64}, MOI.LessThan{Float64}}, ScalarConstantChange{Float64})
+```
+
+## Objective Function
+
+canmodify(model::ModelLike, ::ObjectiveFunction, ::Type{M})::Bool where M<:AbstractFunctionModification
```
mlubin: Missing code formatting?

src/Bridges/bridgeoptimizer.jl
odow: I had a hard time getting rid of method ambiguities since there are quite a few abstract subtypes of ModelLike floating around. With the Union there is an ambiguity with `set!(::ModelLike, ::ConstraintFunction, ...)`. as the first argument is more specific but the second is less specific. I've typed this now and it seems to all work.

src/Utilities/model.jl
```diff
@@ -33,7 +33,7 @@ end
 _modifyconstr(ci::CI{F, S}, f::F, s::S, change::F) where {F, S} = (ci, change, s)
 _modifyconstr(ci::CI{F, S}, f::F, s::S, change::S) where {F, S} = (ci, f, change)
 _modifyconstr(ci::CI{F, S}, f::F, s::S, change::MOI.AbstractFunctionModification) where {F, S} = (ci, modifyfunction(f, change), s)
-function _modifyconstraint!(constrs::Vector{C{F, S}}, ci::CI{F}, i::Int, change) where {F, S}
+function _modify!(constrs::Vector{C{F, S}}, ci::CI{F}, i::Int, change) where {F, S}
```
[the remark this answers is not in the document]
odow: @blegat? The whole `model.jl` file could do with some refactoring as there are quite a few levels of mis-direction. Some arguments could be more verbose as it's not obvious at first glance about the difference between `C{F,S}` and `CI{F,S}`.

src/constraints.jl
```diff
@@ -44,83 +44,81 @@ function addconstraints! end
...
-```julia
-canmodifyconstraint(model, c, ScalarConstantChange{Float64})
-```
+Return a `Bool` indicating whether the set in a constraint of type `F`-in-`S`
+can be replaced by another set of the same type `S` as the original set.
 """
-function canmodifyconstraint end
-canmodifyconstraint(model::ModelLike, c::ConstraintIndex, change) = false
+canset(model::ModelLike, ::ConstraintSet, type_of_constraint_set) = false
+# Note: the above method is deliberately under-typed to avoid method ambiguities
```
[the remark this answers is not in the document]
odow: Added.

src/Bridges/bridgeoptimizer.jl
```diff
@@ -209,25 +209,57 @@ function MOI.addconstraint!(b::AbstractBridgeOptimizer, f::MOI.AbstractFunction,
...
-        MOI.modifyconstraint!(b, bridge(b, ci), change)
-        MOI.modifyconstraint!(b.bridged, ci, change)
+        MOI.modify!(b, bridge(b, ci), change)
+        MOI.modify!(b.bridged, ci, change)
+    else
+        MOI.modify!(b.model, ci, change)
+    end
+end
+
+function MOI.canset(b::AbstractBridgeOptimizer, attr::MOI.ConstraintSet, ::Type{CI{F, S}}) where {F, S}
```
blegat: This can be merged with the `ConstraintFunction` or not finally ? I thought you said in your comment that it was updated

src/Utilities/model.jl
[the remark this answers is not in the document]
blegat: I can do a refactoring adding more doc in a separate PR.

src/Utilities/universalfallback.jl
```diff
@@ -152,12 +152,26 @@ end
 MOI.supportsconstraint(uf::UniversalFallback, ::Type{F}, ::Type{S}) where {F<:MOI.AbstractFunction, S<:MOI.AbstractSet} = true
 MOI.canaddconstraint(uf::UniversalFallback, ::Type{F}, ::Type{S}) where {F<:MOI.AbstractFunction, S<:MOI.AbstractSet} = MOI.canaddconstraint(uf.model, F, S)
 MOI.addconstraint!(uf::UniversalFallback, f::MOI.AbstractFunction, s::MOI.AbstractSet) = MOI.addconstraint!(uf.model, f, s)
-MOI.canmodifyconstraint(uf::UniversalFallback, ci::CI, change) = MOI.canmodifyconstraint(uf.model, ci, change)
-MOI.modifyconstraint!(uf::UniversalFallback, ci::CI, change) = MOI.modifyconstraint!(uf.model, ci, change)
+MOI.canmodify(uf::UniversalFallback, ci::Type{<:CI}, change) = MOI.canmodify(uf.model, ci, change)
+MOI.modify!(uf::UniversalFallback, ci::CI, change) = MOI.modify!(uf.model, ci, change)
+
+function MOI.canset(uf::UniversalFallback, ::MOI.ConstraintSet, ::Type{C}) where C <: CI
```
blegat: Again, do we really need 4 functions instead of two ?

src/constraints.jl
```diff
@@ -44,83 +44,97 @@ function addconstraints! end
...
 If `c` is a `ConstraintIndex{F,Interval}`
 
 ```julia
-modifyconstraint!(model, c, Interval(0, 5))
-modifyconstraint!(model, c, NonPositives) # Error
+set!(model, ConstraintSet(), c, Interval(0, 5))
+set!(model, ConstraintSet(), c, GreaterThan(0.0))  # Error
 ```
+"""
+function set!(model::ModelLike, ::ConstraintSet, constraint_index, set)
```
blegat: We shouldn't define `set!` and `canset` methods with so specific types.
We should do `set!(model::ModelLike, attr::AbstractConstraintAttribute, ci::CI, value) = default_set!(model, attr, ci, value)` and then in the methods of `default_set!` we can do any typing we like without risk of ambiguity

src/Bridges/bridgeoptimizer.jl
odow: It is ambiguous with calls like
https://github.com/JuliaOpt/MathOptInterface.jl/blob/odow/new-modifications/src/Bridges/intervalbridge.jl#L60

blegat: See https://github.com/JuliaOpt/MathOptInterface.jl/pull/388#discussion_r195987435

odow:
- I have moved `canset` and `set!` out of the `constraints.jl` file and added them to the original docstring. 
- I've also implemented @blegat's idea of a `hidden_set!` function which we can overload to throw errors without causing ambiguity. 
- I have also removed a few of the separate `canset(m, ::ConstraintSet` and `canset(m, ::ConstraintFunction` methods in favor of `::Union{ConstraintSet,ConstraintFunction}` methods.

There are still a few unresolved tick-boxes in the description of this PR, but they can be probably be addressed in future PRs.

src/attributes.jl
```diff
@@ -527,13 +578,35 @@ It is guaranteed to be equivalent but not necessarily identical to the function
 """
 struct ConstraintFunction <: AbstractConstraintAttribute end
 
+function hidden_set!(model::ModelLike, ::ConstraintFunction, constraint_index::ConstraintIndex, func::AbstractFunction)
+    if func_type(constraint_index) != typeof(func)
```
blegat: This can be done with multiple dispatch,
`hidden_set!(..., ::ConstraintFunction, ::CI{F, S}, ::F)` -> same constraint set
`hidden_set!(..., ::ConstraintFunction, ::CI, ::AbstractFunction)` -> different constraint set

src/attributes.jl
```diff
@@ -527,13 +578,35 @@ It is guaranteed to be equivalent but not necessarily identical to the function
...
+
 """
     ConstraintSet()
 
 Return the `AbstractSet` object used to define the constraint.
 """
 struct ConstraintSet <: AbstractConstraintAttribute end
 
+function hidden_set!(model::ModelLike, ::ConstraintSet, constraint_index::ConstraintIndex, set::AbstractSet)
+    if set_type(constraint_index) != typeof(set)
```
blegat: Same, here, it is cleaner to do that using multiple dispatch

src/attributes.jl
odow: Oh yeah good point. This is a hold-over from before

src/Utilities/cachingoptimizer.jl
```diff
@@ -227,54 +227,75 @@ function MOI.addconstraint!(m::CachingOptimizer, func::MOI.AbstractFunction, set
...
+    if m.state == AttachedOptimizer && m.mode == Manual
+        if !MOI.canset(m.optimizer, attr, C)
+            return false
+        end
+    end
+    return true
+end
+
+# This function avoids duplicating code, but allows us to strongly type the
+# arguments of the set! functions below.
```
blegat: I didn't get that immediately, mentioning the 4th argument explicitly in the comment would be clearer

docs/src/apimanual.md
```diff
@@ -412,11 +412,210 @@ primal_variable_result = MOI.get(optimizer, MOI.VariablePrimal(), x)
...
+```
+where `m` is our [`ModelLike`](@ref) model. However, the following will fail as
+the new function is of a different type to the original function:
+```julia
+set!(m, ConstraintFunction(), c,
+    ScalarAffineFunction([ScalarAffineTerm(1.0, x)], 0.0)
+)
+```
+
+### In-place Modifications
```
mlubin: Nit: Inconsistent capitalization of words in section titles. (Also "Problem Modification" at the top).

test/model.jl
```diff
@@ -123,3 +123,28 @@ end
     @test MOI.dimension(s) == 1
 
 end
+
+struct Func388 <: MOI.AbstractScalarFunction
```
mlubin: Nit: use a more descriptive name or add a link to the PR if we need to reference the discussion. 388 isn't too meaningful on its own.

PR #389, opened 2018-06-14
Update Model

blegat:
Test still fails for the bridges because of the ambiguity mentioned in https://github.com/JuliaOpt/MathOptInterface.jl/pull/388#discussion_r195321540


blegat:
I think it is a strict improvement. I expect that the cost of doing this change now will be lower than the cost of not having this improvement.
It was confusing that `ConstraintSet` and `ConstraintFunction` where only gettable and it was confusing that `modify` could have three different changes.
Merging `modifyconstraint` and `modifyobjective` is even better since it allow to implements both with one method for layers (e.g. mockoptimizer, cachingoptimizer, bridges, ...)

odow:
Okay, I'm going to merge this across and have a go at getting the tests to pass.

Issue #321, opened 2018-04-19
CachingOptimizer modifyconstraint! performance


tkoolen:
I'm a little worried about the performance of `modifyconstraint!` with `CachingOptimizer`. Maybe the point of worrying about performance hasn't been reached yet, but I still wanted to bring this to your attention.

Here's an example:

```julia
using BenchmarkTools
using OSQP.MathOptInterfaceOSQP
const MOI = MathOptInterface
const MOIU = MOI.Utilities

optimizer = OSQPOptimizer()
MOI.set!(optimizer, OSQPSettings.Verbose(), false)
caching_optimizer = MOIU.CachingOptimizer(OSQPModel{Float64}(), optimizer);
v = MOI.addvariables!(caching_optimizer, 2)
cf = MOI.ScalarAffineFunction(v, [0.0, 0.0], 0.0)
c = MOI.addconstraint!(caching_optimizer, cf, MOI.Interval(-Inf, 1.0))
```

With no optimizer attached:

```julia
julia> @btime MOI.modifyconstraint!($caching_optimizer, $c, $(MOI.Interval(0.0, 2.0)))
  133.940 ns (7 allocations: 176 bytes)
```

With optimizer attached:

```julia
julia> MOIU.attachoptimizer!(caching_optimizer);
julia> @btime MOI.modifyconstraint!($caching_optimizer, $c, $(MOI.Interval(0.0, 2.0)))
  322.491 ns (12 allocations: 272 bytes)
julia> @btime MOI.modifyconstraint!($(caching_optimizer.model_cache), $c, $(MOI.Interval(0.0, 2.0))); # for the record
  26.208 ns (3 allocations: 96 bytes)
```

while for the raw optimizer:

```julia
julia> @btime MOI.modifyconstraint!($(caching_optimizer.optimizer), $(caching_optimizer.model_to_optimizer_map[c]), $(MOI.Interval(0.0, 2.0)))
  11.927 ns (0 allocations: 0 bytes)
2.0
```

I think there are a few opportunities to improve performance:

1. Parameterize `CachingOptimizer` on the types of `model_cache` and `optimizer` (i.e., `CachingOptimizer{M<:MOI.ModelLike, O<:MOI.AbstractOptimizer}`). These fields are currently abstractly typed, so any interaction with them will cause runtime dispatch and allocations. This would mean that the current approach of setting the `optimizer` field to `nothing` to detach the optimizer has to be implemented in a different way.
[...]

PR #323, opened 2018-04-19
Parametrize CachingOptimizer and MockOptimizer on model type

blegat:
Initial benchmark:
```
v0.6
126.218 ns (7 allocations: 176 bytes)
452.726 ns (17 allocations: 416 bytes)
 23.927 ns (3 allocations: 96 bytes)
 40.416 ns (5 allocations: 144 bytes)
v0.7
105.262 ns (5 allocations: 128 bytes)
414.565 ns (12 allocations: 304 bytes)
  9.623 ns (1 allocation: 48 bytes)
 29.113 ns (3 allocations: 96 bytes)
```
After parametrization of the MockOptimizer:
```
v0.6
121.149 ns (7 allocations: 176 bytes)
357.014 ns (14 allocations: 352 bytes)
 24.722 ns (3 allocations: 96 bytes)
 24.712 ns (3 allocations: 96 bytes)
v0.7
104.708 ns (5 allocations: 128 bytes)
311.206 ns (9 allocations: 240 bytes)
  9.306 ns (1 allocation: 48 bytes)
  9.400 ns (1 allocation: 48 bytes)
```
After parametrization of CachingOptimizer
```
v0.6
 38.889 ns (4 allocations: 112 bytes)
252.048 ns (11 allocations: 288 bytes)
 24.412 ns (3 allocations: 96 bytes)
 24.982 ns (3 allocations: 96 bytes)
v0.7
 23.149 ns (2 allocations: 64 bytes)
207.150 ns (6 allocations: 176 bytes)
  9.761 ns (1 allocation: 48 bytes)
  9.591 ns (1 allocation: 48 bytes)
```

Related to #321

PR #390, opened 2018-06-14
Add type parameter for CachingOptimizer

blegat:
This PR adds a type parameter for the optimizer of `CachingOptimizer`. It adds this type parameters in the first place as it will be more useful than the second one (see e.g. `dropoptimizer!`). It also changes the order of the cache and the optimizer fields for consistency.
Note that it does not implement the idea of https://github.com/JuliaOpt/MathOptInterface.jl/issues/321#issuecomment-382729573 of removing the possibility to have no optimizer or to change the optimizer.
Indeed, this would require JuMP to handle these cases and I initially thought that this was the solution because I didn't have the idea of simply using the non concrete type parameter `Union{Nothing, MOI.AbstractOptimizer}` in the case where we want to be able to drop the optimizer and to change it and use the concrete optimizer type when we don't want to change it and the type instability becomes the bottleneck (e.g. https://github.com/JuliaOpt/MathOptInterface.jl/issues/321).

We can now make a similar change in JuMP. Have a parameter for the `moibackend` type that the user can set it to the concrete type of the optimizer to have no overhead (but loosing the features of changing optimizer, droping it, ...)

## Benchmarks
See https://github.com/JuliaOpt/MathOptInterface.jl/pull/323 for benchmarks before this change

With a CachingOptimizer with concrete field type for the optimizer field (i.e. `MockOptimizer{Model{Float64}}`):
```
v0.6
 36.591 ns (4 allocations: 112 bytes)
148.012 ns (8 allocations: 224 bytes)
 24.322 ns (3 allocations: 96 bytes)
 24.871 ns (3 allocations: 96 bytes)
v0.7
 12.130 ns (1 allocation: 48 bytes)
103.417 ns (3 allocations: 112 bytes)
 13.545 ns (1 allocation: 48 bytes)
 11.785 ns
```
With a CachingOptimizer with `Union{Nothing, OptimizerType}` with concrete `OptimizerType` for the `optimizer field:
```
v0.6
 40.083 ns (4 allocations: 112 bytes)
177.939 ns (11 allocations: 288 bytes)
 29.122 ns (3 allocations: 96 bytes)
 25.401 ns (3 allocations: 96 bytes)
v0.7
 13.049 ns (1 allocation: 48 bytes)
109.670 ns (3 allocations: 112 bytes)
 11.501 ns (1 allocation: 48 bytes)
 12.421 ns (1 allocation: 48 bytes)
```

The overhead of the union with `Void` is negligible in v0.7 as suggested by https://github.com/JuliaOpt/MathOptInterface.jl/pull/390#discussion_r195424172 so it seems cleaner to use it

Closes https://github.com/JuliaOpt/MathOptInterface.jl/issues/321

### DIFF
--- a/src/Utilities/cachingoptimizer.jl
+++ b/src/Utilities/cachingoptimizer.jl
@@ -22,9 +22,9 @@ A `CachingOptimizer` has two modes of operation (`CachingOptimizerMode`):
 - `Manual`: The only methods that change the state of the `CachingOptimizer` are [`resetoptimizer!`](@ref), [`dropoptimizer!`](@ref), and [`attachoptimizer!`](@ref). Attempting to perform an operation in the incorrect state results in an error.
 - `Automatic`: The `CachingOptimizer` changes its state when necessary. For example, `optimize!` will automatically call `attachoptimizer!` (an optimizer must have been previously set). Attempting to add a constraint or perform a modification not supported by the optimizer results in a drop to `EmptyOptimizer` mode.
 """
-mutable struct CachingOptimizer{OT, MT<:MOI.ModelLike} <: MOI.AbstractOptimizer
-    optimizer::OT
-    model_cache::MT
+mutable struct CachingOptimizer{OptimizerType, ModelType<:MOI.ModelLike} <: MOI.AbstractOptimizer
+    optimizer::OptimizerType
+    model_cache::ModelType
     state::CachingOptimizerState
     mode::CachingOptimizerMode
     model_to_optimizer_map::IndexMap
--- a/src/Utilities/cachingoptimizer.jl
+++ b/src/Utilities/cachingoptimizer.jl
@@ -23,7 +23,7 @@ A `CachingOptimizer` has two modes of operation (`CachingOptimizerMode`):
 - `Automatic`: The `CachingOptimizer` changes its state when necessary. For example, `optimize!` will automatically call `attachoptimizer!` (an optimizer must have been previously set). Attempting to add a constraint or perform a modification not supported by the optimizer results in a drop to `EmptyOptimizer` mode.
 """
 mutable struct CachingOptimizer{OptimizerType, ModelType<:MOI.ModelLike} <: MOI.AbstractOptimizer
-    optimizer::OptimizerType
+    optimizer::Union{Nothing, OptimizerType}
     model_cache::ModelType
     state::CachingOptimizerState
     mode::CachingOptimizerMode
@@ -35,7 +35,7 @@ mutable struct CachingOptimizer{OptimizerType, ModelType<:MOI.ModelLike} <: MOI.
 end
 
 function CachingOptimizer(model_cache::MOI.ModelLike, mode::CachingOptimizerMode)
-    CachingOptimizer{Union{Nothing,MOI.AbstractOptimizer}, typeof(model_cache)}(nothing, model_cache, NoOptimizer, mode, IndexMap(), IndexMap())
+    CachingOptimizer{MOI.AbstractOptimizer, typeof(model_cache)}(nothing, model_cache, NoOptimizer, mode, IndexMap(), IndexMap())
 end
 
 """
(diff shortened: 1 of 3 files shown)


src/Utilities/cachingoptimizer.jl
```diff
@@ -22,9 +22,9 @@ A `CachingOptimizer` has two modes of operation (`CachingOptimizerMode`):
 - `Manual`: The only methods that change the state of the `CachingOptimizer` are [`resetoptimizer!`](@ref), [`dropoptimizer!`](@ref), and [`attachoptimizer!`](@ref). Attempting to perform an operation in the incorrect state results in an error.
 - `Automatic`: The `CachingOptimizer` changes its state when necessary. For example, `optimize!` will automatically call `attachoptimizer!` (an optimizer must have been previously set). Attempting to add a constraint or perform a modification not supported by the optimizer results in a drop to `EmptyOptimizer` mode.
 """
-mutable struct CachingOptimizer{MT<:MOI.ModelLike} <: MOI.AbstractOptimizer
+mutable struct CachingOptimizer{OT, MT<:MOI.ModelLike} <: MOI.AbstractOptimizer
```
mlubin: Let's try more explicit names for the type parameters. Something like `OptimizerType` and `ModelType` maybe?

src/Utilities/cachingoptimizer.jl
```diff
@@ -34,19 +34,24 @@ mutable struct CachingOptimizer{MT<:MOI.ModelLike} <: MOI.AbstractOptimizer
     # optimizer indices.
 end
 
-CachingOptimizer(model_cache::MOI.ModelLike, mode::CachingOptimizerMode) = CachingOptimizer(model_cache, nothing, NoOptimizer, mode, IndexMap(), IndexMap())
+function CachingOptimizer(model_cache::MOI.ModelLike, mode::CachingOptimizerMode)
+    CachingOptimizer{Union{Nothing,MOI.AbstractOptimizer}, typeof(model_cache)}(nothing, model_cache, NoOptimizer, mode, IndexMap(), IndexMap())
```
mlubin: `Union{Nothing,MOI.AbstractOptimizer}` is a really unusual type parameter. Why not type the `optimizer` as `Union{Nothing, OptimizerType}`? My understanding is that there's very little overhead in 0.7 for `Union{Nothing, T}` if `T` is concrete.

blegat: Indeed, I have updated the description which a benchmark showing this

src/Utilities/cachingoptimizer.jl
```diff
@@ -34,19 +34,24 @@ mutable struct CachingOptimizer{MT<:MOI.ModelLike} <: MOI.AbstractOptimizer
...
 
 """
     CachingOptimizer(model_cache::MOI.ModelLike, optimizer::AbstractOptimizer)
 
 Creates an `CachingOptimizer` in `Automatic` mode, with the optimizer `optimizer`.
 The model_cache manager returned behaves like an `AbstractOptimizer` as long as no
-`CachingOptimizer`-specific functions (e.g. `dropoptimizer!`) are called on it.
+`CachingOptimizer`-specific functions (e.g. `resetoptimizer!`) are called on it.
+The type of the optimizer returned is `CachingOptimizer{typeof(optimizer),
+typeof(model_cache)}` so it does not support the function `dropoptimizer!`
```
mlubin: > so it does not support the function `dropoptimizer!` nor the method `resetoptimizer!.

Is this true? Can't the field be set to nothing?

blegat: Indeed, fixed :)

src/Utilities/cachingoptimizer.jl
```diff
@@ -22,9 +22,9 @@ A `CachingOptimizer` has two modes of operation (`CachingOptimizerMode`):
 - `Manual`: The only methods that change the state of the `CachingOptimizer` are [`resetoptimizer!`](@ref), [`dropoptimizer!`](@ref), and [`attachoptimizer!`](@ref). Attempting to perform an operation in the incorrect state results in an error.
 - `Automatic`: The `CachingOptimizer` changes its state when necessary. For example, `optimize!` will automatically call `attachoptimizer!` (an optimizer must have been previously set). Attempting to add a constraint or perform a modification not supported by the optimizer results in a drop to `EmptyOptimizer` mode.
 """
-mutable struct CachingOptimizer{MT<:MOI.ModelLike} <: MOI.AbstractOptimizer
-    model_cache::MT
-    optimizer::Union{Nothing,MOI.AbstractOptimizer}
+mutable struct CachingOptimizer{OptimizerType, ModelType<:MOI.ModelLike} <: MOI.AbstractOptimizer
+    optimizer::OptimizerType
```
mlubin: Do you plan to update this to `optimizer::Union{OptimizerType, Nothing}`?

blegat: Ups, fixed

Issue #397, opened 2018-06-19
Arbitrary constraints for UniversalFallback


blegat:
I am planning to add this once https://github.com/JuliaOpt/MathOptInterface.jl/pull/388 is merged

PR #399, opened 2018-06-23
update NEWS for 0.4

mlubin:
@blegat @odow @joaquimg let me know if there's a reason to hold off on this release. If we do wait, then we should stick to 0.3 for JuMP-dev.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,17 @@
 MathOptInterface (MOI) release notes
 ====================================
 
+v0.4.0 (June 23, 2018)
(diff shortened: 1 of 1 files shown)


blegat:
No, I may suggest more changes before JuMP-dev (e.g. https://github.com/JuliaOpt/MathOptInterface.jl/issues/397) but they are not breaking so it can be tagged as v0.4.1.
We should definitely tag v0.4 soon if we want solvers to work with v0.4 during the JuMP-dev meetup

PR #409, opened 2018-06-29
Fix names for CachingOptimizer

blegat:
Trying to run unit tests on the caching optimizer showed that there was method missing in the name implementation.
There was also some unit tests that didn't check `config.solve`. I move this check to the subfunction to be sure it is always checked

### DIFF
--- a/src/Test/UnitTests/unit_tests.jl
+++ b/src/Test/UnitTests/unit_tests.jl
@@ -56,6 +56,7 @@ function test_model_solution(model, config;
         constraint_primal = nothing,
         constraint_dual   = nothing
     )
+    config.solve || return
     atol, rtol = config.atol, config.rtol
     MOI.optimize!(model)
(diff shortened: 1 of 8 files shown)


src/Test/UnitTests/unit_tests.jl
```diff
@@ -56,6 +56,7 @@ function test_model_solution(model, config;
         constraint_primal = nothing,
         constraint_dual   = nothing
     )
+    config.solve || return
```
odow: Here is something that should be addressed by the style guide (https://github.com/JuliaOpt/JuMP.jl/pull/1369): I prefer an explicit
```julia
if config.solve
    return
end
```
over the short-circuiting approach.

blegat: I agree, this should be discussed. I am ok with both forms but I think @mlubin likes the `||` and `&&` short circuits.
Let's leave it as is for now until we decide of a convention

PR #411, opened 2018-06-30
Some cleanup

mohamed82008:
Active reading of the package.

### DIFF
--- a/src/MathOptInterface.jl
+++ b/src/MathOptInterface.jl
@@ -47,7 +47,7 @@ function write end
 """
     read!(model::ModelLike, filename::String)
 
-Read the file `filename` into the model `model`. If `m` is non-empty, this may
+Read the file `filename` into the model `model`. If `model` is non-empty, this may
 throw an error.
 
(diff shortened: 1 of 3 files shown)


mohamed82008:
Two comments I made on gitter:
1. Isn't size more appropriate here https://github.com/JuliaOpt/MathOptInterface.jl/blob/585a9593d9d093d60397de22295899b4d2aef6e1/src/sets.jl#L275 to avoid the ambiguity of the use of the word dimension?
2. Unnecessary T? https://github.com/JuliaOpt/MathOptInterface.jl/blob/585a9593d9d093d60397de22295899b4d2aef6e1/src/sets.jl#L400

@odow replied to the second: "I'd say the t is necessary as sos1{int} is not the same as sos1{float64}."

mohamed82008:
So is `SOS1{Int}([1, 2])` different from SOS1{Float64}([1.0, 2.0]) and are these different from `SOS1{Float32}([1f0, 2f0])`?

PR #413, opened 2018-06-30
WIP: VectorAffineBridge

odow:
This PR adds a bridge between `VectorAffineFunction`-in-`{Zeros, Nonnegatives, Nonpositives}` and multiple `ScalarAffineFunction`-in-`{EqualTo, LessThan, GreaterThan}`.

The biggest target for this is LQOI. We should be able to remove a large chunk of code which effectively implements this bridge anyway. I commented out all of the `VectorAffineFunction` code in LQOI, and then ran the GLPK tests with `MOIB.VectorToScalar{Float64}(GLPK.Optimizer())` and everything passed!

- [ ] I'm not sure what the best way to test this is. @blegat?

<s>Just playing around trying to understand the bridge code.

Untested atm.

@blegat is this how it should be?</s>

### DIFF
--- a/src/Bridges/vaftosafbridge.jl
+++ b/src/Bridges/vaftosafbridge.jl
@@ -5,6 +5,10 @@ get_scalar_set(T, ::Type{MOI.Zeros}) = MOI.EqualTo{T}
 get_scalar_set(T, ::Type{MOI.Nonpositives}) = MOI.LessThan{T}
 get_scalar_set(T, ::Type{MOI.Nonnegatives}) = MOI.GreaterThan{T}
 
+get_vector_set(::Type{MOI.EqualTo{T}}) where T = MOI.Zeros
+get_vector_set(::Type{MOI.LessThan{T}}) where T = MOI.Nonpositives
+get_vector_set(::Type{MOI.GreaterThan{T}}) where T = MOI.Nonnegatives
+
 """
     VectorToScalarBridge{S, T}
 
@@ -14,7 +18,8 @@ The `VectorToScalarBridge` splits an `VectorAffineFunction` into a series of
 struct VectorToScalarBridge{T, S<:VectorToScalarBridgeSets} <: AbstractBridge
     constraints::Vector{CI{MOI.ScalarAffineFunction{T}, S}}
 end
-function VectorToScalarBridge(model, func::MOI.VectorAffineFunction{T}, set::S) where {T, S}
+function VectorToScalarBridge(model, func::MOI.VectorAffineFunction{T},
+                              set::S) where {T, S}
     set_type = get_scalar_set(T, S)
     sets = [set_type(-constant) for constant in func.constants]
     functions = [MOI.ScalarAffineFunction{T}(MOI.ScalarAffineTerm{T}[], zero(T))
@@ -25,7 +30,7 @@ function VectorToScalarBridge(model, func::MOI.VectorAffineFunction{T}, set::S) 
         push!(scalar_function.terms, term.scalar_term)
     end
     constraint_indices = MOI.addconstraints!(model, functions, sets)
-    VectorToScalarBridge(constraint_indices)
+    return VectorToScalarBridge(constraint_indices)
 end
 
 function MOI.supportsconstraint(::Type{VectorToScalarBridge{T, S}},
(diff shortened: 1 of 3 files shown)


src/Bridges/vaftosafbridge.jl
```diff
@@ -0,0 +1,86 @@
...
+function VectorAffineBridge{T}(model, f::MOI.VectorAffineFunction{T}, s::VectorAffineBridgeSets) where T
+    set_type = getscalarset(T, typeof(s))
+    sets      = [ set_type(-constant) for constant in f.constants ]
+    functions = [ MOI.ScalarAffineFunction{T}(MOI.ScalarAffineTerm{T}[], zero(T)) for term in f.terms ]
+    @assert dimension(s) == length(sets) == length(functions)
+    for term in f.terms
+        saf = functions[term.output_index]
+        push!(saf.terms, term.scalar_term)
+    end
+    ci = CI{MOI.ScalarAffineFunction{T}, S}[]
```
blegat: could be preallocated

src/Bridges/vaftosafbridge.jl
```diff
@@ -0,0 +1,86 @@
...
+        push!(ci, MOI.addconstraint!(model, func, set))
+    end
+    VectorAffineBridge(ci)
+end
+getscalarset(T, ::Type{MOI.Zeros})        = MOI.EqualTo{T}
+getscalarset(T, ::Type{MOI.Nonpositives}) = MOI.LessThan{T}
+getscalarset(T, ::Type{MOI.Nonnegatives}) = MOI.GreaterThan{T}
+
+
+MOI.supportsconstraint(::Type{VectorAffineBridge{T,S}}, ::Type{MOI.VectorAffineFunction{T}}, ::Type{S}) where {T,S<:VectorAffineBridgeSets} = true
```
blegat: S is different here. The first `S` is the scalar set and the second is the vector set

blegat:
Looks good ! We might wonder whether we should add the bridge with or without the type `S` specified in the LazyBridge. If we decide to add `VectorAffineBridge{T}` in the lazy bridge optimizer, then it would say that it supports the three sets but when you create it you see which set it is and you create a bridge of concrete type.

src/Bridges/vaftosafbridge.jl
```diff
@@ -0,0 +1,86 @@
...
+    sets      = [ set_type(-constant) for constant in f.constants ]
+    functions = [ MOI.ScalarAffineFunction{T}(MOI.ScalarAffineTerm{T}[], zero(T)) for term in f.terms ]
+    @assert dimension(s) == length(sets) == length(functions)
+    for term in f.terms
+        saf = functions[term.output_index]
+        push!(saf.terms, term.scalar_term)
+    end
+    ci = CI{MOI.ScalarAffineFunction{T}, S}[]
+    for (func, set) in zip(functions, sets)
+        push!(ci, MOI.addconstraint!(model, func, set))
```
odow: This should use `MOI.add_constraints!` so that solvers can provide an efficient implementation. @joaquimg

src/Bridges/vaftosafbridge.jl
```diff
@@ -0,0 +1,148 @@
...
+end
+function VectorToScalarBridge{T, ScalarSet}(model,
+                                            func::MOI.VectorAffineFunction{T},
+                                            set::VectorSet) where
+                                            {T, ScalarSet, VectorSet}
+    @assert ScalarSet == get_scalar_set(T, VectorSet)
+    sets = [ScalarSet(-constant) for constant in func.constants]
+    @assert MOI.dimension(set) == length(sets)
+    functions = [MOI.ScalarAffineFunction{T}(MOI.ScalarAffineTerm{T}[], zero(T))
+        for i in 1:MOI.dimension(set)]
```
blegat: Cosmetic/Subjective: I would indent to start `for` just past the `[`

src/Bridges/vaftosafbridge.jl
```diff
@@ -0,0 +1,148 @@
...
+MOI.canget(::MOI.ModelLike, ::MOI.ConstraintSet, ::Type{<:VectorToScalarBridge}) = true
+function MOI.get(model::MOI.ModelLike, ::MOI.ConstraintSet,
+                 bridge::VectorToScalarBridge{T, S}) where {T, S}
+    return get_vector_set(S)
+end
+
+MOI.canget(::MOI.ModelLike, ::MOI.ConstraintFunction, ::Type{<:VectorToScalarBridge}) = true
+function MOI.get(model::MOI.ModelLike, ::MOI.ConstraintFunction,
+                 bridge::VectorToScalarBridge{T, S}) where {T, S}
+    terms = VectorAffineTerm{T}[]
```
blegat: `moivcat` can be helpful here. Using `moivcat` will help adding support for quadratic functions

src/Bridges/vaftosafbridge.jl
```diff
@@ -0,0 +1,148 @@
...
+
+The `VectorToScalarBridge` splits an `VectorAffineFunction` into a series of
+`ScalarAffineFunction`s.
+
+Supported sets are:
+ - `MOI.Zeros` =>`MOI.EqualTo{T}`
+ - `MOI.Nonnegatives` =>`MOI.GreaterThan{T}`
+ - `MOI.Nonpositives` =>`MOI.LessThan{T}`
+"""
+struct VectorToScalarBridge{T, S} <: AbstractBridge
```
blegat: Would be nice to have
```julia
struct VectorToScalarBridge{T, F, S} <: AbstractBridge
    constriants::Vector{CI{F, S}}
end
```
where `F` can also be, e.g. `SingleVariable` and `ScalarQuadraticFunction`.
You can leave it for future work as the tools needed are being added in https://github.com/JuliaOpt/MathOptInterface.jl/pull/455

blegat:
Looks good ! Just missing a few tests :)

src/Bridges/vaftosafbridge.jl
blegat: indenting is where coder can express their artistic inner-self :D 
More seriously, I was confused when reading the line as I thought it was a block (e.g. like `for`, `begin`, `if`, ...) instead of a line-break.
I usually do 4-spaces for blocks and for line-breaks I align with parentheses, brackets, ...

blegat:
@odow This PR has been superseeded by https://github.com/JuliaOpt/MathOptInterface.jl/pull/664

PR #420, opened 2018-07-07
Some cleanup and a few utility functions.

### LINKED ISSUE #422: no-op and redundant code
This part https://github.com/JuliaOpt/MathOptInterface.jl/blob/247c4baf673eb8b1c29177e7f1652a0a78e1d570/src/Utilities/functions.jl#L220 seems like a no-op for all the `moiname`s except for `VectorOfVariables` for which https://github.com/JuliaOpt/MathOptInterface.jl/blob/247c4baf673eb8b1c29177e7f1652a0a78e1d570/src/Utilities/functions.jl#L49 defines the intended behavior.

For instance:
```julia
julia> MOI.Utilities.map_variables(MOI.SingleVariable(MOI.VariableIndex(1)), Dict(MOI.VariableIndex(1)=>MOI.VariableIndex(2)))
MathOptInterface.SingleVariable(MathOptInterface.VariableIndex(1))

julia> MOI.Utilities.map_variables(MOI.VectorOfVariables([MOI.VariableIndex(1)]), Dict(MOI.VariableIndex(1)=>MOI.VariableIndex(2)))
MathOptInterface.VectorOfVariables(MathOptInterface.VariableIndex[MathOptInterface.VariableIndex(2)])
```

mohamed82008:
Well, just some cleanup and a few utility functions, what did you expect!?

Also closes #422.


src/Utilities/functions.jl
```diff
@@ -4,20 +4,20 @@ using Compat.Test
...
     evalvariables(varval::Function, f::AbstractFunction)
 
-Returns the value of function `f` if each variable index `vi` is evaluate as `varval(vi)`.
+Returns the value of function `f` if each variable index `vi` is evaluated as `varval(vi)`.
 """
 function evalvariables end
 evalvariables(varval::Function, f::MOI.SingleVariable) = varval(f.variable)
 evalvariables(varval::Function, f::MOI.VectorOfVariables) = varval.(f.variables)
-evalvariables(varval::Function, f::MOI.ScalarAffineFunction) = sum(evalterm.(varval, f.terms)) + f.constant
+evalvariables(varval::Function, f::MOI.ScalarAffineFunction) = sum(evalterm(varval, t) for t in f.terms) + f.constant
```
odow: Why is this preferred?

src/Utilities/functions.jl
```diff
@@ -135,43 +143,26 @@ If `x` (resp. `y`, `z`) is `VariableIndex(1)` (resp. 2, 3).
...
+    sorted_terms = sort(f.terms, by = termindices)
+    terms = eltype(f.terms)[]
     for t in sorted_terms
-        if !isempty(terms) && t.output_index == last(terms).output_index && t.scalar_term.variable_index == last(terms).scalar_term.variable_index
-            terms[end] = MOI.VectorAffineTerm(t.output_index, MOI.ScalarAffineTerm(terms[end].scalar_term.coefficient + t.scalar_term.coefficient, t.scalar_term.variable_index))
-        elseif !iszero(t.scalar_term.coefficient)
-            if !isempty(terms) && iszero(last(terms).scalar_term.coefficient)
+        last_t = last(terms)
+        tnot0 = !iszero(coefficient(t))
+        if !isempty(terms) && termindices(t) == termindices(last_t) && tnot0
```
odow: Much nicer :)

odow: Although tests are failing somewhere in here

src/Utilities/functions.jl
mohamed82008: It doesn't allocate, it just sums the terms.

src/Utilities/functions.jl
mohamed82008: Should be fixed now.

src/Utilities/functions.jl
mohamed82008: Actually it seems like it does, I will double check.

mohamed82008: Nah was just bechmarking in global scope, the new one doesn't allocate:
```julia
julia> using BenchmarkTools

julia> a = rand(1000);

julia> f(a) = sum(i^2 for i in a)
f (generic function with 1 method)

julia> g(a) = sum(a.^2)
g (generic function with 1 method)

julia> @btime f(a);
  1.422 μs (3 allocations: 48 bytes)

julia> @btime g(a);
  6.199 μs (26 allocations: 8.98 KiB)
```

mohamed82008: Even better is `mapreduce`: 
```julia
julia> @btime mapreduce((x)->(x^2), +, $a)
  89.555 ns (0 allocations: 0 bytes)
329.62559257013555

julia> @btime f($a);
  1.130 μs (3 allocations: 48 bytes)

julia> @btime g($a);
  600.012 ns (1 allocation: 7.94 KiB)
```

mohamed82008:
Seems like it's passing all tests now.

src/Utilities/functions.jl
```diff
@@ -4,20 +4,20 @@ using Compat.Test
...
     evalvariables(varval::Function, f::AbstractFunction)
 
-Returns the value of function `f` if each variable index `vi` is evaluate as `varval(vi)`.
+Returns the value of function `f` if each variable index `vi` is evaluated as `varval(vi)`.
 """
 function evalvariables end
 evalvariables(varval::Function, f::MOI.SingleVariable) = varval(f.variable)
 evalvariables(varval::Function, f::MOI.VectorOfVariables) = varval.(f.variables)
-evalvariables(varval::Function, f::MOI.ScalarAffineFunction) = sum(evalterm.(varval, f.terms)) + f.constant
+evalvariables(varval::Function, f::MOI.ScalarAffineFunction) = mapreduce(t->evalterm(varval, t), +, zero(f.constant), f.terms) + f.constant
```
blegat: There are sometimes issues with `sum` not working over empty collection if I am not mistaken. Is this function tested with empty `ScalarAffineFunction` ?

src/Utilities/functions.jl
```diff
@@ -121,6 +120,15 @@ function Base.getindex(it::ScalarFunctionIterator{MOI.VectorAffineFunction{T}},
     MOI.VectorAffineFunction(terms, constant)
 end
 
+termindices(t::MOI.ScalarAffineTerm) = (t.variable_index.value,)
```
blegat: There should be a comment indicating what it returns

src/functions.jl
```diff
@@ -40,6 +40,7 @@ This function is naturally be used for single variable bounds or integrality con
 struct SingleVariable <: AbstractScalarFunction
     variable::VariableIndex
 end
+output_dimension(::SingleVariable) = 1
```
blegat: It is already implemented for `AbstractScalarFunction` so why is this needed ?

src/Utilities/functions.jl
```diff
@@ -331,11 +308,14 @@ function modifyfunction(f::MOI.ScalarQuadraticFunction, change::MOI.ScalarCoeffi
...
+if VERSION >= v"0.7.0-DEV.2813"
+    using SparseArrays: sparsevec
+end
 function _modifycoefficients(n, terms::Vector{<:MOI.VectorAffineTerm}, variable::MOI.VariableIndex, new_coefficients)
     terms = copy(terms)
     # Maps between rows in the `VectorAffineTerm`s and indices in new_coefficients
-    rowmap = zeros(Int, n)
-    rowmap[map(c -> c[1], new_coefficients)] = 1:length(new_coefficients)
+    # Uses sparse vector since `new_coefficients` also uses a sparse representation 
+    rowmap = sparsevec(getindex.(new_coefficients, 1), 1:length(new_coefficients), n)
```
blegat: The advantage of the approach currently used is the O(1) access to the new coefficient. With a sparse vector, it is O(log(n)) (if it does sorted search which I guess it does). It is also annoying to add a dependency on `SparseArrays` since there are discussion on removing it from Base.
We might use a `Dict` here which offers O(1) access and memory proportional to the number of elements in `new_coefficients`

src/Utilities/functions.jl
```diff
@@ -251,23 +229,22 @@ function test_models_equal(model1::MOI.ModelLike, model2::MOI.ModelLike, variabl
         f2 = MOI.get(model2, MOI.ConstraintFunction(), index2)
         s1 = MOI.get(model1, MOI.ConstraintSet(), index1)
         s2 = MOI.get(model2, MOI.ConstraintSet(), index2)
-        @test isapprox(f1, map_variables(f2, variablemap_2to1))
+        @test isapprox(f1, mapvariables(variablemap_2to1, f2))
```
blegat: Good catch ! Why was this working ? Insufficient testing ?

src/Utilities/functions.jl
```diff
@@ -121,6 +120,15 @@ function Base.getindex(it::ScalarFunctionIterator{MOI.VectorAffineFunction{T}},
     MOI.VectorAffineFunction(terms, constant)
 end
 
+termindices(t::MOI.ScalarAffineTerm) = (t.variable_index.value,)
+termindices(t::MOI.VectorAffineTerm) = (t.output_index, termindices(t.scalar_term)...)
+
+unsafe_add(t1::MOI.ScalarAffineTerm, t2::MOI.ScalarAffineTerm) = MOI.ScalarAffineTerm(t1.coefficient + t2.coefficient, t1.variable_index)
```
blegat: Document what it does and explains that it is unsafe because it assume that the variable index and output_index is the same

src/Utilities/functions.jl
```diff
@@ -135,43 +143,25 @@ If `x` (resp. `y`, `z`) is `VariableIndex(1)` (resp. 2, 3).
...
-    if !isempty(terms) && iszero(last(terms).coefficient)
-        pop!(terms)
-    end
-    SAF{T}(terms, f.constant)
-end
-function canonical(f::VAF{T}) where T
-    sorted_terms = sort(f.terms, by = t -> (t.output_index, t.scalar_term.variable_index.value))
-    terms = MOI.VectorAffineTerm{T}[]
+function canonical(f::Union{SAF, VAF})
+    sorted_terms = sort(f.terms, by = termindices)
```
blegat: Nice work, merging these two methods was definitely something I wanted to happen :)

src/Utilities/functions.jl
blegat: Ups, you are doing the opposite, I misread the diff, you are replacing `sum` with `mapreduce`. Is it because of the issue with empty function ?

src/Utilities/functions.jl
blegat: I see, this is related to https://github.com/JuliaOpt/MathOptInterface.jl/issues/422. You can add `Closes #422` in the comment of this PR then.

odow:
It is probably also good to start adopting https://github.com/JuliaOpt/JuMP.jl/pull/1369

src/Utilities/functions.jl
mohamed82008: Yes but also `mapreduce` is faster for some reason.
```julia
julia> using BenchmarkTools

julia> a = rand(1000);

julia> f(a) = sum(i^2 for i in a)
f (generic function with 1 method)

julia> g(a) = sum(a.^2)
g (generic function with 1 method)

julia> h(a) = mapreduce((x)->(x^2), +, a)
h (generic function with 1 method)

julia> @btime h($a);
  80.205 ns (0 allocations: 0 bytes)

julia> @btime f($a);
  1.094 μs (1 allocation: 16 bytes)

julia> @btime g($a);
  691.091 ns (1 allocation: 7.94 KiB)
```

src/Utilities/functions.jl
mohamed82008: Sure

src/functions.jl
mohamed82008: Oh didn't notice that. I will remove these.

mohamed82008:
Interesting guidelines, some stuff might be negotiable (questionable?) like the ! in function names but overall, pretty nice.

src/Utilities/functions.jl
mohamed82008: Sure

src/Utilities/functions.jl
mohamed82008: Good point, didn't think of that. I will change it to `Dict`.

mohamed82008:
All comments addressed. Not bloat this PR, I will not add more stuff in here.

odow:
> some stuff might be negotiable (questionable?) like the ! in function names

Going forward, no.

mohamed82008:
> Going forward, no.

But this seems to go against `Base`'s assumptions and what most people in the ecosystem do, what's the reason for that?

src/Utilities/functions.jl
blegat: Is it still the case in Julia v0.7 ? The allocation of `g` is also suprisingly high

mohamed82008: The above benchmarks are from v0.7. The allocations are not really that high at all, it is just allocating for the intermediate result of the squares. There are 1000 Float64's, each is 8 bytes, so there should be at least 7.8125 KiB of allocations. Not sure where the remaining comes from, could be some internal broadcast stuff.

mohamed82008:
Kind reminder.

src/Utilities/functions.jl
mlubin: Style nit: use one-line functions only if they fit within one 80-character line. So in this case just do
```julia
function evalvariables(...)
    return ...
end
```

src/Utilities/functions.jl
```diff
@@ -4,20 +4,20 @@ using Compat.Test
...
+evalvariables(varval::Function, f::MOI.ScalarAffineFunction) = mapreduce(t->evalterm(varval, t), +, zero(f.constant), f.terms) + f.constant
 function evalvariables(varval::Function, f::MOI.VectorAffineFunction)
     out = copy(f.constants)
     for t in f.terms
         out[t.output_index] += evalterm(varval, t.scalar_term)
     end
     out
 end
-evalvariables(varval::Function, f::MOI.ScalarQuadraticFunction) = sum(evalterm.(varval, f.affine_terms)) + sum(evalterm.(varval, f.quadratic_terms)) + f.constant
+evalvariables(varval::Function, f::MOI.ScalarQuadraticFunction) = mapreduce(t->evalterm(varval, t), +, zero(f.constant), f.affine_terms) + mapreduce(t->evalterm(varval, t), +, zero(f.constant), f.quadratic_terms) + f.constant
```
mlubin: Same comment about long one-line functions. Prefer to split it up and fit within 80 characters.

src/Utilities/functions.jl
```diff
@@ -121,6 +120,36 @@ function Base.getindex(it::ScalarFunctionIterator{MOI.VectorAffineFunction{T}},
...
+"""
+termindices(t::MOI.ScalarAffineTerm) = (t.variable_index.value,)
+termindices(t::MOI.VectorAffineTerm) = (t.output_index, termindices(t.scalar_term)...)
+
+"""
+    unsafe_add(t1::ScalarAffineTerm, t2::ScalarAffineTerm)
+
+Sums the coefficients of `t1` and `t2` and returns an output `ScalarAffineTerm`. It is unsafe because it uses the `variable_index` of `t1` as the `variable_index` of the output without checking that it is equal to that of `t2`.
+"""
+unsafe_add(t1::MOI.ScalarAffineTerm, t2::MOI.ScalarAffineTerm) = MOI.ScalarAffineTerm(t1.coefficient + t2.coefficient, t1.variable_index)
```
mlubin: Same comment about long one-line functions.

src/Utilities/functions.jl
```diff
@@ -121,6 +120,36 @@ function Base.getindex(it::ScalarFunctionIterator{MOI.VectorAffineFunction{T}},
...
+
+"""
+    unsafe_add(t1::VectorAffineTerm, t2::VectorAffineTerm)
+
+Sums the coefficients of `t1` and `t2` and returns an output `VectorAffineTerm`. It is unsafe because it uses the `output_index` and `variable_index` of `t1` as the `output_index` and `variable_index` of the output term without checking that they are equal to those of `t2`.
+"""
+unsafe_add(t1::MOI.VectorAffineTerm, t2::MOI.VectorAffineTerm) = MOI.VectorAffineTerm(t1.output_index, MOI.ScalarAffineTerm(t1.scalar_term.coefficient + t2.scalar_term.coefficient, t1.scalar_term.variable_index))
```
mlubin: Same comment about long one-line functions.

src/Utilities/functions.jl
```diff
@@ -135,43 +164,25 @@ If `x` (resp. `y`, `z`) is `VariableIndex(1)` (resp. 2, 3).
...
-    end
-    SAF{T}(terms, f.constant)
-end
-function canonical(f::VAF{T}) where T
-    sorted_terms = sort(f.terms, by = t -> (t.output_index, t.scalar_term.variable_index.value))
-    terms = MOI.VectorAffineTerm{T}[]
+function canonical(f::Union{SAF, VAF})
+    sorted_terms = sort(f.terms, by = termindices)
+    terms = eltype(f.terms)[]
     for t in sorted_terms
```
mlubin: This is a confusing and overly complicated loop. What about loading all the terms into a `SortedDict` and then extracting them back?

blegat: When are duplicates, they need to be added together and we need to check that a terme zero coefficient is discarded. Would a dict simplify it so much ?

mohamed82008:
AppVeyor hanging, closing and reopening.

src/Utilities/Utilities.jl
```diff
@@ -14,8 +14,18 @@ const VAF{T} = MOI.VectorAffineFunction{T}
 const SQF{T} = MOI.ScalarQuadraticFunction{T}
 const VQF{T} = MOI.VectorQuadraticFunction{T}
 
+const SAT{T} = MOI.ScalarAffineTerm{T}
```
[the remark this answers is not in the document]
mohamed82008: hmm `ScalarAffineTerm` makes it hard to respect the style guidelines in the function titles, < 80 characters, and makes certain functions asymmetric using abbreviations for some types and full name for others. I am fine with removing them but that may come at the cost of some aesthetics.

src/Utilities/functions.jl
```diff
@@ -4,21 +4,37 @@ using Compat.Test
...
+function evalvariables(varval::Function, f::SQF)
+    init = zero(f.constant)
+    if VERSION >= v"0.7-"
+        lin = mapreduce(t->evalterm(varval, t), +, f.affine_terms, init=init)
+        quad = mapreduce(t->evalterm(varval, t), +, f.quadratic_terms, init=init)
+    else
+        lin = mapreduce(t->evalterm(varval, t), +, init, f.affine_terms)
+        quad = mapreduce(t->evalterm(varval, t), +, init, f.quadratic_terms)
+    end
+    return lin + quad + f.constant
```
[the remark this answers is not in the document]
mohamed82008: It infers fine, but added it anyways.

PR #427, opened 2018-07-18
Add return type annotation to `get(::AbstractModel, ::ListOfConstraintIndices)`

### LINKED ISSUE #426: Inference issues with `broadcastvcat` cause MethodErrors
`MOI.get(::AbstractModel, ::MOI.ListOfConstraintIndices)` calls `broadcastvcat` on the internal model: https://github.com/JuliaOpt/MathOptInterface.jl/blob/851125c68541bfa7582eab3bf03aa84635114bcb/src/Utilities/model.jl#L272

However, it seems that that call is not always inferred correctly, which can result in the returned vector to be a `Vector{Any}`. This causes problems downstream here: https://github.com/JuliaOpt/MathOptInterface.jl/blob/851125c68541bfa7582eab3bf03aa84635114bcb/src/Utilities/copy.jl#L77

The issue is that `cis_src` is passed to `passattributes!`, which expects a `Vector{<:ConstraintIndex}`, not a `Vector{Any}`, and the result is a MethodError. 

I'll try to find a nice reproducible example and figure out what's causing the inference failure.

rdeits:
Fixes #426 

The actual root cause of #426 was that we were accidentally mis-using `MOI.@model`, and we had a model type that tried to support `ScalarQuadraticFunction` but didn't support any scalar sets at all. The result was that the model had a `.scalarquadraticfunction`, but that field itself was completely empty. So `vcat()` across the (empty set of) fields resulted in an un-inferable return type. 

The *correct* fix was to update our `@model` usage. But I still think a return type annotation is useful here, as `broadcastvcat` is doing some pretty complicated things and failing to infer this result type actually causes errors, not just slowness, in the downstream code.

### DIFF
--- a/src/Utilities/model.jl
+++ b/src/Utilities/model.jl
@@ -268,7 +268,7 @@ function MOI.get(model::AbstractModel, loc::MOI.ListOfConstraints)
     broadcastvcat(_getloc, model)
 end
 
-function MOI.get(model::AbstractModel, loc::MOI.ListOfConstraintIndices)
+function (MOI.get(model::AbstractModel, loc::MOI.ListOfConstraintIndices{F, S})::Vector{CI{F, S}}) where {F, S}
     broadcastvcat(constrs -> _getlocr(constrs, loc), model)
 end
 


blegat:
Thanks for digging into this. It would help to have a test and have a comment in the code explaining why the conversion of the return value is needed.
I guess simply calling `ListOfConstraintIndices{F, S}` for `F` and `S` not supported the model would also return a `Vector{Any}`.

In view of this, `broadcastvcat` might instead be `broadcast_append` or `broadcastpush` where the a vector of the right eltype is given as a first argument. This issue might be present for all the uses of `broadcastvcat`.
This might be considered out of scope for this PR though, just throwing out the idea.

mlubin:
Yes, given that we don't use return type annotations consistently in MOI, please add a comment explaining why it's there.

rdeits:
It's not as necessary as I'd thought, so i think I'll just close this to keep things simple.

PR #430, opened 2018-07-19
Add in-place version of canonical() and more tests

### LINKED ISSUE #429: Improving canonicalization
Per some discussion on the jump-dev gitter, there are a number of places where it's easy to make mistakes in solver implementations when affine expressions have duplicate terms (that is, terms with the same variable index). We can remove any duplicates safely by calling `MOI.Utilities.canonical`, but the current version of that function is somewhat slow and allocates a lot of memory. I tried implementing a couple of options to improve it. 

The first idea is the one @mlubin  and I were discussing, using a Dict to hold the variable => coefficient mapping: 

```julia
struct DeduplicationCache{T}
    coefficients::Dict{MOI.VariableIndex, T}
    DeduplicationCache{T}() where {T} = new{T}(Dict{MOI.VariableIndex, T}())
end
function deduplicate_with_cache!(f::MOI.ScalarAffineFunction{T}, cache::DeduplicationCache) where {T}
    empty!(cache.coefficients)
    for term in f.terms
        cache.coefficients[term.variable_index] = term.coefficient + get(cache.coefficients, term.variable_index, zero(valtype(cache.coefficients)))
    end
    empty!(f.terms)
    for (var, coeff) in cache.coefficients
        push!(f.terms, MOI.ScalarAffineTerm(coeff, var))
    end
    f
end
function deduplicate_with_cache(f::MOI.ScalarAffineFunction, cache::DeduplicationCache)
    result = MOI.ScalarAffineFunction(copy(f.terms), f.constant)
    deduplicate_with_cache!(result, cache)
end
```

That turned out to perform worse than I was hoping, and it also doesn't sort the resulting variables so it can't replace `canonical()`. However, I also came up with a method more similar to the current `MOI.Utilities.canonical`, but doing all of its work in-place: 

```julia
function my_canonical(f::MOI.ScalarAffineFunction)
    result = MOI.ScalarAffineFunction(copy(f.terms), f.constant)
    my_canonical!(f)
end

function my_canonical!(f::MOI.ScalarAffineFunction)
    if length(f.terms) > 1
        sort!(f.terms, by=t -> t.variable_index.value, alg=QuickSort)
        i1 = 1
        i2 = 2
        while true
            if f.terms[i1].variable_index.value == f.terms[i2].variable_index.value
                f.terms[i1] = MOI.ScalarAffineTerm(f.terms[i1].coefficient + f.terms[i2].coefficient, f.terms[i1].variable_index)
                i2 += 1
            else
                f.terms[i1 + 1] = f.terms[i2]
                i1 += 1
                i2 += 1
            end
            if i2 > length(f.terms)
                break
            end
        end
        resize!(f.terms, i1)
    end
    f
end
```

Benchmarking the various approaches gives: 

### Julia v0.6.3

Time:

| n | MOI.Utilities.canonical | my_canonical | my_canonical! | deduplicate_with_static_cache | deduplicate_with_static_cache! |
| --- | --- | --- | --- | --- | --- |
| 10 | 2.017 μs | 929.000 ns | 894.000 ns | 4.207 μs | 1.299 μs  |
| 100 | 6.031 μs | 3.963 μs | 3.695 μs | 11.210 μs | 8.966 μs  |
| 1000 | 57.961 μs | 41.801 μs | 39.664 μs | 91.052 μs | 87.788 μs  |
| 10000 | 706.862 μs | 527.163 μs | 507.943 μs | 965.707 μs | 951.336 μs  |
| 100000 | 18.166 ms | 13.026 ms | 7.255 ms | 11.913 ms | 11.616 ms  |
| 1000000 | 172.884 ms | 164.866 ms | 85.844 ms | 232.217 ms | 224.895 ms  |
| 10000000 | 1.529 s | 1.011 s | 896.141 ms | 3.537 s | 3.410 s  |

Memory:

| n | MOI.Utilities.canonical | my_canonical | my_canonical! | deduplicate_with_static_cache | deduplicate_with_static_cache! |
| --- | --- | --- | --- | --- | --- |
| 10 | 896 bytes | 352 bytes | 112 bytes | 592 bytes | 320 bytes  |
| 100 | 5.34 KiB | 1.88 KiB | 112 bytes | 4.92 KiB | 3.13 KiB  |
| 1000 | 56.58 KiB | 15.86 KiB | 112 bytes | 47.03 KiB | 31.25 KiB  |
| 10000 | 491.66 KiB | 156.44 KiB | 112 bytes | 468.86 KiB | 312.50 KiB  |
| 100000 | 4.29 MiB | 1.53 MiB | 112 bytes | 4.58 MiB | 3.05 MiB  |
| 1000000 | 39.89 MiB | 15.26 MiB | 112 bytes | 45.78 MiB | 30.52 MiB  |
| 10000000 | 357.88 MiB | 152.59 MiB | 112 bytes | 457.76 MiB | 305.18 MiB  |

### Julia 0.7.0-beta2.43

Time: 

| n | MOI.Utilities.canonical | my_canonical | my_canonical! | deduplicate_with_static_cache | deduplicate_with_static_cache! |
| --- | --- | --- | --- | --- | --- |
| 10 | 456.000 ns | 284.000 ns | 240.000 ns | 3.825 μs | 1.152 μs  |
| 100 | 5.134 μs | 3.557 μs | 3.185 μs | 12.268 μs | 10.164 μs  |
| 1000 | 50.430 μs | 36.937 μs | 34.843 μs | 77.993 μs | 75.854 μs  |
| 10000 | 636.244 μs | 470.540 μs | 452.377 μs | 856.442 μs | 831.404 μs  |
| 100000 | 8.529 ms | 11.648 ms | 11.412 ms | 10.694 ms | 9.332 ms  |
| 1000000 | 107.026 ms | 139.062 ms | 138.975 ms | 265.215 ms | 196.011 ms  |
| 10000000 | 1.441 s | 915.029 ms | 1.121 s | 3.128 s | 2.987 s  |

Memory:

| n | MOI.Utilities.canonical | my_canonical | my_canonical! | deduplicate_with_static_cache | deduplicate_with_static_cache! |
| --- | --- | --- | --- | --- | --- |
| 10 | 528 bytes | 240 bytes | 0 bytes | 592 bytes | 320 bytes  |
| 100 | 4.98 KiB | 1.77 KiB | 0 bytes | 4.92 KiB | 3.13 KiB  |
| 1000 | 56.22 KiB | 15.75 KiB | 0 bytes | 47.03 KiB | 31.25 KiB  |
| 10000 | 491.30 KiB | 156.33 KiB | 0 bytes | 468.86 KiB | 312.50 KiB  |
| 100000 | 4.29 MiB | 1.53 MiB | 0 bytes | 4.58 MiB | 3.05 MiB  |
| 1000000 | 39.89 MiB | 15.26 MiB | 0 bytes | 45.78 MiB | 30.52 MiB  |
| 10000000 | 357.88 MiB | 152.59 MiB | 0 bytes | 457.76 MiB | 305.18 MiB  |

So it looks like the updated `my_canonical` function is faster than both the existing function and the Dict version. That's particularly nice because it means we don't have to bother with creating and hanging onto that Dict cache. 

Here's a gist to reproduce these results: https://gist.github.com/rdeits/713c3e1eddde8cdc4fee028b2bc0a98c

rdeits:
Fixes #429 

This implements the in-place canonicalization I suggested over in #429 with some cleanup to make it a bit more readable. It adds two new functions: `iscanonical()` which tests if an AffineFunction is canonical and `canonical!` which converts an AffineFunction to canonical form in-place. ~~In Julia v0.7 and above, the in-place `canonical!` should allocate zero memory (on v0.6, it allocates a tiny amount for keyword arguments).~~ Edit: the in-place version now has zero allocations on 0.6 and 0.7 (thanks @tkoolen!) 

Another significant change is that if a function is already in canonical form, it is not copied or modified. This should make things a bit faster for modeling layers like JuMP that already provide canonical functions.

### DIFF
--- a/src/Utilities/functions.jl
+++ b/src/Utilities/functions.jl
@@ -201,13 +201,6 @@ Finds the coefficient associated with the term `t`.
 coefficient(t::MOI.ScalarAffineTerm) = t.coefficient
 coefficient(t::MOI.VectorAffineTerm) = t.scalar_term.coefficient
 
-"""
-    copy(f::Union{ScalarAffineFunction, VectorAffineFunction})
-
-Return a new affine function with a shallow copy of the terms and constant(s)
-from `f`.
-"""
-Base.copy(f::F) where {F <: Union{SAF, VAF}} = F(copy(f.terms), copy(_constant(f)))
 
 """
     iscanonical(f::Union{ScalarAffineFunction, VectorAffineFunction})
--- a/src/Utilities/functions.jl
+++ b/src/Utilities/functions.jl
@@ -122,6 +122,36 @@ function Base.getindex(it::ScalarFunctionIterator{MOI.VectorAffineFunction{T}}, 
 end
 
 """
+    iscanonical(f::AbstractFunction)
+
+Returns a Bool indicating whether the function is in canonical form.
+See [`canonical`](@ref).
+"""
+iscanonical(f::MOI.VectorAffineFunction) = iscanonical(f.terms,
+                                                       t -> (t.output_index, t.scalar_term.variable_index.value),
+                                                       t -> !iszero(t.scalar_term.coefficient))
+iscanonical(f::MOI.ScalarAffineFunction) = iscanonical(f.terms,
+                                                       t -> t.variable_index.value,
+                                                       t -> !iszero(t.coefficient))
+function iscanonical(x::AbstractVector, key, keep)
+    if isempty(x)
+        return true
+    end
+    @inbounds for i in 1:(length(x) - 1)
+        if key(x[i + 1]) <= key(x[i])
+            return false
+        end
+        if !keep(x[i])
+            return false
+        end
+    end
+    if !keep(x[end])
+        return false
+    end
+    return true
+end
+
+"""
     canonical(f::AbstractFunction)
 
 Returns the function in a canonical form, i.e.
@@ -135,44 +165,79 @@ If `x` (resp. `y`, `z`) is `VariableIndex(1)` (resp. 2, 3).
 The canonical representation of `ScalarAffineFunction([y, x, z, x, z], [2, 1, 3, -2, -3], 5)` is `ScalarAffineFunction([x, y], [-1, 2], 5)`.
 
 """
-function canonical(f::SAF{T}) where T
-    sorted_terms = sort(f.terms, by = t -> t.variable_index.value)
-    terms = MOI.ScalarAffineTerm{T}[]
-    for t in sorted_terms
-        if !isempty(terms) && t.variable_index == last(terms).variable_index
-            terms[end] = MOI.ScalarAffineTerm(terms[end].coefficient + t.coefficient, t.variable_index)
-        elseif !iszero(t.coefficient)
-            if !isempty(terms) && iszero(last(terms).coefficient)
-                terms[end] = t
+function canonical(f::MOI.ScalarAffineFunction)
+    if iscanonical(f)
+        f
+    else
+        canonical!(MOI.ScalarAffineFunction(copy(f.terms), f.constant))
+    end
+end
+
+function canonical(f::MOI.VectorAffineFunction)
+    if iscanonical(f)
+        f
+    else
+        canonical!(MOI.VectorAffineFunction(copy(f.terms), f.constants))
+    end
+end
+
+
+"""
+    canonical!(f::AbstractFunction)
+
+Convert a function to canonical form in-place, without allocating a copy to hold the result.
+See [`canonical`](@ref).
+"""
+function canonical!(f::MOI.ScalarAffineFunction)
+    merge_duplicates!(f.terms,
+                      t -> t.variable_index.value,
+                      t -> !iszero(t.coefficient),
+                      (t1, t2) -> MOI.ScalarAffineTerm(t1.coefficient + t2.coefficient, t1.variable_index))
+    f
+end
+
+
+function canonical!(f::MOI.VectorAffineFunction)
+    merge_duplicates!(f.terms,
+                      t -> (t.output_index, t.scalar_term.variable_index.value),
+                      t -> !iszero(t.scalar_term.coefficient),
+                      (t1, t2) -> MOI.VectorAffineTerm(t1.output_index,
+                                                       MOI.ScalarAffineTerm(t1.scalar_term.coefficient + t2.scalar_term.coefficient,
+                                                                            t1.scalar_term.variable_index)))
+    f
+end
+
+function merge_duplicates!(x::AbstractVector, key, keep, reduction, alg = QuickSort)
+    if length(x) == 1
+        if !keep(first(x))
+            empty!(x)
+        end
+    elseif length(x) > 1
+        sort!(x, by = key, alg = alg)
+        i1 = 1
+        i2 = 2
+        @inbounds while i2 <= length(x)
+            if key(x[i1]) == key(x[i2])
+                x[i1] = reduction(x[i1], x[i2])
+                i2 += 1
             else
-                push!(terms, t)
+                if !keep(x[i1])
+                    x[i1] = x[i2]
+                else
+                    x[i1 + 1] = x[i2]
+                    i1 += 1
+                end
+                i2 += 1
             end
         end
-    end
-    if !isempty(terms) && iszero(last(terms).coefficient)
-        pop!(terms)
-    end
-    SAF{T}(terms, f.constant)
-end
-function canonical(f::VAF{T}) where T
-    sorted_terms = sort(f.terms, by = t -> (t.output_index, t.scalar_term.variable_index.value))
-    terms = MOI.VectorAffineTerm{T}[]
-    for t in sorted_terms
-        if !isempty(terms) && t.output_index == last(terms).output_index && t.scalar_term.variable_index == last(terms).scalar_term.variable_index
-            terms[end] = MOI.VectorAffineTerm(t.output_index, MOI.ScalarAffineTerm(terms[end].scalar_term.coefficient + t.scalar_term.coefficient, t.scalar_term.variable_index))
-        elseif !iszero(t.scalar_term.coefficient)
-            if !isempty(terms) && iszero(last(terms).scalar_term.coefficient)
-                terms[end] = t
-            else
-                push!(terms, t)
-            end
+        if !keep(x[i1])
+            i1 -= 1
         end
+        resize!(x, i1)
     end
-    if !isempty(terms) && iszero(last(terms).scalar_term.coefficient)
-        pop!(terms)
-    end
-    VAF{T}(terms, f.constants)
+    x
 end
+
 
 function test_variablenames_equal(model, variablenames)
     seen_name = Dict(name => false for name in variablenames)
--- a/src/Utilities/functions.jl
+++ b/src/Utilities/functions.jl
@@ -202,12 +202,20 @@ coefficient(t::MOI.ScalarAffineTerm) = t.coefficient
 coefficient(t::MOI.VectorAffineTerm) = t.scalar_term.coefficient
 
 """
+    copy(f::Union{ScalarAffineFunction, VectorAffineFunction})
+
+Return a new affine function with a shallow copy of the terms vector and constant(s)
+from `f`.
+"""
+Base.copy(f::F) where {F <: Union{SAF, VAF}} = F(copy(f.terms), copy(_constant(f)))
+
+"""
     iscanonical(f::Union{ScalarAffineFunction, VectorAffineFunction})
 
 Returns a Bool indicating whether the function is in canonical form.
 See [`canonical`](@ref).
 """
-function iscanonical(f::Union{MOI.ScalarAffineFunction, MOI.VectorAffineFunction})
+function iscanonical(f::Union{SAF, VAF})
     iscanonical(f.terms, termindices, t -> !iszero(coefficient(t)))
 end
 function iscanonical(x::AbstractVector, by, keep)
@@ -242,21 +250,7 @@ If `x` (resp. `y`, `z`) is `VariableIndex(1)` (resp. 2, 3).
 The canonical representation of `ScalarAffineFunction([y, x, z, x, z], [2, 1, 3, -2, -3], 5)` is `ScalarAffineFunction([x, y], [-1, 2], 5)`.
 
 """
-function canonical(f::MOI.ScalarAffineFunction)
-    if iscanonical(f)
-        f
-    else
-        canonicalize!(MOI.ScalarAffineFunction(copy(f.terms), f.constant))
-    end
-end
-
-function canonical(f::MOI.VectorAffineFunction)
-    if iscanonical(f)
-        f
-    else
-        canonicalize!(MOI.VectorAffineFunction(copy(f.terms), f.constants))
-    end
-end
+canonical(f::Union{SAF, VAF}) = canonicalize!(copy(f))
 
 """
     canonicalize!(f::Union{ScalarAffineFunction, VectorAffineFunction})
@@ -264,7 +258,7 @@ end
 Convert a function to canonical form in-place, without allocating a copy to hold the result.
 See [`canonical`](@ref).
 """
-function canonicalize!(f::Union{MOI.ScalarAffineFunction, MOI.VectorAffineFunction})
+function canonicalize!(f::Union{SAF, VAF})
     sort_and_compress!(f.terms, termindices, t -> !iszero(coefficient(t)), unsafe_add)
     f
 end
@@ -278,7 +272,7 @@ and remove any entries for which `keep(x[i]) == false`. This may result in `x` b
 a shorter length.
 """
 function sort_and_compress!(x::AbstractVector, by, keep, combine)
-    if length(x) > 1
+    if length(x) > 0
         sort!(x, QuickSort, Base.Order.ord(isless, by, false, Base.Sort.Forward))
         i1 = firstindex(x)
         for i2 in eachindex(x)[2:end]
--- a/src/Utilities/functions.jl
+++ b/src/Utilities/functions.jl
@@ -267,7 +267,6 @@ See [`canonical`](@ref).
 """
 function canonicalize!(f::Union{SAF, VAF})
     sort_and_compress!(f.terms, termindices, t -> !iszero(coefficient(t)), unsafe_add)
-    f
 end
 
 """
@@ -299,7 +298,7 @@ function sort_and_compress!(x::AbstractVector, by, keep, combine)
         end
         resize!(x, i1)
     end
-    x
+    return x
 end
 
 
--- a/src/Utilities/functions.jl
+++ b/src/Utilities/functions.jl
@@ -216,20 +216,27 @@ Returns a Bool indicating whether the function is in canonical form.
 See [`canonical`](@ref).
 """
 function iscanonical(f::Union{SAF, VAF})
-    iscanonical(f.terms, termindices, t -> !iszero(coefficient(t)))
+    is_strictly_sorted(f.terms, termindices, t -> !iszero(coefficient(t)))
 end
-function iscanonical(x::AbstractVector, by, keep)
+
+"""
+    is_strictly_sorted(x::AbstractVector, by, filter)
+
+Returns `true` if `by(x[i]) < by(x[i + 1]` and `filter(x[i]) == true` for
+all indices i.
+"""
+function is_strictly_sorted(x::AbstractVector, by, filter)
     if isempty(x)
         return true
     end
-    if !keep(first(x))
+    if !filter(first(x))
         return false
     end
     for i in eachindex(x)[2:end]
         if by(x[i]) <= by(x[i - 1])
             return false
         end
-        if !keep(x[i])
+        if !filter(x[i])
             return false
         end
     end
--- a/src/Utilities/functions.jl
+++ b/src/Utilities/functions.jl
@@ -133,12 +133,12 @@ iscanonical(f::MOI.VectorAffineFunction) = iscanonical(f.terms,
 iscanonical(f::MOI.ScalarAffineFunction) = iscanonical(f.terms,
                                                        t -> t.variable_index.value,
                                                        t -> !iszero(t.coefficient))
-function iscanonical(x::AbstractVector, key, keep)
+function iscanonical(x::AbstractVector, by, keep)
     if isempty(x)
         return true
     end
     @inbounds for i in 1:(length(x) - 1)
-        if key(x[i + 1]) <= key(x[i])
+        if by(x[i + 1]) <= by(x[i])
             return false
         end
         if !keep(x[i])
--- a/src/Utilities/functions.jl
+++ b/src/Utilities/functions.jl
@@ -260,6 +260,7 @@ See [`canonical`](@ref).
 """
 function canonicalize!(f::Union{SAF, VAF})
     sort_and_compress!(f.terms, termindices, t -> !iszero(coefficient(t)), unsafe_add)
+    return f
 end
 
 """
--- a/src/Utilities/functions.jl
+++ b/src/Utilities/functions.jl
@@ -204,7 +204,7 @@ coefficient(t::MOI.VectorAffineTerm) = t.scalar_term.coefficient
 """
     copy(f::Union{ScalarAffineFunction, VectorAffineFunction})
 
-Return a new affine function with a shallow copy of the terms vector and constant(s)
+Return a new affine function with a shallow copy of the terms and constant(s)
 from `f`.
 """
 Base.copy(f::F) where {F <: Union{SAF, VAF}} = F(copy(f.terms), copy(_constant(f)))
--- a/src/Utilities/functions.jl
+++ b/src/Utilities/functions.jl
@@ -207,18 +207,18 @@ function canonical!(f::MOI.VectorAffineFunction)
     f
 end
 
-function merge_duplicates!(x::AbstractVector, key, keep, reduction, alg = QuickSort)
+function merge_duplicates!(x::AbstractVector, by, keep, combine)
     if length(x) == 1
         if !keep(first(x))
             empty!(x)
         end
     elseif length(x) > 1
-        sort!(x, by = key, alg = alg)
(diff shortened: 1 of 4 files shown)


rdeits:
@tkoolen I think the in-place `MOI.Utilities.canonical!` should work well for SimpleQP, without you needing to keep a `Dict` around.

mohamed82008:
Touches same code as https://github.com/JuliaOpt/MathOptInterface.jl/pull/420.

src/Utilities/functions.jl
```diff
@@ -135,45 +165,80 @@ If `x` (resp. `y`, `z`) is `VariableIndex(1)` (resp. 2, 3).
...
+    merge_duplicates!(f.terms,
+                      t -> (t.output_index, t.scalar_term.variable_index.value),
+                      t -> !iszero(t.scalar_term.coefficient),
+                      (t1, t2) -> MOI.VectorAffineTerm(t1.output_index,
+                                                       MOI.ScalarAffineTerm(t1.scalar_term.coefficient + t2.scalar_term.coefficient,
+                                                                            t1.scalar_term.variable_index)))
+    f
+end
+
+function merge_duplicates!(x::AbstractVector, by, keep, combine)
```
mlubin: Where is this function coming from? (It looks like it's copied from Base julia.) It also needs a docstring.

test/functions.jl
```diff
@@ -204,4 +204,103 @@
...
+        _isapprox(f1::MOI.ScalarAffineFunction, f2::MOI.ScalarAffineFunction) = _var_idx.(f1.terms) == _var_idx.(f2.terms) && _coeff.(f1.terms) ≈ _coeff.(f2.terms) && f1.constant ≈ f2.constant
+        _var_idx(f::MOI.VectorAffineTerm) = f.scalar_term.variable_index
+        _coeff(f::MOI.VectorAffineTerm) = f.scalar_term.coefficient
+        _output_idx(f::MOI.VectorAffineTerm) = f.output_index
+        _isapprox(f1::MOI.VectorAffineFunction, f2::MOI.VectorAffineFunction) = _var_idx.(f1.terms) == _var_idx.(f2.terms) && _output_idx.(f1.terms) == _output_idx.(f2.terms) && _coeff.(f1.terms) ≈ _coeff.(f2.terms) && f1.constants ≈ f2.constants
+
+        srand(1)
+        @testset "ScalarAffineFunction" begin
+            for i in 1:1000
+                f = random_scalar_affine_function(rand(1:1000))
```
mlubin: It's better to test with interesting corner cases with known correct solutions rather than random input comparing with the old version of the code.

This setup has readability issues because it requires a bunch of work to reproduce and understand what happened when a test failed. It doesn't print the offending function and the observed and expected results. The old reference code is already complicated enough that I can't eyeball to verify that it's correct, so a reader shouldn't have confidence that this test is checking what it claims to test.

src/Utilities/functions.jl
```diff
@@ -135,45 +165,80 @@ If `x` (resp. `y`, `z`) is `VariableIndex(1)` (resp. 2, 3).
...
+    if length(x) == 1
+        if !keep(first(x))
+            empty!(x)
+        end
+    elseif length(x) > 1
+        sort!(x, QuickSort, Base.Order.ord(isless, by, false, Base.Sort.Forward))
+        i1 = 1
+        i2 = 2
```
tkoolen: Assumes `1:n` indexing, while `x::AbstractVector`.

src/Utilities/functions.jl
```diff
@@ -135,45 +165,80 @@ If `x` (resp. `y`, `z`) is `VariableIndex(1)` (resp. 2, 3).
...
+        @inbounds while i2 <= length(x)
```
tkoolen: `i2` gets incremented in both branches, so I'd just use `for i2 in eachindex(x)[2 : end]`.

src/Utilities/functions.jl
```diff
@@ -121,6 +121,36 @@ function Base.getindex(it::ScalarFunctionIterator{MOI.VectorAffineFunction{T}},
...
+
+Returns a Bool indicating whether the function is in canonical form.
+See [`canonical`](@ref).
+"""
+iscanonical(f::MOI.VectorAffineFunction) = iscanonical(f.terms,
+                                                       t -> (t.output_index, t.scalar_term.variable_index.value),
+                                                       t -> !iszero(t.scalar_term.coefficient))
+iscanonical(f::MOI.ScalarAffineFunction) = iscanonical(f.terms,
+                                                       t -> t.variable_index.value,
+                                                       t -> !iszero(t.coefficient))
```
tkoolen: What about `ScalarQuadraticFunction`, `VectorQuadraticFunction`?

src/Utilities/functions.jl
tkoolen: `merge_duplicates!` -> `combine_elements!` / `combine_terms!`?

tkoolen: Or `compress!` / `sort_and_compress!`?

rdeits: @mlubin It came from...me writing it yesterday? Sure, I can add a docstring

rdeits: `sort_and_compress!` seems very clear

src/Utilities/functions.jl
mohamed82008: Can use the new `termindices` and `coefficient` functions to make this generic.

src/Utilities/functions.jl
```diff
@@ -135,45 +165,80 @@ If `x` (resp. `y`, `z`) is `VariableIndex(1)` (resp. 2, 3).
...
-                terms[end] = t
+
+
+"""
+    canonical!(f::AbstractFunction)
+
+Convert a function to canonical form in-place, without allocating a copy to hold the result.
+See [`canonical`](@ref).
+"""
+function canonical!(f::MOI.ScalarAffineFunction)
```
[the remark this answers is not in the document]
rdeits: sure

src/Utilities/functions.jl
```diff
@@ -135,45 +165,80 @@ If `x` (resp. `y`, `z`) is `VariableIndex(1)` (resp. 2, 3).
...
     end
-    if !isempty(terms) && iszero(last(terms).coefficient)
-        pop!(terms)
+end
+
+function canonical(f::MOI.VectorAffineFunction)
+    if iscanonical(f)
+        f
+    else
+        canonical!(MOI.VectorAffineFunction(copy(f.terms), f.constants))
```
mohamed82008: Should probably copy `f.constants` too? This part was there before, but might as well change it.

src/Utilities/functions.jl
```diff
@@ -135,45 +165,80 @@ If `x` (resp. `y`, `z`) is `VariableIndex(1)` (resp. 2, 3).
...
+function canonical!(f::MOI.VectorAffineFunction)
+    merge_duplicates!(f.terms,
+                      t -> (t.output_index, t.scalar_term.variable_index.value),
+                      t -> !iszero(t.scalar_term.coefficient),
+                      (t1, t2) -> MOI.VectorAffineTerm(t1.output_index,
+                                                       MOI.ScalarAffineTerm(t1.scalar_term.coefficient + t2.scalar_term.coefficient,
+                                                                            t1.scalar_term.variable_index)))
+    f
+end
+
```
mohamed82008: Using `unsafe_add`, `coefficient` and `termindices` functions, these 2 `canonical!`s can be combined.

src/Utilities/functions.jl
```diff
@@ -121,6 +121,36 @@ function Base.getindex(it::ScalarFunctionIterator{MOI.VectorAffineFunction{T}},
...
+Returns a Bool indicating whether the function is in canonical form.
+See [`canonical`](@ref).
+"""
+iscanonical(f::MOI.VectorAffineFunction) = iscanonical(f.terms,
+                                                       t -> (t.output_index, t.scalar_term.variable_index.value),
+                                                       t -> !iszero(t.scalar_term.coefficient))
+iscanonical(f::MOI.ScalarAffineFunction) = iscanonical(f.terms,
+                                                       t -> t.variable_index.value,
+                                                       t -> !iszero(t.coefficient))
+function iscanonical(x::AbstractVector, by, keep)
```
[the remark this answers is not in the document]
rdeits: In OSQP you can just do 

```julia
MOI.Utilities.sort_and_compress!(f.terms, MOI.Utilities.termindices, t -> true, MOI.Utilities.unsafe_add)
```

src/Utilities/functions.jl
rdeits: fixed

src/Utilities/functions.jl
rdeits: fixed

src/Utilities/functions.jl
rdeits: I don't know what we'd consider canonical form for them, as it was never supported before.

src/Utilities/functions.jl
rdeits: Yup, I just saw those from your PR and rewrote the function to use them :)

src/Utilities/functions.jl
rdeits: Yeah, I was on the fence about that. The constants are never modified during canonicalization, but I guess they might be changed by something downstream. 

I could just implement: 

```
Base.copy(f::ScalarAffineFunction) = <copy terms and constant>
Base.copy(f::VectorAffineFunction) = <copy terms and constants>
```

and then re-write `canonical(f) = canonicalize!(copy(f))` which would work for both

src/Utilities/functions.jl
```diff
@@ -241,65 +246,51 @@ function canonical(f::MOI.ScalarAffineFunction)
...
+        canonicalize!(MOI.ScalarAffineFunction(copy(f.terms), f.constant))
     end
 end
 
 function canonical(f::MOI.VectorAffineFunction)
     if iscanonical(f)
         f
     else
-        canonical!(MOI.VectorAffineFunction(copy(f.terms), f.constants))
+        canonicalize!(MOI.VectorAffineFunction(copy(f.terms), f.constants))
```
mohamed82008: Can use the `_constant` function to merge this and the previous `canonical`s.

src/Utilities/functions.jl
```diff
@@ -121,6 +121,36 @@ function Base.getindex(it::ScalarFunctionIterator{MOI.VectorAffineFunction{T}},
...
+                                                       t -> (t.output_index, t.scalar_term.variable_index.value),
+                                                       t -> !iszero(t.scalar_term.coefficient))
+iscanonical(f::MOI.ScalarAffineFunction) = iscanonical(f.terms,
+                                                       t -> t.variable_index.value,
+                                                       t -> !iszero(t.coefficient))
+function iscanonical(x::AbstractVector, by, keep)
+    if isempty(x)
+        return true
+    end
+    @inbounds for i in 1:(length(x) - 1)
```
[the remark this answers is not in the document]
rdeits: fixed

src/Utilities/functions.jl
rdeits: done (I'm using it in my implementation of Base.copy)

test/functions.jl
rdeits: Sure, we can add explicit test cases and even delete this if you want, but it's been immensely helpful in ensuring that I don't break the implementation while I rewrite it.

src/Utilities/functions.jl
```diff
@@ -242,29 +250,15 @@ If `x` (resp. `y`, `z`) is `VariableIndex(1)` (resp. 2, 3).
...
-end
-
-function canonical(f::MOI.VectorAffineFunction)
-    if iscanonical(f)
-        f
-    else
-        canonicalize!(MOI.VectorAffineFunction(copy(f.terms), f.constants))
-    end
-end
+canonical(f::Union{SAF, VAF}) = canonicalize!(copy(f))
```
mohamed82008: `deepcopy`?

mohamed82008: Nevermind.

rdeits:
@mlubin I've updated the tests with some useful potential corner cases. I kept some limited randomized testing, but I removed the complicated reference implementation, so I'm just testing the invariant that `iscanonical(canonical(f))` is always true in the random cases.

rdeits:
Just snuck in a few more tests to try to get the coverage up

src/Utilities/functions.jl
```diff
@@ -202,7 +202,42 @@ coefficient(t::MOI.ScalarAffineTerm) = t.coefficient
...
+"""
+    iscanonical(f::Union{ScalarAffineFunction, VectorAffineFunction})
+
+Returns a Bool indicating whether the function is in canonical form.
+See [`canonical`](@ref).
+"""
+function iscanonical(f::Union{SAF, VAF})
+    iscanonical(f.terms, termindices, t -> !iszero(coefficient(t)))
+end
+function iscanonical(x::AbstractVector, by, keep)
```
mlubin: It's not obvious from the signature the method does, could you add a docstring or (more informally) a comment?

test/functions.jl
```diff
@@ -156,6 +156,7 @@
             @test f.terms == MOI.ScalarAffineTerm.([2, 3], [x, y])
             f = MOIU.modifyfunction(f, MOI.ScalarCoefficientChange(x, 0))
             @test f.terms == MOI.ScalarAffineTerm.([3], [y])
+
```
mlubin: Unnecessary whitespace change?

test/functions.jl
```diff
@@ -204,4 +205,149 @@
             @test g.affine_terms == MOI.VectorAffineTerm.([1, 2, 1], MOI.ScalarAffineTerm.([3, 1, 1], [x, x, y]))
         end
     end
+    @testset "Conversion to canonical form" begin
+        function _isapprox(f1::T, f2::T) where {T <: Union{MOI.ScalarAffineFunction, MOI.VectorAffineFunction}}
```
mlubin: The difference between `_isapprox` and `isapprox` ([here](https://github.com/JuliaOpt/MathOptInterface.jl/blob/ae87890bec8922c21df4c817b4e847c67b402240/src/functions.jl#L310)) is confusing and not well explained by an underscore. Maybe call this `isapprox_ordered`?

test/functions.jl
```diff
@@ -204,4 +205,149 @@
...
+            ((MOIU.termindices.(f1.terms) == MOIU.termindices.(f2.terms)) &&
+             (MOIU._constant(f1) ≈ MOIU._constant(f2)) &&
+             (MOIU.coefficient.(f1.terms) ≈ MOIU.coefficient.(f2.terms)))
+        end
+        function test_canonicalization(f::T, expected::T) where {T <: Union{MOI.ScalarAffineFunction, MOI.VectorAffineFunction}}
+            @test MOI.Utilities.iscanonical(expected)
```
mlubin: Inconsistent use of `MOIU` and `MOI.Utilities`.

test/functions.jl
```diff
@@ -204,4 +205,149 @@
...
+                MOI.ScalarAffineFunction(MOI.ScalarAffineTerm.([0.5], MOI.VariableIndex.([2])), 5.0),
+                )
+            function random_scalar_affine_function(n)
+                indices = MOI.VariableIndex.([rand(1:n) for i in 1:n])
+                coefficients = randn(n)
+                coefficients[rand(Float64, size(coefficients)) .<= 0.2] .= 0
+                MOI.ScalarAffineFunction(MOI.ScalarAffineTerm.(coefficients, indices), randn())
+            end
+            srand(1)
+            for i in 1:1000
```
mlubin: I see the potential usefulness of the randomized fuzzing, but we don't use this testing style of testing anywhere else in MOI. It may be confusing to someone reading the code why we would do this to test only `iscanonical`. I would lean towards dropping it but could be swayed otherwise.

src/Utilities/functions.jl
mlubin: Sorry, it was a bit early in the morning, didn't mean for it to sound like an accusation if that was the case. I meant that it looks like code that could appear in Base Julia :)

test/functions.jl
rdeits: fixed

test/functions.jl
rdeits: yup, done

test/functions.jl
rdeits: fixed

test/functions.jl
rdeits: Ok, removed

src/Utilities/functions.jl
```diff
@@ -202,7 +202,49 @@ coefficient(t::MOI.ScalarAffineTerm) = t.coefficient
...
+See [`canonical`](@ref).
+"""
+function iscanonical(f::Union{SAF, VAF})
+    is_strictly_sorted(f.terms, termindices, t -> !iszero(coefficient(t)))
+end
+
+"""
+    is_strictly_sorted(x::AbstractVector, by, filter)
+
+Returns `true` if `by(x[i]) < by(x[i + 1]` and `filter(x[i]) == true` for
```
mlubin: Typo: `by(x[i]) < by(x[i + 1])`

rdeits: fixed, thanks

src/Utilities/functions.jl
```diff
@@ -215,27 +250,52 @@ If `x` (resp. `y`, `z`) is `VariableIndex(1)` (resp. 2, 3).
...
+    sort_and_compress!(x::AbstractVector, by::Function, keep::Function, combine::Function)
+
+Sort the vector `x` in-place using `by` as the function from elements to comparable keys, then
+combine all entries for which `by(x[i]) == by(x[j])` using the function `x[i] = combine(x[i], x[j])`,
+and remove any entries for which `keep(x[i]) == false`. This may result in `x` being resized to
+a shorter length.
+"""
+function sort_and_compress!(x::AbstractVector, by, keep, combine)
+    if length(x) > 0
+        sort!(x, QuickSort, Base.Order.ord(isless, by, false, Base.Sort.Forward))
```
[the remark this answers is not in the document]
tkoolen: Quicksort is in-place, fast, and doesn't allocate. I briefly thought about using counting sort, but that's an out-of-place algorithm, so either you need to preallocate memory somewhere (a hassle) or not care about allocating (which makes this a lot less useful for us and probably also negates any performance gains compared to quicksort in practice).

src/Utilities/functions.jl
```diff
@@ -215,27 +250,52 @@ If `x` (resp. `y`, `z`) is `VariableIndex(1)` (resp. 2, 3).
...
+        i1 = firstindex(x)
+        for i2 in eachindex(x)[2:end]
```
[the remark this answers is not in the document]
tkoolen: Yeah, it's a generalization. The implementation is different though:

```julia
function unique(f::Callable, C)
    out = Vector{eltype(C)}()
    seen = Set()
    for x in C
        y = f(x)
        if !in(y, seen)
            push!(seen, y)
            push!(out, x)
        end
    end
    out
end
```

(on 0.6). On 0.7 there is also a `unique!(C)` (but interestingly there is no `unique!(f, C)`). And they still allocate and use a different algorithm.

src/Utilities/functions.jl
```diff
@@ -215,27 +250,52 @@ If `x` (resp. `y`, `z`) is `VariableIndex(1)` (resp. 2, 3).
...
+
+"""
+    canonicalize!(f::Union{ScalarAffineFunction, VectorAffineFunction})
+
+Convert a function to canonical form in-place, without allocating a copy to hold the result.
+See [`canonical`](@ref).
+"""
+function canonicalize!(f::Union{SAF, VAF})
+    sort_and_compress!(f.terms, termindices, t -> !iszero(coefficient(t)), unsafe_add)
+    f
```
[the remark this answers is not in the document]
rdeits: It was unnecessary, so i removed it.

src/Utilities/functions.jl
```diff
@@ -215,27 +250,52 @@ If `x` (resp. `y`, `z`) is `VariableIndex(1)` (resp. 2, 3).
...
-    sorted_terms = sort(f.terms, by = termindices)
-    terms = eltype(f.terms)[]
-    for t in sorted_terms
-        tnot0 = !iszero(coefficient(t))
-        if !isempty(terms) && termindices(t) == termindices(last(terms)) && tnot0
-            terms[end] = unsafe_add(t, last(terms))
-        elseif tnot0
-            if !isempty(terms) && iszero(coefficient(last(terms)))
-                terms[end] = t
+canonical(f::Union{SAF, VAF}) = canonicalize!(copy(f))
```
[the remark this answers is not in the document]
rdeits: For non-isbits types, a shallow copy should still be appropriate, as we never mutate any terms; we just replace elements in the terms vector. Moreover, `deepcopy` is significantly slower than the `copy` implementation here: 

```julia
julia> f = MOI.ScalarAffineFunction(MOI.ScalarAffineTerm.(rand(1000), MOI.VariableIndex.(rand(1:100, 1000))), 1.0);

julia> using BenchmarkTools

julia> @btime copy($f);
  1.045 μs (2 allocations: 15.78 KiB)

julia> @btime deepcopy($f);
  2.058 μs (5 allocations: 16.22 KiB)
```

src/Utilities/functions.jl
```diff
@@ -202,7 +202,42 @@ coefficient(t::MOI.ScalarAffineTerm) = t.coefficient
 coefficient(t::MOI.VectorAffineTerm) = t.scalar_term.coefficient
 
 """
-    canonical(f::AbstractFunction)
+    copy(f::Union{ScalarAffineFunction, VectorAffineFunction})
+
+Return a new affine function with a shallow copy of the terms and constant(s)
+from `f`.
+"""
+Base.copy(f::F) where {F <: Union{SAF, VAF}} = F(copy(f.terms), copy(_constant(f)))
```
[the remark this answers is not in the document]
rdeits: done

src/Utilities/functions.jl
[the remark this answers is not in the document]
rdeits: It's actually different anyway, as `unique` preserves the original order of appearance rather than sorting. So this is similar to `unique!(sort!(x))`, but I really don't think we need to go into that.

test/runtests.jl
```diff
@@ -6,6 +6,7 @@ const MOIB = MathOptInterface.Bridges
 
 using Compat
 using Compat.Test
+using Compat.Random
```
mlubin: Not needed anymore

rdeits: done

PR #435, opened 2018-07-29
Implements SolverName for MockOptimizer

blegat:


src/Utilities/cachingoptimizer.jl
```diff
@@ -395,11 +395,20 @@ function MOI.canset(m::CachingOptimizer, attr::Union{MOI.AbstractVariableAttribu
     return true
 end
 
+function MOI.get(m::CachingOptimizer, attr::MOI.AbstractOptimizerAttribute)
```
mlubin: Nit: Avoid using `m`. (I'm responsible for introducing it in this part of the code...)

Issue #423, opened 2018-07-08
Removing can... functions


blegat:
After a long discussion with @odow and @mlubin . We plan to remove the `canset`, `canaddconstraint`, `canaddvariable`, `canmodify`, `cantransform`, `candelete` functions in favor of throwing an error during `set!`, `addconstraint!`, ...
The CachingOptimizer is using these methods in `Automatic` mode to see if it should change from the `AttachedOptimizer` state to the `EmptyOptimizer` state automatically because the optimizer cannot do the modification in its currently state even it supports it (so by being emptied and then having the model recopied before optimization, it could work).
Instead of using these `can...` functions, the CachingOptimizer will do the `set!`, `addconstraint!`, ... in a `try`-`catch` and detach the optimizer automatically if it throws an error mentioning it cannot handle the operation in its current state.
This `try` `catch` seems to have a fixed 10 ns cost in our benchmark so it shouldn't be noticeable for most usage.
The `Manual` mode won't do any `try`-`catch` so it can be used if someonce needs to use the CachingOptimizer but do not want this 10 ns penalty for every operation.

This would resolve https://github.com/JuliaOpt/MathOptInterface.jl/issues/295, https://github.com/JuliaOpt/MathOptInterface.jl/issues/218, https://github.com/JuliaOpt/MathOptInterface.jl/pull/357#issuecomment-388685219

The `CopyResult` will also disappear and will be replaced by the exceptions thrown by the `set!`, `addconstraint!`, ... calls

This would resolve https://github.com/JuliaOpt/MathOptInterface.jl/issues/243

blegat:
There is a last remaining issue with names where `canset` is used to determine if a name is already in use. What is the replacement for this ?

rdeits:
Is there a standard error that will be thrown when the `get/set!/etc.` fails?

mlubin:
Yes, standard exceptions for unsupported operations are part of this redesign. We want to distinguish `get`/`set!` calls that should fail because the solver doesn't support the operation, and other errors (e.g., bugs in the wrapper) that should propagate up instead of silently being caught.

rdeits:
Ok, sounds great, thanks

Issue #243, opened 2018-02-19
Replace CopyResult with exceptions?


tkoolen:
I find `CopyResult` a little cumbersome to work with.

If you call a bunch of functions that could fail in some circumstances in your `copy!` method, they all have to return a `CopyResult` in addition to their normal output arguments, and then in `copy!` itself you have to continually check whether those subfunctions returned a non-`Success` `CopyResult`.

Less importantly, it's also annoying that you need to pass in a dud index map when you're about to return a bad `CopyResult`, even though the index map is useless since it maps to nothing.

Why not just have some documented `Exception` subtypes in MOI and require that those be thrown if something goes wrong?

mlubin:
What if we move the `indexmap` outside of `CopyResult` and then return a `Union{IndexMapType,CopyResult}`?

Given that Julia doesn't provide a way to catch only certain types of exceptions (afaik), if we switched from result objects to exceptions, the proper way to call `MOI.copy!` would be:
```julia
try
    indexmap = MOI.copy!(a, b)
catch e
    if e isa MOICopyError
        ...
    else
        rethrow()
    end
end
```
That's pretty painful and is bound to not be implemented correctly. OTOH, with result objects it's clear what the caller has to do.

tkoolen:
The lazy implementer will just write
```julia
indexmap = MOI.copy!(a, b)
```
without  the try-catch, in which case the program stops and you get a nice stacktrace showing exactly where things went wrong (another advantage over returning a `CopyResult`). I think this is the most common use case: an error during `copy!` will usually be fatal: the user just did something wrong. The try-catch scenario is much rarer (with https://github.com/JuliaOpt/MathOptInterface.jl/issues/188#issue-286115084 being an example of where it would be needed).

I would actually argue that the current implementation is much more error prone, as the caller is required to handle the `CopyResult` manually. I've personally already been bitten by the [current lack of handling in JuMP](https://github.com/JuliaOpt/JuMP.jl/blob/c58e6f08a91e1fcfbc7cb100c86284207e2c6169/src/optimizerinterface.jl#L26).

mlubin:
I'll let others weigh in here.

odow:
I prefer throwing an (informative) exception if anything goes wrong. It seems simpler and Miles is right, the alternative is not going to be implemented correctly.

blegat:
I am in favor the using Exceptions and have `copy!` just return an index map.
[...]

PR #436, opened 2018-07-29
Replace can... functions by errors

blegat:
- [x] Remove `cantransform`
- [x] Remove `canaddconstraint`
- [x] Remove `canaddvariable`
- [x] Remove `canset`
- [x] Remove `canmodify`
- [x] Remove `candelete`
- [x] Add error section in doc

Closes https://github.com/JuliaOpt/MathOptInterface.jl/issues/423 
Closes https://github.com/JuliaOpt/MathOptInterface.jl/issues/295
Closes https://github.com/JuliaOpt/MathOptInterface.jl/issues/218
Closes https://github.com/JuliaOpt/MathOptInterface.jl/issues/243

### DIFF
--- a/docs/src/apireference.md
+++ b/docs/src/apireference.md
@@ -286,9 +286,11 @@ constraint_expr
 ## Errors
 
 When an MOI call fails on a model, precise errors should be thrown when possible
-instead of simply calling `error` with a message. This allows for instance to
-the CachingOptimizer to automatically detach the optimizer when an
-incremental update of the model is not supported.
+instead of simply calling `error` with a message. The docstrings for the
+respective methods describe the errors that the implementation should thrown in
+certain situations. This error-reporting system allows code to distinguish
+between internal errors (that should be shown to the user) and unsupported
+operations which may have automatic workarounds.
 
 When an invalid index is used in an MOI call, an [`InvalidIndex`](@ref) should
 be thrown:
--- a/docs/src/apireference.md
+++ b/docs/src/apireference.md
@@ -160,7 +160,6 @@ Functions for adding and modifying constraints.
 
 ```@docs
 isvalid(::ModelLike,::ConstraintIndex)
-canaddconstraint
 addconstraint!
 addconstraints!
 transform!
--- a/docs/src/apireference.md
+++ b/docs/src/apireference.md
@@ -23,7 +23,6 @@ Functions for getting and setting attributes.
 canget
 get
 get!
-canset
 set!
 supports
 ```
@@ -42,8 +41,6 @@ Copying
 
 ```@docs
 copy!
-CopyResult
-CopyStatusCode
 ```
 
 List of model attributes
@@ -129,7 +126,6 @@ BasisStatusCode
 ```@docs
 VariableIndex
 ConstraintIndex
-candelete
 isvalid
 delete!(::ModelLike,::Index)
 ```
@@ -139,7 +135,6 @@ delete!(::ModelLike,::Index)
 Functions for adding variables. For deleting, see index types section.
 
 ```@docs
-canaddvariable
 addvariables!
 addvariable!
 ```
@@ -250,7 +245,6 @@ Functions for modifying objective and constraint functions.
 
 ```@docs
 modify!
-canmodify
 AbstractFunctionModification
 ScalarConstantChange
 VectorConstantChange
@@ -287,6 +281,39 @@ eval_hessian_lagrangian
 eval_hessian_lagrangian_product
 objective_expr
 constraint_expr
+```
+
+## Errors
+
+When an MOI call fails on a model, precise errors should be thrown when possible
+instead of simply calling `error` with a message. This allows for instance to
+the CachingOptimizer to automatically detach the optimizer when an
+incremental update of the model is not supported.
+
+When an invalid index is used in an MOI call, an [`InvalidIndex`](@ref) should
+be thrown:
+```@docs
+InvalidIndex
+```
+
+The rest of the errors defined in MOI fall in two categories represented by the
+following two abstract types:
+```@docs
+UnsupportedError
+CannotError
+```
+
+The different [`UnsupportedError`](@ref) and [`CannotError`](@ref) are the
+following errors:
+```@docs
+UnsupportedAttribute
+CannotSetAttribute
+CannotAddVariable
+UnsupportedConstraint
+CannotAddConstraint
+UnsupportedConstraintModification
+UnsupportedObjectiveModification
+UnsupportedDeletion
 ```
 
 ## Bridges
--- a/docs/src/apireference.md
+++ b/docs/src/apireference.md
@@ -164,7 +164,6 @@ canaddconstraint
 addconstraint!
 addconstraints!
 transform!
-cantransform
 supportsconstraint
 ```
 
--- a/src/MathOptInterface.jl
+++ b/src/MathOptInterface.jl
@@ -78,48 +78,17 @@ Empty the model, that is, remove all variables, constraints and model attributes
 function empty! end
 
 """
-    CopyStatusCode
-
-An Enum of possible statuses returned by a `copy!` operation through the `CopyResult` struct.
-
-* `CopySuccess`: The copy was successful.
-* `CopyUnsupportedAttribute`: The copy failed because the destination does not support an attribute present in the source.
-* `CopyUnsupportedConstraint`: The copy failed because the destination does not support a constraint present in the source.
-* `CopyOtherError`: The copy failed for a different reason.
-
-In the failure cases:
-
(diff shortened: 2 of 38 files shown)


mlubin:
Do you plan on addressing https://github.com/JuliaOpt/MathOptInterface.jl/issues/424 in this PR or later?

src/Utilities/cachingoptimizer.jl
```diff
@@ -208,25 +220,27 @@ function MOI.supportsconstraint(m::CachingOptimizer, F::Type{<:MOI.AbstractFunct
...
-    # be in sync.
-    if m.mode == Automatic && m.state == AttachedOptimizer && !MOI.canaddconstraint(m.optimizer, typeof(func), typeof(set))
-        resetoptimizer!(m)
+    if m.state == AttachedOptimizer
+        if m.mode == Automatic
+            try
+                cindex_optimizer = MOI.addconstraint!(m.optimizer, mapvariables(m.model_to_optimizer_map, func), set)
+            catch err
+                if err isa MOI.CannotAddConstraint
+                    # It could be MOI.CannotAddConstraint{F', S'} with F' != F
```
mlubin: I'm not sure I understand. If `F` and `S` don't match, that seems like a leak of an internal implementation detail, like http://www.juliaopt.org/JuMP.jl/latest/style.html#User-facing-MethodError-1.

src/constraints.jl
```diff
@@ -10,11 +10,21 @@ If `F`-in-`S` constraints are only not supported in specific circumstances, e.g.
...
 
 """
-    canaddconstraint(model::ModelLike, ::Type{F}, ::Type{S})::Bool where {F<:AbstractFunction,S<:AbstractSet}
+    UnsupportedConstraint{F, S} <: CannotError
 
-Return a `Bool` indicating whether it is possible to add a constraint ``f(x) \\in \\mathcal{S}`` where ``f`` is of type `F`, and ``\\mathcal{S}`` is of type `S`.
+Constraints of type `F`-in-`S` are not supported by the model.
 """
-canaddconstraint(model::ModelLike, ::Type{<:AbstractFunction}, ::Type{<:AbstractSet}) = false
+struct UnsupportedConstraint{F<:AbstractFunction, S<:AbstractSet} <: CannotError end
```
mlubin: We should consider adding an optional text error string in case the solver has a more specific error message to return.

src/error.jl
```diff
@@ -0,0 +1,29 @@
+"""
+    operation_name(err::Union{UnsupportedError, CannotError})
+
+Return the name of the operation throwing the error in a gerund (i.e. -ing form).
```
mlubin: It's a bit awkward to break up the error message like this. Maybe let each error implement `showerror` instead?

docs/src/apireference.md
```diff
@@ -160,11 +160,9 @@ Functions for adding and modifying constraints.
 
 ```@docs
 isvalid(::ModelLike,::ConstraintIndex)
```
mlubin: Add a new section on errors to the api reference?

src/Utilities/cachingoptimizer.jl
blegat: There is no way around that, the bridges only check that the constraints created are supported but they do not check if it can be added since `canaddconstraint` do not exists anymore. Using `try` `catch` in bridges just for transforming the error doesn't seem like a good idea and it would hide the initial error.
This is unlikely to happen anyway in JuMP since bridges are added on top of the caching optimizer so the user will need to manually add bridges below to caching optimizer for this to happen in JuMP.

src/error.jl
blegat: This is to enforce consistency in the error message. It also ensure the consistency between the abstract type chosen and the error message. It helps when choosing which supertype should be chosen. For instance, for modify I was hesitant between `UnsupportedError` and `CannotError` but the error should be the one of `UnsupportedError` so that's what I have chosen.

blegat:
> Do you plan on addressing https://github.com/JuliaOpt/MathOptInterface.jl/issues/424 in this PR or later?

On another PR, I wrote `canget` on the TODO list but I meant `canset`, I have updated it.

docs/src/apireference.md
```diff
@@ -291,6 +283,39 @@ objective_expr
 constraint_expr
 ```
 
+## Errors
+
+When an MOI call fails on a model, precise errors should be thrown when possible
+instead of simply calling `error` with a message. This allows for instance to
+the CachingOptimizer to automatically detach the optimizer when an
```
mlubin: Mentioning `CachingOptimizer` might be more confusing than helpful in this context. You could say more simply that this error-reporting system allows code to distinguish between internal errors (that should be shown to the user) and unsupported operations which may have automatic workarounds.

src/MathOptInterface.jl
```diff
@@ -78,48 +78,17 @@ Empty the model, that is, remove all variables, constraints and model attributes
...
-    status::CopyStatusCode
-    message::String # Human-friendly explanation why the copy failed
-    indexmap::T     # Only valid if status is CopySuccess
-end
-
-"""
-    copy!(dest::ModelLike, src::ModelLike; copynames=true, warnattributes=true)::CopyResult
+    copy!(dest::ModelLike, src::ModelLike; copynames=true, warnattributes=true)
 
 Copy the model from `src` into `dest`. The target `dest` is emptied, and all
```
mlubin: Discuss which errors can be thrown from `copy!`.

src/Utilities/cachingoptimizer.jl
mlubin: I'm not entirely convinced on this point, maybe I don't understand bridges well enough. But this is relatively minor, we can address it later.

src/attributes.jl
```diff
@@ -35,6 +35,33 @@ abstract type AbstractConstraintAttribute end
...
+        attr::AttrType
+    end
+
+Setting attribute `attr` is not supported by the model.
+"""
+struct UnsupportedAttribute{AttrType<:AnyAttribute} <: UnsupportedError
+    attr::AttrType
+end
+# old message:
+# "ModelLike of type $(typeof(model)) does not support setting the attribute $attr"
```
mlubin: Is this comment meant to be here?

src/error.jl
```diff
@@ -0,0 +1,29 @@
...
+"""
+    CannotError <: Exception
+
+Abstract type for error thrown when an operation is supported but cannot be
+applied in the current state of the model.
+"""
+abstract type CannotError <: Exception end
+
+function Base.showerror(io::IO, err::CannotError)
+    print(io, typeof(err), ": ", operation_name(err), " cannot be done in the current state of the model even if the operation is supported. You may want to use a `CachingOptimizer` in `Automatic` mode or you may need to call `resetoptimizer!` before doing this operation if the `CachingOptimizer` in `Manual` mode.")
```
mlubin: Nit: "if the `CachingOptimizer` *is* in `Manual` mode"

mlubin: Nit: "be done" -> "be performed"

mlubin: The error message is missing a point about the underlying reason that the operation is not supported. This makes me think that we should allow the solvers to attach strings to the errors.

src/indextypes.jl
```diff
@@ -27,11 +27,19 @@ end
 const Index = Union{ConstraintIndex,VariableIndex}
 
 """
-    candelete(model::ModelLike, index::Index)::Bool
+    struct InvalidIndex{IndexType<:Index} <: Exception
+        index::IndexType
+    end
 
-Return a `Bool` indicating whether the object referred to by `index` can be removed from the model `model`.
+The index `index` is invalid.
```
mlubin: An error indicating that ...

src/indextypes.jl
```diff
@@ -41,18 +49,22 @@ Return a `Bool` indicating whether this index refers to a valid object in the mo
 isvalid(model::ModelLike, ref::Index) = false
 
 """
-    delete!(model::ModelLike, index::Index)
+    UnsupportedDeletion{IndexType} <: UnsupportedError
 
-Delete the referenced object from the model.
+Deleting indices of type `IndexType` is not supported by the model.
```
mlubin: An error indicating that ...

src/modifications.jl
```diff
@@ -1,44 +1,45 @@
...
-```
-
-## Objective Function
+    struct UnsupportedConstraintModification{F<:AbstractFunction, S<:AbstractSet,
+                                             C<:AbstractFunctionModification} <: UnsupportedError
+        change::C
+    end
 
-    canmodify(model::ModelLike, ::ObjectiveFunction, ::Type{M})::Bool where M<:AbstractFunctionModification
+Constraints of type `F`-in-`S` do not support the constraint modification `change`.
```
mlubin: An error indicating that ...

odow:
train

PR #444, opened 2018-08-02
Fix method specialization

blegat:
Without the `S<:AbstractSet` in the first method, julia considers that the signature is less specific.

### DIFF
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -592,11 +592,14 @@ A constraint attribute for the `AbstractSet` object used to define the constrain
 """
 struct ConstraintSet <: AbstractConstraintAttribute end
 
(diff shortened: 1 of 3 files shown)


src/attributes.jl
```diff
@@ -575,11 +575,14 @@ It is guaranteed to be equivalent but not necessarily identical to the function
 """
 struct ConstraintFunction <: AbstractConstraintAttribute end
 
-function hidden_set!(model::ModelLike, attr::ConstraintFunction, constraint_index::ConstraintIndex{F,S}, func::F) where {F,S}
+function hidden_set!(::ModelLike, attr::ConstraintFunction,
```
mlubin: I got confused by this name. Why it it hidden? Maybe `set!_fallback_error`

PR #445, opened 2018-08-02
Fix supports for CachingOptimizer with names

blegat:


src/Utilities/cachingoptimizer.jl
```diff
@@ -397,6 +401,20 @@ function MOI.set!(m::CachingOptimizer, attr::Union{MOI.AbstractVariableAttribute
     MOI.set!(m.model_cache, attr, index, value)
 end
 
+# Names are not copied, i.e. we use the option `copynames=false` in
+# `attachoptimizer`, so the optimizer do not need to support names.
```
mlubin: clarify: so the caching optimizer can support names even if the optimizer doesn't.

Issue #442, opened 2018-08-02
More ObjectiveBound tests


blegat:
Is there a reason ObjectiveBound is currently only tested in one MIP test ?
Even for an LP with termination status Success and primal and dual status FeasiblePoint, ObjectiveBound should be accessible and give the value of the dual objective right ?
As a result of the lack of testing, ECOS was passing all the tests while not even implementing ObjectiveBound: https://github.com/JuliaOpt/ECOS.jl/pull/71

odow:
> Even for an LP with termination status Success and primal and dual status FeasiblePoint, ObjectiveBound should be accessible and give the value of the dual objective right ?

Probably. But if you ask Gurobi for the objective bound of an LP, it returns `-1e100` for minimization problems.

mlubin:
LP solvers are not very responsible about computing dual objectives but conic primal-dual solvers should be better. Maybe add this on a exp-cone and SDP problems? Worse case it's a test config option that solvers can disable if they prefer.

odow:
Closed by #446.

PR #447, opened 2018-08-03
Add result fallbacks

blegat:
As discussed in https://github.com/JuliaOpt/MathOptInterface.jl/issues/442 and https://github.com/JuliaOpt/LinQuadOptInterface.jl/issues/41, some solver API do not contain any getters for result informations that can be deduced for variable primals and constraint duals of non-variablewise constraints.

These include the following:
* ObjectiveValue: can be deduced from VariablePrimal and ObjectiveFunction
* ObjectiveBound: can be deduced from ConstraintDual and constants in ConstraintFunction's and ConstraintSet's
* ConstraintPrimal: can be deduced from ConstraintFunction's and VariablePrimal's
* ConstraintDual of variablewise constraints: can be deduced from ConstraintDual of non-variablewise constraints, ObjectiveFunction and ConstraintFunction's.

This PR
* moves the fallback already used by MockOptimizer for ObjectiveValue and ConstraintPrimal to a new `get_fallback` function that is independent of the MockOptimizer.
* creates a fallback for ConstraintDual and use it for MockOptimizer in the tests (it can also be useful for solving https://github.com/JuliaOpt/LinQuadOptInterface.jl/issues/41)
* Introduce `set_dot` and `dot_coefficients` as new set functions that need to be implemented. They have a default implementation for non-triangle sets.
* renames `evalobjective` to `eval_objective_value` in MockOptimizer to make place to `eval_objective_bound`

TODO:

- [x] Document `get_fallback`

### DIFF
--- a/src/Utilities/results.jl
+++ b/src/Utilities/results.jl
@@ -1,6 +1,6 @@
 # This file contains the implementation of different methods for the
 # `get_fallback` function. These methods can be used by solver wrappers as
-# fallbacks for implemented the `get` method when the solver API does not
+# fallbacks for implementing the `get` method when the solver API does not
 # provide the required result. For instance, if the solver does not provide the
 # value of the constraints, the solver wrapper can write
 # ```julia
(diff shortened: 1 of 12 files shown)


src/sets.jl
```diff
@@ -46,6 +46,27 @@ abstract type AbstractVectorSet <: AbstractSet end
 
 dimension(s::AbstractVectorSet) = s.dimension # .dimension field is conventional, overwrite this method if not applicable
 
+"""
+    set_dot(x::Vector, y::Vector, set::AbstractVectorSet)
+
+Return the scalar product between a vector `x` of the set `set` and a vector
+`y` of the dual of the set `s`.
+"""
+function set_dot(x::Vector, y::Vector, set::AbstractVectorSet)
```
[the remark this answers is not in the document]
blegat: If someone defines a new set with a non-trivial inner product, he is forced to define `set_dot` and `dot_coefficients` otherwise all codes using these functions fails. As this is mandatory, this is part of the API, isn't it ?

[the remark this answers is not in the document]
mlubin: Is there code using these functions outside of MOIU? The `get_fallback` should be opt-in by the solver and not automatic.

blegat:
Good to merge ?

src/Utilities/results.jl
```diff
@@ -0,0 +1,283 @@
+# This file contains the implementation of different methods for the
+# `get_fallback` function. These methods can be used by solver wrappers as
+# fallbacks for implemented the `get` method when the solver API does not
```
mlubin: "for implemented" -> "for implementing"

src/Utilities/results.jl
```diff
@@ -0,0 +1,283 @@
...
+                       vi::MOI.VariableIndex,
+                       F::Type{<:Union{MOI.SingleVariable,
+                                       MOI.VectorOfVariables}},
+                       S::Type{<:MOI.AbstractSet})
+    for constraint_index in MOI.get(model, MOI.ListOfConstraintIndices{F, S}())
+        if constraint_index != ci
+            func = MOI.get(model, MOI.ConstraintFunction(), constraint_index)
+            if (F == MOI.SingleVariable && func.variable == vi) ||
+               (F == MOI.VectorOfVariables && vi in func.variables)
+               error("Fallback getter for variable constraint dual does not support other variable-wise constraints on the variable.")
```
mlubin: Nit: line break the error message (using * to concat strings)

src/Utilities/results.jl
```diff
@@ -0,0 +1,283 @@
...
+end
+
+function variable_dual(::MOI.ModelLike,
+                       ::MOI.ConstraintDual,
+                       ::MOI.ConstraintIndex,
+                       ::MOI.VariableIndex,
+                       ::Type{<:Union{MOI.ScalarQuadraticFunction,
+                                      MOI.VectorQuadraticFunction}},
+                       ::Type{<:MOI.AbstractSet})
+    error("Fallback getter for variable constraint dual only supports affine constraint functions.")
```
mlubin: What happens if a user sees this error message? Does it indicate an internal error in the solver wrapper or that the user did something wrong? (Same with similar errors below.)

src/Utilities/results.jl
```diff
@@ -0,0 +1,283 @@
...
+            if (F == MOI.SingleVariable && func.variable == vi) ||
+               (F == MOI.VectorOfVariables && vi in func.variables)
+               error("Fallback getter for variable constraint dual does not support other variable-wise constraints on the variable.")
+            end
+        end
+    end
+    return 0.0
+end
+
+function variable_dual(::MOI.ModelLike,
```
mlubin: This function has lots of methods that aren't obvious to understand. It needs a docstring or at least a comment.

src/attributes.jl
```diff
@@ -380,6 +380,24 @@ it has non-integer coefficient and `F` is `ScalarAffineFunction{Int}`.
 """
 struct ObjectiveFunction{F<:AbstractScalarFunction} <: AbstractModelAttribute end
 
+"""
+    ObjectiveFunctionType()
```
mlubin: Add this to the list of attributes in the docs.

src/Utilities/results.jl
mlubin: This should be reflected in the error message to help users if they see it.

PR #448, opened 2018-08-04
Fixes for Julia v0.7-rc2

blegat:
Blocked by until https://github.com/JuliaLang/Compat.jl/pull/602 is merged and released


mlubin:
Let's drop `depwarn=error` temporarily (with a TODO). That's better than having the CI broken for multiple days.

Issue #453, opened 2018-08-06
Bug in bridge optimizer


odow:
@blegat: Tests are failing with my [Gurobi PR](https://github.com/JuliaOpt/Gurobi.jl/pull/146) on both v0.6 and v0.7 because of 
https://github.com/JuliaOpt/Gurobi.jl/blob/e16363e64f4ba602396c8d5bce9dc897fd44e292/test/MOIWrapper.jl#L15-L18

The MOI test is:
https://github.com/JuliaOpt/MathOptInterface.jl/blob/c164eabcff67e6d96341e69eb053e6fc01b5e3a3/src/Test/UnitTests/constraints.jl#L110-L125
and the error is
```julia
solve_affine_interval: Error During Test at C:\Users\Oscar\.julia\dev\Gurobi\test\MOIWrapper.jl:13
  Got exception KeyError("c") outside of a @test
  KeyError: key "c" not found
  Stacktrace:
   [1] getindex at .\dict.jl:478 [inlined]
   [2] get at C:\Users\Oscar\.julia\dev\LinQuadOptInterface\src\constraints.jl:157 [inlined]
   [3] get at C:\Users\Oscar\.julia\dev\MathOptInterface\src\Bridges\bridgeoptimizer.jl:325 [inlined]
   [4] solve_affine_interval(::MathOptInterface.Bridges.SingleBridgeOptimizer{MathOptInterface.Bridges.SplitIntervalBridge{Float64,F} where F<:MathOptInterface.AbstractScalarFunction,MathOptInterface.Bridges.SplitIntervalInstance{Float64},Gurobi.Optimizer}, ::MathOptInterface.Test.TestConfig) at C:\Users\Oscar\.julia\dev\MathOptInterface\src\Test\UnitTests\constraints.jl:118
   [5] macro expansion at C:\Users\Oscar\.julia\dev\Gurobi\test\MOIWrapper.jl:15 [inlined]
   [6] macro expansion at C:\cygwin\home\Administrator\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v0.7\Test\src\Test.jl:1079 [inlined]
   [7] macro expansion at C:\Users\Oscar\.julia\dev\Gurobi\test\MOIWrapper.jl:15 [inlined]
   [8] macro expansion at C:\cygwin\home\Administrator\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v0.7\Test\src\Test.jl:1079 [inlined]
[...]

Issue #424, opened 2018-07-08
Removing canget


blegat:
We plan to remove the `canset`, `canaddconstraint`, `canaddvariable`, `canmodify`, `cantransform` functions in https://github.com/JuliaOpt/MathOptInterface.jl/issues/423 .
The `canget` function is somewhat different since `get` does not modify the model.
There are currently a few issues with removing `canget`:

* `canget` with `ConstraintPrimal` is currently used to signal that there is no primal solution to return. A new `PrimalStatus` should be created to handle this situation. Same for duals. See https://github.com/JuliaOpt/MathOptInterface.jl/issues/394
* `canget(::ModelLike, ::Type{<:Index}, ::String)` is currently used to check if the name is used. What is the replacement for this ?

This would resolve https://github.com/JuliaOpt/MathOptInterface.jl/issues/302

PR #454, opened 2018-08-06
Fix bug in bridges querying index via string

odow:
Closes https://github.com/JuliaOpt/MathOptInterface.jl/issues/453

- [x] Needs tests

### DIFF
--- a/src/Bridges/bridgeoptimizer.jl
+++ b/src/Bridges/bridgeoptimizer.jl
@@ -345,9 +345,9 @@ function MOI.canget(b::AbstractBridgeOptimizer, IdxT::Type{CI},
 end
 function MOI.get(b::AbstractBridgeOptimizer, IdxT::Type{CI},
                  name::String)
-    try
+    if MOI.canget(b.model, IdxT, name)
         return MOI.get(b.model, IdxT, name)
-    catch
+    else
         return MOI.get(b.bridged, IdxT, name)
     end
 end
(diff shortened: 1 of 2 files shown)


src/Bridges/bridgeoptimizer.jl
```diff
@@ -318,11 +320,36 @@ end
...
+# constraint is in the bridge or the model. Therefore, we try the model first,
+# and then the bridge if that fails. As such, we need canget for either the
+# bridge or the model.
+function MOI.canget(b::AbstractBridgeOptimizer, IdxT::Type{CI},
+                    name::String)
+    return MOI.canget(b.bridged, IdxT, name) || MOI.canget(b.model, IdxT, name)
+end
+function MOI.get(b::AbstractBridgeOptimizer, IdxT::Type{CI},
+                 name::String)
+    try
```
blegat: We should use `canget(b.model, IdxT, name)` instead of `try` `catch` as done in https://github.com/JuliaOpt/MathOptInterface.jl/blob/c164eabcff67e6d96341e69eb053e6fc01b5e3a3/src/Utilities/universalfallback.jl#L229-L235
This need for `canget` is talked about in https://github.com/JuliaOpt/MathOptInterface.jl/issues/424.

blegat: We'll replace it with `try` ... `catch` when we remove `canget` at both places if that is the solution we choose.

Issue #418, opened 2018-07-07
ScalarFunctionIterator has quadratic complexity


mohamed82008:
The current `ScalarFunctionIterator` has a best case quadratic and worst case cubic (if the constraints are dense) complexity. If we do a Radix sort by the `output_index` then it becomes O(nk) where k is the average number of terms per constraint. For sparse constraints, this is linear complexity. This together with the use of `StructsOfArrays.jl` can enable returning a view from `scalar_terms_at_index` which reduces allocations. The catch here is if the user/solver expects the terms to be in a certain order and sorting will just confuse everyone. I haven't seen this assumption made anywhere so I am asking if you know otherwise.

https://github.com/JuliaOpt/MathOptInterface.jl/blob/247c4baf673eb8b1c29177e7f1652a0a78e1d570/src/Utilities/functions.jl#L103

odow:
We should hold off until we can benchmark the performance. Where is this used in the rest of the code?

mohamed82008:
> Where is this used in the rest of the code?

I am still reading so I am not sure but this seems like the kind of thing I would assume is efficient since it is baked in MOI. Maybe others know where it is used.

mohamed82008:
Shouldn't be hard to benchmark this specific part of the code using random sparse matrices of `ScalarAffineTerms` which are converted to `VAF`s, then iterated over. I will try to do it and see what happens.

blegat:
`getindex` used to be called directly on the `AbstractVectorFunction`.
Now that it `eachscalar` is called first, we could add caching in the `ScalarFunctionIterator` to avoid having to redo `findall` for every `getindex`.
We could for instance have a `Vector{Vector{Int}}` stored in cache storing the list of indices of terms for a specific output index. This can be constructed from a single pass through the terms in `eachscalar`.

mohamed82008:
This is a nice middle way solution that does not involve the hidden sorting of the terms. Sorting will probably use less memory but may be an unexpected and unwelcome side effect.

mohamed82008:
If memory is not of primary concern, we can also cache a sorted version of the vector of terms and a vector of end markers for the rows' terms. This will enable using the StructsOfArrays trick, which will save allocations by returning a view into the sorted vector of terms.

blegat:
The `ScalarAffineFunction` terms need to be a `Vector`, not any `AbstractVector` so it cannot be a view, can it ?

mohamed82008:
[...]

PR #455, opened 2018-08-06
Define operate and refactor moivcat

blegat:
Bridges need to do some arithmetic operations like `vcat`, `+`, `-`, ... and need to be able to determine the result of such operation to determine which constraints the bridge generate at compile time.
To achieve this, this PR suggest to define `operate(op::Function, ::Type{T}, args...) that does the operation `op(args...)` with coefficient type `T` and `promote_operation` that gives the type of the output the function `operate` without the need for Julia's inference system.

Currently most bridges only work when the function is `VectorAffineFunction` and convert `VectorOfVariables` to `VectorAffineFunction` while they could also work with `VectorQuadraticFunction` and keep `VectorAffineFunction` unconverted would allow the bridge to create `VectorAffineFunction` functions which could be handled more efficiently by the solver.
Using these functions instead of hardcoding the operations for `VectorAffineFunction` would allow to easily build bridges that work for *any* `AbstractVectorFunction` as input (just like SplitIntervalBridge currently).


blegat:
Ready for review, I am not planning to do any more changes :)

blegat:
> What's the argument for creating our own operator?

We want `[a, b]` where `a` and `b` and `ScalarAffineFunction` to be a `Vector{ScalarAffineFunction}`, not a `VectorAffineFunction` so we define `moivcat` which is now `MOI.operate(vcat, ...)`.
Same thing for `getindex` and `eachscalar` except that for `getindex` we might still use `eachscalar` and not `operate` as `eachscalar` may allow doing some caching to speed up (see https://github.com/JuliaOpt/MathOptInterface.jl/issues/418#issuecomment-403277617).
What would we do for the call `(+, - or *)(::MOI.SingleVariable, MOI.SingleVariable)`, which coefficient type do we use for the result ?

When doing `a * b` in JuMP we have the issue that we don't know if we can overwrite `a` so we have to copy it. Here we have `operate!` and `operate` so the user can use `operate!` if he knows the first argument can be overwritten.

In MOI we do not care too much about being super-user friendly (user friendly is enough).
In JuMP we need to be super-user friendly so we are forced to implement the `Base` methods,
here we have the choice and we can choose to be more explicit about inplace/coefficient type rather than being super-user friendly.

> type promotion system?

`Base.promote_op` should not be used
https://github.com/JuliaLang/julia/pull/26344#issuecomment-371613750

blegat:
As a second thought, it would indeed be nice to also define `Base.:+`, ... methods using `operate`

mlubin:
> We want [a, b] where a and b and ScalarAffineFunction to be a Vector{ScalarAffineFunction}, not a VectorAffineFunction so we define moivcat which is now MOI.operate(vcat, ...).

I wonder if using a macro to switch the context could be useful. Like `[a, b]` returns `Vector{ScalarAffineFunction}` and `@MOI.operate [a, b]` returns `VectorAffineFunction`. But not needed immediately.

PR #460, opened 2018-08-07
add CannotModify{Objective|Constraint} exceptions

### LINKED ISSUE #459: Error types in objective modification
Carried over from discussion on Gitter and https://github.com/oxfordcontrol/OSQP.jl/pull/25#discussion_r208277285

OSQP.jl supports objective function modifications, but only in certain states (i.e. when the model is not empty). So it seems like it would be useful to have a `CannotModifyObjective` exception that `modify!()` could throw.

rdeits:
Fixes #459 

I couldn't find anywhere that the exceptions themselves are tested in the unit tests. I can at least add a test to verify that constructing and calling `Base.showerror` works for these types.

### DIFF
--- a/src/modifications.jl
+++ b/src/modifications.jl
@@ -20,6 +20,29 @@ end
 operation_name(err::UnsupportedConstraintModification{F, S}) where {F, S} = "Modifying `$F`-in-`$S` constraints with $(err.change)"
 
 """
+    struct CannotModifyConstraint{F<:AbstractFunction, S<:AbstractSet,
+                                  C<:AbstractFunctionModification} <: CannotError
+        change::C
+        message::String
+    end
+
+An error indicating that constraints of type `F`-in-`S` do support the constraint
+modification `change` but cannot accomplish it in the current state.
+"""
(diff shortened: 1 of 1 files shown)


blegat:
Some exceptions are tested in `test/attributes.jl` and here:
https://github.com/JuliaOpt/MathOptInterface.jl/blob/84efa5fc49b102b3a49ee50408c41166b1f3e182/test/cachingoptimizer.jl#L139-L161
You also need to update
https://github.com/JuliaOpt/MathOptInterface.jl/blob/84efa5fc49b102b3a49ee50408c41166b1f3e182/src/Utilities/cachingoptimizer.jl#L254
and
https://github.com/JuliaOpt/MathOptInterface.jl/blob/84efa5fc49b102b3a49ee50408c41166b1f3e182/src/Utilities/cachingoptimizer.jl#L305
so that the error is recognized by the caching optimizer as something it can handle and to not have to rethrow

blegat:
Superseeded by https://github.com/JuliaOpt/MathOptInterface.jl/pull/468

PR #465, opened 2018-08-08
update of replace_constraint_function_or_set

### LINKED ISSUE #463: Allow CachingOptimizer to set ConstraintSet
While updating Cbc to MOI v0.5.1, I found out that linear6 test is not passing. Cbc is tested using `CachingOptimizer` in `Automatic` mode. The test basically optimizes then tries to modify the `ConstraintSet` and it fails.

It looks to me that the reason is the new `replace_constraint_function_or_set!`. In v0.4.1 `resetoptimizer!(m)` was called, so the modification was done only on the cache and copy was called later on again. But in v0.5.1 it doesn't because the error thrown when trying to set the `ConstraintSet` in Cbc is `MOI.UnsupportedAttribute` instead of `MOI.CannotSetAttribute`.
 
Here is the function in v0.5.1
```julia
function replace_constraint_function_or_set!(m::CachingOptimizer, attr, cindex, replacement)
    if m.state == AttachedOptimizer
        if m.mode == Automatic
            try
                MOI.set!(m.optimizer, attr, m.model_to_optimizer_map[cindex], replacement)
            catch err
                if err isa MOI.CannotSetAttribute
                    resetoptimizer!(m)
                else
                    rethrow(err)
                end
            end
        else
            MOI.set!(m.optimizer, attr, m.model_to_optimizer_map[cindex], replacement)
        end
    end
    MOI.set!(m.model_cache, attr, cindex, replacement)
end
```

And here is the function in v0.4.1
```julia
function replace_constraint_function_or_set!(m::CachingOptimizer, attr, cindex, replacement)
    if (m.mode == Automatic && m.state == AttachedOptimizer &&
            !MOI.canset(m.optimizer, attr, typeof(cindex)) )
        resetoptimizer!(m)
    end
    MOI.set!(m.model_cache, attr, cindex, replacement)
    if m.state == AttachedOptimizer
        MOI.set!(m.optimizer, attr, m.model_to_optimizer_map[cindex], replacement)
    end
end
```

IssamT:
Cbc now only needs this to be running with MOI v0.5.1

### DIFF
--- a/src/Utilities/cachingoptimizer.jl
+++ b/src/Utilities/cachingoptimizer.jl
@@ -268,9 +268,9 @@ end
 # ConstraintSet and ConstraintFunction methods, but allows us to strongly type
 # the third and fourth arguments of the set! methods so that we only support
 # setting the same type of set or function.
-function replace_constraint_function_or_set!(m::CachingOptimizer, attr, cindex, replacement)                                             
+function replace_constraint_function_or_set!(m::CachingOptimizer, attr, cindex, replacement)
     if m.state == AttachedOptimizer
-        if m.mode == Automatic            
+        if m.mode == Automatic
             try
                 MOI.set!(m.optimizer, attr, m.model_to_optimizer_map[cindex], replacement)
             catch err
--- a/src/Utilities/cachingoptimizer.jl
+++ b/src/Utilities/cachingoptimizer.jl
@@ -270,15 +270,19 @@ end
 # setting the same type of set or function.
 function replace_constraint_function_or_set!(m::CachingOptimizer, attr, cindex, replacement)
     if m.state == AttachedOptimizer
-        if m.mode == Automatic
-            try
-                MOI.set!(m.optimizer, attr, m.model_to_optimizer_map[cindex], replacement)
-            catch err
-                if err isa MOI.CannotSetAttribute
-                    resetoptimizer!(m)
-                else
-                    rethrow(err)
+        if m.mode == Automatic            
+            if MOI.supports(m.optimizer, attr, typeof(cindex))
+                try
+                    MOI.set!(m.optimizer, attr, m.model_to_optimizer_map[cindex], replacement)
+                catch err
+                    if err isa MOI.CannotSetAttribute
+                        resetoptimizer!(m)
+                    else
+                        rethrow(err)
+                    end
                 end
+            else
+                resetoptimizer!(m)
             end
         else
             MOI.set!(m.optimizer, attr, m.model_to_optimizer_map[cindex], replacement)
--- a/src/Utilities/cachingoptimizer.jl
+++ b/src/Utilities/cachingoptimizer.jl
@@ -268,21 +268,17 @@ end
 # ConstraintSet and ConstraintFunction methods, but allows us to strongly type
 # the third and fourth arguments of the set! methods so that we only support
 # setting the same type of set or function.
-function replace_constraint_function_or_set!(m::CachingOptimizer, attr, cindex, replacement)
+function replace_constraint_function_or_set!(m::CachingOptimizer, attr, cindex, replacement)                                             
     if m.state == AttachedOptimizer
         if m.mode == Automatic            
-            if MOI.supports(m.optimizer, attr, typeof(cindex))
-                try
-                    MOI.set!(m.optimizer, attr, m.model_to_optimizer_map[cindex], replacement)
-                catch err
-                    if err isa MOI.CannotSetAttribute
-                        resetoptimizer!(m)
-                    else
-                        rethrow(err)
-                    end
+            try
+                MOI.set!(m.optimizer, attr, m.model_to_optimizer_map[cindex], replacement)
+            catch err
+                if err isa MOI.CannotSetAttribute || err isa MOI.UnsupportedAttribute
+                    resetoptimizer!(m)
+                else
+                    rethrow(err)
                 end
-            else
-                resetoptimizer!(m)
             end
         else
             MOI.set!(m.optimizer, attr, m.model_to_optimizer_map[cindex], replacement)


src/Utilities/cachingoptimizer.jl
```diff
@@ -270,15 +270,19 @@ end
...
-        if m.mode == Automatic
-            try
-                MOI.set!(m.optimizer, attr, m.model_to_optimizer_map[cindex], replacement)
-            catch err
-                if err isa MOI.CannotSetAttribute
-                    resetoptimizer!(m)
-                else
-                    rethrow(err)
+        if m.mode == Automatic            
+            if MOI.supports(m.optimizer, attr, typeof(cindex))
```
blegat: I would rather just add `|| (err isa MOI.UnsupportedAttribute)` in the `catch` call to be consistent with
https://github.com/JuliaOpt/MathOptInterface.jl/blob/45095a5accb1ec6a9f8a37b7cf628df91e9cc9af/src/Utilities/cachingoptimizer.jl#L254

IssamT: But here we have an additional lookup in the dictionary that can be avoided since `cindex` is of the same type as `m.model_to_optimizer_map[cindex]`.

blegat: If `supports` return `false` then `resetoptimizer!` (long) will be called and this happens only once every `optimize` call (rare). We don't want a negligible optimization in a rare case, the case we want to make fast is when this is supported. In that case, it's better not to call `MOI.supports` as it might account for a nonnegligible time

IssamT:
You're right !

src/Utilities/cachingoptimizer.jl
```diff
@@ -268,21 +268,17 @@ end
 # ConstraintSet and ConstraintFunction methods, but allows us to strongly type
 # the third and fourth arguments of the set! methods so that we only support
 # setting the same type of set or function.
-function replace_constraint_function_or_set!(m::CachingOptimizer, attr, cindex, replacement)
+function replace_constraint_function_or_set!(m::CachingOptimizer, attr, cindex, replacement)
```
blegat: trailing spaces here

blegat:
It needs to have a test similar to these ones https://github.com/JuliaOpt/MathOptInterface.jl/blob/c7c4d2db40f3d1622580904c5ae852a838e86471/test/cachingoptimizer.jl#L139-L161

blegat:
There are 3 cases

* The attribute is not supported: `supports` return false, that means that if the attribute is present in `src`, `copy!(model, src)` will fail. In that case, the optimizer should throw `MOI.UnsupportedAttribute <: Exception
* The attribute is supported but `MOI.set(model, attr, ...)` is not supported, this is the case for (https://github.com/JuliaOpt/MathOptInterface.jl/issues/459#issuecomment-411539096 and https://github.com/JuliaOpt/MathOptInterface.jl/issues/463). In that case, the optimizer should throw `MOI.UnsupportedSettingAttribute <: MOI.UnsupportedError`
* The attribute is supported, setting the attribute is supported but not in the current state of the optimizer. In that case, it should throw `MOI.CannotSetAttribute <: MOI.CannotError`

In MOI v0.5.1, the last two cases were covered by `MOI.CannotSetAttribute <: MOI.CannotError` and the first case was covered by `MOI.UnsupportedAttribute <: MOI.UnsupportedError` but the abstract type definitions didn't quite match it and this lead to confusion.

mlubin:
I'm in favor of this solution. The PR is missing:

- To check `supports` before interpreting `MOI.UnsupportedAttribute` as a sign to reset.
- Tests
- Similar handling for the objective(?)

Any objections?

IssamT:
I made this PR because I thought there was a contradiction. But I was wrong, there is no contradiction as I explained here https://github.com/JuliaOpt/MathOptInterface.jl/issues/463#issuecomment-411779220 

So why do we still need this PR?

IssamT:
Do we first agree on this? 

supports == true <=> caching and copying is enough, (no need for a bridge)

Otherwise we should first change the definition

http://www.juliaopt.org/MathOptInterface.jl/latest/apireference.html#MathOptInterface.supports

because this PR does not respect this definition.

mlubin:
> Do we first agree on this?
> supports == true <=> caching and copying is enough, (no need for a bridge)

Yes, I agree with that.

> An UnsupportedAttribute error is thrown if model does not support setting attributes attr and a CannotSetAttribute error is thrown if it supports setting attributes attr but it cannot set this attribute the current state of the model.

The definition doesn't address the option of "setting but not with `set!`" that you raised in https://github.com/JuliaOpt/MathOptInterface.jl/issues/463#issuecomment-411779220 but that could be clarified.

The bigger issue is that by `CannotSet` we really mean `CachingOptimizerShouldReset` (going back to my old point). All the behaviors implemented by `CannotSet` are explained by that name, so why be more transparent about it?

IssamT:
I am afraid I don't understand. This

> supports == true <=> caching and copying is enough, (no need for a bridge)

means that `supports` should include "setting but not with set!"

Or, do you allow `UnsupportedAttribute` to be inconsistent with `supports`?

mlubin:
I think we should consider `supports` as "setting but not with `set!`", however, that's not clear at all from the current text in the documentation.

blegat:
Since ConstraintFunction and ConstraintSet are the only attributes that can be set but that will be set during copy without using set. Therefore this PR fixes the issue as it catch UnsupportedAttribute for these 2 attributes but not for the other ones.

For the objective, it can only be set using `MOI.set!` so the cachingoptimizer should rethrow UnsupportedAttribute.

IssamT:
@blegat, if you say so then you are probably right since you have been the one working a lot with the CachingOptimizer and the bridges. But for me it's unfortunately still confusing and I don't know how I should modify the code/documentation. So feel free to close this PR and rewrite it properly.

mlubin:
> For the objective, it can only be set using MOI.set! so the cachingoptimizer should rethrow UnsupportedAttribute.

What about copy-only solvers?

blegat:
> What about copy-only solvers?

Indded, that is the use issue raised by https://github.com/JuliaOpt/MathOptInterface.jl/issues/459#issuecomment-411539096
That motivates the need of splitting `UnsupportedAttribute` in two as suggested in https://github.com/JuliaOpt/MathOptInterface.jl/pull/465#issuecomment-411556000
We then add the following in MOI
```julia
function set!(model::ModelLike, attr::AbstractModelAttribute)
    if supports(model, attr)
        throw(UnsupportedSettingAttribute(attr))
    else
        throw(UnsupportedAttribute(attr))
    end
end
# define similar set! fallback for other attribute types
function supports(model::ModelLike, attr::Union{ConstraintFunction, ConstraintSet}, ::Type{CI{F, S}})
    return supportsconstraint(model, F, S)
end
```
and then the CachingOptimizer catch `UnsupportedSettingAttribute` and rethrow `UnsupportedAttribute`.

mlubin:
Let's rethink this a bit. As far as I understand it, we need only two types of errors:

1. The call failed because "human readable message". CachingOptimizer sends these errors back to the user.

2. The call failed because "human readable message" and the CachingOptimizer (if there is one) should try to reset the model. If there's no CachingOptimizer then the user sees the human readable message.

For `set!` (1) could be called `AttributeError` and (2) could be called `AttributeErrorTryReset`.

(1) covers both "the attribute is supported but not through `set!`" and "the attribute is not supported at all". Maybe I'm missing something but I don't see the need to distinguish between the two cases programmatically. It's more important for the user to get a human readable message.

blegat:
That is inconsistent with the errors of other functions which fit into `UnsupportedError` and `CannotError`. Moreover, for the user, `TryReset` is less clear than the reason.
With three types of error

* UnsupportedAttribute <: Exception
* UnsupportedSettingAttribute <: UnsupportedError
* CannotSetAtttribute <: CannotError

The user will get a precise error message when not using the caching optimizer or when using it in manual mode and the caching optimizer know that it need to reset in the last two and need to rethrow the first one.
Having error message that are well defined without the caching optimizer seems like a better design that having them defined based on the caching optimizer IMO.

mlubin:
> That is inconsistent with the errors of other functions which fit into UnsupportedError and CannotError. 

I'm suggesting to redo these as well, into `ConstraintError` and `ConstraintErrorTryReset` for example.

> Moreover, for the user, `TryReset` is less clear than the reason.

I don't think the distinction between "unsupported" and "cannot" is especially clear to start out with.

> The user will get a precise error message when not using the caching optimizer or when using it in manual mode and the caching optimizer know that it need to reset in the last two and need to rethrow the first one.

A precise human-readable string is more important than a precise exception type.

> Having error message that are well defined without the caching optimizer seems like a better design that having them defined based on the caching optimizer IMO.

I'm not sure. Not discussing CachingOptimizer just obscures the intent of the error types and leaves solver wrapper authors confused.

blegat:
> A precise human-readable string is more important than a precise exception type.

The message in `showerror` is human-readable isn't it ?

blegat:
We could split UnsupportedError into Unsupported and UnsupportedTryReset

mlubin:
> We could split UnsupportedError into Unsupported and UnsupportedTryReset

Do you mean that `UnsupportedTryReset` would replace `Cannot...`?

blegat:
Superseeded by https://github.com/JuliaOpt/MathOptInterface.jl/pull/468

PR #467, opened 2018-08-10
fix NLP tests

mlubin:


blegat:
Weird, I thoughr I used `grep` to check I deleted all of them

PR #468, opened 2018-08-10
Refactor errors

blegat:
Now `CannotError` is for operations (so if it is not supported, the caching optimizer can try to reset) and `UnsupportedError` is for elelments of a model (so if it is not supported, the copy cannot be done).

- [x] update doc
- [x] make it pass the tests

Closes https://github.com/JuliaOpt/MathOptInterface.jl/issues/459
Closes https://github.com/JuliaOpt/MathOptInterface.jl/issues/463


src/constraints.jl
```diff
@@ -53,7 +55,11 @@ it supports `F`-in-`S` constraints but it cannot add the constraint(s) in its
 current state.
 """
 function addconstraint!(model::ModelLike, func::AbstractFunction, set::AbstractSet)
-    throw(UnsupportedConstraint{typeof(func), typeof(set)}())
+    if supports(model, typeof(func), typeof(set))
+        throw(UnsupportedConstraint{typeof(func), typeof(set)}())
+    else
+        throw(CannotAddConstraint{typeof(func), typeof(set)}())
```
mlubin: This would read better IMO as `CannotAddConstraintTryReset`.

src/constraints.jl
```diff
@@ -53,7 +55,11 @@ it supports `F`-in-`S` constraints but it cannot add the constraint(s) in its
 current state.
 """
 function addconstraint!(model::ModelLike, func::AbstractFunction, set::AbstractSet)
-    throw(UnsupportedConstraint{typeof(func), typeof(set)}())
+    if supports(model, typeof(func), typeof(set))
```
mlubin: If not supports?

mlubin:
Could you update the documentation for `get!` according to this proposal? I don't fully understand it yet.

src/constraints.jl
blegat: The try reset can be in the doc or error message but I don't see why it would help to add it in the name. It is for every cannoterror, not just cannotaddconstraint

mlubin: I would argue that `TryReset` should be in the name of all the "cannot" types.

blegat:
I am on my phone until tomorrow. set! should throw UnsupportedAttribute if supports return false and throw CannotSetAttribute if supports return true but it cannot be set.

supports doc : returns false if the presence of the attribute in ListOf... in src make the copy of src into model impossible

src/constraints.jl
blegat: Indeed ^^

src/constraints.jl
blegat: And the super type is CannotErrorTryReset ?

mlubin: That's one option.

mlubin:
I know this is blocking for updating some of the solver wrappers, but it's important that we get this right and that everyone understands the design. So let's not rush it... it will help to see a full proposal then get comments from a few people.

blegat:
Ready for review :)
@odow @joaquimg @IssamT @tkoolen @rdeits What do you think of the new errors ?

IssamT:
> Now CannotError is for operations (so if it is not supported, the caching optimizer can try to reset) and UnsupportedError is for elelments of a model (so if it is not supported, the copy cannot be done).

This clear separation is a nice improvement compared to the current situation.

blegat:
> CannotTryResetError doesn't parse well. It sounds like it cannot reset something

Indeed, `CannotError` is probably better then I think.
Saying that it is not supported by be confusing since the operation may be supported but cannot be performed for other reason (e.g. it is supported but cannot be done for a specific index, or cannot be done in this state). It seems to me that the clearer thing to do is have `CannotError` in the type, it is better to stay vague that trying to say too much in the name and be unclear. Then the solver wrapper dev will check the doc and the user will be the details in showerror.

About Unsupported, it is really that the element is not supported, that is, there is no way to set it somehow. Then as all elements supported should be supported via copy allows us to say that this is equivalent to saying that it is supported via copy but saying "unsupported via copy" as the definition is confusing as it makes is sounds like it is supported but should be set via a different operation than copy

IssamT:
I also don't like TryReset. But I prefer the names given by Benoit although they have a minor issue: a future developer may want to add UnsupportedModification for example and that's not something we want to do anymore. If I understood well (correct me if i am wrong), we want both stateful and stateless operations to be handled by the same Exceptions so that the only difference is between **Elements** and **operations**.

In other words, If you ask someone who has not read the code "what do you think is the difference between CannotError and UnsupportedError?" he will probably think of the **Stateful vs Stateless** instead of **Element vs Operations** (and that's why we choose them in the beginning). While Element errors are usually stateless, operation errors might be stateful or stateless. I see this as a minor issue as it can easily be clarified in the documentation.

But since we are renaming let's think of more suited words to separate **Element vs Operation** errors

How about:

IncompatibleElementError
RejectedOperationError

with declinations like:

IncompatibleConstraintError
RejectedModificationError

mlubin:
@IssamT I like the distinction you're making. Instead of "rejected" what about "not allowed"? `NotAllowedError`, `AddConstraintNotAllowed`, `ModificationNotAllowed`, `AddVariableNotAllowed` all sound nice.

Changing "unsupported" to "incompatible" seems like a wash. They both mean similar things. I would still argue to rename `supports` to `supports_via_copy` or `copy_supports` because the definition is essentially tied to `copy!`.

IssamT:
`NotAllowed` sounds good to me.

Notice that unlike `Incompatible`, `Unsupported` clearly tells that the error is not due to the state of the model. If we want to give the information that all Element errors are independent of the model state, we should probably keep `Unsupported`.

IssamT:
>  I would still argue to rename supports to supports_via_copy or copy_supports because the definition is essentially tied to copy!.

Imagine I have within my package two formulations for a given problem, one is linear and the other is quadratic. I let the user of the package provide the MOI solver he wants to work with and the formulation is chosen automatically based on `supports` answer. that's not a `copy` issue right?

IssamT:
In the example I mentioned, MOI is used directly (without JuMP and without the caching layer). You can communicate directly with the `Optimizer` once you know what **Elements** it supports. `copy_supports` sounds restrictive to me, since the "copy support" is just one result among others for "element support".

mlubin:
If you're using MOI directly, how can you use the result of `supports` without thinking about `copy!`? I don't see it as restrictive, rather adding a description of the essential part of the definition.

mlubin:
`supportsconstraint` implies only that you can build the formulation in a `ModelLike` and load it into the solver using `copy!`. It does not imply that `add_constraint` works.

blegat:
I am ok with `Unsupported` and `NotAllowed`. About adding copy in the name, I am not yet convinced, it read a bit like "it is supported but it need to be given via copy". Let's do a quick vote to get more feedback. What do you prefer :

* `UnsupportedViaCopyError` and `supports_via_copy` :+1: 
* `CopyUnsupportedError` and `copy_supports` :heart: 
* `UnsupportedError` and `supports` :tada: 

cc @odow @joaquimg @chriscoey

mlubin:
I'm not arguing for `CopyUnsupportedError`, that's a bit silly. "Unsupported" is clear on it's own; it means e.g., that the constraint can't be added using any method, so `UnsupportedError` is fine.

The affirmative `supports_constraint` is more confusing because the current definition implies only that you can add the constraint using `copy!`. It's a necessary but not sufficient condition for being able to add the constraint using `add_constraint` (or `addconstraint!` or whatever we call it). I'm arguing for `supports_constraint` to be renamed to `copy_supports_constraint` or `supports_constraint_via_copy` for this reason.

blegat:
I suggest we move the discussion on renaming copy out of this PR. I have created a separate issue: https://github.com/JuliaOpt/MathOptInterface.jl/issues/469.
I have just renamed Cannot to NotAllowed. Any remaining objection to merge the PR ? :)

src/Utilities/cachingoptimizer.jl
```diff
@@ -168,7 +168,7 @@ function MOI.addvariable!(m::CachingOptimizer)
             try
                 vindex_optimizer = MOI.addvariable!(m.optimizer)
             catch err
-                if err isa MOI.CannotAddVariable
+                if err isa MOI.NotAllowedError
```
mlubin: What's the motivation for not checking for the specific error type like `AddVariableNotAllowed`? If the wrong error type like `DeleteNotAllowed` is thrown, that's an error that shouldn't silently be ignored.

src/Utilities/mockoptimizer.jl
```diff
@@ -19,7 +19,9 @@ mutable struct MockOptimizer{MT<:MOI.ModelLike} <: MOI.AbstractOptimizer
     conattribute::Dict{MOI.ConstraintIndex,Int} # MockConstraintAttribute
     needsallocateload::Bool # Allows to tests the Allocate-Load interface, see copy!
     canaddvar::Bool
-    canaddcon::Bool # If false, the optimizer throws CannotAddConstraint
+    canaddcon::Bool # If false, the optimizer throws AddConstraintNotAllowed
```
mlubin: Nit: For consistency these should be named like `add_con_allowed` instead of `canaddcon`.

test/errors.jl
```diff
@@ -0,0 +1,130 @@
...
+
+@testset "Fallbacks for `set!` methods" begin
+    model = DummyModel()
+
+    @testset "AddVariableNotAllowed" begin
+        @test_throws MOI.AddVariableNotAllowed MOI.addvariable!(model)
+        try
+            MOI.addvariable!(model)
+        catch err
+            @test sprint(showerror, err) == "MathOptInterface.AddVariableNotAllowed: Adding variables cannot be performed. You may want to use a `CachingOptimizer` in `Automatic` mode or you may need to call `resetoptimizer!` before doing this operation if the `CachingOptimizer` is in `Manual` mode."
```
mlubin: Nit: Can these error messages be split across lines using `*`?

src/Utilities/cachingoptimizer.jl
blegat: That may allow the user to create new error types. Also, trying to do one modification might result in another one and hence another error.
For instance, if you want to add a constraint and the bridge try to add a variable then it will throw a `AddVariableNotAllowed`.
We have already discussed this and I know you don't like that but if we start doing try catch everywhere, we loose the advantage of having removed `canget`. The advantage of errors is that you don't have to check that it is allowed and you just do the operation. (Except the cachingoptimizer)

mlubin: I'm not a big fan but ok.

PR #472, opened 2018-08-16
Make models broadcastable without Ref

blegat:
It is quite clear that a model is zero-dimensional. This PR allows to use broadcast with a model without the need to use `Ref`

### DIFF
--- a/src/MathOptInterface.jl
+++ b/src/MathOptInterface.jl
@@ -8,6 +8,9 @@ Abstract supertype for objects that implement the "Model" interface for defining
 an optimization problem.
 """
 abstract type ModelLike end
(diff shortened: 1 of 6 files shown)


src/MathOptInterface.jl
```diff
@@ -8,6 +8,9 @@ Abstract supertype for objects that implement the "Model" interface for defining
 an optimization problem.
 """
 abstract type ModelLike end
+@static if VERSION >= v"0.7-"
```
mlubin: Add comments explaining what this does. Are there any other MOI types we should define this for?

blegat: Yes, I just started with one to see if every one is on the same page

PR #473, opened 2018-08-16
Add DualObjectiveValue

blegat:
Closes https://github.com/JuliaOpt/MathOptInterface.jl/issues/681

### DIFF
--- a/src/Utilities/results.jl
+++ b/src/Utilities/results.jl
@@ -23,6 +23,115 @@ function get_fallback(model::MOI.ModelLike, ::MOI.ObjectiveValue)
     return evalvariables(vi -> MOI.get(model, MOI.VariablePrimal(), vi), f)
 end
 
+scalar_constant(T::Type, ::MOI.SingleVariable) = zero(T)
+scalar_constant(::Type, f::MOI.AbstractScalarFunction) = _constant(f)
+
+function constraint_constant(model::MOI.ModelLike,
+                             ci::MOI.ConstraintIndex{<:MOI.AbstractVectorFunction,
+                                                     <:MOI.AbstractVectorSet},
+                             T::Type)
+    return _constant(MOI.get(model, MOI.ConstraintFunction(), ci))
+end
+function constraint_constant(model::MOI.ModelLike,
+                             ci::MOI.ConstraintIndex{<:MOI.AbstractScalarFunction,
+                                                     <:MOI.AbstractScalarSet},
+                             T::Type)
+    return scalar_constant(T, MOI.get(model, MOI.ConstraintFunction(), ci)) -
+           getconstant(MOI.get(model, MOI.ConstraintSet(), ci))
+end
+
+"""
+    dual_objective_value(model::MOI.ModelLike,
+                         F::Type{<:MOI.AbstractFunction},
+                         S::Type{<:MOI.AbstractSet},
+                         T::Type)
+
+Return the part of `DualObjectiveValue` due to the constraint of index `ci` using
+scalar type `T`.
+"""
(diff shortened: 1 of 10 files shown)


mlubin:
How much breakage will this cause in the solver interfaces? Are we sure that the solvers can implement this? Do we really want the fallback for `ObjectiveBound` to be such an expensive operation? The dual objective isn't even a valid bound on the objective if there are violations of dual feasibility. I'd tempted to say that we should be honest that the solver doesn't have one available.

@joaquimg @odow

blegat:
Ready for review :)

blegat:
Any objection to this new attribute ?

src/Utilities/results.jl
```diff
@@ -23,6 +23,123 @@ function get_fallback(model::MOI.ModelLike, ::MOI.ObjectiveValue)
...
+
+Return the part of `DualObjectiveValue` due to `F`-in-`S` constraints using scalar
+type `T`.
+"""
+function dual_objective_value(model::MOI.ModelLike,
+                         F::Type{<:MOI.AbstractFunction},
+                         S::Type{<:MOI.AbstractSet},
+                         T::Type,
+                         result_index::Integer)
+    value = zero(T) # sum won't work if there are now constraints.
```
mlubin: s/now/no/

src/Utilities/results.jl
```diff
@@ -23,6 +23,123 @@ function get_fallback(model::MOI.ModelLike, ::MOI.ObjectiveValue)
...
+
+function is_ray(status::MOI.ResultStatusCode)
+    return status == MOI.INFEASIBILITY_CERTIFICATE ||
+           status == MOI.NEARLY_INFEASIBILITY_CERTIFICATE
+end
+
+"""
+    get_fallback(model::MOI.ModelLike, ::MOI.DualObjectiveValue, T::Type)::T
+
+Compute the dual objective value of type `T` using the `ConstraintDual` results
```
mlubin: Add a note that the result ignores NLP constraints.

src/Utilities/results.jl
```diff
@@ -23,6 +23,123 @@ function get_fallback(model::MOI.ModelLike, ::MOI.ObjectiveValue)
...
+    value = zero(T) # sum will not work if there are zero constraints
+    for (F, S) in MOI.get(model, MOI.ListOfConstraints())
+        value += dual_objective_value(model, F, S, T, attr.result_index)::T
+    end
+    if MOI.get(model, MOI.ObjectiveSense()) != MOI.MAX_SENSE
+        value = -value
+    end
+    if !is_ray(MOI.get(model, MOI.DualStatus()))
+        # The objective constant should not be present in rays
+        F = MOI.get(model, MOI.ObjectiveFunctionType())
```
mlubin: Is this logic correct for quadratic objectives? Isn't there a contribution from a quadratic objective to the dual? https://en.wikipedia.org/wiki/Quadratic_programming#Lagrangian_duality

src/attributes.jl
```diff
@@ -470,6 +470,17 @@ struct ObjectiveValue <: AbstractModelAttribute
     (::Type{ObjectiveValue})(resultindex=1) = new(resultindex)
 end
 
+"""
+    DualObjectiveValue(result_index::Int=1)
+
+A model attribute for the value of the objective function of the dual problem
+for the `resultindex`th dual result.
```
mlubin: s/resultindex/result_index/

src/Utilities/results.jl
blegat: There is no constant in the objective in the wiki article.
The constant in the primal objective is simply added as a constant in the dual objective

PR #478, opened 2018-08-19
Add SOCtoQuad bridge

blegat:
Adds a SOCtoQuad bridges which allows to use the
```julia
@constraint(model, [t, x, y] in SecondOrderCone())
```
with solvers expecting SOC constraint in quadratic form (e.g. [LQOI](https://github.com/JuliaOpt/LinQuadOptInterface.jl) solvers).

It also add missing methods in MOIU for quadratic functions.
While doing this, I noticed, that the `_pair` and `_canonicalize` functions of `src/functions.jl` were doing the same thing than `MOIU.termindices` and `MOIU.coefficient`. Hence I moved `coefficient` and `termindices` (rename `term_indices` following the style guide) to MOI and renamed and documented `_pair` into `term_pair`.


src/Bridges/soctoquadbridge.jl
```diff
@@ -0,0 +1,83 @@
...
+\\sum x_i^2 \\le t^2
+```
+with ``t \\ge 0``.
+"""
+struct SOCtoQuadBridge{T, F, G} <: AbstractBridge
+    quad::CI{F, MOI.GreaterThan{T}}
+    t_nonneg::CI{G, MOI.GreaterThan{T}}
+end
+function SOCtoQuadBridge{T, F, G}(model::MOI.ModelLike,
+                                  f::MOI.AbstractVectorFunction,
```
mlubin: This transformation is recognized by Gurobi/Cplex/Xpress(?) only when `f` is a `VectorOfVariables`. They can't handle general affine expressions. That creates a tricky situation because we don't have a way to indicate this restriction in MOI afaik, so I'm questioning if bridges are the right place for this transformation to happen. Maybe the solvers need to support `VectorOfVariables`-in-`SecondOrderCone` and then let a bridge handle only the transformation from `VectorAffineFunction`-in-`SecondOrderCone` to `VectorOfVariables`-in-`SecondOrderCone`.

blegat: They also support x^T Q x +q^T x <= b apparently: http://www.gurobi.com/documentation/8.0/refman/constraints.html#subsubsection:QuadraticConstraints
so if it is a VectorAffineFunction in SOC, it will be transformed into `x^T Q x +q^T x <= b` with `Q` PSD an Gurobi will also support it, won't it ?

mlubin: This bridge does not generate a constraint of the form x^T Q x +q^T x <= b where Q is PSD (there's one negative eigenvalue).

mlubin: That works for Gurobi et al, but I think there's a larger design issue. Ipopt supports convex and non-convex quadratic constraints but not SOC constraints (there's no special detection for x^T x <= t^2). It this bridge is applied, Ipopt will treat the constraint as a nonconvex quadratic constraint without any guarantees of global optimality. Unless there are additional properties of this transformation that imply that a KKT point must be globally optimal (even though the problem is nonconvex), this means that the user could write down a convex problem using SOC and get a bad answer back from Ipopt.

joaquimg: just adding info to the discussion: Xpress detects automatically SOCs (and RSOCs), and they can only be added in the form: x^T Q x +q^T x <= b as quadratic constraints.

blegat: > this means that the user could write down a convex problem using SOC and get a bad answer back from Ipopt

If the user use an NLP solver, he shouldn't expect globally optimal solution, even if with a conic program. One could even solve an SDP with an NLP solver, and the solution shouldn't be expected to be optimal.

> Xpress detects automatically SOCs (and RSOCs), and they can only be added in the form: x^T Q x +q^T x <= b as quadratic constraints.

Does it need Q to be PSD, i.e. does it detect `x^2 + y^2 - t^2 <= 0` as a SOC constraint ?

joaquimg: Xpress detects that SOC constraints. No need for PSD matrices.

blegat: We discussed about this issue with @mlubin and @ccoffrin. The approach would be to define this bridge in MOI but not add it in `full_bridge_optimizer` so that the user would have to add it manually with `JuMP.add_bridge`. By choosing which bridge to add, it would either allow `VectorOfVariables`-in-`SOC` or `VectorAffineFunction`-in-`SOC` to add so he is responsible to add the one with which the optimizer works.
IMO the optimizer could also simply apply the bridge automatically by itself with a `SingleBridgeOptimizer`. Choosing to build the API on top of quadratic functions instead of SOC constraints was a bad design choice and this is the reason this bridge seems weird because it destroys a structure that the optimizers try to recover afterwards.

joaquimg: Since the three solvers support SOC´s as special quadratic constraints, the bridge could be very useful. Given Ipopt's problem, we could have it non-default, and add a note about the issue.

mlubin: @joaquimg IMO the right solution for this is for Gurobi/Xpress/CPLEX to explicitly support VoV/VAF-in-SOC in their interfaces. We have no other way to know if a solver secretly supports SOC through quadratic constraints.

ccoffrin: > this means that the user could write down a convex problem using SOC and get a bad answer back from Ipopt.

Regarding this point.  As I understand the theory, it does not matter in what form a convex constraint is given to a IPM solver.  The KKT convergence guarantees will be the same; so I don't think this a major issue for the case of targeting NLP solvers.

PR #479, opened 2018-08-20
Remove canget

blegat:
Here are the following changes:

* If a constraint or variable attribute is set to a strict subset of the variables/constraints, it used to return `canget` equal to false and not be included in `ListOf...AttributesSet`. No, it is included in the list independently on the number of variables/constraints with this attribute set.
* To check if a name was already present, one could use `canget` and this was used by the bridgeoptimizer and universalfallback to check whether it is the name of a bridged constraints or of a constraint of the internal model. This required 2 dictionary lookup: 1 for canget and 1 for get. If get is called with a name that does not exists an error was thrown. Now with this PR, `get` returns `nothing` in case the name does not exists. This allows bridges and fallback to have almost no overhead in case the name not the name of a bridged constraint as this is [efficiently handled on Julia v1.0](https://julialang.org/blog/2018/08/union-splitting)
* `NoSolution` has been added as a new result status which specifies that the result has no primal or dual solution, e.g. a result could have only a dual solution in which case the primal status is `NoSolution`.

Closes https://github.com/JuliaOpt/MathOptInterface.jl/issues/424
Closes https://github.com/JuliaOpt/MathOptInterface.jl/issues/302

### DIFF
--- a/src/Test/contconic.jl
+++ b/src/Test/contconic.jl
@@ -47,31 +47,22 @@ function _lin1test(model::MOI.ModelLike, config::TestConfig, vecofvars::Bool)
     if config.solve
         MOI.optimize!(model)
 
-        @test MOI.canget(model, MOI.TerminationStatus())
         @test MOI.get(model, MOI.TerminationStatus()) == MOI.Success
 
-        @test MOI.canget(model, MOI.PrimalStatus())
         @test MOI.get(model, MOI.PrimalStatus()) == MOI.FeasiblePoint
         if config.duals
-            @test MOI.canget(model, MOI.DualStatus())
             @test MOI.get(model, MOI.DualStatus()) == MOI.FeasiblePoint
         end
 
-        @test MOI.canget(model, MOI.ObjectiveValue())
         @test MOI.get(model, MOI.ObjectiveValue()) ≈ -11 atol=atol rtol=rtol
 
-        @test MOI.canget(model, MOI.VariablePrimal(), MOI.VariableIndex)
         @test MOI.get(model, MOI.VariablePrimal(), v) ≈ [1, 0, 2] atol=atol rtol=rtol
 
-        @test MOI.canget(model, MOI.ConstraintPrimal(), typeof(vc))
         @test MOI.get(model, MOI.ConstraintPrimal(), vc) ≈ [1, 0, 2] atol=atol rtol=rtol
-        @test MOI.canget(model, MOI.ConstraintPrimal(), typeof(c))
         @test MOI.get(model, MOI.ConstraintPrimal(), c) ≈ zeros(2) atol=atol rtol=rtol
 
         if config.duals
-            @test MOI.canget(model, MOI.ConstraintDual(), typeof(vc))
             @test MOI.get(model, MOI.ConstraintDual(), vc) ≈ [0, 2, 0] atol=atol rtol=rtol
-            @test MOI.canget(model, MOI.ConstraintDual(), typeof(c))
             @test MOI.get(model, MOI.ConstraintDual(), c) ≈ [-3, -1] atol=atol rtol=rtol
         end
     end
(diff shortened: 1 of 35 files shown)


src/Test/contconic.jl
```diff
@@ -225,17 +203,16 @@ function lin3test(model::MOI.ModelLike, config::TestConfig)
...
 
-        @test MOI.canget(model, MOI.TerminationStatus())
         if config.infeas_certificates
             @test MOI.get(model, MOI.TerminationStatus()) == MOI.Success
         else
             @test MOI.get(model, MOI.TerminationStatus()) in [MOI.InfeasibleNoResult, MOI.InfeasibleOrUnbounded]
         end
-        if MOI.canget(model, MOI.PrimalStatus())
-            @test MOI.get(model, MOI.PrimalStatus()) == MOI.InfeasiblePoint
+        if config.infeas_certificates || MOI.get(model, MOI.ResultCount()) > 0
```
mlubin: The 2 "or" statements here are a bit confusing: `config.infeas_certificates || MOI.get(model, MOI.ResultCount()) > 0`
and `MOI.get(model, MOI.PrimalStatus()) in (MOI.NoSolution, MOI.InfeasiblePoint)`.

Could you untangle what is expected in what cases and/or add a comment about why these different cases can happen?

Doesn't `config.infeas_certificates` imply that we expect `MOI.get(model, MOI.ResultCount()) > 0` and PrimalStatus to be `NoSolution`?

Same comment applies right below.

src/Test/contconic.jl
```diff
@@ -932,32 +836,23 @@ function _geomean1test(model::MOI.ModelLike, config::TestConfig, vecofvars, n=3)
...
-        @test MOI.canget(model, MOI.VariablePrimal(), MOI.VariableIndex)
         @test MOI.get(model, MOI.VariablePrimal(), x) ≈ ones(n) atol=atol rtol=rtol
 
-        @test MOI.canget(model, MOI.ConstraintPrimal(), typeof(gmc))
         @test MOI.get(model, MOI.ConstraintPrimal(), gmc) ≈ ones(n+1) atol=atol rtol=rtol
 
-        @test MOI.canget(model, MOI.ConstraintPrimal(), typeof(c))
         @test MOI.get(model, MOI.ConstraintPrimal(), c) ≈ n atol=atol rtol=rtol
 
     #    if config.duals
```
mlubin: Since you're touching nearby code, could you add a comment explaining why this block is commented?

src/Test/contquadratic.jl
```diff
@@ -294,13 +260,10 @@ function qcp1test(model::MOI.ModelLike, config::TestConfig)
     #
```
mlubin: Why is this code commented?

src/Utilities/cachingoptimizer.jl
```diff
@@ -428,52 +428,78 @@ function MOI.supports(m::CachingOptimizer, attr::MOI.AbstractModelAttribute)
...
+                                     MOI.DualStatus,
+                                     MOI.VariablePrimal,
+                                     MOI.ConstraintBasisStatus,
+                                     MOI.ConstraintPrimal,
+                                     MOI.ConstraintDual,
+                                     MOI.ConstraintBasisStatus})
+    return true
+end
+function MOI.get(model::CachingOptimizer, attr::MOI.AbstractModelAttribute)
+    if is_result_attribute(attr)
```
mlubin: How would solver-specific result attributes work? Like `GurobiIIS`?

src/attributes.jl
```diff
@@ -174,52 +174,6 @@ function get!(output, model::ModelLike, attr::AnyAttribute, args...)
     throw(ArgumentError("ModelLike of type $(typeof(model)) does not support accessing the attribute $attr"))
 end
```
mlubin: (Applies a few lines above.) The documentation for `get` needs to be updated to describe when it now returns `nothing`.

src/Utilities/cachingoptimizer.jl
blegat: They should implement is_result_attribute.

mlubin: What about using `supports` to decide whether to ask the optimizer or the model cache (with priority to the optimizer)? I can't easily think of a case where this would give the wrong answer, and is much simpler.

blegat: `supports` is not currently defined for attributes that never appear in `ListOf...AttributesSet` so it does not make sense for result attibutes which are not settable

mlubin: This points to a larger design issue that needs to be resolved before tagging 0.6, but maybe this PR isn't the best place to discuss. It's definitely not ok for solvers to be required to overload functions in MOIU to get correct behavior.

blegat: Discussion moved to https://github.com/JuliaOpt/MathOptInterface.jl/issues/480 for more visibility

PR #483, opened 2018-08-23
Add QuadtoSOC bridge

blegat:
Currently, it only bridges `x^T Q x + a^T x + b <= c` for PSD matrix Q and does not support `x^T Q x` with one negative eigenvalue, the reason is discussed in the docstring.

Closes https://github.com/JuliaOpt/MathOptInterfaceBridges.jl/issues/93


src/Bridges/quadtosocbridge.jl
```diff
@@ -0,0 +1,179 @@
...
+# References
+function MOI.delete!(model::MOI.ModelLike, bridge::QuadtoSOCBridge)
+    MOI.delete!(model, bridge.soc)
+end
+
+# Attributes, Bridge acting as a constraint
+function MOI.get(model::MOI.ModelLike, attr::MOI.ConstraintPrimal,
+                 bridge::QuadtoSOCBridge)
+    soc = MOI.get(model, attr, bridge.soc)
+    @show soc
```
mlubin: Debugging output?

src/Bridges/quadtosocbridge.jl
```diff
@@ -0,0 +1,179 @@
...
+    if !less_than
+        set_constant = -set_constant
+    end
+    Q, index_to_variable_map = matrix_from_quadratic_terms(func.quadratic_terms)
+    @show Q
+    if !less_than
+        rmul!(Q, -1)
+    end
+    @show Q
+    U = cholesky(Symmetric(Q)).U
```
mlubin: How are errors handled if `Q` is not PSD?

mlubin: I'm concerned about the use of a dense factorization. This is probably the wrong thing to do when `Q` isn't trivially small. (< 100x100). The fact that this bridge will be applied automatically means that users are going to run into a surprise performance bomb with no feedback about what was wrong.

blegat:
@mlubin All your comments have been addressed :)

chriscoey:
It would be cool if Julia could automatically use sparse cholesky if Q is very sparse but dense cholesky otherwise. There has been discussion of making the dense and sparse factorization interfaces fully consistent but no one has stepped up to the plate yet.

blegat:
I also got confused by the inconsistency. For sparse cholesky, there is a permutation matrix and you have `A = P' * L * L' * P` while for dense cholesky, you have `A = L * L'`. Since `P` is identity most of the time, that could easily lead to code that works fine most of the time and suddenly fail. It is unclear from the `cholesky` doc if the permutation matrix can be non-identity if the `perm` argument is not provided. Should check in the SuiteSparse doc, I have opened an issue here: https://github.com/JuliaOpt/MathOptInterface.jl/issues/538

src/Bridges/quadtosocbridge.jl
```diff
@@ -0,0 +1,197 @@
...
+get the opposite of the rotated second-order cone. Note that `y` and `z` have
+the same sign since `yz` is nonnegative hence ``x + y \\ge 0`` is equivalent to
+``x \\ge 0`` and ``y \\ge 0``.
+
+### Note
+
+The check for existence of bound can be implemented (but inefficiently) with the
+current interface but if bound is removed or transformed (e.g. `≤ 0` transformed
+into `≥ 0`) then the bridge is no longer valid. For this reason the homogeneous
+version of the bridge is not implemented yet.
```
mlubin: This sentence at the end is the only hint in this docstring about what the bridge actually does. Say more explicitly at the beginning which transformations are implemented.

src/Bridges/quadtosocbridge.jl
```diff
@@ -0,0 +1,197 @@
...
+            sparse(F[:L]), F[:p]
+        end
+    catch err
+        if err isa PosDefException
+            error("The optimizer supports second-order cone constraints and",
+                  " not quadratic constraints but you entered a quadratic",
+                  " constraint of type: `$(typeof(func))`-in-`$(typeof(set))`.",
+                  " A bridge attempted to transform the quadratic constraint",
+                  " to a second order cone constraint but the constraint is not",
+                  " convex.")
```
mlubin: This is incorrect. `cholfact` will fail when `Q` is only positive semi-definite, but that's a perfectly valid convex constraint.

chriscoey: You could try LDLT instead right? Or bunch-kaufman if dense

src/Bridges/quadtosocbridge.jl
```diff
@@ -86,8 +87,9 @@ function QuadtoSOCBridge{T}(model, func::MOI.ScalarQuadraticFunction{T},
                   " not quadratic constraints but you entered a quadratic",
                   " constraint of type: `$(typeof(func))`-in-`$(typeof(set))`.",
                   " A bridge attempted to transform the quadratic constraint",
-                  " to a second order cone constraint but the constraint is not",
-                  " convex.")
+                  " to a second order cone constraint but the constraint is",
+                  " not strongly convex, i.e., the symmetric matrix of"
+                  " quadratic coefficients is not positive definite.")
```
mlubin: This wording suggests that the user did something wrong, while the issue is in the implementation of the bridge instead. Maybe "... not yet supported" is more appropriate.

PR #490, opened 2018-08-24
Support to bridges with external sets/functions

IssamT:
It is working for what I needed but I'd consider the PR as a WIP. There are still two issues mainly:

- The sets/functions need to be explicitly imported in the current scope.
- I had no clue how to name the new equivalent functions.

### DIFF
--- a/src/Utilities/model.jl
+++ b/src/Utilities/model.jl
@@ -406,7 +406,7 @@ end
 # Expr(:., MOI, :($(QuoteNode(s)))) is Expr(:., MOI, :(:EqualTo)) <- what we want
 
 # (MOI, :Zeros) -> :(MOI.Zeros)
-_mod(m::Module, s::Symbol) = Expr(:., m, :($(QuoteNode(s))))
+_mod(m::Module, s::Symbol) = s
 # (:Zeros) -> :(MOI.Zeros)
 _moi(s::Symbol) = _mod(MOI, s)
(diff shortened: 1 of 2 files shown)


src/Utilities/model.jl
```diff
@@ -410,14 +410,31 @@ _mod(m::Module, s::Symbol) = Expr(:., m, :($(QuoteNode(s))))
 # (:Zeros) -> :(MOI.Zeros)
 _moi(s::Symbol) = _mod(MOI, s)
```
blegat: Why not simply removing this function and force the user to write
```julia
@model LPModel () (MOI.EqualTo, MOI.GreaterThan, MOI.LessThan, MOI.Interval) (MOI.Zeros, MOI.Nonnegatives, MOI.Nonpositives) () (MOI.SingleVariable,) (MOI.ScalarAffineFunction,) (MOI.VectorOfVariables,) (MOI.VectorAffineFunction,)
```

blegat: If you remove this _moi function and all calls to it then it should work

mlubin: FWIW the escaping/scoping rules for macros changed between 0.6 and 0.7. You might run into this.

odow: Forcing the module prefix sounds like the most robust way to go.

IssamT:
replaced by #510

PR #491, opened 2018-08-24
Rename copy! -> copy_to

blegat:
Also rename `defaultcopy!` to `default_copy_to`.

Related to https://github.com/JuliaOpt/MathOptInterface.jl/issues/475


mlubin:
Style question: should we define `copy_to` or extend `Base.copyto!`?

odow:
Define `copy_to`. Internal consistency over using multiple dispatch just for the sake of it.

blegat:
The copy here is not really identical since we need to return a variable map. If some copies are done silently because we overwrite `Base.copyto!` it might indicate an unexpected copy and hence unexpected bugs and/or inefficiency.

PR #493, opened 2018-08-25
Rename allocateload

blegat:


src/Utilities/copy.jl
```diff
@@ -128,13 +128,13 @@ end
...
-    needsallocateload(model::MOI.ModelLike)::Bool
+    needs_allocate_load(model::MOI.ModelLike)::Bool
 
 Return a `Bool` indicating whether `model` does not support `add_variables`/`add_constraint`/`set` but supports `allocatevariables!`/`allocateconstraint!`/`allocate!`/`loadvariables!`/`loadconstraint!`/`load!`.
 That is, the allocate-load interface need to be used to copy an model to `model`.
 """
-function needsallocateload end
-needsallocateload(::MOI.ModelLike) = false
+function needs_allocate_load end
+needs_allocate_load(::MOI.ModelLike) = false
```
odow: Why do we need two definitions? Isn't the second sufficient?

PR #495, opened 2018-08-25
Fix #494

### LINKED ISSUE #494: copynames defaults to false in bridge optimizer
The keyword argument copynames is set to false here:
https://github.com/JuliaOpt/MathOptInterface.jl/blob/615a40b5111e5e69d14550882935066c9440810b/src/Bridges/bridgeoptimizer.jl#L90
while it should be true by default

blegat:
Closes https://github.com/JuliaOpt/MathOptInterface.jl/issues/494

### DIFF
--- a/src/Bridges/bridgeoptimizer.jl
+++ b/src/Bridges/bridgeoptimizer.jl
@@ -87,7 +87,7 @@ function MOI.supports(b::AbstractBridgeOptimizer,
     return MOI.supports(b.model, attr)
 end
 function MOI.copy_to(b::AbstractBridgeOptimizer, src::MOI.ModelLike;
-                   copynames=false)
+                   copynames = true)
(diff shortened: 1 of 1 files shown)


src/Bridges/bridgeoptimizer.jl
```diff
@@ -87,7 +87,7 @@ function MOI.supports(b::AbstractBridgeOptimizer,
     return MOI.supports(b.model, attr)
 end
 function MOI.copy_to(b::AbstractBridgeOptimizer, src::MOI.ModelLike;
-                   copynames=false)
+                   copynames = true)
```
mlubin: copy_names?

blegat: Yes, we should rename that too, I planned to do this in the next PR

Issue #499, opened 2018-08-25
InexactError needs three arguments in Julia 1.0


lostella:
An error is raised in Julia 1.0 by [this line](https://github.com/JuliaOpt/MathOptInterface.jl/blob/84efa5fc49b102b3a49ee50408c41166b1f3e182/src/Utilities/model.jl#L232): no-arguments `InexactError` was deprecated in Julia 0.7, hence was removed from Julia 1.0.

PR #502, opened 2018-08-26
Test InexactError thrown for bad objective function type

blegat:
Closes https://github.com/JuliaOpt/MathOptInterface.jl/issues/499


blegat:
Yes, I use it locally

PR #504, opened 2018-08-26
Define convert

blegat:
It will be quite hard to implement `MOI.get(::OptimizerType, ::ObjectiveFunction{T})` passing the unit test added in https://github.com/JuliaOpt/MathOptInterface.jl/pull/502 without the convert functions added in this PR.

The two conversions that were present in MOIU were moved to MOI as they do not add utilites but rather define constructor of MOI types.

### DIFF
--- a/src/functions.jl
+++ b/src/functions.jl
@@ -381,6 +381,10 @@ function VectorAffineFunction{T}(f::VectorOfVariables) where T
     return VectorAffineFunction(map(i -> VectorAffineTerm(i, ScalarAffineTerm(one(T), f.variables[i])), 1:n), zeros(T, n))
 end
 
+if VERSION < v"0.7-"
+    isone(x::T) = x == one(T)
+end
+
 # Conversion between scalar functions
 # Conversion to SingleVariable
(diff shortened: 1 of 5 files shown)


src/functions.jl
```diff
@@ -369,3 +369,59 @@ Return a new quadratic function with a shallow copy of the terms and constant(s)
 from `func`.
 """
 Base.copy(func::F) where {F <: Union{ScalarQuadraticFunction, VectorQuadraticFunction}} = F(copy(func.affine_terms), copy(func.quadratic_terms), copy(_constant(func)))
+
+# Define shortcuts for
+# SingleVariable -> ScalarAffineFunction
+function ScalarAffineFunction{T}(f::SingleVariable) where T
```
mlubin: Is it still Julia convention (post-1.0) to add constructor methods for conversions like this?

blegat: If I understand correctly, since constructors are the low level functions  for building objects, convert should redirect to constructors not the opposite

mlubin:
Please fix the CI failure before merging.

PR #510, opened 2018-08-27
Support of bridges with external sets/funcs

IssamT:
Simpler solution as suggested by @blegat. Of course we can now delete _mod function and get rid of its calls.

However even with this solution, we still need to import sets/functions to be used by @bridge

If we want the modules passed to @bridge like `MOI.xxxfunction/MOI.xxxset`, we need more modifications in @model since it uses the names passed to create field names of the new structure. Do we go for that?

### DIFF
--- a/src/Bridges/singlebridgeoptimizer.jl
+++ b/src/Bridges/singlebridgeoptimizer.jl
@@ -17,7 +17,7 @@ is_bridged(b::SingleBridgeOptimizer, ::Type{<:MOI.AbstractFunction}, ::Type{<:MO
 bridge_type(b::SingleBridgeOptimizer{BT}, F::Type{<:MOI.AbstractFunction}, S::Type{<:MOI.AbstractSet}) where BT = BT
 
 # :((Zeros, SecondOrderCone)) -> (:(MOI.Zeros), :(MOI.SecondOrderCone))
-_tuple_prefix_moi(t) = MOIU._moi.(t.args)
+_tuple_prefix_moi(t) = t.args
 
 """
 macro bridge(modelname, bridge, scalarsets, typedscalarsets, vectorsets, typedvectorsets, scalarfunctions, typedscalarfunctions, vectorfunctions, typedvectorfunctions)
@@ -39,8 +39,8 @@ will additionally support `ScalarAffineFunction`-in-`Interval`.
 """
 macro bridge(modelname, bridge, ss, sst, vs, vst, sf, sft, vf, vft)
     bridged_model_name = Symbol(string(modelname) * "Instance")
-    bridged_funs = :(Union{$(_tuple_prefix_moi(sf)...), $(_tuple_prefix_moi(sft)...), $(_tuple_prefix_moi(vf)...), $(_tuple_prefix_moi(vft)...)})
-    bridged_sets = :(Union{$(_tuple_prefix_moi(ss)...), $(_tuple_prefix_moi(sst)...), $(_tuple_prefix_moi(vs)...), $(_tuple_prefix_moi(vst)...)})
+    bridged_funs = :(Union{$((sf.args)...), $((sft.args)...), $((vf.args)...), $((vft.args)...)})
+    bridged_sets = :(Union{$((ss.args)...), $((sst.args)...), $((vs.args)...), $((vst.args)...)})
 
     esc(quote
         $MOIU.@model $bridged_model_name $ss $sst $vs $vst $sf $sft $vf $vft
--- a/src/Bridges/singlebridgeoptimizer.jl
+++ b/src/Bridges/singlebridgeoptimizer.jl
@@ -45,7 +45,7 @@ macro bridge(modelname, bridge, ss, sst, vs, vst, sf, sft, vf, vft)
     esc(quote
         $MOIU.@model $bridged_model_name $ss $sst $vs $vst $sf $sft $vf $vft
         const $modelname{T, OT<:MOI.ModelLike} = $MOIB.SingleBridgeOptimizer{$bridge{T}, $bridged_model_name{T}, OT}
-        is_bridged(::$modelname, ::Type{<:$bridged_funs}, ::Type{<:$bridged_sets}) = true
+        MathOptInterface.Bridges.is_bridged(::$modelname, ::Type{<:$bridged_funs}, ::Type{<:$bridged_sets}) = true
         supports_bridging_constraint(::$modelname, ::Type{<:$bridged_funs}, ::Type{<:$bridged_sets}) = true
     end)
 end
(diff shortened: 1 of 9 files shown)


blegat:
Yes, you can create unique field names with `gensym`.

src/Bridges/singlebridgeoptimizer.jl
```diff
@@ -45,7 +45,7 @@ macro bridge(modelname, bridge, ss, sst, vs, vst, sf, sft, vf, vft)
     esc(quote
         $MOIU.@model $bridged_model_name $ss $sst $vs $vst $sf $sft $vf $vft
         const $modelname{T, OT<:MOI.ModelLike} = $MOIB.SingleBridgeOptimizer{$bridge{T}, $bridged_model_name{T}, OT}
-        is_bridged(::$modelname, ::Type{<:$bridged_funs}, ::Type{<:$bridged_sets}) = true
+        MathOptInterface.Bridges.is_bridged(::$modelname, ::Type{<:$bridged_funs}, ::Type{<:$bridged_sets}) = true
```
blegat: You can use `$MOI`

IssamT:
> If we want the modules passed to @bridge like MOI.xxxfunction/MOI.xxxset, we need more modifications in @model since it uses the names passed to create field names of the new structure.

I guess I should have tested first. I was surprised that this is not needed and the reason is: 

```julia
julia> struct abc
           my.name::String
       end
ERROR: syntax: field name "my.name" is not a symbol
```

As Expected. But using metaprogramming you can put a dot inside the name of a field, although it s not clean at all.

```julia
julia> s = Symbol("my.name")
Symbol("my.name")

julia> code = quote
              struct abc
                 $(s)::String
              end
              end
quote
    #= REPL[24]:2 =#
    struct abc
        #= REPL[24]:3 =#
        my.name::String
    end
end

julia> eval(code)

julia> abc("Bob")
abc("Bob")

Maybe I will transform it to underscore to keep readable names? `gensym` will make debugging hard. No?

**Comment by blegat** [comment]
@IssamT The fields are not shown in printing so I have never seen the fields in fact. If you can make it work with underscores, it is fine too.

File: src/Utilities/model.jl [hunk 2]
```diff
@@ -409,9 +409,9 @@ end
 # Expr(:., MOI, :($(QuoteNode(s)))) is Expr(:., MOI, :(:EqualTo)) <- what we want
 
 # (MOI, :Zeros) -> :(MOI.Zeros)
-_mod(m::Module, s::Symbol) = Expr(:., m, :($(QuoteNode(s))))
+_mod(m::Module, s::Union{Symbol,Expr}) = s
 # (:Zeros) -> :(MOI.Zeros)
-_moi(s::Symbol) = _mod(MOI, s)
+_moi(s::Union{Symbol,Expr}) = _mod(MOI, s)
 _set(s::SymbolSet) = _moi(s.s)
```
blegat: Let's remove `_moi` and `_mod`

src/Utilities/model.jl
```diff
@@ -393,12 +393,12 @@ abstract type Constraints{F} end
 
 abstract type SymbolFS end
 struct SymbolFun <: SymbolFS
-    s::Symbol
+    s::Union{Symbol,Expr}
```
odow: Nit: missing a space `Union{Symbol, Expr}`

src/Utilities/model.jl
```diff
@@ -579,14 +577,14 @@ macro model(modelname, ss, sst, vs, vst, sf, sft, vf, vft)
     end
 
     for (func, T) in ((:_add_constraint, CI), (:_modify, CI), (:_delete, CI), (:_getindex, CI), (:_getfunction, CI), (:_getset, CI), (:_getnoc, MOI.NumberOfConstraints))
-        funct = _mod(MOIU, func)
+        funct = func
```
odow: Why have this extra variable?

src/Utilities/model.jl
```diff
@@ -624,6 +622,6 @@ macro model(modelname, ss, sst, vs, vst, sf, sft, vf, vft)
         $MOI.supports_constraint(model::$modelname{T}, ::Type{<:Union{$(_typedfun.(vectorfuns)...)}}, ::Type{<:Union{$(_typedset.(vectorsets)...)}}) where T = true
 
         $code
-
-    end)
+    end
+    esc(code)
```
odow: Style: missing a `return`

test/runtests.jl
```diff
@@ -20,16 +20,16 @@ end
...
-               (EqualTo, GreaterThan, LessThan, Interval, Semicontinuous, Semiinteger),
-               (Reals, Zeros, Nonnegatives, Nonpositives, SecondOrderCone, RotatedSecondOrderCone, GeometricMeanCone, ExponentialCone, DualExponentialCone, PositiveSemidefiniteConeTriangle, PositiveSemidefiniteConeSquare, RootDetConeTriangle, RootDetConeSquare, LogDetConeTriangle, LogDetConeSquare),
-               (PowerCone, DualPowerCone, SOS1, SOS2),
-               (SingleVariable,),
-               (ScalarAffineFunction, ScalarQuadraticFunction),
-               (VectorOfVariables,),
-               (VectorAffineFunction, VectorQuadraticFunction))
+               (MOI.ZeroOne, MOI.Integer),
+               (MOI.EqualTo, MOI.GreaterThan, MOI.LessThan, MOI.Interval, MOI.Semicontinuous, MOI.Semiinteger),
+               (MOI.Reals, MOI.Zeros, MOI.Nonnegatives, MOI.Nonpositives, MOI.SecondOrderCone, MOI.RotatedSecondOrderCone, MOI.GeometricMeanCone, MOI.ExponentialCone, MOI.DualExponentialCone, MOI.PositiveSemidefiniteConeTriangle, MOI.PositiveSemidefiniteConeSquare, MOI.RootDetConeTriangle, MOI.RootDetConeSquare, MOI.LogDetConeTriangle, MOI.LogDetConeSquare),
```
odow: Style: wrap really long lines? (Lots of places with this in the model definitions.)

src/Bridges/singlebridgeoptimizer.jl
```diff
@@ -17,7 +17,7 @@ is_bridged(b::SingleBridgeOptimizer, ::Type{<:MOI.AbstractFunction}, ::Type{<:MO
 bridge_type(b::SingleBridgeOptimizer{BT}, F::Type{<:MOI.AbstractFunction}, S::Type{<:MOI.AbstractSet}) where BT = BT
 
 # :((Zeros, SecondOrderCone)) -> (:(MOI.Zeros), :(MOI.SecondOrderCone))
-_tuple_prefix_moi(t) = MOIU._moi.(t.args)
+_tuple_prefix_moi(t) = t.args
```
blegat: It seems the function is not used anymore, let's remove it

PR #512, opened 2018-08-28
Remove implementation and testing of supports for ConstraintSet

blegat:
`ConstraintSet` and `ConstraintFunction` are not copyable (see `is_copyable`) so they should not implement `supports`.


mlubin:
Please fix CI failures.

PR #513, opened 2018-08-28
Broadcast attributes

blegat:
I had to put `Ref`'s in ECOS because of broadcasting attributes and now that I am updating SDOI, I need them again... Probably better to just define the attributes here as scalar :)

### DIFF
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -36,6 +36,12 @@ abstract type AbstractConstraintAttribute end
 
 const AnyAttribute = Union{AbstractOptimizerAttribute, AbstractModelAttribute, AbstractVariableAttribute, AbstractConstraintAttribute}
 
+@static if VERSION >= v"0.7-"
+    # This allows to use attributes in broadcast calls without the need to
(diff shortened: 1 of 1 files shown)


src/attributes.jl
```diff
@@ -183,7 +189,7 @@ function get end
 # We want to avoid being too specific in the type arguments to avoid method ambiguity.
 # For model, get(::ModelLike, ::AbstractVariableAttribute, ::Vector{VariableIndex}) would not allow
 # to define get(::SomeModel, ::AnyAttribute, ::Vector)
-get(model::ModelLike, attr::AnyAttribute, idxs::Vector) = get.(model, Ref(attr), idxs)
+get(model::ModelLike, attr::AnyAttribute, idxs::Vector) = get.(model, attr, idxs)
```
mlubin: Do we still need this fallback or should users just broadcast themselves?

blegat: Some solvers may do something faster,  e.g. if they have a fixed cost that can be shared when getting by group

Issue #530, opened 2018-09-29
allocate_constraint and load_constraint should throw UnsupportedConstraint error


blegat:
For `add_constraint` we have a nice fallback:

https://github.com/JuliaOpt/MathOptInterface.jl/blob/38a4e883672641da575b284813254a61747c5854/src/constraints.jl#L58-L62

But not for `allocate_constraint` and `load_constraint` which leads to cryptic errors such as https://github.com/JuliaOpt/SCS.jl/issues/112

PR #534, opened 2018-10-09
Add errors for Allocate-Load and bad constraint

blegat:
This PR brings two improvements to user friendliness:

* Allocate-Load equivalents to `set` and `add_constraint` now returns nice user-facing errors instead of method errors
* `add_constraint` with VectorFunction-in-ScalarSet or ScalarFunction-in-VectorSet now returns an error instead of an UnsupportedError relying on the fact that the model implements `supports_constraint` correctly or an `AddConstraintNotAllowed` otherwise which are both less user-friendly

Closes https://github.com/JuliaOpt/MathOptInterface.jl/issues/530

### DIFF
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -278,28 +278,28 @@ function set(model::ModelLike,
 end
 
 function set(model::ModelLike, attr::AnyAttribute, args...)
-    set_fallback_error(model, attr, args...)
+    throw_set_error_fallback(model, attr, args...)
 end
-# set_fallback_error is included so that we can return type-specific error
+# throw_set_error_fallback is included so that we can return type-specific error
 # messages without needing to overload set and cause ambiguity errors. For
-# examples, see ConstraintSet and ConstraintFunction. set_fallback_error should
+# examples, see ConstraintSet and ConstraintFunction. throw_set_error_fallback should
 # not be overloaded by users of MOI.
-function set_fallback_error(model::ModelLike,
-                            attr::Union{AbstractModelAttribute,
-                                        AbstractOptimizerAttribute},
-                            value;
-                            error_if_supported = SetAttributeNotAllowed(attr))
+function throw_set_error_fallback(model::ModelLike,
+                                  attr::Union{AbstractModelAttribute,
+                                              AbstractOptimizerAttribute},
+                                  value;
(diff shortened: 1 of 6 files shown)


src/attributes.jl
```diff
@@ -285,21 +285,23 @@ end
 # examples, see ConstraintSet and ConstraintFunction. set_fallback_error should
 # not be overloaded by users of MOI.
 function set_fallback_error(model::ModelLike,
```
mlubin: This name could be misunderstood as setting a property of the model call the "fallback error". Maybe `throw_set_error_fallback`?

src/constraints.jl
```diff
@@ -54,9 +54,51 @@ An [`UnsupportedConstraint`](@ref) error is thrown if `model` does not support
...
+        throw()
+    else
+        throw(UnsupportedConstraint{typeof(func), typeof(set)}())
+    end
+end
+
+# add_constraint_fallback_error checks whether func and set are both scalar
+# or both vector. If it is the case, it calls
+# `correct_add_constraint_fallback_error`
+function add_constraint_fallback_error(model::ModelLike,
```
mlubin: Same comment about the name.

test/errors.jl
```diff
@@ -92,12 +117,14 @@
     end
 
     @testset "UnsupportedAttribute" begin
-        @test_throws MOI.UnsupportedAttribute begin
-            MOI.set(model, MOI.ObjectiveFunction{MOI.SingleVariable}(),
-                     MOI.SingleVariable(vi))
-        end
-        @test_throws MOI.UnsupportedAttribute begin
-            MOI.set(model, MOI.ConstraintDualStart(), ci, 0.0)
+        for f in (MOI.set, MOIU.load, MOIU.allocate)
```
mlubin: Consider `@testset "Unsupported attribute with $f" for f in (MOI.set, MOIU.load, MOIU.allocate)`

test/errors.jl
```diff
@@ -110,6 +137,22 @@
         end
     end
 
+    for f in (MOIU.load, MOIU.allocate)
+        @testset "$f errors" begin
```
mlubin: `@testset "$f errors" for f in (MOIU.load, MOIU.allocate)`

PR #535, opened 2018-10-11
RFC: construct the name to object dicts lazily

### LINKED ISSUE #536: Duplicate names
We want to stop requiring models to error when a duplicate name is set, see https://github.com/JuliaOpt/MathOptInterface.jl/pull/535.
What should be the requirement now ?
Maybe we can require it to error when the user does `MOI.get(model, MOI.VariableIndex, name)` where two or more variables have the name `name` and if there is only one variable having the name `name` but some other names have duplicates then it might error or not, we leave the choice in the implementation (in model, it errors when building the dict: https://github.com/JuliaOpt/MathOptInterface.jl/pull/535)
What do you think ?

We should also decide whether a model is *allowed* to error on `MOI.set` even if it is not *required*

mlubin:
Following up on the discussion in https://github.com/JuliaOpt/JuMP.jl/issues/1402. This change avoids creating the `names_to_var` dictionary until it's needed.

Using @IainNZ's script that benchmarks variable creation time,

Before:
```
10^1:  0.000 s, 3981.300000 ns/var, 286 a, 28.600000 a/var
10^2:  0.000 s, 1016.850000 ns/var, 844 a, 8.440000 a/var
10^3:  0.001 s, 703.424000 ns/var, 6756 a, 6.756000 a/var
10^4:  0.007 s, 664.980500 ns/var, 69774 a, 6.977400 a/var
10^5:  0.100 s, 1003.817160 ns/var, 699808 a, 6.998080 a/var
10^6:  1.558 s, 1558.397552 ns/var, 6999859 a, 6.999859 a/var
```

After:
```
10^1:  0.000 s, 3638.900000 ns/var, 286 a, 28.600000 a/var
10^2:  0.000 s, 829.310000 ns/var, 838 a, 8.380000 a/var
10^3:  0.001 s, 520.511000 ns/var, 6742 a, 6.742000 a/var
10^4:  0.005 s, 491.230300 ns/var, 69754 a, 6.975400 a/var
10^5:  0.077 s, 766.043240 ns/var, 699776 a, 6.997760 a/var
10^6:  1.005 s, 1004.825593 ns/var, 6999809 a, 6.999809 a/var
```
so we get the ~0.5 seconds back as predicted.

These results are with an immutable `VariableInfo` struct (https://github.com/JuliaOpt/JuMP.jl/issues/1525).

Closes #536

### DIFF
--- a/src/Utilities/model.jl
+++ b/src/Utilities/model.jl
@@ -195,13 +195,13 @@ function MOI.get(model::AbstractModel, ::Type{VI}, name::String)
     if model.name_to_var === nothing
         # Rebuild the map.
         model.name_to_var = Dict{String, VI}()
-        for (var, name) in model.var_to_name
-            isempty(name) && continue
-            if haskey(model.name_to_var, name)
-                error("Variable $var and $(model.name_to_var[name]) have the " *
-                      "same name.")
+        for (var, var_name) in model.var_to_name
+            isempty(var_name) && continue
+            if haskey(model.name_to_var, var_name)
+                error("Variable $var and $(model.name_to_var[var_name]) have " *
+                      "the same name.")
             end
-            model.name_to_var[name] = var
+            model.name_to_var[var_name] = var
         end
     end
     return get(model.name_to_var, name, nothing)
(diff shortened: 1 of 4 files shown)


src/Utilities/model.jl
```diff
@@ -168,48 +168,70 @@ function check_can_assign_name(model::MOI.ModelLike, IndexType::Type{<:MOI.Index
...
 
 function MOI.get(model::AbstractModel, ::Type{VI}, name::String)
-    return get(model.namesvar, name, nothing)
+    if model.name_to_var === nothing
+        # Rebuild the map.
+        model.name_to_var = Dict{String, VI}()
+        for (var, var_name) in model.var_to_name
+            isempty(var_name) && continue
+            if haskey(model.name_to_var, var_name)
+                error("Variable $var and $(model.name_to_var[var_name]) have " *
```
blegat: We should do `model.name_to_var = nothing` before erroring otherwise, we leave the model in an incorrect state

PR #537, opened 2018-10-13
Allow all correct solution for solve_blank_obj

blegat:
CSDP, OSQP and ECOS fail this tests because `x` is not 1.0 but there is no reason why `x` should exactly be 1.0


chriscoey:
bump

PR #542, opened 2018-10-24
Prepare NEWS for MOI v0.6.2

blegat:
I have used unicode symbols to help the user distinguish between bug fixes (🐛), performance improvements (⚡️), usability improvements (🚸) and new features (✨).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,17 @@
 MathOptInterface (MOI) release notes
 ====================================
 
+v0.6.2 (October 24, 2018)
+---------------------------
+
+- 🐛 Fix bug in UniversalFallback attribute getter (#540).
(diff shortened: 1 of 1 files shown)


joaquimg:
Using them consistently means using them always in the news.md right? And not making it mandatory in commits right?

I like these emojis, I am not 100% convinced. Certainly this classification is very good.

Will it generate a dependency? If the emojis change it will affect older news.md?

mlubin:
The division into bug fixes/performance/new features is useful, but I think a similar benefit with less indirection could be gained by writing the words out explicitly as @odow suggested.

ccoffrin:
I'll also vote for words over icons.  But still a stalwart supporter of gitmoji in commits and PRs.

blegat:
The division can be done in a separate PR that does it for all NEWS.md items

PR #543, opened 2018-10-25
🐛 Fix error with UnknownSet

blegat:
This error makes LQOI fails with MOI v0.6.2 (see https://github.com/JuliaOpt/LinQuadOptInterface.jl/issues/67). The PR adds a test to catch the error and fixes it.

The error comes from the fact that `UnknownSet` is neither an `AbstractScalarSet`  nor an `AbstractVectorSet` hence a method error is thrown, see:
https://github.com/JuliaOpt/MathOptInterface.jl/blob/00b5965ea1491a893af01c0bf8b5501a0f11c4bb/src/constraints.jl#L65-L88

### DIFF
--- a/test/dummy.jl
+++ b/test/dummy.jl
@@ -1,19 +1,25 @@
-# DummyModel{true} implements add_variable and add_constraint
-struct DummyModel{A} <: MOI.ModelLike
-end
-MOI.add_variable(::DummyModel{true}) = MOI.VariableIndex(0)
-function MOI.empty!(::DummyModel) end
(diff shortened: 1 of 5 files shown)


test/dummy.jl
```diff
@@ -1,9 +1,19 @@
-struct DummyModel <: MOI.ModelLike
+# DummyModel{true} implements add_variable and add_constraint
+struct DummyModel{A} <: MOI.ModelLike
```
mlubin: What about an abstract dummy model and two concrete subclasses to distinguish between the versions that implement `add_variable` or not.

Issue #523, opened 2018-09-19
Lighter bridges for a brighter life


blegat:
When writing a solver wrapper, we often wonder whether to use bridges or not for a given constraint types that is not natively supported by the solver.
Using a bridge has the advantage of making the solver wrapper simpler and more maintainable but sometimes, when the constraint can be transformed to a supported one easily, we are tempted to support this constraint type too even if it is not supported by the solver to make sure that the transformation is efficiently.

Let's take ECOS to illustrate this. ECOS natively supports 4 types of constraints: VectorAffineFunction in Zeros, Nonnegatives, SecondOrderCone and ExponentialCone.

a) However, we may be tempted to also support VectorOfVariables in these sets since the transformation from VectorOfVariables to VectorAffineFunction is trivial:
https://github.com/JuliaOpt/ECOS.jl/blob/db356d4771c6f6f89e440db72be845ae57c4dcf9/src/MOIWrapper.jl#L170
It is one line but we need to allocate a new vector affine function so in fact it may be not that efficient, see below.
b) We may also be tempted to support Nonpositives since we just need to flip the sign to obtain Nonnegatives:
https://github.com/JuliaOpt/ECOS.jl/blob/db356d4771c6f6f89e440db72be845ae57c4dcf9/src/MOIWrapper.jl#L133
Again, we do `-coeffficients` with the vector of coefficients so it actually does a copy of the vector of coefficients which is not that efficient.
c) We can then also support scalar constraints such as SingleVariable, ScalarAffineFunction in EqualTo, LessThan, GreaterThan.
https://github.com/JuliaOpt/ECOS.jl/blob/db356d4771c6f6f89e440db72be845ae57c4dcf9/src/MOIWrapper.jl#L148-L169
Here we need 20 lines (plus a few lines here and there) so here efficiency is not questioned but it take many lines of code to write, test and maintain.

As we see with the ECOS example, not using bridges either adds a lot of code (c) or adds small inefficiencies that can only be resolved by careful analysis and adding complicated code (a) and b)).

So should bridges be used for such simple transformations ?

Suppose we have a bridge optimizer on top of an optimizer. When adding a bridge,

1) a copy of the function and set is stored in a model cache
2) functions and sets for the constraints generated by the transformation are computed

Both 1) and 2) are O(N) in terms of time and space where N is the number of terms in the function.
[...]

PR #549, opened 2018-10-27
Allow duplicate variable names

### LINKED ISSUE #548: uniqueness of names
On the call yesterday we discussed the issue about names for display purposes being different than names that one would want to use as unique lookup keys. I proposed allowing two different names per variable/constraint, a lookup name and a display name. Thinking about it more, this seems too complicated.

New proposal: We allow variables to have arbitrary non-unique names. On lookup, if there are two variables with the same name, we throw an error. That's all. 

Pros:
- This is overall less breaking from 0.18
- Simple rules
- We already allow variables to have non-unique lookup names (the empty string), so this removes that exception

Cons:
- Authors of code that generates JuMP sub-models will have to decide if they prefer simple names that print well or names that are guaranteed to be unique and useful for lookup

@ccoffrin @jd-lara @IainNZ 

https://github.com/JuliaOpt/JuMP.jl/issues/1292

mlubin:
Throw an error only when looking up a string that corresponds to multiple
variables.

Closes #548
Closes JuliaOpt/JuMP.jl#1292

### DIFF
--- a/src/Utilities/universalfallback.jl
+++ b/src/Utilities/universalfallback.jl
@@ -14,8 +14,8 @@ mutable struct UniversalFallback{MT} <: MOI.ModelLike
     model::MT
     constraints::Dict{Tuple{DataType, DataType}, Dict} # See https://github.com/JuliaOpt/JuMP.jl/issues/1152
     nextconstraintid::Int64
-    connames::Dict{CI, String}
-    namescon::Dict{String, CI}
+    con_to_name::Dict{CI, String}
+    name_to_con::Union{Dict{String, MOI.ConstraintIndex}, Nothing}
     optattr::Dict{MOI.AbstractOptimizerAttribute, Any}
     modattr::Dict{MOI.AbstractModelAttribute, Any}
     varattr::Dict{MOI.AbstractVariableAttribute, Dict{VI, Any}}
@@ -25,7 +25,7 @@ mutable struct UniversalFallback{MT} <: MOI.ModelLike
                            Dict{Tuple{DataType, DataType}, Dict}(),
                            0,
                            Dict{CI, String}(),
-                           Dict{String, CI}(),
+                           nothing,
                            Dict{MOI.AbstractOptimizerAttribute, Any}(),
                            Dict{MOI.AbstractModelAttribute, Any}(),
                            Dict{MOI.AbstractVariableAttribute, Dict{VI, Any}}(),
@@ -39,8 +39,8 @@ function MOI.empty!(uf::UniversalFallback)
     MOI.empty!(uf.model)
     empty!(uf.constraints)
     uf.nextconstraintid = 0
-    empty!(uf.connames)
-    empty!(uf.namescon)
+    empty!(uf.con_to_name)
+    uf.name_to_con = nothing
     empty!(uf.optattr)
     empty!(uf.modattr)
     empty!(uf.varattr)
@@ -65,10 +65,10 @@ function MOI.delete(uf::UniversalFallback, ci::CI{F, S}) where {F, S}
             throw(MOI.InvalidIndex(ci))
         end
         delete!(uf.constraints[(F, S)], ci)
-        if haskey(uf.connames, ci)
-            delete!(uf.namescon, uf.connames[ci])
-            delete!(uf.connames, ci)
+        if haskey(uf.con_to_name, ci)
+            delete!(uf.con_to_name, ci)
         end
+        uf.name_to_con = nothing
     end
     for d in values(uf.conattr)
         delete!(d, ci)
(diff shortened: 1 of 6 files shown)


src/Utilities/universalfallback.jl
```diff
@@ -173,41 +173,76 @@ end
...
+            uf.name_to_con[con_name] = con
+        end
     end
+    return
 end
 
 MOI.get(uf::UniversalFallback, ::Type{VI}, name::String) = MOI.get(uf.model, VI, name)
 function MOI.get(uf::UniversalFallback, ::Type{CI{F, S}}, name::String) where {F, S}
+    if uf.name_to_con === nothing
+        build_name_to_con_map(uf)
```
blegat: Why isn't this only called in the `else` clause below in the case the constraint is not supported ?

src/Utilities/universalfallback.jl
```diff
@@ -173,41 +173,76 @@ end
...
 function MOI.get(uf::UniversalFallback, attr::MOI.ConstraintName, ci::CI{F, S}) where {F, S}
     if MOI.supports_constraint(uf.model, F, S)
         return MOI.get(uf.model, attr, ci)
     else
-        return get(uf.connames, ci, EMPTYSTRING)
+        return get(uf.con_to_name, ci, EMPTYSTRING)
+    end
+end
+
+function build_name_to_con_map(uf::UniversalFallback)
```
blegat: Couldn't this function be shared with `MOIU.AbstractModel` ?

src/Utilities/universalfallback.jl
```diff
@@ -173,41 +173,76 @@ end
...
+    uf.name_to_con = Dict{String, CI}()
+    for (con, con_name) in uf.con_to_name
+        if haskey(uf.name_to_con, con_name)
+            # -1 is a special value that means this string does not map to
+            # a unique constraint name.
+            uf.name_to_con[con_name] = CI{Nothing, Nothing}(-1)
```
blegat: This line is not covered by tests, this can be resolved either by adding a test or by sharing the code with `MOIU.AbstractModel`

src/Utilities/universalfallback.jl
```diff
@@ -173,41 +173,76 @@ end
...
+        if ci !== nothing && haskey(uf.name_to_con, name)
+            error("Multiple constraints have the name $name.")
+        end
+        return ci
     else
-        ci = get(uf.namescon, name, nothing)
-        if ci isa CI{F, S}
+        ci = get(uf.name_to_con, name, nothing)
+        if ci == CI{Nothing, Nothing}(-1)
+            error("Multiple constraints have the name $name.")
```
blegat: This line is not covered by the tests. This could be resolved by sharing this  code with `MOIU.AbstractModel`.

src/Utilities/universalfallback.jl
```diff
@@ -173,41 +173,76 @@ end
...
+    if uf.name_to_con === nothing
+        build_name_to_con_map(uf)
+    end
+
     ci = MOI.get(uf.model, CI, name)
     if ci === nothing
-        return get(uf.namescon, name, nothing)
+        uf_ci = get(uf.name_to_con, name, nothing)
+        if uf_ci == CI{Nothing, Nothing}(-1)
+            error("Multiple constraints have the name $name.")
```
blegat: This line is not covered

src/Utilities/universalfallback.jl
mlubin: Now covered

src/Utilities/universalfallback.jl
mlubin: Now covered

src/Utilities/universalfallback.jl
mlubin: Now covered

src/Utilities/universalfallback.jl
mlubin: Sharing internal functions across files without a proper interface (the function depends on an object having certain fields) makes the code harder to understand IMO. You shouldn't need to understand or think about implementation details of `Model` to understand `UniversalFallback`.

blegat: You can slightly change the function, i.e. pass the fields instead of the model

blegat: With [light bridges](https://github.com/JuliaOpt/MathOptInterface.jl/issues/523), bridges will not have an internal MOIU model but they will still need to handle names.
So this handling of names will be in UniversalFallback, AbstractModel and AbstractBridgeOptimizer.
If we choose to do the same thing in all case, having a separate utilities might help, it will also show that the name treatment done in each case is generic and not specific to the model.

mlubin: Done

PR #553, opened 2018-11-01
fix logdet cone definition (needed perspective), bridge, and tests

### LINKED ISSUE #546: logdet cone definition is wrong (needs perspective variable)
the logdet function is not homogeneous, hence its hypograph (t,X) : t <= logdet(X) is convex but not conic. the "logdet cone" is therefore a misnomer. we should redefine the logdet cone as (x, y, Z) : x <= y*logdet(Z/y).

I'm happy to go through and do this.

chriscoey:
fixes #546 
the bridge and mock optimizer code is difficult to understand. please check my code carefully.

### DIFF
--- a/src/Bridges/detbridge.jl
+++ b/src/Bridges/detbridge.jl
@@ -83,7 +83,7 @@ end
 function LogDetBridge{T}(model, f::MOI.VectorAffineFunction{T}, s::MOI.LogDetConeTriangle) where T
     d = s.side_dimension
     tu, D, Δ, sdindex = extract_eigenvalues(model, f, d, 2)
-    t, u = tu[1], tu[2]
+    t, u = tu
     l = MOI.add_variables(model, d)
     lcindex = [sublog(model, l[i], u, D[i], T) for i in eachindex(l)]
     tlindex = subsum(model, copy(t), l, T)
--- a/src/Bridges/detbridge.jl
+++ b/src/Bridges/detbridge.jl
@@ -7,27 +7,29 @@ function trimap(i::Integer, j::Integer)
 end
 
 """
-    extract_eigenvalues(model, f::MOI.VectorAffineFunction{T}, d::Int) where T
+    extract_eigenvalues(model, f::MOI.VectorAffineFunction{T}, d::Int, offset::Int) where T
 
-The vector `f` contains `t` followed by the matrix `X` of dimension `d`.
-This functions extracts the eigenvalues of `X` and returns `t`,
+The vector `f` contains `t` (if `offset = 1`) or `(t, u)` (if `offset = 2`)
+followed by the matrix `X` of dimension `d`.
+This functions extracts the eigenvalues of `X` and returns a vector containing `t` or `(t, u)`,
 a vector `MOI.VariableIndex` containing the eigenvalues of `X`,
 the variables created and the index of the constraint created to extract the eigenvalues.
 """
-function extract_eigenvalues(model, f::MOI.VectorAffineFunction{T}, d::Int) where T
-    n = trimap(d, d)
-    N = trimap(2d, 2d)
-
-    Δ = MOI.add_variables(model, n)
-
+function extract_eigenvalues(model, f::MOI.VectorAffineFunction{T}, d::Int, offset::Int) where T
     f_scalars = MOIU.eachscalar(f)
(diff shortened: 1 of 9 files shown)


src/Bridges/detbridge.jl
```diff
@@ -78,32 +82,33 @@ function LogDetBridge{T}(model, f::MOI.VectorOfVariables, s::MOI.LogDetConeTrian
 end
 function LogDetBridge{T}(model, f::MOI.VectorAffineFunction{T}, s::MOI.LogDetConeTriangle) where T
     d = s.side_dimension
-    t, D, Δ, sdindex = extract_eigenvalues(model, f, d)
+    tu, D, Δ, sdindex = extract_eigenvalues(model, f, d, 2)
+    t, u = tu[1], tu[2]
```
blegat: Why not `t, u = tu` ?

src/Bridges/detbridge.jl
```diff
@@ -134,9 +139,10 @@ end
 function MOI.get(model::MOI.ModelLike, a::MOI.ConstraintPrimal, c::LogDetBridge)
     d = length(c.lcindex)
     Δ = MOI.get(model, MOI.VariablePrimal(), c.Δ)
-    t = MOI.get(model, MOI.ConstraintPrimal(), c.tlindex) - sum(log.(Δ[trimap.(1:d, 1:d)]))
+    t = MOIU.evalvariables(vi -> MOI.get(model, MOI.VariablePrimal(), vi), c.t)
```
blegat: Why not do compute it `t` from the `ConstraintPrimal` of `tlindex` and the `ConstraintPrimal` and the values of the `x`s in `lcindex`s ?
The value of `u` can be computed from one of the `ConstraintPrimal` of the `lcindex`.
This would allow to avoid storing `t` and `u` in the bridge.

src/Test/contconic.jl
```diff
@@ -1428,6 +1437,10 @@ function _det1test(model::MOI.ModelLike, config::TestConfig, vecofvars::Bool, de
 
         @test MOI.get(model, MOI.VariablePrimal(), t) ≈ expectedobjval atol=atol rtol=rtol
 
+        if logdet
+            @test MOI.get(model, MOI.VariablePrimal(), u) ≈ 1. atol=atol rtol=rtol
```
blegat: The JuMP style guide prefers `1.0` over `1.`

src/Utilities/results.jl
```diff
@@ -291,11 +291,14 @@ function set_dot(x::Vector, y::Vector,
...
 end
 
-function set_dot(x::Vector, y::Vector, set::Union{MOI.LogDetConeTriangle,
-                                                  MOI.RootDetConeTriangle})
+function set_dot(x::Vector, y::Vector, set::MOI.RootDetConeTriangle)
     return x[1] * y[1] + triangle_dot(x, y, set.side_dimension, 1)
 end
 
+function set_dot(x::Vector, y::Vector, set::MOI.LogDetConeTriangle)
+    return x[1] * y[1] + triangle_dot(x, y, set.side_dimension, 2)
```
blegat: Why isn't there `x[2] + y[2]` in the sum ? This is the generalization of the ExponentialCone so the scalar product should match the exponential cone one for 1x1 matrices

src/Bridges/detbridge.jl
chriscoey: This part caused me the most confusion. I think I don't fully understand the purpose of getting constraint primals. Could you update the PR with your suggestion? I'll address your other comments today.

src/Test/contconic.jl
chriscoey: I was following the style in surrounding code. I can do a PR for style consistency throughout moi after informs next week. If there's one thing I'm good at, it's being anal.

src/Utilities/results.jl
chriscoey: Correct, I'll fix

src/Test/contconic.jl
blegat: Yes, the surrounding code was written pre-JuMP style guide and should be updated. I agree that it fits a separate PR

chriscoey:
@blegat if you get a chance, could you make the remaining change you requested (which I'm not sure how to do) and merge?

PR #556, opened 2018-11-03
add Manifest and Project files

chriscoey:
to make running tests with latest Julia easy

### DIFF
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -1,0 +1,92 @@
+[[Base64]]
+uuid = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
+
+[[Compat]]
+deps = ["Base64", "Dates", "DelimitedFiles", "Distributed", "InteractiveUtils", "LibGit2", "Libdl", "LinearAlgebra", "Markdown", "Mmap", "Pkg", "Printf", "REPL", "Random", "Serialization", "SharedArrays", "Sockets", "SparseArrays", "Statistics", "Test", "UUIDs", "Unicode"]
+git-tree-sha1 = "2d9e14d19bad3f9ad5cc5e4cffabc3cfa59de825"
+uuid = "34da2185-b29b-5c13-b0c7-acf172513d20"
+version = "1.3.0"
+
+[[Dates]]
+deps = ["Printf"]
+uuid = "ade2ca70-3891-5945-98fb-dc099432e06a"
+
+[[DelimitedFiles]]
+deps = ["Mmap"]
+uuid = "8bb1440f-4735-579b-a4ab-409b98df4dab"
+
+[[Distributed]]
+deps = ["LinearAlgebra", "Random", "Serialization", "Sockets"]
+uuid = "8ba89e20-285c-5b6f-9357-94700520ee1b"
+
+[[InteractiveUtils]]
+deps = ["LinearAlgebra", "Markdown"]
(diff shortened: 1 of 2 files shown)


blegat:
Why do we have so much entries in Manifest.toml ? Is that Compat dependencies ?

chriscoey:
yeah I think so (it was automatically generated by pkg)

mlubin:
Hold off on merging, I'd like to take a look first.

Project.toml
```diff
@@ -0,0 +1,7 @@
+name = "MathOptInterface"
+uuid = "b8f27783-ece8-5eb3-8dc8-9495eed66fee"
+version = "0.6.2"
+
+[deps]
+Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
+Unicode = "4ec0a83e-493e-50e2-b9ac-8f72acf5a8f5"
```
odow: Where/why do we need `Unicode`?

Manifest.toml
```diff
@@ -0,0 +1,92 @@
+[[Base64]]
+uuid = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
+
+[[Compat]]
+deps = ["Base64", "Dates", "DelimitedFiles", "Distributed", "InteractiveUtils", "LibGit2", "Libdl", "LinearAlgebra", "Markdown", "Mmap", "Pkg", "Printf", "REPL", "Random", "Serialization", "SharedArrays", "Sockets", "SparseArrays", "Statistics", "Test", "UUIDs", "Unicode"]
+git-tree-sha1 = "2d9e14d19bad3f9ad5cc5e4cffabc3cfa59de825"
+uuid = "34da2185-b29b-5c13-b0c7-acf172513d20"
+version = "1.3.0"
```
mlubin: What is the procedure for maintaining this file? Do we have to manually `Pkg.update()` when Compat gets updated?

Project.toml
```diff
@@ -0,0 +1,7 @@
+name = "MathOptInterface"
+uuid = "b8f27783-ece8-5eb3-8dc8-9495eed66fee"
+version = "0.6.2"
```
mlubin: How does this version get updated when we tag a new MOI?

Project.toml
```diff
@@ -0,0 +1,7 @@
...
+
```
mlubin: Missing a [compatibility](https://docs.julialang.org/en/v1/stdlib/Pkg/#Compatibility-1) section specifying the allowed versions of Compat and julia.

chriscoey:
I have all the same questions. And yes perhaps we don't need the Manifest.

Anyone feel free to modify or replace this PR. Probably worth pinging Pkg3 devs on Slack.

Manifest.toml
odow: This gets updated automatically when you run `Pkg` operations locally in this project.

mlubin: So yes, this requires manual updates. In what workflow do you need a manifest file for the package? `Pkg.test` doesn't require it.

odow: I use a `Manifest.toml` so that my travis builds against JuMP master. I couldn't figure out how to do this in `Project.toml`:
https://github.com/odow/Kokako.jl/blob/d6e48925b450de46f3fa17adf494c10deca59471/Manifest.toml#L86-L92

chriscoey:
Closing in favor of #714

Issue #555, opened 2018-11-01
External set bridging broken on v0.6.2


IssamT:
I think #544 broke this example of external set bridging. It works with MOI v0.6.1

https://github.com/atoptima/RCSP.jl/blob/master/src/simplebridge.jl

I think it's a scoping issue, but I am not sure about the exact reason.

Here is the error I get:

```
ERROR: LoadError: LoadError: LoadError: UndefVarError: SGT not defined
Stacktrace:
 [1] top-level scope at /Users/itahiri/.julia/packages/MathOptInterface/QO3Mk/src/Utilities/model.jl:636
 [2] top-level scope at /Users/itahiri/.julia/packages/MathOptInterface/QO3Mk/src/Bridges/singlebridgeoptimizer.jl:45
 [3] include at ./boot.jl:317 [inlined]
 [4] include_relative(::Module, ::String) at ./loading.jl:1038
 [5] include at ./sysimg.jl:29 [inlined]
 [6] include(::String) at /Users/itahiri/.julia/dev/RCSP/src/RCSP.jl:1
 [7] top-level scope at none:0
 [8] include at ./boot.jl:317 [inlined]
 [9] include_relative(::Module, ::String) at ./loading.jl:1038
 [10] include(::Module, ::String) at ./sysimg.jl:29
 [11] include(::String) at ./client.jl:388
 [12] top-level scope at none:0
 [13] include at ./boot.jl:317 [inlined]
 [14] include_relative(::Module, ::String) at ./loading.jl:1038
 [15] include(::Module, ::String) at ./sysimg.jl:29
 [16] include(::String) at ./client.jl:388
 [17] top-level scope at none:0
in expression starting at /Users/itahiri/.julia/dev/RCSP/src/simplebridge.jl:68
in expression starting at /Users/itahiri/.julia/dev/RCSP/src/RCSP.jl:21
```

blegat:
I guess the sets and functions should be escaped with `esc`. We should have a test with the `@model` macro and external function/set

IssamT:
Sure. I will start by adding a test for that in MOI.

PR #557, opened 2018-11-05
Adding tests for external set model and bridge

IssamT:
related to #555


blegat:
Could you wrap it in a module and avoid defining the constants `MOI` and `MOIU` to make sure the macro doesn't use the fact that these constants are defined.

blegat:
As a fix you can try adding `esc` here:
https://github.com/JuliaOpt/MathOptInterface.jl/blob/34269f1cc714d8d34f998c81b2586f8042930608/src/Utilities/model.jl#L416-L417

PR #558, opened 2018-11-05
Fix off-diagonal entries for QuadtoSOC

blegat:


test/bridge.jl
```diff
@@ -274,6 +274,32 @@ end
                                    MOI.LessThan(0.0))
             end
         end
+        MOIU.set_mock_optimize!(mock,
```
mlubin: Consider moving into a separate testset. This code block is cluttered and it's hard to see what's being tested. Does `solve_qcp_edge_cases` test something or is it just meant to set up a problem that's later queried to see if the bridge is correct?

blegat: Yes, it should test ConstraintPrimal for quad constraint of more than one variable

PR #559, opened 2018-11-09
📝 Fail build when doc has error

blegat:


blegat:
It is currently failing but it went unnoticed https://travis-ci.org/JuliaOpt/MathOptInterface.jl/jobs/451887132#L482-L493

PR #561, opened 2018-11-09
Add supports_default_copy_to

blegat:
As discussed, we plan to make bridging happen after caching hence the need for an additional caching after the bridges to cache the bridge model.
If the optimizer returns `true` for `supports_incremental_copy` then this second cache will not be added.
If it returns `true` even if `copy_names` is true we could even get rid of the first cache.
With MOI v0.7, the second cache could also be removed if `supports_allocate_load` is true and `supports_incremental_copy` is false.
This PR should be non-breaking even if it adds this new function so it can be part of MOI v0.6.3

### DIFF
--- a/docs/src/apireference.md
+++ b/docs/src/apireference.md
@@ -359,3 +359,37 @@ Bridges.SOCtoPSDBridge
 Bridges.RSOCtoPSDBridge
 ```
 For each bridge defined in this package, a corresponding bridge optimizer is available with the same name without the "Bridge" suffix, e.g., `SplitInterval` is an `SingleBridgeOptimizer` for the `SplitIntervalBridge`.
+
+## Allocate-Load API
+
+The Allocate-Load API allows to implement [`copy_to`](@ref) in a way that still
+allows transformations to be applied in the copy between the cache and the
+model if the transformations are implemented as MOI layers implemented the
+Allocate-Load API.
+
+Allocate-Load Interface: 2-pass copy of a MathOptInterface model
+Some solver wrappers (e.g. SCS, ECOS, SDOI) do not supporting copying an optimization model using `MOI.add_constraints`, `MOI.add_variables` and `MOI.set`
+as they first need to figure out some information about a model before being able to pass the problem data to the solver.
+
+During the first pass (called allocate) : the model collects the relevant information about the problem so that
+on the second pass (called load), the constraints can be loaded directly to the solver (in case of SDOI) or written directly into the matrix of constraints (in case of SCS and ECOS).
+
+To support `MOI.copy_to` using this 2-pass mechanism, implement the allocate-load interface defined below and do:
+MOI.copy_to(dest::ModelType, src::MOI.ModelLike) = MOIU.allocate_load(dest, src)
+In the implementation of the allocate-load interface, it can be assumed that the different functions will the called in the following order:
+1) `allocate_variables`
+2) `allocate` and `allocate_constraint`
+3) `load_variables` and `allocate_constraint`
+4) `load` and `load_constraint`
+The interface is not meant to be used to create new constraints with `allocate_constraint` followed by `load_constraint` after a solve, it is only meant for being used in this order to implement `MOI.copy_to`.
+
+
+```@docs
+allocate_variables
+allocate
+allocate_constraint
+load_variables
+load
+load_constraint
+allocate_load
+```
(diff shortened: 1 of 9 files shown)


docs/src/apireference.md
```diff
@@ -359,3 +359,66 @@ Bridges.SOCtoPSDBridge
...
+transformations to be applied in the copy between the cache and the
+model if the transformations are implemented as MOI layers implementing the
+Allocate-Load API, see [Implementing copy](@ref) for more details.
+
+Loading a model using the Allocate-Load interface consists of two passes
+through the model data:
+1) the allocate pass where the model typically records the necessary information
+   about the constraints and attributes such as their number and size.
+   This information may be used by the solver to allocate datastructures of
+   appriate size.
```
odow: s/appriate/appropriate

docs/src/apireference.md
```diff
@@ -359,3 +359,66 @@ Bridges.SOCtoPSDBridge
...
+2) the load pass where the model typically loads the constraints and attributes
+   data to the model.
```
odow: "constraints and attributes data" should just be "constraint and attribute data"

docs/src/apireference.md
```diff
@@ -359,3 +359,66 @@ Bridges.SOCtoPSDBridge
...
+
+The description above only gives a suggestion of what to achieve in each pass.
+In fact the exact same constraints and attributes are passes in each pass so
```
odow: "In fact, the same constraint and attribute data is provided to each pass, so an implementation of ..."

docs/src/apireference.md
```diff
@@ -359,3 +359,66 @@ Bridges.SOCtoPSDBridge
...
+an implementation of the Allocate-Load API is free to do whatever is more
+convenient in each pass.
+
+The main difference between each pass, appart from the fact that one is
```
odow: s/appart/apart

docs/src/apireference.md
```diff
@@ -359,3 +359,66 @@ Bridges.SOCtoPSDBridge
...
+executed before the other during a copy is that the allocate pass needs to
+create and return new variable and constraint indices while during the load pass
+the appropriate constraint indices are provided.
```
odow: We need some commas in this sentence.
- after "during a copy"
- before "while during"

src/Utilities/copy.jl
```diff
@@ -1,3 +1,69 @@
...
+
+"""
+    supports_incremental_copy(model::ModelLike, copy_names::Bool)
+
+Return a `Bool` indicating whether the model `model` supports
+[`default_copy_to(model, src, copy_names=copy_names)`](@ref) if all the
+attributes set to `src` and constraints added to `src` are supported by `model`.
+
+This function can be used to determine whether a model can be loaded into
+`model` incrementally or whether it should be cached and copied at once instead.
```
mlubin: "Incremental copy" is not very clear to me because `copy_to` is just one function call, so how can it be incremental? What about `supports_default_copy_to`?

src/Utilities/copy.jl
```diff
@@ -1,3 +1,69 @@
...
+This is used by JuMP to determine whether to add a cache or not.
+If this function returns false whatever the value of `copy_names` is then
```
mlubin: Could this paragraph on what happens for different return values be formatted as a table or bullet points? I'm having trouble keeping all the possible conditions straight.

src/Utilities/copy.jl
```diff
@@ -22,23 +22,22 @@ function automatic_copy_to(dest::MOI.ModelLike, src::MOI.ModelLike;
...
-JuMP stores a cache in a `CachingOptimizer` handling the names. Then if bridges
-are used, a cache of the bridged model is stored in another `CachingOptimizer`.
-If `supports_incremental_copy(optimizer, false)` is `true` then no cache is
-used by default for the bridged model. If
-`supports_incremental_copy(optimizer, true)` is `true` then no caching is used
-by default to store the model and handling names as names are handled by
-`optimizer`.
+This is used by JuMP to determine whether to add a cache or not in two
+situations:
+1) A first cache can be used to store the model as entered by the user as well
```
mlubin: Is `1)` recognized for lists in markdown? I thought you need to do `1.`.

src/Utilities/copy.jl
```diff
@@ -22,23 +22,22 @@ function automatic_copy_to(dest::MOI.ModelLike, src::MOI.ModelLike;
...
+   as the names of variables and constraints. This cache is created if this
+   function returns `false` when `copy_names` is `true`.
+2) If bridges are used, then a second cache can be used to store of the bridged
```
mlubin: s/of//

src/Utilities/copy.jl
blegat: Yes, it also renders as `1.`. I have turned it into `1.` to make it consistent

PR #563, opened 2018-11-10
✨ Add support for copying attributes not set to all indices

### LINKED ISSUE #545: Return `nothing` for missing attributes
Return `nothing` when getting attributes for index for which it is not set for the index.
If the index is invalid, it should still throw `InvalidIndex`.
Related to https://github.com/JuliaOpt/JuMP.jl/issues/1460

blegat:
Closes #545
Closes https://github.com/JuliaOpt/JuMP.jl/issues/1460

### DIFF
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -143,7 +143,9 @@ Return an attribute `attr` of the model `model`.
 
     get(model::ModelLike, attr::AbstractVariableAttribute, v::VariableIndex)
 
-Return an attribute `attr` of the variable `v` in model `model`.
+If the attribute `attr` is set for the variable `v` in the model `model`, return
+return an empty string if `attr` is [`ConstraintName`](@ref) and `nothing`
+otherwise.
 
     get(model::ModelLike, attr::AbstractVariableAttribute, v::Vector{VariableIndex})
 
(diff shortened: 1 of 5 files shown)


src/attributes.jl
```diff
@@ -143,15 +143,19 @@ Return an attribute `attr` of the model `model`.
 
     get(model::ModelLike, attr::AbstractVariableAttribute, v::VariableIndex)
 
-Return an attribute `attr` of the variable `v` in model `model`.
+If the attribute `attr` is set for the variable `v` in the model `model`, return
+return an empty string if `attr` is [`ConstraintName`](@ref) and `nothing`
```
mlubin: This behavior is better documented on the `ConstraintName` and `VariableName` attributes by saying that they have a default value of "" if not set.

blegat:
> Do existing solvers still pass the tests? This is a breaking change in behavior.

The behavior was undefined for accessing missing attributes hence it was not tested.
Now there is a test with `start_values_test` but solvers should manually call it so it won't make their tests fail.
I just run SCS, LQOI, GLPK, MOIMosek tests and they all pass apart from https://github.com/JuliaOpt/LinQuadOptInterface.jl/issues/71 but this is not related to this PR.

src/attributes.jl
```diff
@@ -143,15 +143,17 @@ Return an attribute `attr` of the model `model`.
 
     get(model::ModelLike, attr::AbstractVariableAttribute, v::VariableIndex)
 
-Return an attribute `attr` of the variable `v` in model `model`.
+If the attribute `attr` is set for the variable `v` in the model `model`, return
```
mlubin: Should we mention that `get` should return an error instead of `nothing` for unsupported attributes?

PR #565, opened 2018-11-10
allow accessing optimizer attributes through CachingOptimizer

mlubin:
And implement SolverName for MockOptimizer

Closes #435

### DIFF
--- a/src/Utilities/cachingoptimizer.jl
+++ b/src/Utilities/cachingoptimizer.jl
@@ -470,9 +470,14 @@ function MOI.get(m::CachingOptimizer, IdxT::Type{<:MOI.Index}, name::String)
     return MOI.get(m.model_cache, IdxT, name)
 end
 
+# TODO: MOI.set for MOI.AbstractOptimizerAttribute.
+
 function MOI.get(model::CachingOptimizer, attr::MOI.AbstractOptimizerAttribute)
     # TODO: Better error message.
     @assert model.state != NoOptimizer
+    # TODO: Copyable attributes (e.g., TimeLimit) could also be stored in the
+    # cache. When MOI.set is implemented for MOI.AbstractOptimizerAttribute,
+    # make sure this case is handled correctly.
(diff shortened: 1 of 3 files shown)


src/Utilities/cachingoptimizer.jl
```diff
@@ -470,6 +470,13 @@ function MOI.get(m::CachingOptimizer, IdxT::Type{<:MOI.Index}, name::String)
     return MOI.get(m.model_cache, IdxT, name)
 end
 
+function MOI.get(model::CachingOptimizer, attr::MOI.AbstractOptimizerAttribute)
+    # TODO: Better error message.
+    @assert model.state != NoOptimizer
+    return attribute_value_map(model.optimizer_to_model_map,
```
blegat: What about solver attribute such as TimeLimit that might be set to JuMP before the optimizer is given (e.g. it is given at `JuMP.optimize!`) ?
In that case it should be stored in the cache and then copied to the optimizer.
I tried something with canget in https://github.com/JuliaOpt/MathOptInterface.jl/pull/435 but it didn't seem right. Now that canget is gone and we have `is_copyable` we can do somethink like if is is copyable then get it in the cache and otherwise get it in the optimizer (SolverName is not copyable)
We could also try to get it in the cache and if it fails try to get it in the optimizer (but only do this if it is copyable and supported by the cache I'd say)

mlubin: Trying to get it from the cache seems premature because there's no method in CachingOptimizer to set optimizer attributes :). I was trying to make a relatively small change so that https://github.com/JuliaOpt/JuMP.jl/pull/1602 works but your suggestions sound reasonable as a more correct solution.

blegat: Yes, it is a bit premature, we can leave a TODO and/or an open issue

mlubin: I added TODOs.

Issue #566, opened 2018-11-12
MOIU.Model allocates in add_constraint


mlubin:
```julia
using MathOptInterface
const MOI = MathOptInterface
using BenchmarkTools


MOI.Utilities.@model(Model,
                     (MOI.ZeroOne, MOI.Integer),
                     (),
                     (),
                     (),
                     (MOI.SingleVariable,),
                     (),
                     (),
                     ())

function add_bin_constraints(N)
    model = Model{Float64}()
    x = MOI.add_variables(model, N)
    for var in x
        MOI.add_constraint(model, MOI.SingleVariable(var), MOI.ZeroOne())
    end
end

@btime add_bin_constraints(1_000_000) 

function add_bin_constraints_fast(N)
    model = Model{Float64}()
    x = MOI.add_variables(model, N)
    categories = fill(:Cont, length(x))
    for var in x
        categories[var.value] = :Bin
    end
end

@btime add_bin_constraints_fast(1_000_000)
```
yields
```
  231.084 ms (2000105 allocations: 809.08 MiB)
  64.763 ms (69 allocations: 49.76 MiB)
```

The 64 ms time basically matches JuMP 0.18 (https://github.com/JuliaOpt/JuMP.jl/issues/1607). Using `add_constraint` is *a lot* slower and allocates twice per call. It not easy to understand and tweak the `MOIU.Model` code to see what's going on.

blegat:
It seems that `deepcopy`
https://github.com/JuliaOpt/MathOptInterface.jl/blob/fdb7ec993d11fe120efe2a64388a875dfb1fa301/src/Utilities/model.jl#L276
is the part that allocates the most.
```julia
module M
using MathOptInterface
const MOI = MathOptInterface

MOI.Utilities.@model(Model,
                     (MOI.ZeroOne, MOI.Integer),
                     (),
                     (),
                     (),
                     (MOI.SingleVariable,),
                     (),
                     (),
                     ())

const model = Model{Float64}()
const x = MOI.add_variable(model)
const f = MOI.SingleVariable(x)
const s = MOI.ZeroOne()
function a()
    MOI.add_constraint(model, f, s)
end
const cons = MOI.Utilities.C{typeof(f), typeof(s)}[]
function b()
    cref = MOI.ConstraintIndex{typeof(f), typeof(s)}(1)
    push!(cons, (cref, f, s))
end
function c()
[...]

PR #567, opened 2018-11-12
Use copy instead of deepcopy in MOIU model

blegat:
In the benchmark of https://github.com/JuliaOpt/MathOptInterface.jl/issues/566#issue-379603471:

Before this PR:
```julia
julia> @btime M.add_bin_constraints(1_000_000)
267.887 ms (2000105 allocations: 809.08 MiB)
```

After this PR:
```julia
julia> @btime M.add_bin_constraints(1_000_000)
80.247 ms (105 allocations: 68.14 MiB)
```

Closes https://github.com/JuliaOpt/MathOptInterface.jl/issues/566


src/Utilities/model.jl
```diff
@@ -270,10 +270,11 @@ function MOI.add_constraint(model::AbstractModel, f::F, s::S) where {F<:MOI.Abst
     if MOI.supports_constraint(model, F, S)
         # We give the index value `nextconstraintid + 1` to the new constraint.
         # As the same counter is used for all pairs of F-in-S constraints,
-        # the index value is unique across all constraint types as mentionned in `@model`'s doc.
+        # the index value is unique across all constraint types as mentionned in
```
mlubin: Typo: mentioned

PR #569, opened 2018-11-12
Document storage of constraints

blegat:
This is a first step to help readability of this file.


src/Utilities/model.jl
```diff
@@ -1,3 +1,19 @@
+## Storage of constraints
+#
+# All `F`-in-`S` constraints are stored in a vector of `C{F, S}`. The index in
```
mlubin: `C` isn't a very readable name, any possible improvements?

PR #577, opened 2018-11-18
Add `OptimizeNotCalled` termination status with tests and docs

### LINKED ISSUE #571: Caching optimizer should provide status
CachingOptimizer errors when asked for a `TerminationStatus` (and `PrimalStatus`, etc.) when `.optimizer` field is `nothing`. 

See https://github.com/JuliaOpt/JuMP.jl/issues/1618

### LINKED ISSUE #572: "NotSolved" TerminationStatus (or equivalent)
As raised by https://github.com/JuliaOpt/JuMP.jl/issues/1618, what should optimizers return as `TerminationStatus` when the problem hasn't been solved yet?

```julia
model = Model()
JuMP.termination_status(model)
```

Here is the list of current `TerminationStatus`:
https://github.com/JuliaOpt/MathOptInterface.jl/blob/81c115594f6cc0dae3d288d69ffa428eadc0f5b7/src/attributes.jl#L722

odow:
Also fixes default statuses in CachingOptimizer.

Closes #571 
Closes #572

### DIFF
--- a/test/Utilities/mockoptimizer.jl
+++ b/test/Utilities/mockoptimizer.jl
@@ -1,5 +1,6 @@
 @testset "Default statuses" begin
     model = MOIU.MockOptimizer(ModelForMock{Float64}())
+    MOIT.default_status_test(model)
     MOI.empty!(model)
     MOIT.default_status_test(model)
 end
--- a/test/Utilities/mockoptimizer.jl
+++ b/test/Utilities/mockoptimizer.jl
@@ -1,3 +1,9 @@
+@testset "Default statuses" begin
+    model = MOIU.MockOptimizer(ModelForMock{Float64}())
+    MOI.empty!(model)
+    MOIT.default_status_test(model)
+end
+
 @testset "Mock optimizer name test" begin
     MOIT.nametest(MOIU.MockOptimizer(ModelForMock{Float64}()))
 end
--- a/src/Test/modellike.jl
+++ b/src/Test/modellike.jl
@@ -3,6 +3,12 @@
 struct UnknownScalarSet <: MOI.AbstractScalarSet end
 struct UnknownVectorSet <: MOI.AbstractVectorSet end
 
+function default_status_test(model::MOI.ModelLike)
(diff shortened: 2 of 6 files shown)


test/Utilities/mockoptimizer.jl
```diff
@@ -1,3 +1,9 @@
+@testset "Default statuses" begin
+    model = MOIU.MockOptimizer(ModelForMock{Float64}())
+    MOI.empty!(model)
+    MOIT.default_status_test(model)
```
blegat: I would also add `MOIT.default_status_test(model)` before `MOI.empty!(model)`

odow:
From the caching optimizer perspective, previous errors now either still error with a nice error message, or provide a sane default instead of erroring.

For solvers this probably is breaking, since many probably return `UnknownResultStatus` when they mean `NoSolution`. And for `TerminationStatus`, ¯\\\_(ツ)\_/¯

mlubin:
> For solvers this probably is breaking, since many probably return UnknownResultStatus when they mean NoSolution. And for TerminationStatus, ¯\_(ツ)_/¯

So will we have reverse dependency issues when we tag this as MOI 0.6.4?

odow:
We shouldn't have any reverse dependency issues. The solvers will need to opt-in to `default_status_test`. 

The current solver implementations are arguably not correct at present. For example, GLPK [returns `UnknownResultStatus`](https://github.com/JuliaOpt/GLPK.jl/blob/4b21e26e81eef5f1d95db3f015c17999a30e1bcf/src/MOIWrapper.jl#L607) when it should return `NoSolution`, likely because it was written prior to the introduction of `NoSolution` and we didn't have these tests. 

This causes a problem for JuMP and `has_values` which [only checks for `!=NoSolution`](https://github.com/JuliaOpt/JuMP.jl/blob/f8d8687b24c58f48f57bf9bed27c5f36f355fafd/src/variables.jl#L556).

blegat:
The breaking part will be https://github.com/JuliaOpt/MathOptInterface.jl/issues/583

PR #578, opened 2018-11-18
Add guideline for naming solvers.

### LINKED ISSUE #425: capitalization of solver names
**First** (and most importantly): let's use consistent capitalization _within_ solver packages. CPLEX.jl does not - I always forget whether to write
```
using CPLEX
solver = CplexSolver()
```
or 
```
using Cplex
solver = CPLEXSolver()
```
and have to go look it up. Knitro/KNITRO is also inconsistent. Others too probably but I didn't check them all. In "MathOptInterfaceCPLEX.jl", which seems dead at the moment, "CPLEXOptimizer" is used - that fixes the internal consistency issue there at least.

**Second** (and less importantly): let's have consistency _across_ different solver packages. On the MOSEK website, it seems the name is officially "MOSEK", yet Mosek.jl uses a consistent "Mosek". For CBC (Coin Branch & Cut), Cbc.jl uses a consistent "Cbc". Similarly for CLP. So maybe we should use "Cplex", "Ecos", "Scs", "Csdp", "Knitro" etc. Why should Cbc and Clp and Mosek be not in all caps when these other packages are? So it would be nice to stick to a convention. I think I have prefer not all caps - so keep "Cbc" and "Clp" and "Mosek" and change "CPLEX" etc.

odow:
Closes #425

### DIFF
--- a/docs/src/apimanual.md
+++ b/docs/src/apimanual.md
@@ -936,12 +936,22 @@ users to avoid extra copies in this case.
 
 Solver wrappers should document how the low-level solver statuses map to the MOI statuses. In particular, the characterization of a result with status `FeasiblePoint` and termination status `Success` is entirely solver defined. It may or may not be a globally optimal solution. Solver wrappers are not responsible for verifying the feasibility of results. Statuses like `NearlyFeasiblePoint`, `InfeasiblePoint`, `NearlyInfeasiblePoint`, and `NearlyReductionCertificate` are designed to be used when the solver explicitly indicates as much.
 
-### Package Naming
+### Naming
 
-MOI solver interfaces may be in the same package as the solver itself (either the C wrapper if the solver is accessible through C, or the Julia code if the solver is written in Julia, for example).
-The guideline for naming the file containing the MOI wrapper is `src/MOIWrapper.jl` and `test/MOIWrapper.jl` for the tests.
-If the MOI wrapper implementation is spread in several files, they should be stored in a `src/MOIWrapper` folder and included by a `src/MOIWrapper/MOIWrapper.jl` file.
-In some cases it may be more appropriate to host the MOI wrapper in its own package; in this case it is recommended that the MOI wrapper package be named `MathOptInterfaceXXX` where `XXX` is the solver name.
+MOI solver interfaces may be in the same package as the solver itself (either
+the C wrapper if the solver is accessible through C, or the Julia code if the
+solver is written in Julia, for example). The guideline for naming the file
+containing the MOI wrapper is `src/MOIWrapper.jl` and `test/MOIWrapper.jl` for
+the tests. If the MOI wrapper implementation is spread in several files, they
+should be stored in a `src/MOIWrapper` folder and included by a
+`src/MOIWrapper/MOIWrapper.jl` file. In some cases it may be more appropriate to
+host the MOI wrapper in its own package; in this case it is recommended that the
+MOI wrapper package be named `MathOptInterfaceXXX` where `XXX` is the solver
+name.
+
+By convention, optimizers should not be exported and should be named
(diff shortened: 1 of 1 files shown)


docs/src/apimanual.md
```diff
@@ -936,12 +936,22 @@ users to avoid extra copies in this case.
...
-MOI solver interfaces may be in the same package as the solver itself (either the C wrapper if the solver is accessible through C, or the Julia code if the solver is written in Julia, for example).
-The guideline for naming the file containing the MOI wrapper is `src/MOIWrapper.jl` and `test/MOIWrapper.jl` for the tests.
-If the MOI wrapper implementation is spread in several files, they should be stored in a `src/MOIWrapper` folder and included by a `src/MOIWrapper/MOIWrapper.jl` file.
-In some cases it may be more appropriate to host the MOI wrapper in its own package; in this case it is recommended that the MOI wrapper package be named `MathOptInterfaceXXX` where `XXX` is the solver name.
+### Naming
+
+MOI solver interfaces may be in the same package as the solver itself (either
+the C wrapper if the solver is accessible through C, or the Julia code if the
+solver is written in Julia, for example). The guideline for naming the file
+containing the MOI wrapper is `src/MOIWrapper.jl` and `test/MOIWrapper.jl` for
```
mlubin: Shouldn't this be `MOI_wrapper.jl` or `moi_wrapper.jl` according to http://www.juliaopt.org/JuMP.jl/dev/style/#Naming-1?

odow: Good point. I changed to `MOI_wrapper.jl` since `moi_wrapper.jl` looked weird, and `MOI` is consistent with how we use it in all other contexts.

docs/src/apimanual.md
```diff
@@ -936,12 +936,22 @@ users to avoid extra copies in this case.
...
+containing the MOI wrapper is `src/MOI_wrapper.jl` and `test/MOI_wrapper.jl` for
+the tests. If the MOI wrapper implementation is spread in several files, they
+should be stored in a `src/MOI_wrapper` folder and included by a
+`src/MOI_wrapper/MOI_wrapper.jl` file. In some cases it may be more appropriate
+to host the MOI wrapper in its own package; in this case it is recommended that
+the MOI wrapper package be named `MathOptInterfaceXXX` where `XXX` is the solver
+name.
+
+By convention, optimizers should not be exported and should be named
+`PackageName.Optimizer`. For example, `Cplex.Optimizer`, `Gurobi.Optimizer`, and
```
mlubin: Nit: It's CPLEX.jl so `CPLEX.optimizer`

blegat: `CPLEX.Optimizer` with capital `O`

docs/src/apimanual.md
blegat: If it was a submodule, it would be names `MOIWrapper` because it should be capitalized right ?

blegat:
@odow @mlubin I am not sure about the change `MOIWrapper` -> `MOI_wrapper`. If one package implement it as a submodule, the module should be named `MOIWrapper` so the file should be named `MOIWrapper.jl`.

mlubin:
Sorry, I missed this discussion when I decided to merge. If there is *no* submodule then we should follow the underscore style. We don't really have well-developed style guidelines for submodules but yes that's a good point. Does the submodule name need to be `MOIWrapper`? That's the only case where `MOIWrapper.jl` would make sense.

blegat:
Yes but it would be nice if packages using a submodule have the same names that packages that does not

mlubin:
I think following standard style guidelines for naming files based on their content is more important. The effect of "other packages have files with this name" is too non-local and less obvious. IMO it doesn't merit an exception.

PR #580, opened 2018-11-18
Document solver-specific attributes.

### LINKED ISSUE #139: Solver-dependent options
Should we have a way to deal with solver specific attributes?
like:
`setattribute!(m::AbstractInstance, ::SolverSpecificAttribute, parameter::Any, value::Any)`

So that the parameter could be a String, Symbol, Integer, whatever the solver interface understands.

would be some kind of MOI shortcut for the `setparam!` and friends...

This way even jump would have a "standard" way to have the user passing parameters (outside `Model(param = val)`). So that the user does not have to use `solverX.setparam!(jump.moi.inner, param, val)`

odow:
Closes #139

### DIFF
--- a/docs/src/apimanual.md
+++ b/docs/src/apimanual.md
@@ -852,6 +852,7 @@ Then, the user can write:
 model = MyPackage.Optimizer()
 MOI.set(model, MyPackage.PrintLevel(), 0)
 ```
+
 ### Implementing copy
 
 Avoid storing extra copies of the problem when possible. This means that solver wrappers should not use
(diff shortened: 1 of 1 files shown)


docs/src/apimanual.md
```diff
@@ -835,6 +835,23 @@ will additionally support `ScalarAffineFunction`-in-`Interval`.
...
+struct PrintLevel <: MOI.AbstractSolverAttribute end
+function MOI.set(model::Optimizer, ::PrintLevel, level::Int)
+    # ... set the print level ...
+end
+```
+Then, the user can write:
+```julia
+model = MyPackage.Optimizer()
+MOI.set(model, MyPackage.PrintLevel(), 0)
+```
```
mlubin: Nit: Should there be a newline before the new section?

PR #582, opened 2018-11-19
Fix error with external functions and sets.

### LINKED ISSUE #555: External set bridging broken on v0.6.2

odow:
Closes #555
Closes #557 

- [x] still need to test bridges

### DIFF
--- a/test/Utilities/model.jl
+++ b/test/Utilities/model.jl
@@ -2,8 +2,7 @@
 
 # We need to test this in a module at the top level because it can't be defined
 # in a testset. If it runs without error, then we're okay.
-module TestExternal
-    using Base: copy
+module TestExternalModel
     using MathOptInterface
     struct NewSet <: MathOptInterface.AbstractScalarSet end
     struct NewFunction <: MathOptInterface.AbstractScalarFunction end
@@ -20,16 +19,17 @@ module TestExternal
         ()
     )
 end
-@testset "External functions and sets" begin
-    model = TestExternal.ExternalModel{Float64}()
+
+@testset "External @model" begin
+    model = TestExternalModel.ExternalModel{Float64}()
     c = MOI.add_constraint(
-        model, TestExternal.NewFunction(), TestExternal.NewSet())
-    @test typeof(c) ==
-        MOI.ConstraintIndex{TestExternal.NewFunction, TestExternal.NewSet}
+        model, TestExternalModel.NewFunction(), TestExternalModel.NewSet())
+    @test typeof(c) == MOI.ConstraintIndex{TestExternalModel.NewFunction,
+        TestExternalModel.NewSet}
     c2 = MOI.add_constraint(
-        model, TestExternal.NewFunction(), MOI.ZeroOne())
+        model, TestExternalModel.NewFunction(), MOI.ZeroOne())
     @test typeof(c2) ==
-        MOI.ConstraintIndex{TestExternal.NewFunction, MOI.ZeroOne}
+        MOI.ConstraintIndex{TestExternalModel.NewFunction, MOI.ZeroOne}
 end
 
 @testset "Name test" begin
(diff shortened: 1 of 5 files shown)


test/Utilities/model.jl
```diff
@@ -1,5 +1,37 @@
 # TODO: It's hard to find where Model is defined!
 
+# We need to test this in a module at the top level because it can't be defined
+# in a testset. If it runs without error, then we're okay.
+module TestExternal
+    using Base: copy
```
blegat: this line shouldn't be needed

test/bridge.jl
```diff
@@ -397,3 +397,57 @@ end
...
     end
 end
+
+# We need to test this in a module at the top level because it can't be defined
+# in a testset. If it runs without error, then we're okay.
+module TestExternalBridge
+    using MathOptInterface
+
+    struct StrictlyGreaterThan <: MathOptInterface.AbstractScalarSet end
+    struct StrictlyGreaterBridge{T} <: MathOptInterface.Bridges.AbstractBridge end
```
odow: @blegat do bridges have to have a type parameter `T`? This cause me no end of problems. The only error you get is the unhelpful `Unionall` one.

test/bridge.jl
```diff
@@ -397,3 +397,57 @@ end
...
+        ()
+    )
+end
+
+@testset "@bridge with external components" begin
+    model = SimpleModel{Float64}();
+    @test MOI.supports_constraint(model, MOI.SingleVariable, MOI.GreaterThan{Float64})
+    @test !MOI.supports_constraint(model, MOI.SingleVariable, TestExternalBridge.StrictlyGreaterThan)
+
+    bridge = TestExternalBridge.StrictlyGreater{Float64}(model);
```
odow: Follow-up from the above comment. Is it related to the fact that this defines a model using `@model`?

test/bridge.jl
blegat: Yes, it has to, see https://github.com/JuliaOpt/MathOptInterface.jl/blob/b0477a680fc7ce274af60b0004fc4653cc03bee3/src/Bridges/singlebridgeoptimizer.jl#L46

test/bridge.jl
blegat: Yes, this is the same type as the MOIU `@model`. This design may be questionable but with light bridges there won't be an internal model anymore

PR #587, opened 2018-11-20
Add allow_modify_function to add_constraint

blegat:
When this keyword argument is set, the model does not have to copy the function in argument if it want to store it or modify it. It is similar to
https://github.com/JuliaOpt/MathOptInterface.jl/blob/6dcee0bda0c870380576ddd56194dbd21efe3798/src/Utilities/constraints.jl#L5
For instance, in `JuMP.add_constraint`, an MOI function is created, sent to the model which copies it but then JuMP discards the original function. The function is therefore copied for nothing!
`MOIU.AbstractModel` implements `add_constraint` with this keyword argument but the rest of `MOIU` avoids using it at places where the solver is used to avoid breakage.
I have left TODO for these spots so that we can change it before we release MOI v0.7.

Here is a benchmark which shows the speedup.
```julia
function time_add(model, f, s, c)
    @btime MOI.add_constraint($model, $f, $s, allow_modify_function=$c)
end
function bench()
    model = Model()
    x = @variable(model)
    y = @variable(model)
    f = JuMP.moi_function(x + y)
    s = MOI.EqualTo(0.0)
    time_add(JuMP.backend(model), f, s, true)
    time_add(JuMP.backend(model), f, s, false)
    return
end
bench()

# output
  67.566 ns (3 allocations: 64 bytes)
  159.491 ns (5 allocations: 208 bytes)
```

### DIFF
--- a/docs/src/apimanual.md
+++ b/docs/src/apimanual.md
@@ -839,10 +839,10 @@ will additionally support `ScalarAffineFunction`-in-`Interval`.
 
 Solver-specific attributes should either be passed to the optimizer on creation,
 e.g., `MyPackage.Optimizer(PrintLevel = 0)`, or through a sub-type of
-[`AbstractSolverAttribute`](@ref). For example, inside `MyPackage`, we could add
+[`AbstractOptimizerAttribute`](@ref). For example, inside `MyPackage`, we could add
 the following:
 ```julia
-struct PrintLevel <: MOI.AbstractSolverAttribute end
+struct PrintLevel <: MOI.AbstractOptimizerAttribute end
 function MOI.set(model::Optimizer, ::PrintLevel, level::Int)
     # ... set the print level ...
 end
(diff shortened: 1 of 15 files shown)


blegat:
We might need to turn it into a positional argument for performance reason.

mlubin:
It's unfortunate that Julia doesn't have a way to talk about transferring ownership of an array. A safer approach that still doesn't seem possible would be capture the memory of the user-provided array and leave the user with an empty array after the call (see https://discourse.julialang.org/t/swap-array-contents/7774).

> We might need to turn it into a positional argument for performance reason.

If we're going to do that, what about
```julia
MOI.add_constraint(model, MayModify(f), s)
```
instead of
```julia
MOI.add_constraint(model, f, s, true) # what does this mean?
```

joaquimg:
What if we have a Model Attribute AllowModifyFunction, that is false by default and the user is allowed to turn on?

odow:
Closing as out-dated. We can come back to this post MOI 1.0.

PR #595, opened 2018-12-08
✅ Test Status before optimize!

blegat:
Closes https://github.com/JuliaOpt/MathOptInterface.jl/issues/583

### DIFF
--- a/src/Test/contconic.jl
+++ b/src/Test/contconic.jl
@@ -46,6 +46,10 @@ function _lin1test(model::MOI.ModelLike, config::TestConfig, vecofvars::Bool)
     MOI.set(model, MOI.ObjectiveSense(), MOI.MinSense)
 
     if config.solve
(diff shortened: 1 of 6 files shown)


src/Test/contconic.jl
```diff
@@ -45,6 +45,10 @@ function _lin1test(model::MOI.ModelLike, config::TestConfig, vecofvars::Bool)
     MOI.set(model, MOI.ObjectiveSense(), MOI.MinSense)
 
     if config.solve
+        @test MOI.get(model, MOI.TerminationStatus()) == MOI.OptimizeNotCalled
+        @test MOI.get(model, MOI.PrimalStatus()) == MOI.NoSolution
```
mlubin: I'm a bit confused now. Do solvers need to return a value for `PrimalStatus()` even if `ResultCount()` is zero?
Do solvers need to return a value for `PrimalStatus(N)` if `ResultCount() < N`?

PR #596, opened 2018-12-08
redesign TerminationStatus

mlubin:
This is a breaking change to `TerminationStatus` following the discussion in https://github.com/JuliaOpt/JuMP.jl/issues/1656.

The main change is that `TerminationStatus` now distinguishes between optimal, locally optimal, infeasible, locally infeasible, and dual infeasible, addressing a usability issue in JuMP where users had to query both termination status and result statuses in order to interpret what happened in very basic cases. We had originally planned to abstract this away at the JuMP level, but this was hard to do because it requires additional knowledge about the solver than what's available through MOI.

I've made only the documentation changes. If we agree on this I'll get tests passing. If you can think of additional interesting corner cases, I'll add them to the tables in the docs (screenshots below). The tables seem useful for justifying why we need to put so much effort into reporting statuses correctly.

image

image

image

@ccoffrin @chkwon @tkoolen @ulfworsoe @frapac

### DIFF
--- a/docs/src/apimanual.md
+++ b/docs/src/apimanual.md
@@ -414,8 +414,9 @@ and certificates.
 
 ### Common status situations
 
-The sections below describe how to interpret typical status cases for three
-common classes of solvers. Solver wrappers may provide additional information on
+The sections below describe how to interpret typical or interesting status cases
+for three common classes of solvers. The example cases are illustrative, not
+comprehensive. Solver wrappers may provide additional information on
 how the solver's statuses map to MOI statuses.
 
 `?` in the tables indicate that multiple different values are possible.
--- a/docs/src/apimanual.md
+++ b/docs/src/apimanual.md
@@ -414,7 +414,7 @@ and certificates.
 
 ### Common status situations
 
-The sections below describe how to interpret different status cases for three
+The sections below describe how to interpret typical status cases for three
 common classes of solvers. Solver wrappers may provide additional information on
 how the solver's statuses map to MOI statuses.
 
@@ -424,12 +424,13 @@ how the solver's statuses map to MOI statuses.
 
 Linear programming and conic optimization solvers fall into this category.
 
-| What happended?                   | `TerminationStatus()` | `ResultCount()` | `PrimalStatus()`                         | `DualStatus()`                           |
-| --------------------------------- | --------------------- | --------------- | ---------------------------------------- | ---------------------------------------- |
-| Proved optimality                 | `Optimal`             | 1               | `FeasiblePoint`                          | `FeasiblePoint`                          |
-| Proved infeasible                 | `Infeasible`          | 1               | `NoSolution`                             | `InfeasibilityCertificate`               |
-| Optimal within relaxed tolerances | `AlmostOptimal`       | 1               | `FeasiblePoint` or `AlmostFeasiblePoint` | `FeasiblePoint` or `AlmostFeasiblePoint` |
-| Stall                             | `SlowProgress`        | 1               | ?                                        | ?                                        |
+| What happended?                         | `TerminationStatus()` | `ResultCount()` | `PrimalStatus()`                         | `DualStatus()`                           |
+| --------------------------------------- | --------------------- | --------------- | ---------------------------------------- | ---------------------------------------- |
+| Proved optimality                       | `Optimal`             | 1               | `FeasiblePoint`                          | `FeasiblePoint`                          |
+| Proved infeasible                       | `Infeasible`          | 1               | `NoSolution`                             | `InfeasibilityCertificate`               |
+| Optimal within relaxed tolerances       | `AlmostOptimal`       | 1               | `FeasiblePoint` or `AlmostFeasiblePoint` | `FeasiblePoint` or `AlmostFeasiblePoint` |
+| Detected an unbounded ray of the primal | `DualInfeasible`      | 1               | `InfeasibilityCertificate`               | `NoSolution`                             |
+| Stall                                   | `SlowProgress`        | 1               | ?                                        | ?                                        |
 
 #### Global branch-and-bound solvers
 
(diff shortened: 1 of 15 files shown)


tkoolen:
Thanks for the heads up!

docs/src/apimanual.md
```diff
@@ -410,31 +412,56 @@ See also the attributes [`ConstraintPrimal`](@ref), and
...
+Linear programming and conic optimization solvers fall into this category.
 
-#### Global mixed-integer or nonconvex solver
+| What happended?                   | `TerminationStatus()` | `ResultCount()` | `PrimalStatus()`                         | `DualStatus()`                           |
+| --------------------------------- | --------------------- | --------------- | ---------------------------------------- | ---------------------------------------- |
+| Proved optimality                 | `Optimal`             | 1               | `FeasiblePoint`                          | `FeasiblePoint`                          |
+| Proved infeasible                 | `Infeasible`          | 1               | `NoSolution`                             | `InfeasibilityCertificate`               |
+| Optimal within relaxed tolerances | `AlmostOptimal`       | 1               | `FeasiblePoint` or `AlmostFeasiblePoint` | `FeasiblePoint` or `AlmostFeasiblePoint` |
+| Stall                             | `SlowProgress`        | 1               | ?                                        | ?                                        |
```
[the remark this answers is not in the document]
mlubin: Done

docs/src/apimanual.md
```diff
@@ -410,31 +412,56 @@ See also the attributes [`ConstraintPrimal`](@ref), and
 See [Duals](@ref) for a discussion of the MOI conventions for primal-dual pairs
 and certificates.
 
-
-
 ### Common status situations
 
-The sections below describe how to interpret different status cases for three common classes of solvers. Most importantly, it is essential to know if a solver is expected to provide a global or only locally optimal solution when interpreting the result statuses. Solver wrappers may provide additional information on how the solver's statuses map to MOI statuses.
+The sections below describe how to interpret different status cases for three
```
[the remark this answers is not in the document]
mlubin: Done

ccoffrin:
This is sounding good to me.  Just want to confirm that the general philosophy is to move away from the abstract `success` terminology in favor of encoding a little bit of the solution information into the termination status, correct?

I also recomend adding unbounded to the primal-dual example table.

Questions:
- In the primal-dual example with infeasible, why is result count 1?
- In the case when the solver hits a limit, what is the expected workflow for the user to check if any feasible solution was found or not?
- Is there an easy way for the user to check the broad categories of the termination status?  For example, checking the categories Ok, Limits and Problematic that mentioned in the docs?

A few cases to test the current proposal,
- NL-OA with NLP constraints (e.g. Pavito) converges with a feasible solution, what termination status does it report?  With the current NLP interface, I presume Pavito cannot assume NLP constraint convexity.
- NL-BB with a non-convex NLP constraints (e.g. Juniper) finishes it's tree search.  In the case of finding a feasible solution and not finding a feasible solution, what termination statuses does it report? 
- A solver implements some ad-hoc greedy heuristic without any quality guarantees. In the case of finding a feasible solution and not finding a feasible solution, what termination statuses does it report?

joaquimg:
> In the primal-dual example with infeasible, why is result count 1?

It is 1 if the solver has a Infeasibility Certificate, 0 otherwise.

> In the case when the solver hits a limit, what is the expected workflow for the user to check if any feasible solution was found or not?

Check `PrimalStatus()`.

> Is there an easy way for the user to check the broad categories of the termination status? For example, checking the categories Ok, Limits and Problematic that mentioned in the docs?

It would be nice to have functions for this, like:
```julia
limit_termination(status) = status in [TimeLimit, IterationLimit, NormLimit, ...]
```
Same for problematic (`bad_termination`) and Ok (`successful_termination`)

leethargo:
Are the above tables in the screenshot comprehensive, or do they only cover the `OK` cases from the previous `TerminationStatus`? I wonder why there are no results with `ResultCount == 0`, e.g., for linear programming solvers with a timelimit and no warmstart.

mlubin:
> Are the above tables in the screenshot comprehensive

No, not at all. Timing out without a solution is covered in the global solvers table (because it happens more typically with MIP solvers), but it also applies to any solver. [CPLEX has 133 statuses](https://www.ibm.com/support/knowledgecenter/SSSA5P_12.6.3/ilog.odms.cplex.help/refcallablelibrary/macros/Solution_status_codes.html) and we won't be listing all of their translations in the MOI docs.

ccoffrin:
Sounds good to me.  Agreed that `LocallySolved` is more inclusive for local search algorithms.

chkwon:
Everything looks reasonable to me. Thanks for the changes.

blegat:
The new statuses look good to me.

mlubin:
It was easier than I thought to get tests passing. Maybe I missed something.

On thing we'll have to address is testing local solvers on the tests where we expect `Optimal` not `LocallySolved`. Probably needs another field in the test config.

docs/src/apimanual.md
```diff
@@ -410,31 +406,67 @@ See also the attributes [`ConstraintPrimal`](@ref), and
...
+| What happened?                                   | `TerminationStatus()`   | `ResultCount()` | `PrimalStatus()`  | `DualStatus()` |
+| ------------------------------------------------ | ----------------------- | --------------- | ----------------- | -------------- |
+| Proved optimality                                | `Optimal`               | 1               | `FeasiblePoint`   | `NoSolution`   |
+| Presolve detected infeasibility or unboundedness | `InfeasibleOrUnbounded` | 0               | `NoSolution`      | `NoSolution`   |
+| Proved infeasibility                             | `Infeasible`            | 0               | `NoSolution`      | `NoSolution`   |
+| Timed out (no solution)                          | `TimeLimit`             | 0               | `NoSolution`      | `NoSolution`   |
+| Timed out (with a solution)                      | `TimeLimit`             | 1               | `FeasiblePoint`   | `NoSolution`   |
+| `CPXMIP_OPTIMAL_INFEAS`                          | `AlmostOptimal`         | 1               | `InfeasiblePoint` | `NoSolution`   |
```
odow: I can see someone reading this and thinking, didn't it find an `AlmostFeasiblePoint`? The answer is "it depends" but it probably goes nicely with `AlmostOptimal`.

mlubin: I almost put `Optimal` | `InfeasiblePoint` for this case, because that's what the solver actually says. I stepped back because you could make an argument that it would be easy to see `Optimal` and forget to check the `PrimalStatus` and miss the issue. `AlmostOptimal` and `AlmostFeasiblePoint` aren't strictly correct because the solver never reported that relaxed tolerances are satisfied. Not sure...

odow: I mean, the [docs say](https://www.ibm.com/support/knowledgecenter/SSSA5P_12.6.3/ilog.odms.cplex.help/refcallablelibrary/macros/CPXMIP_OPTIMAL_INFEAS.html) "Problem is optimal with unscaled infeasibilities." So I think either is correct. Let's just leave as is for now, and revisit it if we need to in the future.

PR #599, opened 2018-12-12
Add config.optimal_status

### LINKED ISSUE #597: Handling local sovlers in tests
After #596, we'll need a few small changes to make it possible for local solvers to pass the tests that check for `Optimal` or `Infeasible` status.

https://github.com/JuliaOpt/MathOptInterface.jl/pull/596#issuecomment-445580360

mlubin:
So that local solvers can pass tests.

Closes #597.


blegat:
We should also change this line:
https://github.com/JuliaOpt/MathOptInterface.jl/blob/013565e7216a4cd23f9c32961b8d50311cb79133/src/Test/UnitTests/unit_tests.jl#L62

PR #610, opened 2018-12-28
WIP: Slack bridges

joaquimg:
- Still not sure about how to test the bridges. Can I create a model with `@model` that supports `ScalarAffineFunction`-in-`EqualTo` but not -in-`LessThan`?

- `operate` does not work on vector functions for addition, what should I do?


blegat:
> Still not sure about how to test the bridges. Can I create a model with `@model` that supports `ScalarAffineFunction`-in-`EqualTo` but not -in-`LessThan`?

You can run it through some `contlinear` tests. The underlying model of the mockoptimizer may support the constraint that is bridged that is not an issue as the `SingleBridgeOptimizer` bridges all the constraints it can bridge regardless of whether it is supported or not (which is appropriate for testing), as opposed to `LazyBridgeOptimizer` which is used in JuMP.

> `operate` does not work on vector functions for addition, what should I do?

The missing `operate` methods should be added

src/Bridges/slackbridge.jl
```diff
@@ -0,0 +1,145 @@
...
+end
+function ScalarSlackBridge{T, F, S}(model, f::F, s::S) where {T, F<:NonSingleVariable{T}, S<:ScalarSlackSet{T}}
+    slack = MOI.add_variable(model)
+    new_f = MOIU.operate(-, T, f, MOI.SingleVariable(slack))
+    slack_in_set = MOI.add_constraint(model, MOI.SingleVariable(slack), s)
+    equality = MOI.add_constraint(model, new_f, MOI.EqualTo(0.0))
+    return ScalarSlackBridge{T, F, S}(slack, slack_in_set, equality)
+end
+
+MOI.supports_constraint(::Type{ScalarSlackBridge{T, F, S}}, ::Type{<:NonSingleVariable{T}}, ::Type{<:ScalarSlackSet{T}}) where {T, F, S} = true
```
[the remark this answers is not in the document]
joaquimg: agreed

src/Bridges/slackbridge.jl
```diff
@@ -0,0 +1,145 @@
...
+const ScalarSlackSet{T} = Union{MOI.Interval{T}, MOI.GreaterThan{T}, MOI.LessThan{T}}
+const NonSingleVariable{T} = Union{MOI.ScalarAffineFunction{T}, MOI.ScalarQuadraticFunction{T}}
+
+"""
+    ScalarSlackBridge{T, F, S}
+
+The `ScalarSlackBridge` converts a constraint `F`-in-`S` where `F` is a function different
+ from `SingleVariable` into the constraints ``F in EqualTo{T}`` and `SingleVariable`-in-`S`.
+"""
+struct ScalarSlackBridge{T, F<:NonSingleVariable{T}, S<:ScalarSlackSet{T}} <: AbstractBridge
```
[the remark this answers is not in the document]
joaquimg: Except for `EqualTo` but I will use the above strategy

src/Utilities/functions.jl
```diff
@@ -599,27 +605,51 @@ function operate_term(::typeof(*), t1::MOI.ScalarAffineTerm,
...
 
 # Avoid a copy in the case of +
 function operate_terms(::typeof(+),
                        terms::Vector{<:Union{MOI.ScalarAffineTerm,
                                              MOI.ScalarQuadraticTerm}})
     return terms
 end
+function operate_terms(::typeof(+),
+        terms::Vector{<:Union{MOI.VectorAffineTerm,
+                            MOI.VectorQuadraticTerm}})
```
blegat: Missing two space in indenting here

joaquimg:
PS there is a bunch of indentation mistakes.
VSCode is very annoying, I wanted to have an idea if this is in the right direction and then I will fix style :)

joaquimg:
It should be almost finished, I might need to fix style and add operate tests

src/Bridges/Bridges.jl
```diff
@@ -59,6 +61,26 @@ end
 
 include("vectorizebridge.jl")
 @bridge Vectorize VectorizeBridge () (MOI.EqualTo, MOI.LessThan, MOI.GreaterThan,) () () (MOI.SingleVariable,) (MOI.ScalarAffineFunction, MOI.ScalarQuadraticFunction) () ()
+include("slackbridge.jl")
+@bridge ScalarSlack ScalarSlackBridge () (MOI.Interval, MOI.LessThan, MOI.GreaterThan) () () () (MOI.ScalarAffineFunction, MOI.ScalarQuadraticFunction) () ()
+@bridge(VectorSlack, VectorSlackBridge, 
+(),
```
blegat: Indent and merge some lines if it fits 80 chars. Having to list all sets is not ideal but the need will be removed by light bridges

src/Bridges/slackbridge.jl
```diff
@@ -0,0 +1,162 @@
...
+                        ::Type{<:MOI.SingleVariable},
+                        ::Type{<:MOI.EqualTo}) where {T, F, S} = false
+MOI.supports_constraint(::Type{ScalarSlackBridge{T, F, S}},
+                        ::Type{<:MOI.SingleVariable},
+                        ::Type{<:MOI.AbstractScalarSet}) where {T, F, S} = false
+MOI.supports_constraint(::Type{ScalarSlackBridge{T, F, S}},
+                        ::Type{<:MOI.AbstractScalarFunction},
+                        ::Type{<:MOI.EqualTo}) where {T, F, S} = false
+function added_constraint_types(::Type{ScalarSlackBridge{T, F, S}}) where {T, F, S}
+    return [(F, MOI.EqualTo{T}), (MOI.SingleVariable, S)]
```
blegat: F should be `promote_operation(-, T F, SingleVariable)`

src/Bridges/slackbridge.jl
```diff
@@ -0,0 +1,162 @@
...
+end
+function concrete_bridge_type(::Type{<:ScalarSlackBridge{T}},
+                              F::Type{<:MOI.AbstractScalarFunction},
+                              S::Type{<:MOI.AbstractScalarSet}) where T
+    return ScalarSlackBridge{T, F, S}
```
blegat: Same here, `F` should be `promote_operation(-, T F, SingleVariable)`

src/Bridges/slackbridge.jl
```diff
@@ -0,0 +1,162 @@
...
+    return
+end
+
+# Attributes, Bridge acting as a constraint
+function MOI.get(model::MOI.ModelLike, attr::MOI.ConstraintPrimal, c::VectorSlackBridge)
+    # due to equality, slacks should have the same value as original affine function
+    return MOI.get(model, attr, c.slacks_in_set)
+end
+function MOI.get(model::MOI.ModelLike, a::MOI.ConstraintDual, c::VectorSlackBridge)
+    error("VectorSlackBridge is not returning duals for now")
```
blegat: The dual constraint on `slack` (since it is free) is `-dual_slack_in_set  + dual_equality = 0` so the two duals are equal and we can return either one of them. I would return the dual of `slack_in_set`

src/Bridges/slackbridge.jl
```diff
@@ -0,0 +1,162 @@
...
+end
+
+# Constraints
+function MOI.modify(model::MOI.ModelLike, c::VectorSlackBridge, change::MOI.AbstractFunctionModification)
+    MOI.modify(model, c.equality, change)
+end
+
+function MOI.set(model::MOI.ModelLike, ::MOI.ConstraintFunction,
+                  c::VectorSlackBridge{T, F, S}, func::F) where {T, F, S}
```
blegat: one space of indent that should be removed

src/Bridges/slackbridge.jl
```diff
@@ -0,0 +1,162 @@
...
+                        ::Type{<:MOI.VectorOfVariables},
+                        ::Type{<:MOI.Zeros}) where {T, F, S} = false
+MOI.supports_constraint(::Type{VectorSlackBridge{T, F, S}},
+                        ::Type{<:MOI.AbstractVectorFunction},
+                        ::Type{<:MOI.Zeros}) where {T, F, S} = false
+MOI.supports_constraint(::Type{VectorSlackBridge{T, F, S}},
+                        ::Type{<:MOI.VectorOfVariables},
+                        ::Type{<:MOI.AbstractVectorSet}) where {T, F, S} = false
+function added_constraint_types(::Type{VectorSlackBridge{T, F, S}}) where {T, F<:MOI.AbstractVectorFunction, S}
+    return [(F, MOI.Zeros), (MOI.VectorOfVariables, S)]
```
blegat: `F` should be computed with `promote_operation` like in the scalar version

src/Bridges/slackbridge.jl
```diff
@@ -0,0 +1,162 @@
...
+end
+function concrete_bridge_type(::Type{<:VectorSlackBridge{T}},
+                              F::Type{<:MOI.AbstractVectorFunction},
+                              S::Type{<:MOI.AbstractVectorSet}) where T
+    return VectorSlackBridge{T, F, S}
```
blegat: `F` should be computed with `promote_operation` like in the scalar version

src/Bridges/slackbridge.jl
blegat: Using just `F` is sufficient for the three scalar functions types that we have in MOI but this bridge might be used by functions defined outside of MOI for which a `promote_operation` is necessary

test/Utilities/functions.jl
```diff
@@ -492,3 +492,56 @@
...
     end
 end
+
+@testset "Vector operate tests" begin
+    w = MOI.VariableIndex(0)
+    x = MOI.VariableIndex(1)
+    y = MOI.VariableIndex(2)
+
+    T = Float64
+    @test MOIU.promote_operation(+, T, MOI.VectorAffineFunction{T}, MOI.VectorAffineFunction{T}) == MOI.VectorAffineFunction{T}
```
blegat: I would also test `promote_operation` with `VectorOfVariables

src/Bridges/slackbridge.jl
joaquimg: By the way, this function is not being hit, should I test it explicitly?

joaquimg: Actually here it should still be F with no promote_operation, such promotion occurred before creating the bridge instance.

blegat: Yes but it needs to determine the constraints being created

joaquimg: Both `concrete_bridge_type` and `added_constraint_types` are called before the bridge constructor?
I agree that we need `promote_operation` inside `concrete_bridge_type`.
But I don't think it should be inside `added_constraint_types`, just like the other bridges.

blegat: You are right `F` is already the type computed with `promote_operation` in `conrete_type_bridge` so no need to change it

test/bridge.jl
```diff
@@ -341,7 +341,113 @@ end
...
+        ci = MOI.add_constraint(bridgedmock, f, MOI.Nonpositives(1))
+        @test MOI.get(bridgedmock, MOI.ConstraintFunction(), ci) ≈ f
+        newf = MOI.VectorAffineFunction(MOI.VectorAffineTerm.(1, MOI.ScalarAffineTerm.([2.0, 1.0], [x, y])), [0.0])
+        MOI.set(bridgedmock, MOI.ConstraintFunction(), ci, newf)
+        @test MOI.get(bridgedmock, MOI.ConstraintFunction(), ci) ≈ newf
+        @test MOI.get(bridgedmock, MOI.ConstraintSet(), ci) == MOI.Nonpositives(1)
+        MOI.modify(bridgedmock, ci, MOI.VectorConstantChange([1.0]))
+        @test MOI.get(bridgedmock, MOI.ConstraintFunction(), ci) ≈ 
+            MOI.VectorAffineFunction(MOI.VectorAffineTerm.(1, MOI.ScalarAffineTerm.([2.0, 1.0], [x, y])), [1.0])
+        # This power cone test is failing
```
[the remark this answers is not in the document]
joaquimg: I was about to ask you,
Its a TODO, hopefully easy to fix now.
I will paste here the error later.
It turns out that replacing a PowerCone is not done anywhere. The error might be something to do with paramtric types. I think its the only parametric type in vector cones.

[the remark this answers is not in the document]
joaquimg: At some point it complains about the impossibility to convert PowerCone into PoweCone{Float64} or vice versa.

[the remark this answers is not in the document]
joaquimg: At some point it complains about the impossibility to convert PowerCone into PoweCone{Float64} or vice versa.

src/Bridges/Bridges.jl
```diff
@@ -59,6 +61,16 @@ end
 
 include("vectorizebridge.jl")
 @bridge Vectorize VectorizeBridge () (MOI.EqualTo, MOI.LessThan, MOI.GreaterThan,) () () (MOI.SingleVariable,) (MOI.ScalarAffineFunction, MOI.ScalarQuadraticFunction) () ()
+include("slackbridge.jl")
+@bridge ScalarSlack ScalarSlackBridge () (MOI.Interval, MOI.LessThan, MOI.GreaterThan) () () () (MOI.ScalarAffineFunction, MOI.ScalarQuadraticFunction) () ()
+@bridge(VectorSlack, VectorSlackBridge,  (), (),
+    (MOI.Nonnegatives, MOI.Nonpositives, MOI.SecondOrderCone,
+    MOI.RotatedSecondOrderCone, MOI.GeometricMeanCone,
+    MOI.PositiveSemidefiniteConeSquare, MOI.LogDetConeTriangle,
+    MOI.RootDetConeTriangle, MOI.PowerCone, MOI.DualPowerCone),
```
blegat: The `PowerCone` should be in the next parenteses as `(MOI.PowerCone,)`

test/bridge.jl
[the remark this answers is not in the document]
blegat: There error is in the `@bridge` macro then, see my other comment

src/Bridges/Bridges.jl
joaquimg: Good catch, thanks!

joaquimg:
There is something about `test_delete_bridge`.
It assumes no variable was added by the bridge?

blegat:
> There is something about `test_delete_bridge`.
> It assumes no variable was added by the bridge?

You need to implement `NumberOfVariables` to so that the bridgeoptimizer can adjust the number of variables to ignore the ones created by the bridge. See `GeoMeanBridge`.

joaquimg:
It's alive!

Issue #612, opened 2018-12-31
32-bit problems


odow:
Explicit references to 64 bit integers cause failures like https://ci.appveyor.com/project/JuliaOpt/cbc-jl/builds/21300111/job/wodb6qccxfv8cgpi

Here is one offender:
https://github.com/JuliaOpt/MathOptInterface.jl/blob/999247c5e548c2235e49bec196809dc679c1948e/src/Utilities/model.jl#L591

@blegat any reason not to make these `Int`s?

mlubin:
Do we want to limit the number of variables allowed on a 32-bit system? Seems arbitrary to do so, unless it actually doesn't work.

blegat:
We use `Int64` in `VariableIndex` and `ConstraintIndex` to allow having more than 2^31 variables/constraint on 32-bit systems. If we remove it from MOIU model, then it is a bit useless to keep it in `VariableIndex` and `ConstraintIndex`.
I think it is better to fix Cbc.

PR #613, opened 2018-12-31
Replace Int64 with Int in .Utilities

odow:
Not overly sure what the impact of this is, or why it was chosen to be `Int64` in the first place.


mlubin:
https://github.com/JuliaOpt/MathOptInterface.jl/issues/612#issuecomment-450606719

Issue #614, opened 2019-01-02
VectorizeBridge and ScalarizeBridge


blegat:
- [x] VectorizeBridge: `AbstractScalarFunction`-in-`Union{EqualTo, LessThan, GreaterThan}` -> `AbstractVectorFunction`-in-`Union{Zeros, Nonpositives, Nonnegatives}`
- [x] ScalarizeBridge: `AbstractVectorFunction`-in-`Union{Zeros, Nonpositives, Nonnegatives}` -> `AbstractScalarFunction`-in-`Union{EqualTo, LessThan, GreaterThan}`

PR #615, opened 2019-01-02
✨ Add VectorizeBridge

blegat:
One can get a taste of how some wrapper can be simplified thanks to this bridge with : https://github.com/blegat/SeDuMi.jl/pull/6

This PR also fix and tests (with linear14test) a bug of bridges with deletion of variables.

First item of https://github.com/JuliaOpt/MathOptInterface.jl/issues/614

### DIFF
--- a/docs/src/apireference.md
+++ b/docs/src/apireference.md
@@ -317,7 +317,8 @@ InvalidIndex
 ```
 
 As discussed in [JuMP mapping](@ref), for scalar constraint with a nonzero
-function constant, a [`ScalarFunctionConstantNotZero`](@ref) should be thrown:
+function constant, a [`ScalarFunctionConstantNotZero`](@ref) exception may be
+thrown:
 ```@docs
 ScalarFunctionConstantNotZero
 ```
--- a/docs/src/apireference.md
+++ b/docs/src/apireference.md
@@ -316,6 +316,12 @@ be thrown:
 InvalidIndex
 ```
 
+As discussed in [JuMP mapping](@ref), for scalar constraint with a nonzero
+function constant, a [`ScalarFunctionConstantNotZero`](@ref) should be thrown:
+```@docs
+ScalarFunctionConstantNotZero
+```
+
 The rest of the errors defined in MOI fall in two categories represented by the
 following two abstract types:
 ```@docs
--- a/docs/src/apireference.md
+++ b/docs/src/apireference.md
@@ -349,6 +349,7 @@ Bridges.add_bridge
 
 Below is the list of bridges implemented in this package.
 ```@docs
+Bridges.VectorizeBridge
 Bridges.SplitIntervalBridge
 Bridges.RSOCBridge
 Bridges.GeoMeanBridge
--- a/src/Bridges/vectorizebridge.jl
+++ b/src/Bridges/vectorizebridge.jl
@@ -16,18 +16,22 @@ Transforms a constraint `AbstractScalarFunction`-in-`S` where `S <: LPCone` to
 """
 mutable struct VectorizeBridge{T, F, S} <: AbstractBridge
     vector_constraint::CI{F, S}
-    constant::T
-    set_constant::T # Need to store it as it is added in ConstraintPrimal
+    set_constant::T # constant in scalar set
 end
 function VectorizeBridge{T, F, S}(model::MOI.ModelLike,
                                   f::MOI.AbstractScalarFunction,
                                   set::MOI.AbstractScalarSet) where {T, F, S}
     set_constant = MOIU.getconstant(set)
     g = MOIU.operate(-, T, f, set_constant)
-    constant = MOI._constant(g)[1]
+    if -set_constant != MOI._constant(g)[1]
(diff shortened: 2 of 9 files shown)


src/Bridges/vectorizebridge.jl
```diff
@@ -0,0 +1,102 @@
...
+
+"""
+    VectorizeBridge{T}
+
+Transforms a constraint `AbstractScalarFunction`-in-`S` where `S <: LPCone` to
+`AbstactVectorFunction`-in-`vector_set_type(S)`.
+"""
+mutable struct VectorizeBridge{T, F, S} <: AbstractBridge
+    vector_constraint::CI{F, S}
+    constant::T
```
mlubin: What's the difference between `constant` and `set_constant`?

src/Bridges/vectorizebridge.jl
```diff
@@ -0,0 +1,102 @@
...
+    set_constant::T # Need to store it as it is added in ConstraintPrimal
+end
+function VectorizeBridge{T, F, S}(model::MOI.ModelLike,
+                                  f::MOI.AbstractScalarFunction,
+                                  set::MOI.AbstractScalarSet) where {T, F, S}
+    set_constant = MOIU.getconstant(set)
+    g = MOIU.operate(-, T, f, set_constant)
+    constant = MOI._constant(g)[1]
```
mlubin: Why do we have both `MOIU.getconstant` and `MOI._constant`? How is a reader supposed to distinguish these?

src/Bridges/vectorizebridge.jl
```diff
@@ -0,0 +1,102 @@
...
+    h = MOIU.operate(vcat, T, g)
+    vector_constraint = MOI.add_constraint(model, h, vector_set(set))
+    VectorizeBridge{T, F, S}(vector_constraint, constant, set_constant)
```
mlubin: Nit: missing `return`

src/Bridges/vectorizebridge.jl
```diff
@@ -0,0 +1,102 @@
...
+end
+function added_constraint_types(::Type{VectorizeBridge{T, F, S}}) where {T, F, S}
+    return [(F, S)]
+end
+function concrete_bridge_type(::Type{<:VectorizeBridge{T}},
+                              F::Type{<:MOI.AbstractScalarFunction},
+                              S::Type{<:LPCone{T}}) where T
+    G = MOIU.promote_operation(-, T, F, T)
+    H = MOIU.promote_operation(vcat, T, G)
+    VectorizeBridge{T, H, vector_set_type(S)}
```
mlubin: Nit: missing `return`

test/bridge.jl
```diff
@@ -215,6 +215,68 @@ end
...
+                (MOI.VectorAffineFunction{Float64}, MOI.Nonnegatives) => [[0], [1]]))
+        MOIT.linear2test(bridgedmock, config)
+
+        MOIU.set_mock_optimize!(mock,
+            (mock::MOIU.MockOptimizer) -> MOIU.mock_optimize!(mock, [0, 0]),
+            (mock::MOIU.MockOptimizer) -> MOIU.mock_optimize!(mock, [100, 0]),
+            (mock::MOIU.MockOptimizer) -> MOIU.mock_optimize!(mock, [100, -100]))
+        MOIT.linear4test(bridgedmock, config)
+
+        function set_mock_optimize_linear5Test!(mock)
```
mlubin: Nit: T should be lower case

src/Bridges/vectorizebridge.jl
blegat: Yes naming are very bad, we need to change these https://github.com/JuliaOpt/MathOptInterface.jl/issues/618

docs/src/apireference.md
```diff
@@ -316,6 +316,12 @@ be thrown:
 InvalidIndex
 ```
 
+As discussed in [JuMP mapping](@ref), for scalar constraint with a nonzero
+function constant, a [`ScalarFunctionConstantNotZero`](@ref) should be thrown:
```
mlubin: From this statement and the test below, does this mean that solvers are obligated to throw an error in this case? This is a pretty big shift in design and needs a bit more discussion.

@odow @joaquimg

blegat: I wanted to write `may`, in other places, I write `may` IIRC

src/Test/modellike.jl
```diff
@@ -344,7 +344,7 @@ function orderedindicestest(model::MOI.ModelLike)
     @test MOI.get(model, MOI.ListOfConstraintIndices{MOI.SingleVariable, MOI.LessThan{Float64}}()) == [c2, c3]
 end
 
-# Test that `MOI.ScalarFunctionConstantNotZero` is thrown when constraint with
+# Test that `MOI.ScalarFunctionConstantNotZero` is thrown when a constraint with
 # a function with nonzero constant is added
```
mlubin: Clarify that this test is optional because solvers could choose to support scalar functions with nonzero constants.

PR #617, opened 2019-01-05
Use UniversalFallback in fullbridgeoptimizer

blegat:
This is required by https://github.com/JuliaOpt/JuMP.jl/pull/1733 as it needs to be able to hold `Nonnegative`

### DIFF
--- a/src/Bridges/Bridges.jl
+++ b/src/Bridges/Bridges.jl
@@ -41,7 +41,8 @@ MOIU.@model(AllBridgedConstraints,
 Returns a `LazyBridgeOptimizer` bridging `model` for every bridge defined in this package and for the coefficient type `T`.
 """
 function fullbridgeoptimizer(model::MOI.ModelLike, ::Type{T}) where T
-    bridgedmodel = MOIB.LazyBridgeOptimizer(model, AllBridgedConstraints{T}())
+    cache = MOIU.UniversalFallback(AllBridgedConstraints{T}())
+    bridgedmodel = MOIB.LazyBridgeOptimizer(model, cache)
     add_bridge(bridgedmodel, MOIB.VectorizeBridge{T})
     add_bridge(bridgedmodel, MOIB.SplitIntervalBridge{T})
     add_bridge(bridgedmodel, MOIB.QuadtoSOCBridge{T})
--- a/src/Bridges/bridgeoptimizer.jl
+++ b/src/Bridges/bridgeoptimizer.jl
@@ -309,10 +309,12 @@ end
 function MOI.add_constraint(b::AbstractBridgeOptimizer, f::MOI.AbstractFunction,
                             s::MOI.AbstractSet)
     if is_bridged(b, typeof(f), typeof(s))
-        # We compute `BridgeType` first as might throw an
-        # `UnsupportedConstraint` error in which case, we do not want any
-        # modification to have been done
+        # We compute `BridgeType` first as `concrete_bridge_type` calls
+        # `bridge_type` which might throw an `UnsupportedConstraint` error in
+        # which case, we do not want any modification to have been done
         BridgeType = concrete_bridge_type(b, typeof(f), typeof(s))
+        # `add_constraint` might throw an `UnsupportedConstraint` but no
+        # modification has been done in the previous line
         ci = MOI.add_constraint(b.bridged, f, s)
         @assert !haskey(b.bridges, ci)
         b.bridges[ci] = BridgeType(b, f, s)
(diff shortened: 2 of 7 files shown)


src/Bridges/bridgeoptimizer.jl
```diff
@@ -296,9 +296,13 @@ end
 function MOI.add_constraint(b::AbstractBridgeOptimizer, f::MOI.AbstractFunction,
                             s::MOI.AbstractSet)
     if is_bridged(b, typeof(f), typeof(s))
+        # We compute `BridgeType` first as might throw an
```
mlubin: What might throw? `add_constraint`?

src/Bridges/lazybridgeoptimizer.jl
```diff
@@ -105,5 +105,7 @@ function supports_bridging_constraint(b::LazyBridgeOptimizer, F::Type{<:MOI.Abst
 end
 function bridge_type(b::LazyBridgeOptimizer{BT}, F::Type{<:MOI.AbstractFunction}, S::Type{<:MOI.AbstractSet}) where BT
     update_constraint!(b, F, S)
-    b.best[(F, S)]
+    get(b.best, (F, S)) do
```
mlubin: Nit: missing `return`

mlubin: This use of the `do` syntax seems a bit unusual. IMO the following would be easier to read:
```julia
result = get(b.best, (F, S), nothing)
if result !== nothing
   return result
else
   throw(...)
end
```

src/Bridges/Bridges.jl
```diff
@@ -41,7 +41,8 @@ MOIU.@model(AllBridgedConstraints,
 Returns a `LazyBridgeOptimizer` bridging `model` for every bridge defined in this package and for the coefficient type `T`.
 """
 function fullbridgeoptimizer(model::MOI.ModelLike, ::Type{T}) where T
-    bridgedmodel = MOIB.LazyBridgeOptimizer(model, AllBridgedConstraints{T}())
+    cache = MOIU.UniversalFallback(AllBridgedConstraints{T}())
```
mlubin: Are there any performance implications of adding another model layer?

blegat: The UniversalFallback is designed to have almost no overhead for supported constraints.
By the way, this internal model will go away with light bridges so it does not have any long term implication.

src/Bridges/lazybridgeoptimizer.jl
blegat: I am worried that with this construct the compiler might fail to deduce that the function never returns `nothing`. With this, the inferrability of the result is certain.
If followed the docstring of `get`:
```
get(f::Function, collection, key)

  Return the value stored for the given key, or if no mapping for the key is
  present, return f(). Use get! to also store the default value in the
  dictionary.

  This is intended to be called using do block syntax

  get(dict, key) do
      # default value calculated here
      time()
  end
```

blegat: Indeed, `typeof(f())` is `Nothing` so it might be equivalent

src/Bridges/lazybridgeoptimizer.jl
```diff
@@ -105,5 +105,9 @@ function supports_bridging_constraint(b::LazyBridgeOptimizer, F::Type{<:MOI.Abst
 end
 function bridge_type(b::LazyBridgeOptimizer{BT}, F::Type{<:MOI.AbstractFunction}, S::Type{<:MOI.AbstractSet}) where BT
     update_constraint!(b, F, S)
-    b.best[(F, S)]
+    BridgeType = get(b.best, (F, S), nothing)
```
mlubin: Hmm. There's a naming inconsistency. Local variables should be named like `bridge_type`, not BridgeType.

src/Bridges/lazybridgeoptimizer.jl
```diff
@@ -105,9 +105,9 @@ function supports_bridging_constraint(b::LazyBridgeOptimizer, F::Type{<:MOI.Abst
 end
 function bridge_type(b::LazyBridgeOptimizer{BT}, F::Type{<:MOI.AbstractFunction}, S::Type{<:MOI.AbstractSet}) where BT
     update_constraint!(b, F, S)
-    BridgeType = get(b.best, (F, S), nothing)
-    if BridgeType === nothing
+    bridge_type = get(b.best, (F, S), nothing)
```
mlubin: I'm assuming this doesn't cause an issue with having the same name as the function itself?

PR #624, opened 2019-01-10
Add expression graphs for NLP tests

odow:


src/Test/nlp.jl
```diff
@@ -174,12 +188,22 @@ end
...
-        return [:Grad, :Jac, :Hess]
+        return [:Grad, :Jac, :Hess, :ExprGraph]
     else
-        return [:Grad, :Jac]
+        return [:Grad, :Jac, :ExprGraph]
     end
 end
 
+function MOI.objective_expr(d::FeasibilitySenseEvaluator)
+    return :(0 * x[1])
```
mlubin: This seems hacky. What about returning 0.0 instead?

odow: I think there is a bug somewhere if you pass 0.0. It's on my list of things to investigate today.

src/Test/nlp.jl
```diff
@@ -174,12 +188,22 @@ end
...
-        return [:Grad, :Jac, :Hess]
+        return [:Grad, :Jac, :Hess, :ExprGraph]
     else
-        return [:Grad, :Jac]
+        return [:Grad, :Jac, :ExprGraph]
     end
 end
 
+function MOI.objective_expr(d::FeasibilitySenseEvaluator)
+    return :(0.0)
```
mlubin: `:(0.0)` parses exactly the same as `0.0`.

src/Test/nlp.jl
chriscoey: worth merging in master and trying again?

PR #626, opened 2019-01-12
add test for solving with a partial starting point

mlubin:
Ref https://github.com/JuliaOpt/Ipopt.jl/issues/155

### DIFF
--- a/src/Test/contlinear.jl
+++ b/src/Test/contlinear.jl
@@ -1472,8 +1472,12 @@ function partial_start_test(model::MOI.ModelLike, config::TestConfig)
     MOI.empty!(model)
     @test MOI.is_empty(model)
 
+    @test MOI.supports(model, MOI.VariablePrimalStart(), MOI.VariableIndex)
+
     x = MOI.add_variable(model)
     y = MOI.add_variable(model)
+
+    MOI.set(model, MOI.VariablePrimalStart(), x, 1.0)
 
     MOI.add_constraint(model, x, MOI.GreaterThan(0.0))
     MOI.add_constraint(model, y, MOI.GreaterThan(0.0))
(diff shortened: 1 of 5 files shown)


src/Test/contlinear.jl
```diff
@@ -1459,6 +1459,41 @@ function linear15test(model::MOI.ModelLike, config::TestConfig)
     end
 end
 
+# Exclude this test if the solver isn't intended to support incomplete starting
```
blegat: Where is the starting point set ?

mlubin: Good catch! :sweat_smile:

blegat:
We won't be able to include it in MOI v0.8.2 as it will make almost all solver tests fail or we will have to force it on METADATA.
It would be nice if contlinear didn't use primal start do that all solvers can pass all the tests.
To me the new test feel more like a unit test.

mlubin:
METADATA isn't currently running reverse dependency tests, so we're not blocked by that.

Yes, I'd guess most solvers fail this test with an unhelpful error message, but that's a pretty significant usability issue and likely regression form JuMP 0.18 that we should be aware of and try to address. We can either add the test here so we're forced to deal with it, or open issues on all of the solver wrappers.

blegat:
What I mean is that some solvers do not even support starting points.
For solvers that support it but wrappers fail when it is partially set, I agree that their tests should fail.
I am concerned that this prevents solvers not supporting starting points to pass the tests. Maybe a `config` parameter could help

mlubin:
> I am concerned that this prevents solvers not supporting starting points to pass the tests. 

They can always exclude the test, but should have a good reason to do so. Erroring when the solver does not support warm starts is a regression from JuMP 0.18. This was previously a warning: https://github.com/JuliaOpt/JuMP.jl/blob/d8f526d48f9c79f6aa27ef45aea381b4b7e63f32/src/solvers.jl#L413.

mlubin:
> Maybe we should implement the warning in MOI then instead of an ArgumentError.

That makes sense.

blegat:
Then we don't have to exclude this test from `test/Utilities/...`, `test/bridge.jl`, ...

PR #629, opened 2019-01-14
✨ Add SpecificTerminationStatus

blegat:
Closes https://github.com/JuliaOpt/MathOptInterface.jl/issues/99

### DIFF
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -789,6 +789,13 @@ This group of statuses means that something unexpected or problematic happened.
     OTHER_ERROR
 )
 
+"""
+    SpecificTerminationStatus()
+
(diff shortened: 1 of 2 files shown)


src/attributes.jl
```diff
@@ -789,6 +789,13 @@ This group of statuses means that something unexpected or problematic happened.
     OTHER_ERROR
 )
 
+"""
+    SpecificTerminationStatus()
```
mlubin: Shortening "solver-specific" to "specific" makes this a bit ambiguous to me. What about `SolverStatusString()` or `SolverTerminationStatus()` or `RawTerminationStatus()` or `RawStatusString()`?

blegat: I like `RawStatusString`

PR #630, opened 2019-01-14
Only warn if starting values not supported

blegat:
I am still unsure whether we should add this in copy or do it by adding a custom `set` methods.
The advantage of doing it in `copy` is that we can easily add a keyword argument to disable the warning (normally all wrappers pass the keyword arguments with `kws...` so it won't imply any change for them).
The disadvantage is that it will not work in JuMP direct mode as it does not use copy. So in direct mode, adding starting values to an optimizer not supporting it gives an error. Maybe we could say in direct mode, the user should explicitly check if it is supported with `supports`.
Note that as starting values are not supported by the `MOIU` `@model` yet (it should IMO), the wrappers should use an `UniversalFallback` in the tests to pass the tests with starting values added in https://github.com/JuliaOpt/MathOptInterface.jl/pull/626.
See e.g.
https://github.com/blegat/SDPA.jl/blob/1d562d175febf90cd661a136bf103f04878620a1/test/MOI_wrapper.jl#L17-L18

### DIFF
--- a/src/Test/contlinear.jl
+++ b/src/Test/contlinear.jl
@@ -1472,8 +1472,6 @@ function partial_start_test(model::MOI.ModelLike, config::TestConfig)
     MOI.empty!(model)
     @test MOI.is_empty(model)
 
-    @test MOI.supports(model, MOI.VariablePrimalStart(), MOI.VariableIndex)
-
     x = MOI.add_variable(model)
     y = MOI.add_variable(model)
 
--- a/src/Utilities/copy.jl
+++ b/src/Utilities/copy.jl
@@ -117,15 +117,27 @@ function pass_attributes(dest::MOI.ModelLike, src::MOI.ModelLike, copy_names::Bo
     _pass_attributes(dest, src, copy_names, idxmap, attrs, (CI{F, S},), (cis_src,), (cis_dest,), passattr!)
 end
 
-function _pass_attributes(dest::MOI.ModelLike, src::MOI.ModelLike, copy_names::Bool, idxmap::IndexMap, attrs, canargs, getargs, setargs, passattr!::Function=MOI.set)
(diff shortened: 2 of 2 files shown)


src/Test/contlinear.jl
```diff
@@ -1472,8 +1472,6 @@ function partial_start_test(model::MOI.ModelLike, config::TestConfig)
     MOI.empty!(model)
     @test MOI.is_empty(model)
 
-    @test MOI.supports(model, MOI.VariablePrimalStart(), MOI.VariableIndex)
```
mlubin: Hmm? This shouldn't be removed.

src/Utilities/copy.jl
```diff
@@ -117,16 +117,28 @@ function pass_attributes(dest::MOI.ModelLike, src::MOI.ModelLike, copy_names::Bo
...
-                mapped_value = attribute_value_map(idxmap, value)
-                passattr!(dest, attr, setargs..., mapped_value)
+        if !copy_names && (attr isa MOI.Name || attr isa MOI.VariableName || attr isa MOI.ConstraintName)
+            continue
+        end
+        if attr isa MOI.VariablePrimalStart || attr isa MOI.ConstraintPrimalStart || attr isa MOI.ConstraintDualStart
+            # As starting values are simply *hints* for the optimization, not
+            # supporting them gives a warning, not an error
+            if !MOI.supports(dest, attr, supports_args...)
+                Compat.@warn("$attr is not supported by $(typeof(dest)).")
```
mlubin: Mention that the information will be discarded.

src/Test/contlinear.jl
blegat: This allow solvers not supporting variable primal start to pass the test with a warning

PR #634, opened 2019-01-17
✨ Add zero for functions

blegat:

### DIFF
--- a/src/Utilities/functions.jl
+++ b/src/Utilities/functions.jl
@@ -547,6 +547,14 @@ end
 
 # Arithmetic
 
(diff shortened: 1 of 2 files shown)


mlubin:
I should have taken a look at this earlier. We should not be using `Base.zero()` for mutable types (https://github.com/JuliaOpt/JuMP.jl/issues/1151). Consider creating a new function instead of extending `Base.zero`.

blegat:
Sorry for merging without review, I needed the changes of this PR and https://github.com/JuliaOpt/MathOptInterface.jl/pull/610 at the same time.

PR #635, opened 2019-01-18
make read/write to/from file use IO instead of filename string

### LINKED ISSUE #633: Read/write from/to file should be IO not filenames
Currently we have
https://github.com/JuliaOpt/MathOptInterface.jl/blob/b63ab44e22cd23d86296aff7f465bb95f32a85e2/src/MathOptInterface.jl#L43

but it would be better change this to `read_from_file(model, io::IO)` and to have a fallback
```julia
function read_from_file(model::ModelLike, io::IO)
    error("read_from_file not implemented for model of type $(typeof(model))")
end

function read_from_file(model::ModelLike, filename::String)
    open(filename, "r") do io
        read_from_file(model, io)
    end
end
```
The motivation is to enable MathOptFormat to support GZipped files https://github.com/odow/MathOptFormat.jl/issues/39. cc @chriscoey

chriscoey:
please review / change as necessary, as I am not sure I formatted the interface definitions and docstrings correctly.

fixes #633 

needed for https://github.com/odow/MathOptFormat.jl/pull/40

### DIFF
--- a/src/MathOptInterface.jl
+++ b/src/MathOptInterface.jl
@@ -44,9 +44,9 @@ function write_to_file(model::ModelLike, io::IO)
     error("write_to_file not implemented for model of type $(typeof(model))")
 end
 
-function write_to_file(model::MOI.ModelLike, filename::String)
+function write_to_file(model::ModelLike, filename::String)
     open(filename, "w") do io
-        MOI.write_to_file(model, io)
+        write_to_file(model, io)
     end
 end
 
@@ -65,7 +65,7 @@ Once the contents of the file are loaded into the model, users can query the var
 `get(model, ListOfVariableIndices())`. However, some filetypes, such as LP files, do not
 maintain an explicit ordering of the variables, so the returned list may be in an
 arbitrary order. In this case, users should look up each variable index by name:
-`get(model, VariableIndex, "name")`. 
+`get(model, VariableIndex, "name")`.
 """
 function read_from_file end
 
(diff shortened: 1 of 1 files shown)


src/MathOptInterface.jl
```diff
@@ -33,14 +33,26 @@ function optimize! end
 
 """
     write_to_file(model::ModelLike, filename::String)
+    write_to_file(model::ModelLike, io::IO)
```
mlubin: This isn't well defined. `write_to_file` extracts the file format from the file extension. An `IO` object doesn't specify the file format.

chriscoey: see discussion in https://github.com/odow/MathOptFormat.jl/issues/39
and new function at https://github.com/odow/MathOptFormat.jl/pull/40/files#diff-8da2c3201bdc254700cc0b2b20f2fc8eR26
it "works" in MOF tests because the model type itself specifies the file type
@odow thoughts?

odow: It seems like `write_to_file` and `read_from_file` should only ever be defined for concrete `model` arguments. Depending on parsing the format from the extension is just looking for trouble.

Alternatively, we could have something like:
```julia
module MOI

abstract type AbstractFileType end

"""
    write_to_file(model::ModelLike, filename::String)
    write_to_file(model::ModelLike, io::IO, file_type::AbstractFileType)
"""
function write_to_file end

end
```
```julia
module MathOptFormat

struct MOFFile <: MOI.AbstractFileType
struct MOFModel <: MOI.ModelLike end

function MOI.write_to_file(model::MOFModel, io, ::MOFFile)
    # ...
end

function MOI.write_to_file(model::MOFModel, filename)
    if endswith(filename, ".mof.json")
        open(filename, "r") do io
            MOI.write_to_file(model, io, MOFFile())
        end
    end
end

end
```

mlubin: What about using Julia's pre-existing show/display functions with MIME types (https://docs.julialang.org/en/v1/base/io-network/index.html#Multimedia-I/O-1)?

odow:
Closed by #969

Issue #636, opened 2019-01-18
Should we extend Base.zero for ScalarAffineFunction and ScalarQuadraticFunction


blegat:
See https://github.com/JuliaOpt/MathOptInterface.jl/pull/634#issuecomment-455572745

mlubin:
No, we shouldn't. This is a mistake in JuMP that can lead to hard to find bugs.

blegat:
I would argue that the case is a bit different from JuMP where JuMP expressions are meant to be userfriendly and the user should be able to manipulate just like numbers.
On the other hand, MOI function utilities are primarily used by bridges and we may expect more from the user (and some say that [`zeros` should be thought as a short-hand for `fill`](https://github.com/JuliaLang/julia/issues/29168#issuecomment-421059562) hence it's ok to define `zero` for mutable struct). We could also overload `zeros` on MOI functions and throw an error suggesting to use `fill` instead (as `zeros(::{ScalarAffineFunction{T}}, n)` may seem ambiguous on whether the user is trying to build a `VectorAffineFunction`).

The reason I would like `zero` to be defined is that it allows MOI functions to be used in a code that doesn't know anything about MOI and simply uses standard arithmetic operations. This is needed for SumOfSquares which relies on MultivariatePolynomials which does not assumes anything about the coefficient type. The coefficients used to be JuMP expressions but now with the bridges, it will be MOI functions.

mlubin:
> The reason I would like zero to be defined is that it allows MOI functions to be used in a code that doesn't know anything about MOI and simply uses standard arithmetic operations. 

That's exactly the issue. Code that doesn't know anything about MOI probably wasn't written with mutable number types in mind and isn't tested with them. We shouldn't implement the `zero` method if we violate the assumptions of the interface. This is about correctness and avoiding traps, which applies regardless of how advanced the users are. If `MultivariatePolynomials` is aware of mutable zeros, it can define and use `maybe_mutable_zero(T)` that is extensible by callers.

I don't know if implementing `zeros` is sufficient. We don't know where else this assumption is used.

blegat:
> That's exactly the issue. Code that doesn't know anything about MOI probably wasn't written with mutable number types in mind and isn't tested with them.

[...]

PR #643, opened 2019-01-25
✨ Add iszero

blegat:
Fallbacks to `zero(f)` by default which is not implemented (only implemented for its type) and is disputed https://github.com/JuliaOpt/MathOptInterface.jl/issues/636.

### DIFF
--- a/src/Utilities/functions.jl
+++ b/src/Utilities/functions.jl
@@ -354,6 +354,24 @@ function test_constraintnames_equal(model, constraintnames)
 end
 
 """
+    all_coefficients(p::Function, f::MOI.AbstractFunction)
+
+Determine whether predicate `p` returns `true` for all coefficients of `f`,
+returning `false` as soon as the first coefficient of `f` for which `p`
+returns `false` is encountered (short-circuiting). Similar to `all`.
+"""
+function all_coefficients end
+
+function all_coefficients(p::Function, f::MOI.ScalarAffineFunction)
+    p(f.constant) && all(t -> p(MOI.coefficient(t)), f.terms)
+end
+function all_coefficients(p::Function, f::MOI.ScalarQuadraticFunction)
+    p(f.constant) &&
+        all(t -> p(MOI.coefficient(t)), f.affine_terms) &&
+        all(t -> p(MOI.coefficient(t)), f.quadratic_terms)
+end
+
+"""
     isapprox_zero(f::MOI.AbstractFunction, tol)
 
 Return a `Bool` indicating whether the function `f` is approximately zero using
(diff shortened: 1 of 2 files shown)


src/Utilities/functions.jl
```diff
@@ -379,16 +372,28 @@ then `isapprox_zero(f)` is `false` but `isapprox_zero(MOIU.canonical(f))` is
...
+function isapprox_zero(f::MOI.AbstractFunction, tol)
+    return _iszero(f, α -> isapprox_zero(α, tol))
+end
+
+function Base.iszero(f::Union{MOI.ScalarAffineFunction{T},
+                              MOI.ScalarQuadraticFunction{T}}) where T
+    return _iszero(canonical(f), iszero)
+end
+
+function _iszero(f::MOI.ScalarAffineFunction, _is_zero::Function)
```
mlubin: Please improve the naming :)
Local variables shouldn't begin with underscores. And the distinction between `_iszero` and `_is_zero` is confusing.

src/Utilities/functions.jl
```diff
@@ -353,11 +353,22 @@ function test_constraintnames_equal(model, constraintnames)
...
+    all_coefficients(p::Function, f::MOI.AbstractFunction)
+
+Determine whether predicate `p` returns `true` for all coefficients of `f`,
+returning `false` as soon as the first coefficient of `f` for which `p`
+returns `false` is encountered (short-circuiting). Similar to `all`.
+"""
+function all_coefficients end
+
+function all_coefficients(p::Function, f::MOI.ScalarAffineFunction)
+    p(f.constant) && all(t -> p(MOI.coefficient(t)), f.terms)
```
odow: Missing explicit `return`

src/Utilities/functions.jl
```diff
@@ -353,11 +353,22 @@ function test_constraintnames_equal(model, constraintnames)
...
+end
+function all_coefficients(p::Function, f::MOI.ScalarQuadraticFunction)
+    p(f.constant) &&
+        all(t -> p(MOI.coefficient(t)), f.affine_terms) &&
+        all(t -> p(MOI.coefficient(t)), f.quadratic_terms)
```
odow: ditto

src/Utilities/functions.jl
```diff
@@ -379,16 +390,15 @@ then `isapprox_zero(f)` is `false` but `isapprox_zero(MOIU.canonical(f))` is
 """
 function isapprox_zero end
 
-function isapprox_zero(f::MOI.ScalarAffineFunction, tol)
-    isapprox_zero(f.constant, tol) && all(t -> isapprox_zero(t, tol),
-                                           f.terms)
+isapprox_zero(α::AbstractFloat, tol) = -tol <= α <= tol
```
odow: Maybe a separate PR, but this should be named `is_approx_zero`?

blegat: This is best discussed in a separate issue but I would argue that isapprox_zero has the advantage to be consistent with Base.isapprox

PR #644, opened 2019-01-25
✨ Add unary +

blegat:

### DIFF
--- a/src/Utilities/functions.jl
+++ b/src/Utilities/functions.jl
@@ -778,6 +778,11 @@ function operate(op::typeof(+), ::Type{T}, f, g, h, args...) where T
     return operate!(+, T, operate(+, T, f, g), h, args...)
 end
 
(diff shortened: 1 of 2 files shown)


src/Utilities/functions.jl
```diff
@@ -778,6 +778,11 @@ function operate(op::typeof(+), ::Type{T}, f, g, h, args...) where T
     return operate!(+, T, operate(+, T, f, g), h, args...)
 end
 
+# Unary +
+function operate(op::typeof(+), ::Type{T}, f::MOI.AbstractFunction) where T
+    return f
```
mlubin: The docstring for `operate` doesn't say what `T` is used for. Why is it ignored most of the time?

blegat: It is the coefficient type. It is useful when summing two SingleVariable for instance

PR #647, opened 2019-01-25
🐛 Fix lazy bridge check for supported constraints

blegat:
Fix a StackOverflow when it re-triggers the algo when checking if a constraint is supported.

### DIFF
--- a/src/Bridges/lazybridgeoptimizer.jl
+++ b/src/Bridges/lazybridgeoptimizer.jl
@@ -43,7 +43,7 @@ function update_dist!(b::LazyBridgeOptimizer, constraints)
         changed = false
         for BT in b.bridgetypes
             for (F, S) in constraints
-                if MOI.supports_constraint(BT, F, S) && all(C -> MOI.supports_constraint(b, C[1], C[2]), added_constraint_types(BT, F, S))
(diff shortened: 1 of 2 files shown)


src/Bridges/lazybridgeoptimizer.jl
```diff
@@ -99,6 +99,11 @@ end
 function is_bridged(b::LazyBridgeOptimizer, F::Type{<:MOI.AbstractFunction}, S::Type{<:MOI.AbstractSet})
     !MOI.supports_constraint(b.model, F, S)
 end
+# Same as supports_constraint but do not trigger `update_constraint!`. This is
```
mlubin: Where is the implementation of `supports_constraint` for `LazyBridgeOptimizer`?

blegat:
I need the graph of bridges to contains a cycle for this bug to occur so I am waiting for https://github.com/JuliaOpt/MathOptInterface.jl/pull/658 to be merged to create a test

src/Bridges/lazybridgeoptimizer.jl
blegat: in `bridgeoptimizer.jl`

Issue #511, opened 2018-08-28
What happens when deleting a variable in a constraint?


odow:
As [raised by Miles](https://github.com/JuliaOpt/JuMP.jl/pull/1445#discussion_r213150825), what should happen when the user tries to delete a variable that still exists in a constraint?

It seems like there are a few possible answers:
1. the variable gets deleted and the constraints are modified. This is easy for scalar terms, but less obvious for quadratic terms. Do they become linear?
    - [ ] todo: check what happens in Gurobi etc.
2. the solver throws and error and says delete constraints associated with the variable first
3. solver-specific -- some solvers allow, others will throw an error.

Thoughts?

joaquimg:
I don't think its a nice optimization API if you need to delete or modify constraints explicitly when you delete a variable.
I don't know about quadratics, maybe thats why Xpress does not separates them. Also, Xpress has a method to identify constraints with quadratic terms, once all variables in quadratic terms for a constraint are deletes the constraint is identified as a linear constraint.

blegat:
If one deletes a variable part of a VectorOfVariables constraint, it should be an error except when it deletes a vector of variables and all the variables of the VectorOfVariables constraint are deleted

joaquimg:
> If one deletes a variable part of a VectorOfVariables constraint, it should be an error except when it deletes a vector of variables and all the variables of the VectorOfVariables constraint are deleted

this makes a lot of sense, but I am afraid that going for consistency might harm a lot the interface. Maybe we need a new kind of delete?

blegat:
Like one `delete` that errors if constraints are still using the variable and one `delete_clean` that also sets the variable to zero in all functions ?

joaquimg:
thats a possibility

odow:
When Gurobi deletes a variable, it deletes the corresponding column from the linear A matrix, and deletes the row and column from any quadratic Q matrix. So if there are `x*y` cross-terms, these are deleted as well; they are not converted to linear terms.

Gist: https://gist.github.com/odow/8a4b4b529ae47c18b64b57feebd45a47

odow:
There are so many edge-cases to consider. 
- What happens if I have `ObjectiveFunction{SingleVariable}` and then delete that variable? What type is the objective function?
- If I have a quadratic objective with `x^2`, and then I delete `x`, is the objective linear? Or quadratic with no terms? 
[...]

PR #652, opened 2019-01-31
adding clarification to docs related to variable deletion

IssamT:
Is this valid? 

My understanding is based on LQOI [interpretation](https://github.com/JuliaOpt/LinQuadOptInterface.jl/blob/master/src/variables.jl#L177-L180).


blegat:
It is discussed in https://github.com/JuliaOpt/MathOptInterface.jl/issues/511

IssamT:
Thanks.

Issue #528, opened 2018-09-23
Slack and Unslack bridge


blegat:
- [x]  SlackBridge: Transforms a constraint `F`-in-`S` where `F` is AbstractFunction different from `SingleVariable` and `VectorOfVariables` and `S` is any set into a constraint `F`-in-`Zeros` and a constraint `SingleVariable` (or `VectorOfVariables`) in `S` (https://github.com/JuliaOpt/MathOptInterface.jl/pull/610).
- [x] VectorOfVariablesBridge (resp. SingleVariableBridge): Transforms a constraint `VectorOfVariables`-in-`AbstractVectorSet` (resp. `SingleVariable`-in-`AbstractScalarSet`) into `VectorAffineFunction`-in-`AbstractVectorSet` (resp. `ScalarAffineFunction`-in-`AbstractScalarSet`).

Not yet totally convinced by the names so any better idea would be appreciated :)

joaquimg:
@blegat have you started working on it? I was planning to give it a try myself.

blegat:
Not at all, feel free :)

joaquimg:
My first idea was to create a single bridge for scalar and vector functions, but due to type instability it might be much harder to write a single bridge. So I decided to split and the scalar one would go like this:
```julia
struct ScalarSlackBridge{T, F<:MOI.AbstractScalarFunction, S<:MOI.AbstractScalarSet} <: AbstractBridge
    slack::MOI.VariableIndex
    slack_in_set::CI{MOI.SingleVariable, S}
    equality::CI{F, MOI.EqualTo{T}}
end
```
My first draft is here: https://github.com/JuliaOpt/MathOptInterface.jl/commit/fbb946d4e5196f618da5678e5b2ddfe7257e6831
Suggestions appreciated :)

joaquimg:
I am, but I plan to finish the slack bridge this week only. My schedule would be to do the unslack before the end of january.
If you are in need go for it.

joaquimg:
Now I am going for the second group

blegat:
Unslack done in https://github.com/JuliaOpt/MathOptInterface.jl/pull/659

PR #666, opened 2019-02-07
Add Bridges.ListOfNonstandardBridges attribute

blegat:
This attributes allow an optimizer to list a few bridges that should automatically be used by JuMP.
Here are a few use cases:
* Solvers that requires SOC constraint to be provided as ScalarQuadraticFunction-in-LessThan can add the SOCtoQuad bridge in this list in the prefered form (either supporting VectorOfVariables-in-SOC or VectorAffineFunction-in-SOC, since for some solver, the quadratic function obtained is recognized only if it is computated from the correct one, and we can go from one form to another with the [Slack and VectorOfVariables bridges](https://github.com/JuliaOpt/MathOptInterface.jl/issues/528)), see https://github.com/JuliaOpt/MathOptInterface.jl/pull/478#discussion_r211120417 (@ccoffrin this would resolve the issue we discussed with @mlubin )
* Solvers that use non-standard form for a set can create a custom set and a bridge inside the wrapper:
    - [ECOS](https://github.com/JuliaOpt/ECOS.jl) exponential cone which does not has the same order
    - [SeDuMi](https://github.com/blegat/SeDuMi.jl) PSDSquare-like cone
    - [CDCS](https://github.com/blegat/CDCS.jl) PSDSquare-like cone (not clear if it exactly the PSDSquare yet, see https://github.com/oxfordcontrol/CDCS/issues/13)

- [x] Needs tests

### DIFF
--- a/src/Bridges/Bridges.jl
+++ b/src/Bridges/Bridges.jl
@@ -23,16 +23,69 @@ include("debug.jl")
 """
     full_bridge_optimizer(model::MOI.ModelLike, ::Type{T}) where {T}
 
-Returns a `LazyBridgeOptimizer` bridging `model` for every bridge defined in
-this package and for the coefficient type `T`.
+Returns a [`LazyBridgeOptimizer`](@ref) bridging `model` for every bridge
+defined in this package (see below for the few exceptions) and for the
+coefficient type `T` in addition to the bridges in the list returned by
+`MOI.get(model, MOI.Bridges.ToAdd{T}())`.
+
+See also [`ToAdd`](@ref).
+
+!!! note
+    The following bridges are not added by `full_bridge_optimizer` except if
+    they are in the list returned by `MOI.get(model, MOI.Bridges.ToAdd{T}())`
+    (see the docstrings of the corresponding bridge for the reason they are not
+    added):
+    * [`Constraint.SOCtoNonConvexQuad`](@ref),
+      [`Constraint.RSOCtoNonConvexQuad`](@ref) and
+      [`Constraint.SOCtoPSDBridge`](@ref).
+    * The subtypes of [`Constraint.AbstractToIntervalBridge`](@ref) (i.e.
+      [`Constraint.GreaterToIntervalBridge`](@ref) and
+      [`Constraint.LessToIntervalBridge`](@ref)) if `T` is not a subtype of
+      `AbstractFloat`.
 """
 function full_bridge_optimizer(model::MOI.ModelLike, ::Type{T}) where {T}
     bridged_model = LazyBridgeOptimizer(model)
+    for BT in MOI.get(model, ToAdd{T}())
+        MOIB.add_bridge(bridged_model, BT)
+    end
     Variable.add_all_bridges(bridged_model, T)
     Constraint.add_all_bridges(bridged_model, T)
     Objective.add_all_bridges(bridged_model, T)
     return bridged_model
 end
+
+"""
+    ToAdd{T}()
+
+A list of additional bridges that should be added to the
+[`LazyBridgeOptimizer`](@ref) in `full_bridge_optimizer(optimizer, T)`.
+
+See also [`full_bridge_optimizer`](@ref).
+
+## Examples
+
+Suppose an optimizer can exploit specific structure of a constraint.
+For instance, suppose it can exploit the structure of the matrix `A`
+in the linear system of equations `A * x = b`.
+The optimizer can create a function:
+```julia
+struct MatrixAffineFunction{T,M<:AbstractMatrix{T}} <: MOI.AbstractVectorFunction
+    A::M
+    b::Vector{T}
+end
+```
+and then a bridge from `VectorAffineFunction{T}` to
+`MatrixAffineFunction{T,SparseMatrixCSC{T,Int}}`.
+The bridge is then included in the list returned as the value of the
+`ToAdd` attribute. That way, the optimizer can exploit a specific matrix
+structure if the user uses a `MatrixAffineFunction` and in case the user uses
+`VectorAffineFunction` then it's automatically bridged.
+"""
+struct ToAdd{T} <: MOI.AbstractOptimizerAttribute end
+
+MOI.is_copyable(::ToAdd) = false
+
+MOI.get(model::MOI.ModelLike, ::ToAdd) = []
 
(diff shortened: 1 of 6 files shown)


src/attributes.jl
```diff
@@ -332,6 +332,18 @@ An optimizer attribute for the string identifying the solver/optimizer.
 """
 struct SolverName <: AbstractOptimizerAttribute end
 
+"""
+    RequiredBridges()
+
+A list of bridges that are required when using the optimizer. For instance, if
```
mlubin: What does "required" mean? Required to pass tests? What does this imply for direct mode?

blegat: > What does "required" mean? Required to pass tests?

I was hesitant for the naming, may be recommended (and RecommendedBridges) is better ?

> What does this imply for direct mode?

If the user uses a `LazyBridgeOptimizer` in direct mode then the bridges will be automatically added. The bridges are added in the `LazyBridgeOptimizer` constructor, not just in the `full_bridge_optimizer` function.

mlubin: I don't think `RecommendedBridges` is any clearer. Recommended for what? How is someone looking at this description supposed to know how it fits into `LazyBridgeOptimizer`?

If a solver has required bridges, maybe it shouldn't publicly return its MOI optimizer and instead return a `LazyBridgeOptimizer` that wraps the optimizer with the required bridges? It doesn't make sense to me to have required bridges while the default way of constructing the solver doesn't include them.

blegat: There is indeed the other option of returning the optimizer wrapped in a `SingleBridgeOptimizer` (like we currently do in tests). As the bridge is required there is no need to include it in the Bellman-Ford computation.
What would be the recommendation for naming the constructor and the optimizer ?
Adding an underscore before `Optimizer` ?
```julia
struct _Optimizer <: MOI.AbstractOptimizer
end
Optimizer(...) = CustomBridge(_Optimizer(...))
```

mlubin: > Adding an underscore before `Optimizer` ?

Yes, that makes sense.

src/Bridges/Bridges.jl
```diff
@@ -23,17 +23,70 @@ include("debug.jl")
...
+end
+```
+and then a bridge from `VectorAffineFunction{T}` to
+`MatrixAffineFunction{T,SparseMatrixCSC{T,Int}}`.
+The bridge is then included in the list returned as the value of the
+`ToAdd` attribute. That way, the optimizer can exploit a specific matrix
+structure if the user uses a `MatrixAffineFunction` and in case the user uses
+`VectorAffineFunction` then it's automatically bridged.
+"""
+struct ToAdd{T} <: MOI.AbstractOptimizerAttribute end
```
odow, line 84: I'm not a fan of `ToAdd`. How about

````Julia
"""
    ListOfNonstandardBridges{T}() <: MOI.AbstractOptimizerAttribute

Any optimizer can be wrapped in a [`LazyBridgeOptimizer`](@ref) using 
[`full_bridge_optimizer`](@ref). However, by default [`LazyBridgeOptimizer`](@ref)
uses a limited set of bridges that are:

  1. implemented in `MOI.Bridges`
  2. generally applicable for all optimizers.

For some optimizers however, it is useful to add additional bridges, such as
those that are implemented in external packages (e.g., within the solver package
itself) or only apply in certain circumstances (e.g., 
[`Constraint.SOCtoNonConvexQuad`](@ref)).

Optimizers should implement the `ListOfNonstandardBridges` attribute to return a
vector of bridge types that are added by [`full_bridge_optimizer`](@ref) in
addition to the list of default bridges.

Note that optimizers implementing `ListOfNonstandardBridges` may be unusable
ifthe non-standard bridges are not added.

## Examples

### An optimizer using a non-default bridge in `MOI.Bridges`

Solvers supporting [`MOI.ScalarQuadraticFunction`](@ref) can support
[`MOI.SecondOrderCone`](@ref) and [`MOI.RotatedSecondOrderCone`](@ref) by
defining:
```julia
function MOI.get(::MyQuadraticOptimizer, ::ListOfNonstandardBridges{Float64})
    return Type[
        MOI.Bridges.Constraint.SOCtoNonConvexQuad{Float64},
        MOI.Bridges.Constraint.RSOCtoNonConvexQuad{Float64},
    ]
end
```

### An optimizer defining an internal bridge

Suppose an optimizer can exploit specific structure of a constraint, e.g., it
can exploit the structure of the matrix `A` in the linear system of equations
`A * x = b`.

The optimizer can define the function:
```julia
struct MatrixAffineFunction{T} <: MOI.AbstractVectorFunction
    A::SomeStructuredMatrixType{T}
    b::Vector{T}
end
```
and then a bridge
```julia
struct MatrixAffineFunctionBridge{T} <: MOI.Constraint.AbstractBridge
    f::MOI.VectorAffineFunction{T}
end
```
from `VectorAffineFunction{T}` to the `MatrixAffineFunction`. Finally, it
defines:
```julia
function MOI.get(::Optimizer{T}, ::ListOfNonstandardBridges{T}) where {T}
    return Type[MatrixAffineFunctionBridge{T}]
end
```
"""
````

File: src/Bridges/Bridges.jl [hunk 3]
```diff
@@ -23,17 +23,111 @@ include("debug.jl")
...
+those that are implemented in external packages (e.g., within the solver package
+itself) or only apply in certain circumstances (e.g.,
+[`Constraint.SOCtoNonConvexQuadBridge`](@ref)).
+
+Such optimizers should implement the `ListOfNonstandardBridges` attribute to
+return a vector of bridge types that are added by [`full_bridge_optimizer`](@ref)
+in addition to the list of default bridges.
+
+Note that optimizers implementing `ListOfNonstandardBridges` may be unusable
+if the non-standard bridges are not added.
```
odow, line 77: I know I wrote this, but it might be too strong language.

```
Note that optimizers implementing `ListOfNonstandardBridges` may require
package-specific functions or sets to be used if the non-standard bridges
are not added. Therefore, you are recommended to use
`model = MOI.instantiate(Package.Optimizer; with_bridge_type = T)` instead of 
`model = MOI.instantiate(Package.Optimizer)`.
```

odow (approved):
LGTM, other than the comment about language.

src/Bridges/Bridges.jl
blegat: Indeed, it's better to point to instantiate

PR #673, opened 2019-02-19
Fix minor typos in error.jl

Peiffap:


blegat:
Thanks for the fixes ! You now need to update the error messages in the tests, e.g.
https://github.com/JuliaOpt/MathOptInterface.jl/blob/05b79b7ca59049c930cfdd4476fa06ab0d447559/test/errors.jl#L40

Issue #655, opened 2019-02-02
`VariableBasisStatus` and `ConstraintBasisStatus` not supported


EdvinAblad:
While testing JuMP 0.19 beta2 I found that non of the solvers I tested (Clp, GLPK, Gurobi) supported the
`MOI.get(model, MOI.ConstraintBasisStatus(), index)` or `MOI.get(model, MOI.VariableBasisStatus(), index)`
where model is backend of the JuMP model created by the with_optimizer and no additional input. 
The model was a LP model.

I search the JuliaOpt Repo and found `ConstraintBasisStatus` mentioned only in three places: in attributes.jl, bridge.jl, and in apireference.md (all in the MOI project). 

To me this seems like no solver supports these functions: are they obsolete? or are there plans to implement them? 
(For Clp at least the corresponding functions exists in MathProgBase interface)

mlubin:
These are not implemented yet, and I don't consider them blocking for the 0.19 release. It's up for grabs by whoever needs this functionality. Given the existing MPB examples, it's probably not hard to implement the MOI versions for solvers of interest.

blegat:
Closing as it's now implemented in at least Clp, GLPK and Gurobi which are mentioned in the first comment.

PR #678, opened 2019-02-23
Add tests for get basis status for variables and constraints

EdvinAblad:
Some choices were made that might be incorrect:
basis::Bool = false, i.e., solvers needs to explicitly test this by setting basis=true.

Assume solver returns NONBASIC_AT_LOWER or NONBASIC_AT_UPPER if applicable. 
Currently NONBASIC is not tested. Should it occur only for fixed variables/equality constraints? Or is it simply a less specific attribute than NONBASIC_AT_X and can be returned by solvers lacking the support of the more specific type? 
If the first case is true it could be renamed to e.g., NONBASIC_FIXED
If the second case is true the tests should also accept NONBASIC.
related to #655 and  JuliaOpt/Clp.jl#54

### DIFF
--- a/src/Bridges/intervalbridge.jl
+++ b/src/Bridges/intervalbridge.jl
@@ -52,11 +52,11 @@ function MOI.get(model::MOI.ModelLike, ::MOI.ConstraintBasisStatus, c::SplitInte
     lower_stat = MOI.get(model, MOI.ConstraintBasisStatus(), c.lower)
     upper_stat = MOI.get(model, MOI.ConstraintBasisStatus(), c.upper)
     if lower_stat == MOI.NONBASIC_AT_LOWER
-        Compat.@warn("Single sided constraints should not have basis status:" *
+        Compat.@warn("GreaterThan constraints should not have basis status:" *
             " NONBASIC_AT_LOWER, instead use NONBASIC.")
     end
     if upper_stat == MOI.NONBASIC_AT_UPPER
-        Compat.@warn("Single sided constraints should not have basis status:" *
+        Compat.@warn("LessThan constraints should not have basis status:" *
             " NONBASIC_AT_UPPER, instead use NONBASIC.")
     end
     if lower_stat == MOI.NONBASIC
@@ -66,7 +66,7 @@ function MOI.get(model::MOI.ModelLike, ::MOI.ConstraintBasisStatus, c::SplitInte
         return MOI.NONBASIC_AT_UPPER
     end
     if lower_stat != upper_stat
-        Compat.@warn("Basis status of lower ($lower_stat) and upper ($upper_stat) constraint are inconsistent," *
+        Compat.@warn("Basis status of lower (`$lower_stat`) and upper (`$upper_stat`) constraint are inconsistent," *
             " both should be basic or super basic.")
     end
     return lower_stat
--- a/src/Bridges/intervalbridge.jl
+++ b/src/Bridges/intervalbridge.jl
@@ -48,6 +48,19 @@ function MOI.get(model::MOI.ModelLike, a::MOI.ConstraintDual, c::SplitIntervalBr
     return lower_dual > -upper_dual ? lower_dual : upper_dual
 end
 
+function MOI.get(model::MOI.ModelLike, ::MOI.ConstraintBasisStatus, c::SplitIntervalBridge)
+    lower_stat = MOI.get(model, MOI.ConstraintBasisStatus(), c.lower)
+    if lower_stat == MOI.NONBASIC || lower_stat == MOI.NONBASIC_AT_LOWER
+        return lower_stat
+    end
+    upper_stat = MOI.get(model, MOI.ConstraintBasisStatus(), c.upper)
+    if upper_stat == MOI.NONBASIC || upper_stat == MOI.NONBASIC_AT_UPPER
+        return upper_stat
+    end
+    # lower_stat == upper_stat should hold
+    return lower_stat
+end
+
 # Constraints
 function MOI.modify(model::MOI.ModelLike, c::SplitIntervalBridge, change::MOI.AbstractFunctionModification)
     MOI.modify(model, c.lower, change)
--- a/src/Bridges/intervalbridge.jl
+++ b/src/Bridges/intervalbridge.jl
@@ -51,11 +51,11 @@ end
 function MOI.get(model::MOI.ModelLike, ::MOI.ConstraintBasisStatus, c::SplitIntervalBridge)
     lower_stat = MOI.get(model, MOI.ConstraintBasisStatus(), c.lower)
     if lower_stat == MOI.NONBASIC || lower_stat == MOI.NONBASIC_AT_LOWER
-        return lower_stat
+        return MOI.NONBASIC_AT_LOWER
     end
     upper_stat = MOI.get(model, MOI.ConstraintBasisStatus(), c.upper)
     if upper_stat == MOI.NONBASIC || upper_stat == MOI.NONBASIC_AT_UPPER
-        return upper_stat
+        return MOI.NONBASIC_AT_UPPER
     end
     # lower_stat == upper_stat should hold
     return lower_stat
--- a/src/Bridges/intervalbridge.jl
+++ b/src/Bridges/intervalbridge.jl
@@ -50,14 +50,21 @@ end
 
 function MOI.get(model::MOI.ModelLike, ::MOI.ConstraintBasisStatus, c::SplitIntervalBridge)
     lower_stat = MOI.get(model, MOI.ConstraintBasisStatus(), c.lower)
-    if lower_stat == MOI.NONBASIC || lower_stat == MOI.NONBASIC_AT_LOWER
+    upper_stat = MOI.get(model, MOI.ConstraintBasisStatus(), c.upper)
+    if lower_stat == MOI.NONBASIC_AT_LOWER || upper_stat == MOI.NONBASIC_AT_UPPER
+        Compat.@warn("Single sided constraints should not have basis status:" *
+        " NONBASIC_AT_LOWER or NONBASIC_AT_UPPER, instead use NONBASIC")
+    end
+    if lower_stat == MOI.NONBASIC
         return MOI.NONBASIC_AT_LOWER
     end
-    upper_stat = MOI.get(model, MOI.ConstraintBasisStatus(), c.upper)
-    if upper_stat == MOI.NONBASIC || upper_stat == MOI.NONBASIC_AT_UPPER
+    if upper_stat == MOI.NONBASIC
         return MOI.NONBASIC_AT_UPPER
     end
-    # lower_stat == upper_stat should hold
+    if lower_stat != upper_stat
+        Compat.@warn("Basis status of lower and upper constraint are inconsistent, " *
+                    "both should be basic or super basic.")
+    end
     return lower_stat
 end
 
--- a/src/Bridges/intervalbridge.jl
+++ b/src/Bridges/intervalbridge.jl
@@ -51,9 +51,13 @@ end
 function MOI.get(model::MOI.ModelLike, ::MOI.ConstraintBasisStatus, c::SplitIntervalBridge)
     lower_stat = MOI.get(model, MOI.ConstraintBasisStatus(), c.lower)
     upper_stat = MOI.get(model, MOI.ConstraintBasisStatus(), c.upper)
-    if lower_stat == MOI.NONBASIC_AT_LOWER || upper_stat == MOI.NONBASIC_AT_UPPER
+    if lower_stat == MOI.NONBASIC_AT_LOWER
         Compat.@warn("Single sided constraints should not have basis status:" *
-        " NONBASIC_AT_LOWER or NONBASIC_AT_UPPER, instead use NONBASIC")
+            " NONBASIC_AT_LOWER, instead use NONBASIC.")
+    end
+    if upper_stat == MOI.NONBASIC_AT_UPPER
+        Compat.@warn("Single sided constraints should not have basis status:" *
+            " NONBASIC_AT_UPPER, instead use NONBASIC.")
     end
     if lower_stat == MOI.NONBASIC
         return MOI.NONBASIC_AT_LOWER
@@ -62,8 +66,8 @@ function MOI.get(model::MOI.ModelLike, ::MOI.ConstraintBasisStatus, c::SplitInte
         return MOI.NONBASIC_AT_UPPER
     end
     if lower_stat != upper_stat
-        Compat.@warn("Basis status of lower and upper constraint are inconsistent, " *
-                    "both should be basic or super basic.")
+        Compat.@warn("Basis status of lower ($lower_stat) and upper ($upper_stat) constraint are inconsistent," *
+            " both should be basic or super basic.")
     end
     return lower_stat
 end
--- a/src/Test/config.jl
+++ b/src/Test/config.jl
@@ -11,8 +11,8 @@ struct TestConfig
     optimal_status::MOI.TerminationStatusCode
     basis::Bool # can get variable and constraint basis status
     function TestConfig(;atol::Float64 = 1e-8, rtol::Float64 = 1e-8, solve::Bool = true, query::Bool = true,
-                        modify_lhs::Bool = true, duals::Bool = true, infeas_certificates::Bool =
-                        true, optimal_status = MOI.OPTIMAL, basis::Bool = false)
+                        modify_lhs::Bool = true, duals::Bool = true, infeas_certificates::Bool = true,
+                        optimal_status = MOI.OPTIMAL, basis::Bool = false)
         new(
             atol,
             rtol,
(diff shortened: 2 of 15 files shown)


mlubin:
Good point about the ambiguity. We need to clarify this in the docs. I would say that for one-sided constraints (i.e., with `LessThan`, `GreaterThan`, or `EqualTo` sets), the value should be either `BASIC` or `NONBASIC`.

The `NONBASIC_AT_LOWER` and `NONBASIC_AT_UPPER` statuses should be used *only* for two-sided constraints with the `Interval` set. In this case the distinction is necessary because either one of two bounds could be active.

What do you think?

EdvinAblad:
> 
> 
> Good point about the ambiguity. We need to clarify this in the docs. I would say that for one-sided constraints (i.e., with `LessThan`, or `GreaterThan`), the value should be either `BASIC` or `NONBASIC`. (If I remember simplex correctly, `EqualTo` sets should be always `BASIC`.)
> 
> The `NONBASIC_AT_LOWER` and `NONBASIC_AT_UPPER` statuses should be used _only_ for two-sided constraints with the `Interval` set. In this case the distinction is necessary because either one of two bounds could be active.
> 
> What do you think?

That would simplify usage, good point. (I make changes accordingly)

EqualTo constraint should imply the associated slack being zero, i.e., NONBASIC. (and NONBASIC_AT_UPPER/NONBASIC_AT_LOWER ;) )
Similarly for variables fixed (at bound) should imply that it's not in the basis.
(I think)

EdvinAblad:
@mlubin ok, change to use NONBASIC whenever its a single sided constraint/variable. 
And corresponding changes in JuliaOpt/Clp.jl#54, 
(all tests succeds on my machine at least).

src/Bridges/intervalbridge.jl
```diff
@@ -48,6 +48,19 @@ function MOI.get(model::MOI.ModelLike, a::MOI.ConstraintDual, c::SplitIntervalBr
     return lower_dual > -upper_dual ? lower_dual : upper_dual
 end
 
+function MOI.get(model::MOI.ModelLike, ::MOI.ConstraintBasisStatus, c::SplitIntervalBridge)
+    lower_stat = MOI.get(model, MOI.ConstraintBasisStatus(), c.lower)
+    if lower_stat == MOI.NONBASIC || lower_stat == MOI.NONBASIC_AT_LOWER
```
mlubin: `lower` is a `GreaterThan` constraint so should never return `NONBASIC_AT_LOWER`. It helps to check for this value but we should print a warning if we see it. Similar comment for `upper`.

src/Bridges/intervalbridge.jl
```diff
@@ -48,6 +48,19 @@ function MOI.get(model::MOI.ModelLike, a::MOI.ConstraintDual, c::SplitIntervalBr
...
+        return MOI.NONBASIC_AT_LOWER
+    end
+    upper_stat = MOI.get(model, MOI.ConstraintBasisStatus(), c.upper)
+    if upper_stat == MOI.NONBASIC || upper_stat == MOI.NONBASIC_AT_UPPER
+        return MOI.NONBASIC_AT_UPPER
+    end
+    # lower_stat == upper_stat should hold
```
mlubin: Maybe warn if they're not?

src/Test/config.jl
```diff
@@ -9,9 +9,10 @@ struct TestConfig
     # The expected "optimal" status returned by the solver. Either
     # MOI.OPTIMAL or MOI.LOCALLY_SOLVED.
     optimal_status::MOI.TerminationStatusCode
+    basis::Bool # can get variable and constraint basis status
     function TestConfig(;atol::Float64 = 1e-8, rtol::Float64 = 1e-8, solve::Bool = true, query::Bool = true,
                         modify_lhs::Bool = true, duals::Bool = true, infeas_certificates::Bool = true,
-                        optimal_status = MOI.OPTIMAL)
+                        optimal_status = MOI.OPTIMAL, basis::Bool = false)
```
mlubin: `false` is the correct default value to avoid breaking all the solvers that run tests.

src/Test/contlinear.jl
```diff
@@ -459,6 +464,9 @@ function linear3test(model::MOI.ModelLike, config::TestConfig)
         @test MOI.get(model, MOI.ObjectiveValue()) ≈ 3 atol=atol rtol=rtol
 
         @test MOI.get(model, MOI.VariablePrimal(), x) ≈ 3 atol=atol rtol=rtol
+        if config.basis
+            @test MOI.get(model, MOI.VariableBasisStatus(), x) == MOI.BASIC
```
mlubin: This should test also the status of the constraints.

src/Test/contlinear.jl
```diff
@@ -1086,6 +1097,13 @@ function linear9test(model::MOI.ModelLike, config::TestConfig)
         @test MOI.get(model, MOI.ObjectiveValue()) ≈ 79e4/11 atol=atol rtol=rtol
         @test MOI.get(model, MOI.VariablePrimal(), x) ≈ 650/11 atol=atol rtol=rtol
         @test MOI.get(model, MOI.VariablePrimal(), y) ≈ 400/11 atol=atol rtol=rtol
+        if config.basis
+            @test MOI.get(model, MOI.VariableBasisStatus(), x) == MOI.BASIC
+            @test MOI.get(model, MOI.VariableBasisStatus(), y) == MOI.BASIC
+            @test MOI.get(model, MOI.ConstraintBasisStatus(), c1[1]) == MOI.BASIC
+            @test MOI.get(model, MOI.ConstraintBasisStatus(), c23[1]) == MOI.NONBASIC
```
mlubin: Why not test all constraints?

src/Test/contlinear.jl
```diff
@@ -1135,6 +1153,11 @@ function linear10test(model::MOI.ModelLike, config::TestConfig)
             @test MOI.get(model, MOI.DualStatus()) == MOI.FEASIBLE_POINT
             @test MOI.get(model, MOI.ConstraintDual(), c) ≈ -1 atol=atol rtol=rtol
         end
+        if config.basis
+            @test (MOI.get(model, MOI.VariableBasisStatus(), x) == MOI.BASIC ||
```
mlubin: Is this because there are multiple optimal bases? Please add a comment explaining.

src/Test/contlinear.jl
```diff
@@ -411,6 +411,11 @@ function linear2test(model::MOI.ModelLike, config::TestConfig)
             @test MOI.get(model, MOI.ConstraintDual(), vc1) ≈ 0 atol=atol rtol=rtol
             @test MOI.get(model, MOI.ConstraintDual(), vc2) ≈ 1 atol=atol rtol=rtol
         end
+        if config.basis
```
mlubin: Nit: There's a newline before the `if config.duals` but not before `if config.basis`. I don't have a preference but this should be consistent at least.

EdvinAblad:
Made a draft on how basis status can be retrieved for single variable constraints.
I put the implementation in the solver see JuliaOpt/Clp.jl#54

In short, MOI.get(m, ConstraintBasisStatus(), var_constr_index)
returns MOI.get(m, VariableBasisStatus(), var_index)
But converting the output to use NONBASIC if var_constr is single sided.

Still think its a good point to keep VariableBasisStatus(), due to usability, but I have no strong opinion about it.

mlubin:
@EdvinAblad Do you mind rebasing the PR now that `VariableBasisStatus` is removed?

EdvinAblad:
> 
> 
> @EdvinAblad Do you mind rebasing the PR now that `VariableBasisStatus` is removed?

Will do, don't got time atm...

src/Bridges/intervalbridge.jl
```diff
@@ -48,6 +48,26 @@ function MOI.get(model::MOI.ModelLike, a::MOI.ConstraintDual, c::SplitIntervalBr
...
+    end
+    if lower_stat == MOI.NONBASIC
+        return MOI.NONBASIC_AT_LOWER
+    end
+    if upper_stat == MOI.NONBASIC
+        return MOI.NONBASIC_AT_UPPER
+    end
+    if lower_stat != upper_stat
+        Compat.@warn("Basis status of lower and upper constraint are inconsistent, " *
+                    "both should be basic or super basic.")
```
blegat: Missing space in indent. Print the values of `lower_stat` and `upper_stat` in message

src/Bridges/intervalbridge.jl
```diff
@@ -48,6 +48,26 @@ function MOI.get(model::MOI.ModelLike, a::MOI.ConstraintDual, c::SplitIntervalBr
     return lower_dual > -upper_dual ? lower_dual : upper_dual
 end
 
+function MOI.get(model::MOI.ModelLike, ::MOI.ConstraintBasisStatus, c::SplitIntervalBridge)
+    lower_stat = MOI.get(model, MOI.ConstraintBasisStatus(), c.lower)
+    upper_stat = MOI.get(model, MOI.ConstraintBasisStatus(), c.upper)
+    if lower_stat == MOI.NONBASIC_AT_LOWER || upper_stat == MOI.NONBASIC_AT_UPPER
+        Compat.@warn("Single sided constraints should not have basis status:" *
+        " NONBASIC_AT_LOWER or NONBASIC_AT_UPPER, instead use NONBASIC")
```
blegat: It's probably more helpful to have a different warning for each case so that the user sees in which case he is

EdvinAblad:
Added a suggestion incorporating ConstraintBasisStatus in the mockoptimizer.
To be complete it might need some rewording of mock_condual!() (or other modification), suggestions are welcome =)

src/Test/contlinear.jl
```diff
@@ -1153,6 +1184,13 @@ function linear10test(model::MOI.ModelLike, config::TestConfig)
             @test MOI.get(model, MOI.DualStatus()) == MOI.FEASIBLE_POINT
             @test MOI.get(model, MOI.ConstraintDual(), c) ≈ 1 atol=atol rtol=rtol
         end
+
+        if config.basis
+            # multiple optimal basis, x or y can in the optimal basis
```
mlubin: Nit: "There are multiple optimal bases. Either x or y can be in the optimal basis." (http://www.juliaopt.org/JuMP.jl/dev/style/#Comments-1)
Applies below also.

src/Utilities/mockoptimizer.jl
```diff
@@ -389,13 +395,14 @@ The primal values of the variables in the order returned by `ListOfVariableIndic
 If `termstatus` is missing, it is assumed to be `MOI.OPTIMAL`.
 If `primstatus` is missing, it is assumed to be `MOI.FEASIBLE_POINT`.
 If `dualstatus` is missing, it is assumed to be `MOI.FEASIBLE_POINT` if there is a primal solution and `primstatus` is not `MOI.INFEASIBLE_POINT`, otherwise it is `MOI.INFEASIBILITY_CERTIFICATE`.
+The bases status are set to the status specified by `conbasis`. Each pair is of the form `(F,S)=>[...]` where `[...]` is the the vector of basis status for the constraints `F`-in-`S` in the order returned by `ListOfConstraintIndices{F,S}`.
 The dual values are set to the values specified by `conduals`. Each pair is of the form `(F,S)=>[...]` where `[...]` is the the vector of dual values for the constraints `F`-in-`S` in the order returned by `ListOfConstraintIndices{F,S}`.
 """
-function mock_optimize!(mock::MockOptimizer, termstatus::MOI.TerminationStatusCode, primal, dual...)
+function mock_optimize!(mock::MockOptimizer, termstatus::MOI.TerminationStatusCode, primal, basis_dual...)
```
mlubin: What about using a keyword argument to specify basis statuses? Mixing duals and basis values in one argument is confusing.

src/Utilities/mockoptimizer.jl
```diff
@@ -41,6 +41,7 @@ mutable struct MockOptimizer{MT<:MOI.ModelLike} <: MOI.AbstractOptimizer
     # constraints having the variable in the function. See `get_fallback`.
     eval_variable_constraint_dual::Bool
     condual::Dict{MOI.ConstraintIndex,Any}
+    conbasis::Dict{MOI.ConstraintIndex,MOI.BasisStatusCode}
```
mlubin: Nit: I understand this is consistent with the surrounding code, but the JuMP style guide recommends using underscores to separate words in a name (http://www.juliaopt.org/JuMP.jl/dev/style/#Naming-1), so `con_basis`. We'll eventually get back to this code and fix the other names, but when making additions we should follow the style guide.

src/Utilities/mockoptimizer.jl
```diff
@@ -389,13 +395,14 @@ The primal values of the variables in the order returned by `ListOfVariableIndic
 If `termstatus` is missing, it is assumed to be `MOI.OPTIMAL`.
 If `primstatus` is missing, it is assumed to be `MOI.FEASIBLE_POINT`.
 If `dualstatus` is missing, it is assumed to be `MOI.FEASIBLE_POINT` if there is a primal solution and `primstatus` is not `MOI.INFEASIBLE_POINT`, otherwise it is `MOI.INFEASIBILITY_CERTIFICATE`.
+The bases status are set to the status specified by `conbasis`. Each pair is of the form `(F,S)=>[...]` where `[...]` is the the vector of basis status for the constraints `F`-in-`S` in the order returned by `ListOfConstraintIndices{F,S}`.
```
mlubin: The function signature in the docstring above doesn't include `conbasis`, so what does this refer to?

src/Bridges/intervalbridge.jl
```diff
@@ -48,6 +48,30 @@ function MOI.get(model::MOI.ModelLike, a::MOI.ConstraintDual, c::SplitIntervalBr
     return lower_dual > -upper_dual ? lower_dual : upper_dual
 end
 
+function MOI.get(model::MOI.ModelLike, ::MOI.ConstraintBasisStatus, c::SplitIntervalBridge)
+    lower_stat = MOI.get(model, MOI.ConstraintBasisStatus(), c.lower)
+    upper_stat = MOI.get(model, MOI.ConstraintBasisStatus(), c.upper)
+    if lower_stat == MOI.NONBASIC_AT_LOWER
+        Compat.@warn("Single sided constraints should not have basis status:" *
```
blegat: I would say `GreaterThan` instead of "Single sided". If a solver relies on the Interval bridge it means that it does not support Interval so you are not thinking about two-sided vs one-sided when you are implementing MOi for this solver, you are maybe not even aware that two-sided constraints exist

src/Bridges/intervalbridge.jl
```diff
@@ -48,6 +48,30 @@ function MOI.get(model::MOI.ModelLike, a::MOI.ConstraintDual, c::SplitIntervalBr
...
+            " NONBASIC_AT_UPPER, instead use NONBASIC.")
+    end
+    if lower_stat == MOI.NONBASIC
+        return MOI.NONBASIC_AT_LOWER
+    end
+    if upper_stat == MOI.NONBASIC
+        return MOI.NONBASIC_AT_UPPER
+    end
+    if lower_stat != upper_stat
+        Compat.@warn("Basis status of lower ($lower_stat) and upper ($upper_stat) constraint are inconsistent," *
```
blegat: Use backsticks around lower_stat and upper_stat since it's Julia constant, not english words

EdvinAblad:
I introduced a new test with an inactive range constraint (its `ConstraintBasisStatus` is `BASIC`) 
The test is a minor modification of linear10, thus I named it linear10b.

src/Utilities/mockoptimizer.jl
```diff
@@ -278,6 +282,7 @@ function MOI.empty!(mock::MockOptimizer)
     mock.dualstatus = MOI.NO_SOLUTION
     mock.varprimal = Dict{MOI.VariableIndex,Float64}()
     mock.condual = Dict{MOI.ConstraintIndex,Any}()
+    mock.con_basis = Dict{MOI.ConstraintIndex,MOI.BasisStatusCode}()
     return
 end
```
blegat: You also need to modify `is_empty` below

src/Utilities/mockoptimizer.jl
```diff
@@ -390,15 +396,19 @@ If `termstatus` is missing, it is assumed to be `MOI.OPTIMAL`.
...
     MOI.set(mock, MOI.ResultCount(), 1)
     mock_primal!(mock, primal)
     mock_dual!(mock, dual...)
+    for con_basis_pair in con_basis
+        mock_basis_status!(mock, con_basis_pair)
+    end
 end
 # Default termination status
-mock_optimize!(mock::MockOptimizer, primdual...) = mock_optimize!(mock, MOI.OPTIMAL, primdual...)
+mock_optimize!(mock::MockOptimizer, primal, dual...; con_basis = []) = mock_optimize!(mock, MOI.OPTIMAL, primal, con_basis = con_basis, dual...)
```
blegat: Why not simply do `mock_optimize!(mock::MockOptimizer, primdual...; kws...) = mock_optimize!(mock, MOI.OPTIMAL, primdual...; kws...)` ?

src/Test/contlinear.jl
```diff
@@ -1180,6 +1225,70 @@ function linear10test(model::MOI.ModelLike, config::TestConfig)
...
+end
+
+# inactive ranged constraints
+function linear10btest(model::MOI.ModelLike, config::TestConfig)
+    atol = config.atol
+    rtol = config.rtol
+    #   minimize x + y
+    #
+    #       s.t.  -1 <= x + y <= 10
+    #                  x,  y >= 0
```
Peiffap:
```suggestion
    #                   x,  y >= 0
```

Issue #570, opened 2018-11-13
Indices of SingleVariable constraints


mlubin:
I propose that the constraint index of a single variable constraint should always match the index of the corresponding variable. That is,
```julia
x = add_variable(model)
ci = add_constraint(model, SingleVariable(x), ZeroOne())
# Then ci.value must equal x.value.
```
With this change, the question "Does `x` have a lower bound?" becomes `is_valid(ConstraintIndex{SingleVariable, GreaterThan{Float64}}(x.value))`, removing the need for JuMP to keep dictionaries (https://github.com/JuliaOpt/JuMP.jl/issues/1607#issuecomment-438219901).

It would also encourage implementations to store SingleVariable constraints in parallel arrays with the variables, which is the sane thing to do and will probably recover most of the performance from 0.18 (https://github.com/JuliaOpt/JuMP.jl/issues/1607).

This would disallow multiple SingleVariable constraints of the same type for a given variable, but that's an error on most solvers anyway.

odow:
This seems like a reasonable proposal.

blegat:
One issue is that you can no longer have integer variables with bounds. Or we move integrality to an attribute ?

mlubin:
> One issue is that you can no longer have integer variables with bounds. Or we move integrality to an attribute ?

Why not? `ConstraintIndex{SingleVariable, GreaterThan{Float64}}(10)` and `ConstraintIndex{SingleVariable, Integer}(10)` refer to different constraints.

blegat:
That would work. Do we want to make this change before JuMP v0.19 ? It would mean tagging MOI v0.7.

blegat:
No, the JuMP API shouldn't be affected

blegat:
I have changed my mind about this. This change allows to store the `SingleVariable` constraints in a `Vector` instead of `Dict`. While this makes adding `SingleVariable` constraints faster, it increases the time of `add_variable` and the memory footprint of each variable.
Indeed, the `Vector` needs to have the same size than the number of variables. If we have one `Vector` for each `MOI.AbstractScalarSet`, that will be a lot of memory for each variables.

mlubin:
I disagree that memory use is an issue with this design. JuMP 0.18 stores the lower bound, upper bound, and category for each variable, so we shouldn't be worried about storing 3 values per variable. I claim that with 2 `Float64`s and 1 `UInt64` we can store all that we care about.

[...]

PR #679, opened 2019-02-24
Remove VariableBasisStatus and clarify when NONBASIC_AT_* should be used

mlubin:
Ref https://github.com/JuliaOpt/MathOptInterface.jl/pull/678#issuecomment-466648971
CC @EdvinAblad

### DIFF
--- a/docs/src/apireference.md
+++ b/docs/src/apireference.md
@@ -133,8 +133,9 @@ ResultStatusCode
 
 ### Basis Status
 
-The `BasisStatus` attribute of a variable or constraint describes its status with respect to a basis, if one is known.
-The value of the attribute is of type `BasisStatusCode`.
+The `BasisStatus` attribute of a constraint describes its status with respect to
+a basis, if one is known. The value of the attribute is of type
+`BasisStatusCode`.
 
 ```@docs
 BasisStatusCode
@@ -165,7 +166,6 @@ Calls to `get` and `set` should include as an argument a single `VariableIndex` 
 VariableName
 VariablePrimalStart
 VariablePrimal
-VariableBasisStatus
 ```
 
 ### Constraints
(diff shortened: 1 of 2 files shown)


mlubin:
I'm looking at this again and I'm questioning why we have a `VariableBasisStatus()` attribute. Shouldn't this be the `BasisStatus()` on the single-variable constraints instead? As it is currently, this is the only place in MOI where "variable bounds" need to be discussed as a concept, which seems inconsistent.

EdvinAblad:
Consistency is important, however, I think that the most common use case is for a user to query whether a variable is basic or non-basic, not if a constraint is...

Is it possible (and sane) to have both? 
E.g., calling MOI.get(m, MOI.BasisStatus(), var_index), is basically an indirect call to MOI.get(m, MOI.BasisStatus(), var_constr). (where var_constr is the applicable one)

Also I'm not quite sure how these single variable constraints works if they are called multiple times, e.g. x => 0, x >= 3, x =>2. In JuMP they seems to be available from set_lower_bound, in LQOI adding a single-variable bound also (seems to) replace the previous ones. Is it possible for multiple variable bounds to be present? If so, the basis status of the dominated constraints should always return MOI.BASIC (where should this domination check be done, LQOI?).

blegat:
> Is it possible for multiple variable bounds to be present?

No, see https://github.com/JuliaOpt/MathOptInterface.jl/issues/570 The `LowerBoundRef` gives in JuMP gives you the reference of the last bound set

mlubin:
I updated the PR to remove `VariableBasisStatus`. It's not only inconsistent with how we report dual solutions, it's also redundant because you can ask for the `ConstraintBasisStatus` of a `SingleVariable` constraint anyway. Given that there's a direct mapping between the two, we shouldn't force solver interfaces to implement both. Convenience functions can be added at the JuMP level like we do for lower and upper bounds.

@odow @joaquimg

PR #689, opened 2019-03-04
Fix LazyBridgeOptimizer with bridge that add no constraints

blegat:
See https://github.com/JuliaOpt/SumOfSquares.jl/blob/586b55eccd4e63afa45f2f9d6875e13c401a7f5b/src/empty_bridge.jl#L16-L17

### DIFF
--- a/src/Bridges/lazybridgeoptimizer.jl
+++ b/src/Bridges/lazybridgeoptimizer.jl
@@ -45,7 +45,9 @@ function update_dist!(b::LazyBridgeOptimizer, constraints)
             for (F, S) in constraints
                 if MOI.supports_constraint(BT, F, S) && all(C -> supports_constraint_no_update(b, C[1], C[2]), added_constraint_types(BT, F, S))
                     # Number of bridges needed using BT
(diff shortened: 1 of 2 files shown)


src/Bridges/lazybridgeoptimizer.jl
```diff
@@ -35,6 +35,9 @@ function _dist(b::LazyBridgeOptimizer, F::Type{<:MOI.AbstractFunction}, S::Type{
     end
 end
 
+# Compute the sum of `Int`.
+int_sum(f::Function, x) = isempty(x) ? 0 : sum(f, x)
```
mlubin: This naming is confusing because nothing in the code implies that the function works only for `Int`s. What about using `reduce(+,  _dist(b, C[1], C[2]) for C in added_constraint_types(BT, F, S), init=0)` instead of defining a helper function?
(Compat provides this version of `reduce` in 0.6.)

blegat: It didn't work on Julia v0.6 but I didn't with the Compat prefix

Issue #663, opened 2019-02-05
Make LogLevel a required solver option


ndinsmore:
Across the Julia ecosystem LogLevel is fairly standard as the verbosity identifying option. 

By making this part of the interface it would remove a lot of the difficulty in controlling the output to STDOUT.

odow:
We previously had two solver-independent parameters:
`silent` and `time_limit`
https://mathprogbasejl.readthedocs.io/en/latest/solverinterface.html?highlight=timelimit#setparameters!

See also https://github.com/JuliaOpt/MathOptInterface.jl/issues/61

blegat:
> Across the Julia ecosystem LogLevel is fairly standard as the verbosity identifying option.

What are the packages sharing the LogLevel convention ? What would be the difference between LogLevel 1 and 2 ?

PR #695, opened 2019-03-11
Add Silent and implement setting optimizer attributes in caching and mock optimizers

blegat:
- [x] Needs tests

Closes https://github.com/JuliaOpt/MathOptInterface.jl/issues/663

### DIFF
--- a/src/Utilities/cachingoptimizer.jl
+++ b/src/Utilities/cachingoptimizer.jl
@@ -506,7 +506,6 @@ function MOI.get(m::CachingOptimizer, IdxT::Type{<:MOI.Index}, name::String)
     return MOI.get(m.model_cache, IdxT, name)
 end
 
-# TODO: MOI.set for MOI.AbstractOptimizerAttribute.
 function MOI.set(model::CachingOptimizer, attr::MOI.AbstractOptimizerAttribute,
                  value)
     optimizer_value = attribute_value_map(model.model_to_optimizer_map, value)
--- a/src/Utilities/cachingoptimizer.jl
+++ b/src/Utilities/cachingoptimizer.jl
@@ -517,12 +517,13 @@ end
 
 function MOI.get(model::CachingOptimizer, attr::MOI.AbstractOptimizerAttribute)
     if state(model) == NO_OPTIMIZER
+        # TODO: Copyable attributes (e.g., Silent, TimeLimit) are also be stored
+        # in the cache so we could return the value stored in the cache instead.
+        # However, for non-copyable attributes( e.g. `SolverName`) the error is
+        # appropriate.
         error("Cannot query $(attr) from caching optimizer because no " *
               "optimizer is attached.")
     end
-    # TODO: Copyable attributes (e.g., TimeLimit) could also be stored in the
-    # cache. When MOI.set is implemented for MOI.AbstractOptimizerAttribute,
-    # make sure this case is handled correctly.
     return attribute_value_map(model.optimizer_to_model_map,
                                MOI.get(model.optimizer, attr))
 end
--- a/src/Utilities/cachingoptimizer.jl
+++ b/src/Utilities/cachingoptimizer.jl
@@ -80,6 +80,11 @@ function reset_optimizer(m::CachingOptimizer, optimizer::MOI.AbstractOptimizer)
     @assert MOI.is_empty(optimizer)
     m.optimizer = optimizer
     m.state = EMPTY_OPTIMIZER
+    for attr in MOI.get(m.model_cache, MOI.ListOfOptimizerAttributesSet())
+        value = MOI.get(m.model_cache, attr)
+        optimizer_value = attribute_value_map(m.model_to_optimizer_map, value)
+        MOI.set(m.optimizer, attr, optimizer_value)
+    end
     return
 end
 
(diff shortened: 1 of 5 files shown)


src/attributes.jl
```diff
@@ -337,6 +337,17 @@ An optimizer attribute for the string identifying the solver/optimizer.
 """
 struct SolverName <: AbstractOptimizerAttribute end
 
+"""
+    LogLevel(value::Int)
+
+An optimizer attribute for level of verbosity of the optimizer. `0` means
```
mlubin: Which option has supremacy if a user sets log level here and also a solver-dependent version?

src/attributes.jl
```diff
@@ -337,6 +337,17 @@ An optimizer attribute for the string identifying the solver/optimizer.
...
+no output and for positive values, the higher the values the more output is
+allowed.
```
mlubin: Say explicitly that the meaning of the output levels is solver-specific except for `0`?

src/Utilities/cachingoptimizer.jl
```diff
@@ -502,6 +507,14 @@ function MOI.get(m::CachingOptimizer, IdxT::Type{<:MOI.Index}, name::String)
 end
 
 # TODO: MOI.set for MOI.AbstractOptimizerAttribute.
```
mlubin: Remove the TODO?

src/attributes.jl
blegat: Good question, the one set last I'd say.

src/attributes.jl
```diff
@@ -337,6 +337,18 @@ An optimizer attribute for the string identifying the solver/optimizer.
 """
 struct SolverName <: AbstractOptimizerAttribute end
 
+"""
+    LogLevel(value::Int)
+
+An optimizer attribute for level of verbosity of the optimizer. `0` means
+no output and the meaning is solver specific for positive values.
+If a solver specific attribute controlling the output level is also set, the
+last attributes set gets supremacy over the other one.
```
mlubin: How does this rule interact with the caching optimizer? The order that the user set the attributes might not match the order that they're copied to the optimizer.

blegat: Good point. It's kind of weird for caching optimizer to store RawParameter though

src/attributes.jl
```diff
@@ -337,6 +337,18 @@ An optimizer attribute for the string identifying the solver/optimizer.
 """
 struct SolverName <: AbstractOptimizerAttribute end
 
+"""
+    LogLevel(value::Int)
+
+An optimizer attribute for level of verbosity of the optimizer. `0` means
+no output and the meaning is solver specific for positive values.
```
odow: Before we just had `silent`. It was unambiguous. Is that better than a weird solver-specific parameter?

blegat: Good point, I am not sure what's best. The idea of `LogLevel` is that most solver have this functionality, just with a different name. I usually check the MOI tests of the solver to see which name is the log-level.

src/attributes.jl
mlubin: > It's kind of weird for caching optimizer to store RawParameter though

I don't think it's weird. Won't this be necessary for users to pass RawParameters from JuMP to solvers in general?

src/attributes.jl
mlubin: `Silent` addresses the question of "how do I turn the solver output off?", but not "how do I turn solver output on?". `LogLevel` covers both.

odow: `Silent(false)`. I'm fine with loglevel though.

mlubin: What's the definition of `Silent(false)`? Would it turn on output or just not silence it?

src/Utilities/cachingoptimizer.jl
```diff
@@ -501,16 +506,24 @@ function MOI.get(m::CachingOptimizer, IdxT::Type{<:MOI.Index}, name::String)
...
+    optimizer_value = attribute_value_map(model.model_to_optimizer_map, value)
+    if model.optimizer !== nothing
+        MOI.set(model.optimizer, attr, optimizer_value)
+    end
+    MOI.set(model.model_cache, attr, value)
+end
 
 function MOI.get(model::CachingOptimizer, attr::MOI.AbstractOptimizerAttribute)
     if state(model) == NO_OPTIMIZER
+        # TODO: Copyable attributes (e.g., Silent, TimeLimit) are also be stored
```
mlubin: s/are also be/should also be/

src/attributes.jl
```diff
@@ -337,6 +337,27 @@ An optimizer attribute for the string identifying the solver/optimizer.
 """
 struct SolverName <: AbstractOptimizerAttribute end
 
+"""
+    Silent
+
+An optimizer attribute for silencing the output of an optimizer. When `set`
+to `true`, it takes precedence over any other attribute controlling verbosity
+and requires the solver to produce no output. The default value is `false`
+which has no effect, the verbosity is controlled by other attributes.
```
mlubin: Run-on sentence:
The default value is `false` which has no effect. In this case the verbosity is controlled by other attributes.

src/attributes.jl
```diff
@@ -337,6 +337,27 @@ An optimizer attribute for the string identifying the solver/optimizer.
...
+
+
+## Note
+
+Every optimizer should have verbosity on by default. For instance, if a solver
+has a solver-specific attribute `LogLevel` which is `0` by default, the MOI
```
mlubin: `LogLevel` is not the actual name of an attribute, so this could be a bit confusing. I would just use English, "if a solver has a solver-specific log level attribute, the MOI implementation should set it to 1 by default." ...

Issue #106, opened 2017-08-03
Where should be caching?


IssamT:
I agree that the current strategy of JuMP is useful but I don't like the idea of cache being only in JuMP... I understand and accept the argument of having MOI a pure API but It would be nice to have the MOIU instance shared by JuMP and the solver wrapper. Example: if a solver wrapper wants to answer the query about MOI.ConstraintSet it will need to store it's own copy or build the set on the fly which would be a waste since the set exists and is stored in JuMP.

The problem with sharing is that I don't know if it is possible with julia to deny the solvers from modifying the instance once it is shared.

One possible solution: given a valid MOIU instance there are queries that do not need a solver to be answered. so wrappers do not implement those methods and we move them to MOIU. Everything using MOIU instance goes to MOIU to keep it separated from what solvers should implement. Then I see methods in MOIU like

````julia
MOIU.addconstraint!(m::MOI.AbstractSolverInstance, func::F, set::S, <:MOIU.AbstractInstance{T}, updateform::Bool)
#MOIU update its instance and if updateform == false
#MOIU keeps track of what is not yet passed through vectors for each combination F,S as discussed
#here https://github.com/JuliaOpt/MathOptInterface.jl/issues/63#issuecomment-317526848 

MOIU.update_solver_instance() 
#calls MOI.addconstraints! and deletes the vectors but keeps the instance.

MOIU.getattribute(m::MOIU.AbstractInstance{T}, ::MOI.ConstraintSet}, ref)
#replacing what was before in MOI and implemented by the solvers 
#MOI.getattribute(m::MOI.SolverInstance{T}, ::MOI.ConstraintSet}, ref)
````

As @blegat mentioned, this will not work in the following case:
if the solver wrapper wants to have a mode on the fly where every info required for a single constraint is added and deleted right afterwards (being only kept inside the solver). 

Such solver wrappers will be denied to have an answer to the following query since now it s the responsability of MOIU instance. 

````julia
MOI.getattribute(m::MOI.SolverInstance{T}, ::MOI.ConstraintSet}, ref)
````

To solve this issue, solvers need to have access to the instance (bad thing). and all those methods will not be replaced but only get a default implementation that uses MOIU instance (when available):

So Instead of this

````julia
MOIU.getattribute(m::MOIU.AbstractInstance{T}, ::MOI.ConstraintSet, ref)
# replacing what was before in MOI and implemented by the solvers
[...]

Issue #529, opened 2018-09-27
Objective function bridge


blegat:
It would be nice to have a bridge that transforms an unsupported objective into a SingleVariable objective with a LessThan (or GreaterThan depending on objective sense) constraint.
See https://github.com/JuliaOpt/SCS.jl/pull/111#issuecomment-425214874

blegat:
It could also bridge SingleVariable to ScalarAffineFunction, it would resolve https://github.com/JuliaOpt/Cbc.jl/issues/96

leethargo:
Nice, we have the same issue in [SCIP.jl](https://github.com/SCIP-Interfaces/SCIP.jl/issues/80#issuecomment-453662154).

blegat:
The short term solution is simply converting it to `MOI.ScalarAffineFunction` like what is done in https://github.com/JuliaOpt/SCS.jl/pull/116. `MOI.ObjectiveFunctionType` is wrong though for SCS since it does not remember that it was set as `MOI.SingleVariable` but that is acceptable as a short term solution.

arnavs:
I just wanted to follow up on this issue and see if there's been any movement. As it stands currently, my Mosek license is basically unusable in Julia, so it would be great to merge stuff even if it's a bit clunky.

odow:
> As it stands currently, my Mosek license is basically unusable in Julia

It looks like you successfully followed the instructions here https://github.com/JuliaOpt/MosekTools.jl/issues/2? 

> so it would be great to merge stuff even if it's a bit clunky.

This isn't implemented yet. If you would like to have a go, we can point you in the right direction.

arnavs:
Yeah, the issue with that workaround is that it only holds for problems with all positive eigenvalues, as @blegat points out. 

Yeah, would be happy to. If you could briefly describe what needs to be done, I'll give it a shot.

odow:
> Yeah, the issue with that workaround is that it only holds for problems with all positive eigenvalues

Mosek is a convex solver. If the matrix isn't PSD (i.e., it has negative eigenvalues), Mosek can't solve it.

arnavs:
Just to be clear, the problem we're discussing is a quadratic objective `z' L z + z' q` for some vectors `z, q`, and some matrix `L`, with (in my setup) zero or one negative eigenvalues. 

I thought I was able to solve this problem before, but maybe we used different parameters. Maybe I was hallucinating.

PR #697, opened 2019-03-13
Add tests, extend docstring for ScalarQuadraticFunction.

leethargo:
We just discovered a bug in SCIP.jl ([#106](https://github.com/SCIP-Interfaces/SCIP.jl/issues/106)) in the handling of quadratic coefficients off the diagonal, based on my misunderstanding of the MOI docs.

To add prevent this kind of problem, I've added a small example to the docstring and two QCP tests.

### DIFF
--- a/src/Test/contquadratic.jl
+++ b/src/Test/contquadratic.jl
@@ -385,9 +385,61 @@ function qcp3test(model::MOI.ModelLike, config::TestConfig)
     end
 end
 
+function qcp4test(model::MOI.ModelLike, config::TestConfig)
+    atol = config.atol
+    rtol = config.rtol
+    # Max 2x + y
+    # s.t. x*y <= 4 (c)
+    #      x, y >= 1
+
+    @test MOIU.supports_default_copy_to(model, #=copy_names=# false)
+    @test MOI.supports(model, MOI.ObjectiveFunction{MOI.ScalarAffineFunction{Float64}}())
+    @test MOI.supports_constraint(model, MOI.ScalarQuadraticFunction{Float64},MOI.LessThan{Float64})
+
+    MOI.empty!(model)
+    @test MOI.is_empty(model)
+
+    x = MOI.add_variable(model)
+    y = MOI.add_variable(model)
+    @test MOI.get(model, MOI.NumberOfVariables()) == 2
+
+    MOI.add_constraint(model, MOI.SingleVariable(x), MOI.GreaterThan(1.0))
+    MOI.add_constraint(model, MOI.SingleVariable(y), MOI.GreaterThan(1.0))
+
+    cf = MOI.ScalarQuadraticFunction([MOI.ScalarAffineTerm(0.0, x)], [MOI.ScalarQuadraticTerm(1.0, x, y)], 0.0)
+    c = MOI.add_constraint(model, cf, MOI.LessThan(4.0))
+    @test MOI.get(model, MOI.NumberOfConstraints{MOI.ScalarQuadraticFunction{Float64}, MOI.LessThan{Float64}}()) == 1
+
+    MOI.set(model, MOI.ObjectiveFunction{MOI.ScalarAffineFunction{Float64}}(), MOI.ScalarAffineFunction(MOI.ScalarAffineTerm.([2.0, 1.0], [x, y]), 0.0))
+    MOI.set(model, MOI.ObjectiveSense(), MOI.MAX_SENSE)
+    @test MOI.get(model, MOI.ObjectiveSense()) == MOI.MAX_SENSE
+
+    if config.query
+        @test cf ≈ MOI.get(model, MOI.ConstraintFunction(), c)
+    end
+
+    if config.solve
+        @test MOI.get(model, MOI.TerminationStatus()) == MOI.OPTIMIZE_NOT_CALLED
+
+        MOI.optimize!(model)
+
+        @test MOI.get(model, MOI.TerminationStatus()) == config.optimal_status
+
+        @test MOI.get(model, MOI.PrimalStatus()) == MOI.FEASIBLE_POINT
+
+        @test MOI.get(model, MOI.ObjectiveValue()) ≈ 9.0 atol=atol rtol=rtol
+
+        @test MOI.get(model, MOI.VariablePrimal(), x) ≈ 4.0 atol=atol rtol=rtol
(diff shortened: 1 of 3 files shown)


leethargo:
By the way, I don't know the process for adding tests to MOI. Should these pass with some subset of solvers? Is this considered a breaking change that requires a major version bump?

blegat:
It's not considered a breaking change. The solvers should be passing the new tests, otherwise it means they have a bug.

leethargo:
> The solvers should be passing the new tests, otherwise it means they have a bug.

That makes sense, but it also requires that tests should be highly configurable in terms of what the solvers support. For example, SCIP.jl excludes one test of the QCP suite, because it requires `VectorAffineFunctions`. If somebody added another test of that kind, it would fail on SCIP.jl, but I would not consider that as evidence for a bug.

blegat:
> For example, SCIP.jl excludes one test of the QCP suite, because it requires `VectorAffineFunctions`

 It requires `VectorAffineFunction`s instead of what ?

leethargo:
> It requires `VectorAffineFunction`s instead of what ?

The other QCP tests are written with `ScalarAffineFunction` (and `ScalarQuadraticFunction`).
So, a bridged SCIP solver could support them, I guess.

blegat:
Yes, you could use the `ScalarizeBridge`

src/Test/contquadratic.jl
```diff
@@ -385,9 +385,113 @@ function qcp3test(model::MOI.ModelLike, config::TestConfig)
...
+function qcp5test(model::MOI.ModelLike, config::TestConfig)
+    atol = config.atol
+    rtol = config.rtol
+    # Find x,y
+    # s.t. x*y == 4 (c)
+    #      x*x == 4 (c2)
+    #      x, y >= 0
+
+    @test MOIU.supports_default_copy_to(model, #=copy_names=# false)
+    @test MOI.supports_constraint(model, MOI.ScalarQuadraticFunction{Float64},MOI.EqualTo{Float64})
```
blegat: Missing a space after the comma

src/Test/contquadratic.jl
```diff
@@ -385,9 +385,113 @@ function qcp3test(model::MOI.ModelLike, config::TestConfig)
...
+function qcp4test(model::MOI.ModelLike, config::TestConfig)
+    atol = config.atol
+    rtol = config.rtol
+    # Max 2x + y
+    # s.t. x*y <= 4 (c)
+    #      x, y >= 1
+
+    @test MOIU.supports_default_copy_to(model, #=copy_names=# false)
+    @test MOI.supports(model, MOI.ObjectiveFunction{MOI.ScalarAffineFunction{Float64}}())
+    @test MOI.supports_constraint(model, MOI.ScalarQuadraticFunction{Float64},MOI.LessThan{Float64})
```
blegat: Missing space after comma

leethargo:
Sorry, this is kind of off-topic here, but let's clarify it anyways:

> Yes, you could use the `ScalarizeBridge`

And we do: [SCIP/test/MOI_wrapper.j](https://github.com/SCIP-Interfaces/SCIP.jl/blob/master/test/MOI_wrapper.jl#L46-L55).
I found that there is value in testing the directly supported features without any bridges.

Also, even using the `ScalarizeBridge` will not make these tests future-proof: Maybe somebody will add more tests that use quadratic objective functions which would not be bridged properly, so the new tests would fail in SCIP.jl.

src/Test/contquadratic.jl
leethargo: Fixed (within my patch only).

blegat:
> I found that there is value in testing the directly supported features without any bridges.

Indeed, if you do that, you might be easily be broken by new tests. With the addition of Scalarize, Functionize, FlipSign and Slack bridges, solvers can have a very limited set of supported constraints and may not pass many tests without bridges. I would recommend running all tests with the optimizer bridged with `full_bridge_optimizer` as I just did in https://github.com/JuliaOpt/SCS.jl/pull/142
Making a breaking release everytime we add a test wouldn't be practical.
If you want to run tests without bridges, I would suggest calling a specific test instead of running a test-set with an exclude list.

> Maybe somebody will add more tests that use quadratic objective functions which would not be bridged properly, so the new tests would fail in SCIP.jl.

This should be resolved by https://github.com/JuliaOpt/MathOptInterface.jl/issues/529 so it shouldn't weight in the long term decision on whether new tests mean breaking change :-P

leethargo:
> If you want to run tests without bridges, I would suggest calling a specific test instead of running a test-set with an exclude list.

That's a good suggestion which I will implement now :+1:

src/Test/contquadratic.jl
```diff
@@ -385,9 +385,113 @@ function qcp3test(model::MOI.ModelLike, config::TestConfig)
     end
 end
 
+function qcp4test(model::MOI.ModelLike, config::TestConfig)
+    atol = config.atol
+    rtol = config.rtol
+    # Max 2x + y
+    # s.t. x*y <= 4 (c)
```
blegat: Need space after and before `*`

src/Test/contquadratic.jl
```diff
@@ -385,9 +385,113 @@ function qcp3test(model::MOI.ModelLike, config::TestConfig)
...
+        @test MOI.get(model, MOI.ConstraintPrimal(), c) ≈ 4.0 atol=atol rtol=rtol
+    end
+end
+
+function qcp5test(model::MOI.ModelLike, config::TestConfig)
+    atol = config.atol
+    rtol = config.rtol
+    # Find x,y
+    # s.t. x*y == 4 (c)
+    #      x*x == 4 (c2)
```
blegat: Same here

src/Test/contquadratic.jl
leethargo: Done.

blegat:
These two tests do not correspond to convex qcp for qcp4, it is quadratic-in-lessthan with a non-psd matrix and for qcp5 it is quadratic-in-equalto. We should probably make a different testset for non-convex or a sub-testset so that SOC solvers can easily exclude them all.

leethargo:
Fair enough, it was not obvious that the QCP testset should be limited to convex problems. Actually, the intent of these cases was not to test nonconvex constraints, but quadratic constraints with off-diagonal nonzeros, because they are treated differently w.r.t. to the scaling factor 2. So maybe it would be better to turn them into convex problems instead.

Issue #684, opened 2019-03-04
Add support for starting values in bridges


blegat:
Updated list

| Constraint Bridge | ConstraintPrimal | ConstraintDual |
| ----------------- | ---------------- | -------------- |
| GreaterToIntervalBridge | x | x |
| LessToIntervalBridge    | x | x |
| GreaterToLessBridge     | x | x |
| LessToGreaterBridge     | x | x |
| NonnegToNonposBridge | x | x |
| NonposToNonnegBridge | x | x |
| VectorizeBridge       | x | x |
| ScalarizeBridge       | x | x |
| ScalarSlackBridge     | x | x |
| VectorSlackBridge     | x | x |
| ScalarFunctionizeBridge | x | x |
| VectorFunctionizeBridge | x | x |
| SplitComplexEqualToBridge | x | x |
| SplitComplexZerosBridge | x | x |
| SplitIntervalBridge | x | x |
| SOCtoRSOCBridge | x | x |
| RSOCtoSOCBridge | x | x |
| SOCtoNonConvexQuadBridge | NA | NA |
| RSOCtoNonConvexQuadBridge | NA | NA |
| QuadtoSOCBridge | NA | NA |
| SOCtoPSDBridge | | |
| RSOCtoPSDBridge | | |
| NormInfinityBridge | x | x |
| NormOneBridge | x | x |
| GeoMeantoRelEntrBridge | x | x |
| GeoMeanBridge | | |
| RelativeEntropyBridge | x | x |
| NormSpectralBridge | | |
| NormNuclearBridge | x | x |
| SquareBridge | | |
| RootDetBridge | | |
| LogDetBridge | x | x |
| IndicatorActiveOnFalseBridge | NA | NA |
| IndicatorSOS1Bridge | NA | NA |
| SemiToBinaryBridge | NA | NA |
| ZeroOneBridge | NA | NA |
| AllDifferentToCountDistinctBridge | NA | NA |
| BinPackingToMILPBridge | NA | NA |
| CircuitToMILPBridge | NA | NA |
| CountAtLeastToCountBelongsBridge | NA | NA |
| CountBelongsToMILPBridge | NA | NA |
| CountDistinctToMILPBridge | NA | NA |
| CountGreaterThanToMILPBridge | NA | NA |
| TableToMILPBridge | NA | NA |

| Variable Bridge       | VariablePrimal | ConstraintPrimal | ConstraintDual |
[...]

PR #699, opened 2019-03-19
overload MOI.set & supports for ConstraintAttribute in bridgeoptimizer

guimarqu:
Hi, this PR follows my message in gitter. 

I'm working on a package that creates variable/constraint attributes in the JuMP model. Then, another package retrieves these attributes through the MOI interface. In the second package, I overloaded `MOI.set` and `MOI.get` to copy/get variable & constraint attributes. 

These are the definitions of MOI.set & MOI.get in my packages :
```julia
MOI.get(dest::MOIU.UniversalFallback, attr::BD.ConstraintDecomposition, ci::MOI.ConstraintIndex)
MOI.set(dest::MOIU.UniversalFallback, attr::BD.ConstraintDecomposition, ci::MOI.ConstraintIndex, annotation)
```

During the copy process, `MOI.set` of `VariableAttribute` receives a `MOIU.CachingOptimizer`as first argument whereas `MOI.set` of `ConstraintAttribute` receives a `MOIU.Bridges.LazyBridgeOptimizer`. Since LazyBridge is not a subtype of UniversalFallback, MOI throws an error. Looking in file bridgeoptimizer.jl, the following functions are defined for `VariableAttribute` but not for `ConstraintAttribute`.

```julia
function MOI.set(b::AbstractBridgeOptimizer,
                  attr::MOI.AbstractVariableAttribute,
                  index::MOI.Index, value)
    return MOI.set(b.model, attr, index, value)
end
function MOI.set(b::AbstractBridgeOptimizer,
                  attr::MOI.AbstractVariableAttribute,
                  indices::Vector{<:MOI.Index}, values::Vector)
    return MOI.set(b.model, attr, indices, values)
end
```

I overloaded `MOI.set` for `ConstraintAttributes`in bridge optimizer to pass the `MOIU.CachingOptimizer` as first argument.

I also overloaded `MOI.support` because there is no definition of it for `ConstraintAttribute`. However, I'm not sure of it is needed and I don't know how to test it. (I can get the attributes without overloading supports in my packages).

I can retrieve constraints attributes with these changes but is this the reason of my bug?

### DIFF
--- a/src/Bridges/bridgeoptimizer.jl
+++ b/src/Bridges/bridgeoptimizer.jl
@@ -262,7 +262,7 @@ function MOI.get(b::AbstractBridgeOptimizer, attr::MOI.ConstraintName,
 end
 
 function MOI.supports(b::AbstractBridgeOptimizer, attr::MOI.ConstraintName,
-                      Index::Type{<:CI})
+                      Index::Type{CI{F,S}}) where {F,S}
     if is_bridged(b, Index)
         return MOI.supports(b.bridged, attr, Index)
     else
--- a/src/Bridges/bridgeoptimizer.jl
+++ b/src/Bridges/bridgeoptimizer.jl
@@ -227,25 +227,16 @@ function MOI.get(b::AbstractBridgeOptimizer,
                  attr::MOI.AbstractConstraintAttribute,
                  ci::CI)
     if is_bridged(b, typeof(ci))
-        if attr isa MOI.ConstraintName || attr isa MOI.ConstraintFunction ||
-                attr isa MOI.ConstraintSet
-            MOI.get(b.bridged, attr, ci)
-        else
-            MOI.get(b, attr, bridge(b, ci))
-        end
+        return MOI.get(b, attr, bridge(b, ci))
     else
-        MOI.get(b.model, attr, ci)
+        return MOI.get(b.model, attr, ci)
     end
 end
 function MOI.supports(b::AbstractBridgeOptimizer,
                       attr::MOI.AbstractConstraintAttribute,
                       IndexType::Type{CI{F, S}}) where {F,S}
     if is_bridged(b, IndexType)
-        if attr isa MOI.ConstraintName
-            return MOI.supports(b.bridged, attr, IndexType)
-        else
-            return MOI.supports(b, attr, concrete_bridge_type(b, F, S))
-        end
+        return MOI.supports(b, attr, concrete_bridge_type(b, F, S))
     else
         return MOI.supports(b.model, attr, IndexType)
     end
@@ -255,32 +246,53 @@ function MOI.set(b::AbstractBridgeOptimizer,
                  attr::MOI.AbstractConstraintAttribute,
                  index::CI, value)
     if is_bridged(b, typeof(index))
-        if attr isa MOI.ConstraintName
-            return MOI.set(b.bridged, attr, index, value)
-        else
-            return MOI.set(b, attr, bridge(b, index), value)
-        end
+        return MOI.set(b, attr, bridge(b, index), value)
     else
         return MOI.set(b.model, attr, index, value)
     end
 end
-## Setting names
-# function MOI.supports(b::AbstractBridgeOptimizer, attr::MOI.ConstraintName,
-#                       Index::Type{<:CI})
-#     if is_bridged(b, Index)
-#         return MOI.supports(b.bridged, attr, Index)
-#     else
-#         return MOI.supports(b.model, attr, Index)
-#     end
-# end
-# function MOI.set(b::AbstractBridgeOptimizer, attr::MOI.ConstraintName,
-#                   constraint_index::CI, name::String)
-#     if is_bridged(b, typeof(constraint_index))
-#         MOI.set(b.bridged, attr, constraint_index, name)
-#     else
-#         MOI.set(b.model, attr, constraint_index, name)
-#     end
-# end
+## Getting and Setting names
+function MOI.get(b::AbstractBridgeOptimizer, attr::MOI.ConstraintName,
+                 constraint_index::CI)
+    if is_bridged(b, typeof(constraint_index))
+        return MOI.get(b.bridged, attr, constraint_index)
+    else
+        return MOI.get(b.model, attr, constraint_index)
+    end
+end
+
+function MOI.supports(b::AbstractBridgeOptimizer, attr::MOI.ConstraintName,
+                      Index::Type{CI{F, S}}) where {F,S}
+    if is_bridged(b, Index)
+        return MOI.supports(b.bridged, attr, Index)
+    else
+        return MOI.supports(b.model, attr, Index)
+    end
+end
+function MOI.set(b::AbstractBridgeOptimizer, attr::MOI.ConstraintName,
+                  constraint_index::CI, name::String)
+    if is_bridged(b, typeof(constraint_index))
+        MOI.set(b.bridged, attr, constraint_index, name)
+    else
+        MOI.set(b.model, attr, constraint_index, name)
+    end
+end
+## Getting functions and sets
+function MOI.get(b::AbstractBridgeOptimizer, attr::MOI.ConstraintSet, ci::CI)
+    if is_bridged(b, typeof(ci))
+        return MOI.get(b.bridged, attr, ci)
+    else
+        return MOI.get(b.model, attr, ci)
+    end
+end
+function MOI.get(b::AbstractBridgeOptimizer, attr::MOI.ConstraintFunction, 
+        ci::CI)
+    if is_bridged(b, typeof(ci))
+        return MOI.get(b.bridged, attr, ci)
+    else
+        return MOI.get(b.model, attr, ci)
+    end
+end    
 ## Setting functions and sets
 function MOI.set(b::AbstractBridgeOptimizer, ::MOI.ConstraintSet,
                   constraint_index::CI{F, S}, set::S) where {F, S}
--- a/src/Bridges/bridgeoptimizer.jl
+++ b/src/Bridges/bridgeoptimizer.jl
@@ -262,7 +262,7 @@ function MOI.get(b::AbstractBridgeOptimizer, attr::MOI.ConstraintName,
 end
 
 function MOI.supports(b::AbstractBridgeOptimizer, attr::MOI.ConstraintName,
-                      Index::Type{CI{F, S}}) where {F,S}
+                      Index::Type{<:CI})
     if is_bridged(b, Index)
         return MOI.supports(b.bridged, attr, Index)
     else
@@ -286,7 +286,7 @@ function MOI.get(b::AbstractBridgeOptimizer, attr::MOI.ConstraintSet, ci::CI)
     end
 end
 function MOI.get(b::AbstractBridgeOptimizer, attr::MOI.ConstraintFunction, 
-        ci::CI)
+                 ci::CI)
     if is_bridged(b, typeof(ci))
         return MOI.get(b.bridged, attr, ci)
     else
--- a/src/Bridges/bridgeoptimizer.jl
+++ b/src/Bridges/bridgeoptimizer.jl
@@ -222,15 +222,21 @@ function MOI.set(b::AbstractBridgeOptimizer,
     return MOI.set(b.model, attr, indices, values)
 end
 
+# function MOI.get(::MOI.ModelLike, attr::MOI.AbstractConstraintAttribute,
+#     bridge::AbstractBridge)
+# throw(ArgumentError("Constraint bridge of type `$(typeof(bridge))` does not support accessing the attribute `$attr`."))
+# end
+
 # Constraint attributes
 function MOI.get(b::AbstractBridgeOptimizer,
                  attr::MOI.AbstractConstraintAttribute,
                  ci::CI)
     if is_bridged(b, typeof(ci))
-        if MOI.is_set_by_optimize(attr)
-            MOI.get(b, attr, bridge(b, ci))
-        else
+        if attr isa MOI.ConstraintName || attr isa MOI.ConstraintFunction ||
+                attr isa MOI.ConstraintSet
             MOI.get(b.bridged, attr, ci)
+        else
+            MOI.get(b, attr, bridge(b, ci))
         end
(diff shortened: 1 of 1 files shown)


blegat:
These two methods are indeed missing, thanks for the PR! However, the implementation you suggest will only work if the constraint is not bridged, the implementation should branch on whether it is bridged or not, see, e.g.:
https://github.com/JuliaOpt/MathOptInterface.jl/blob/f18603addbcab092aeda76df1ff9112a1d69e6c2/src/Bridges/bridgeoptimizer.jl#L225-L238
To test the implementation, you could implement starting values for one of the bridges (see https://github.com/JuliaOpt/MathOptInterface.jl/issues/684) and test that `get`, `set` and `supports` work correctly for starting values of constraints bridges using this bridge.
Note that `MOI.get` should be modified as
https://github.com/JuliaOpt/MathOptInterface.jl/blob/f18603addbcab092aeda76df1ff9112a1d69e6c2/src/Bridges/bridgeoptimizer.jl#L233
should only be called for names.
https://github.com/JuliaOpt/MathOptInterface.jl/blob/f18603addbcab092aeda76df1ff9112a1d69e6c2/src/Bridges/bridgeoptimizer.jl#L230
should be replaced by `if !(attr isa MOI.ConstraintName)` (note that the `bridged` field is going away with https://github.com/JuliaOpt/MathOptInterface.jl/issues/523).
All constraint attributes (except constraint names of bridged constraints) should be transferred to `b.model` either directly if the constraint is not bridged or indirectly if it is bridged (similarly to what is done with `ConstraintPrimal` and `ConstraintDual`).

guimarqu:
So now I have to implement starting values for constraint bridges and the test should be something like 
```julia
MOI.set(mock_optimizer, ConstraintPrimalStart(), index_of_a_bridged_constraint, value)
```
and when I'll do
```julia
MOI.get(mock_optimizer, ConstraintPrimalStart(), index_of_a_bridged_constraint)
```
I should be able to retrieve `value`.

Did I understand ? Where should I do the tests ?

src/Bridges/bridgeoptimizer.jl
```diff
@@ -239,12 +239,25 @@ end
...
+        return MOI.supports(b.model, attr, IndexType)
+    end
 end
+
 function MOI.set(b::AbstractBridgeOptimizer,
                  attr::MOI.AbstractConstraintAttribute,
                  index::MOI.Index, value)
-    return MOI.set(b.model, attr, index, value)
+    if is_bridged(b, typeof(index))
+        if MOI.is_set_by_optimize(attr)
```
blegat: This should be `if !(attr isa MOI.ConstraintName)` (also modify in `MOI.get`)

src/Bridges/bridgeoptimizer.jl
```diff
@@ -239,12 +239,25 @@ end
 function MOI.supports(b::AbstractBridgeOptimizer,
                       attr::MOI.AbstractConstraintAttribute,
                       IndexType::Type{<:MOI.Index})
-    return MOI.supports(b.model, attr, IndexType)
+    if is_bridged(b, IndexType)
+        return MOI.supports(b.bridged, attr, IndexType)
```
blegat: This is only if `attr isa MOI.ConstraintName`, otherwise, it should be `MOI.supports(b, attr, concrete_bridge_type(b, F, S))`

src/Bridges/bridgeoptimizer.jl
```diff
@@ -239,12 +239,25 @@ end
 function MOI.supports(b::AbstractBridgeOptimizer,
                       attr::MOI.AbstractConstraintAttribute,
                       IndexType::Type{<:MOI.Index})
```
blegat: Use `Type{MOI.ConstraintIndex{F, S}}`

src/Bridges/bridgeoptimizer.jl
```diff
@@ -239,12 +239,25 @@ end
...
+    if is_bridged(b, IndexType)
+        return MOI.supports(b.bridged, attr, IndexType)
+    else
+        return MOI.supports(b.model, attr, IndexType)
+    end
 end
+
 function MOI.set(b::AbstractBridgeOptimizer,
                  attr::MOI.AbstractConstraintAttribute,
                  index::MOI.Index, value)
```
blegat: Use `::MOI.ConstraintIndex` to avoid method ambiguity

blegat:
You can for instance implement it to `FlipSign` and then add a testset here:
https://github.com/JuliaOpt/MathOptInterface.jl/blob/f18603addbcab092aeda76df1ff9112a1d69e6c2/test/bridge.jl#L409
Trying `set` and `get` looks good, you should also `get` it from the `mock` so check that is was not just set to the `bridged` field but it was transmitted to the solver. Also check `supports`

src/Bridges/bridgeoptimizer.jl
guimarqu: `MOI.get` throws an error when I try to retrieve attributes `ConstraintFunction` and `ConstraintSet`. Following code works : 
```julia
    if is_bridged(b, typeof(ci))
        if attr isa MOI.ConstraintName || attr isa MOI.ConstraintFunction ||
                attr isa MOI.ConstraintSet
            MOI.get(b.bridged, attr, ci)
        else
            MOI.get(b, attr, bridge(b, ci))
        end
    else #...
```

src/Bridges/bridgeoptimizer.jl
guimarqu: Done

src/Bridges/bridgeoptimizer.jl
guimarqu: Done

src/Bridges/bridgeoptimizer.jl
```diff
@@ -222,37 +222,72 @@ function MOI.set(b::AbstractBridgeOptimizer,
     return MOI.set(b.model, attr, indices, values)
 end
 
+# function MOI.get(::MOI.ModelLike, attr::MOI.AbstractConstraintAttribute,
```
guimarqu: To be deleted (I forgot to remove this comment)

src/Bridges/bridgeoptimizer.jl
guimarqu: Error message I got when I run the tests
> ArgumentError: Constraint bridge of type `MathOptInterface.Bridges.SplitIntervalBridge{Float64,MathOptInterface.ScalarAffineFunction{Float64}}` does not support accessing the attribute `MathOptInterface.ConstraintSet()`.

src/Bridges/bridgeoptimizer.jl
```diff
@@ -222,37 +222,72 @@ function MOI.set(b::AbstractBridgeOptimizer,
...
+
 # Constraint attributes
 function MOI.get(b::AbstractBridgeOptimizer,
                  attr::MOI.AbstractConstraintAttribute,
                  ci::CI)
     if is_bridged(b, typeof(ci))
-        if MOI.is_set_by_optimize(attr)
-            MOI.get(b, attr, bridge(b, ci))
-        else
+        if attr isa MOI.ConstraintName || attr isa MOI.ConstraintFunction ||
```
blegat: Indeed, ConstraintFunction and ConstraintSet currently need to be obtained from the `bridged` model (this is what will change with light bridges, then the bridges will have to implement these getters).

src/Bridges/bridgeoptimizer.jl
```diff
@@ -222,37 +222,72 @@ function MOI.set(b::AbstractBridgeOptimizer,
...
-## Setting names
-function MOI.supports(b::AbstractBridgeOptimizer, attr::MOI.ConstraintName,
-                      Index::Type{<:CI})
-    if is_bridged(b, Index)
-        return MOI.supports(b.bridged, attr, Index)
+function MOI.supports(b::AbstractBridgeOptimizer,
+                      attr::MOI.AbstractConstraintAttribute,
+                      IndexType::Type{CI{F, S}}) where {F,S}
+    if is_bridged(b, IndexType)
+        if attr isa MOI.ConstraintName || attr isa MOI.ConstraintFunction ||
```
blegat: `ConstraintFunction` and `ConstraintSet` are not copyable (see `is_copyable`) so `supports(model, MOI.ConstraintFunction(), ...)` is not defined. We should not worry about these two attributes here

src/Bridges/bridgeoptimizer.jl
```diff
@@ -222,37 +222,72 @@ function MOI.set(b::AbstractBridgeOptimizer,
...
-function MOI.set(b::AbstractBridgeOptimizer, attr::MOI.ConstraintName,
-                  constraint_index::CI, name::String)
-    if is_bridged(b, typeof(constraint_index))
-        MOI.set(b.bridged, attr, constraint_index, name)
+
+function MOI.set(b::AbstractBridgeOptimizer,
+                 attr::MOI.AbstractConstraintAttribute,
+                 index::CI, value)
+    if is_bridged(b, typeof(index))
+        if attr isa MOI.ConstraintName || attr isa MOI.ConstraintFunction ||
```
blegat: For `ConstraintFunction` and `ConstraintSet`, they should be set both to `bridged` and to the bridge.
With the current implementation, the user modifies the constraint but the modification is not reflected to the solver model.

src/Bridges/bridgeoptimizer.jl
guimarqu: Ok so I leave this condition.

src/Bridges/bridgeoptimizer.jl
guimarqu: I removed `ConstraintFunction` and `ConstraintSet` from the condition.

src/Bridges/bridgeoptimizer.jl
guimarqu: I removed `ConstraintFunction` and `ConstraintSet` from the condition. There were useless because there are specific definitions of `MOI.set` for these attributes.

src/Bridges/bridgeoptimizer.jl
```diff
@@ -227,32 +227,60 @@ function MOI.get(b::AbstractBridgeOptimizer,
...
+## Setting names
+# function MOI.supports(b::AbstractBridgeOptimizer, attr::MOI.ConstraintName,
+#                       Index::Type{<:CI})
+#     if is_bridged(b, Index)
+#         return MOI.supports(b.bridged, attr, Index)
+#     else
+#         return MOI.supports(b.model, attr, Index)
+#     end
+# end
+# function MOI.set(b::AbstractBridgeOptimizer, attr::MOI.ConstraintName,
```
blegat: It might be more efficient to re-enable these methods and letting multiple dispatch do the work instead of using `isa`. in the generic `set` and `supports`

src/Bridges/bridgeoptimizer.jl
```diff
@@ -227,32 +227,60 @@ function MOI.get(b::AbstractBridgeOptimizer,
                  attr::MOI.AbstractConstraintAttribute,
                  ci::CI)
     if is_bridged(b, typeof(ci))
-        if MOI.is_set_by_optimize(attr)
-            MOI.get(b, attr, bridge(b, ci))
-        else
+        if attr isa MOI.ConstraintName || attr isa MOI.ConstraintFunction ||
```
blegat: For consistency with `set` and `supports`, it might be better to create a specific method for `ConstraintName` and a specific method for `ConstraintFunction`/`ConstraintSet`.

src/Bridges/bridgeoptimizer.jl
```diff
@@ -227,18 +227,42 @@ function MOI.get(b::AbstractBridgeOptimizer,
...
+    if is_bridged(b, typeof(constraint_index))
+        return MOI.get(b.bridged, attr, constraint_index)
+    else
+        return MOI.get(b.model, attr, constraint_index)
+    end
+end
+
 function MOI.supports(b::AbstractBridgeOptimizer, attr::MOI.ConstraintName,
-                      Index::Type{<:CI})
+                      Index::Type{CI{F, S}}) where {F,S}
```
blegat: You can leave `<:CI`

src/Bridges/bridgeoptimizer.jl
```diff
@@ -253,6 +277,22 @@ function MOI.set(b::AbstractBridgeOptimizer, attr::MOI.ConstraintName,
...
+## Getting functions and sets
+function MOI.get(b::AbstractBridgeOptimizer, attr::MOI.ConstraintSet, ci::CI)
+    if is_bridged(b, typeof(ci))
+        return MOI.get(b.bridged, attr, ci)
+    else
+        return MOI.get(b.model, attr, ci)
+    end
+end
+function MOI.get(b::AbstractBridgeOptimizer, attr::MOI.ConstraintFunction, 
+        ci::CI)
```
blegat: indent issue

src/Bridges/bridgeoptimizer.jl
guimarqu: Julia throws an "ambiguous name" method error if I leave `<:CI`.

src/Bridges/bridgeoptimizer.jl
guimarqu: Done

guimarqu:
Hi,

I try writing a basic test : 

```julia
struct FlipSign <: MOI.AbstractConstraintAttribute end

MOI.set(bridged_mock, FlipSign(), ci, true)
@test MOI.get(bridged_mock, FlipSign(), ci)
```
The constraint is bridged so it calls `MOI.set(bridged_mock, FlipSign(), bridge(bridged_mock, ci), true)`.

But `bridge(bridged_mock, ci)` returns a  `MathOptInterface.Bridges.GreaterToLessBridge{Float64,MathOptInterface.ScalarAffineFunction{Float64}}`. 

I'm stuck because I'm not sure what is going on here. What should I do with this constraint bridge? Should I overload MOI.set for all types of constraint bridge?

blegat:
Yes an set method should be added for every bridge.
For this PR I suggest to implement ConstraintPrimalStart and ConstraintDualStart for one bridge only. This is enough to test the changes of this PR

blegat:
Let's add the tests in a separate PR, thanks for the contribution !

PR #701, opened 2019-03-29
Add missing sets for functionize bridges

blegat:


blegat:
Rendered useless by https://github.com/JuliaOpt/MathOptInterface.jl/pull/722

PR #702, opened 2019-03-30
Only ask for dual infeasibility certificate if duals is true

blegat:
I was under the impression that `config.infeas_certificates` means that we can expect `INFEASIBILITY_CERTIFICATE` both for `DUAL_INFEASIBLE` and `INFEASIBLE`.
So for a solver that support infeasibility certificates for `DUAL_INFEASIBLE` but not for `INFEASIBLE`, it should just set `duals` to `false` and the tests shouldn't expect infeasibility certificates for `INFEASIBLE`.
What do you think ?

### DIFF
--- a/src/Test/config.jl
+++ b/src/Test/config.jl
@@ -5,7 +5,7 @@ struct TestConfig
     query::Bool # can get objective function, and constraint functions, and constraint sets
     modify_lhs::Bool # can modify function of a constraint
     duals::Bool # test dual solutions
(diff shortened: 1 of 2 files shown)


mlubin:
If we're changing or clarifying the meaning of `infeas_certificates`, then we should update the comments at https://github.com/JuliaOpt/MathOptInterface.jl/blob/f18603addbcab092aeda76df1ff9112a1d69e6c2/src/Test/config.jl#L8.

Issue #547, opened 2018-10-26
consistency of MOI cone definitions


chriscoey:
to make it easier to remember the order of the variables in the various cone definitions, I think we should follow a convention of (epi/hypograph, perspective, argument(s)). note that a epi/hypograph variable and/or a perspective variable may not be present. 

all the cones already follow this convention (though it's not immediately obvious from their names/definitions)
- SOC: x >= norm2(z)
- RSOC: x >= y*sumabs2(z/y)/2 (equivalent to current definition but maybe a little clearer)
- PSD: 0 <= eigmin(Z), Z symmetric
- Power: x >= y*abs(z/y)^q (equivalent to current definition where q = 1/exponent is positive)
- Exponential: x <= y*log(z/y), z >= 9
- RootDet: x <= (det(Z))^1/n, Z PSD
- LogDet: x <= y*log(det(Z/y)), Z PSD (see https://github.com/JuliaOpt/MathOptInterface.jl/issues/546)
- Geomean: x <= prod(z)^1/n, z >= 0 (perhaps replace with generalized version below)
- generalized Geomean: x <= prod_i z_i^alpha_i, z >= 0 (for alpha on unit simplex; sometimes called power cone in literature, with x or abs(x) on LHS)

so to make things clear I would recommend
- noting the (epi/hypograph, perspective, argument(s)) ordering convention in the documentation
- changing docs to use the less arbitrary descriptions above
- potentially changing the names or adding name aliases for some cones, like `MOI.EpiPerspSquare = MOI.RotatedSecondOrderCone` (since this cone is the epigraph of the perspective of the 1/2 square function) and `MOI.HypoPerspLog = MOI.ExponentialCone` (since it is the (closure of the) hypograph of the perspective of the log function)

I am happy to make these non-breaking changes.

chriscoey:
another advantage is that these definitions make it easier to derive the dual cones of whatever cones you define, via the construction on p19 of http://www.isye.umn.edu/courses/ie8534/pdf/Lecture-4.pdf

furthermore, these cone definitions are more intuitive from a modeling perspective.

chriscoey:
Bit late for this to be feasible, so closing

PR #709, opened 2019-04-11
Indicator constraint support

matbesancon:
First draft for the addition of indicator constraints to MOI. There is evidence for multiple solvers handling them:
[Gurobi](http://www.gurobi.com/documentation/7.5/refman/constraints.html)
[CPLEX](https://www.ibm.com/support/knowledgecenter/SSSA5P_12.8.0/ilog.odms.cplex.help/CPLEX/UsrMan/topics/discr_optim/indicator_constr/04_Interactive.html)
[SCIP](https://scip.zib.de/doc/html/cons__indicator_8h.php)

One thing we would want to add (suggestion from @leethargo) is a bridge for solvers supporting SOS1 but not indicator constraints natively. From SCIP documentation:
> This constraint is equivalent to a linear constraint ax−s≤b and an SOS1 constraint on z and s (at most one should be nonzero). In the indicator context we can, however, separate more inequalities.


src/sets.jl
```diff
@@ -438,6 +438,18 @@ Base.isapprox(a::SOS2{T}, b::SOS2{T}; kwargs...) where T = isapprox(a.weights, b
...
+"""
+Set of (x,y) that satisfy the indicator constraint:
+    y ∈ 𝔹, y = 1 ==> a^T x ∈ S
+
+S can be one of `SmallerThan, GreaterThan, EqualTo`
+"""
+struct IndicatorSet{T <: Real, S <:Union{SmallerThan,GreaterThan,EqualTo}} <: MOI.AbstractVectorSet
+    a::Vector{T}
+    s::S
+end
```
odow: Another approach that might work is
```julia
"""
    IndicatorSet(y::VariableIndex, set::AbstractSet)

The set `{y \in {0, 1}, y=1 ==> S, y=0 ==> R^n}`.
"""
struct IndicatorSet{S}
    y::MOI.VariableIndex
    set::S
end
```
Then solvers could implement
```julia
add_constraint(::ModelLike, ::ScalarAffineFunction, ::IndicatorSet)
```

src/sets.jl
```diff
@@ -438,6 +438,18 @@ Base.isapprox(a::SOS2{T}, b::SOS2{T}; kwargs...) where T = isapprox(a.weights, b
 
 dimension(s::Union{SOS1, SOS2}) = length(s.weights)
 
+"""
+Set of (x,y) that satisfy the indicator constraint:
+    y ∈ 𝔹, y = 1 ==> a^T x ∈ S
+
+S can be one of `SmallerThan, GreaterThan, EqualTo`
```
odow: s/SmallerThan/LessThan

Documentation nits: 
 - use proper punctuation (e.g., end sentences with a period)
 - begin docstrings with the name of the type.
 - use backticks for things like `S`

src/sets.jl
matbesancon: Ok I see, I had a mental model of the set being independent from the variables.
Another one I was thinking of was using an `IndicatorFunction` with one index for y and the `ScalarAffineTerms` forx

odow: Although `VectorOfVariables-in-IndicatorSet` is also good. Should `y` already be binary? Or is it the solvers job to add this constraint when an indicator is applied.

odow: > I had a mental model of the set being independent from the variables

I'm not sure if we ever settled on this.

matbesancon: the pain with the set-based design and a vector function is that the user must remember to put the binary first, far from ideal

odow: It is somewhat consistent with the cones requiring the user to put the epigraph variable first: https://github.com/JuliaOpt/MathOptInterface.jl/issues/547

leethargo: I think the current proposal is more easily generalized to a disjunction of linear inequalities, that is `ax ≤ b v cx ≤ d`, where `a` and `c`  with `b` and `d` could be stored in the `Set` while using a `Vector` function.

robertschwarz:
What about having a set parameter for whether `y` should be `0` or `1` to activate the constraint?

Then the disjunction `ax ≤ b v cx ≤ d` could be expressed by the two constraints

    (y, x) in IndicatorSet(1, a, LessThan(b))
    (y, x) in IndicatorSet(0, c, LessThan(d))

reusing the `y` variable. Otherwise, another `z` would have to be introduced with `y + z == 1`?

src/sets.jl
```diff
@@ -438,6 +438,23 @@ Base.isapprox(a::SOS2{T}, b::SOS2{T}; kwargs...) where T = isapprox(a.weights, b
 
 dimension(s::Union{SOS1, SOS2}) = length(s.weights)
 
+"""
+    IndicatorSet{T <: Real, S <:Union{SmallerThan,GreaterThan,EqualTo}}
+
+Set of (x,y) that satisfy the indicator constraint:
+    y ∈ 𝔹, y = 1 ==> a^T x ∈ S
+
+`S` can be one of `LessThan`, `GreaterThan` or `EqualTo`
```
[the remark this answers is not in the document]
matbesancon: I was taking the strict thing supported by solver, I think other constraints will require a bridge

src/sets.jl
```diff
@@ -449,7 +466,7 @@ function Base.copy(set::Union{Reals, Zeros, Nonnegatives, Nonpositives,
                               PositiveSemidefiniteConeSquare,
                               LogDetConeTriangle, LogDetConeSquare,
                               RootDetConeTriangle, RootDetConeSquare,
-                              Integer, ZeroOne, Semicontinuous, Semiinteger})
+                              Integer, ZeroOne, Semicontinuous, Semiinteger, IndicatorSet})
```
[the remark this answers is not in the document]
matbesancon: You're right, I'll re-implement copy. Should we make the vector type generic?

matbesancon:
Should we make the indicator set generic on vector type?
`IndicatorSet{T,S <: ScalarSet,V <: AbstractVector{<:Real}}`

mlubin:
What are the implications of having non `isbits` sets? How does this interact with MOF?

One way to avoid storing the `a` vector in the set would be to store it in the function, e.g.,
```
Set of (y, x) that satisfy the indicator constraint:
    y ∈ 𝔹, y = 1 ==> x ∈ S
```
Then use VAF to write down the constraints, e.g.,
```julia
(1.0 * y, 2*x + 3*z) in IndicatorSet(1, a, LessThan(b))
```
It's an error if the first component of the VAF isn't 1.0 times some variable.

odow:
The SOS sets already have weight vectors, so this isn't breaking new ground. No problem on the MOF front

matbesancon:
see last version for generic vector

odow:
Oh, wait. I misread miles' concern. Not sure how abstractvector would play with MOF. I'd prefer the vectoraffinefunction in set approach.

matbesancon:
problem with `VectorAffine in Indicator` is that it requires `Set` to hold variables. Also keeping a vector of `a` is consistent with `SOS`

mlubin:
> VectorAffine in Indicator is that it requires Set to hold variable

What do you mean by this?

matbesancon:
Either `VectorAffine` has a weird special first term `(1.0, y)` or the `y` needs to be specified in the set

mlubin:
The weird special term is fine with me. That's what we do for cones.

We have machinery already to translate JuMP AffExprs into VAF, so today you could write:
```julia
@constraint(model, [y, 2x + 3z] in IndicatorSet(LessThan(b))
```
OTOH,
```julia
@constraint(model, [y, x, z] in IndicatorSet([2.0, 3.0], LessThan(b))
```
seems like a regression in modeling because you have to manually create a list of coefficients.

matbesancon:
Ok I see. `[y, 2x + 3z]` would be specially parsed at the JuMP level? (Meaning, different from `[y, 2x, 3z]` or `[y + 2x + 3z]`

mlubin:
> [y, 2x + 3z] would be specially parsed at the JuMP level? (Meaning, different from [y, 2x, 3z] or [y + 2x + 3z]

Yes, those are all parsed differently.

matbesancon:
one thing is that the set cannot determine the dimension of the expected function

mlubin:
The dimension of the set I'm proposing (under the MOI definition) would always be 2.

matbesancon:
ok I hadn't thought of counting like this, so the whole affine term `a^T x` count as dim 1 right?

odow:
What about 
```julia
@constraint(model, 2x1 + x2 in MOI.IndicatorSet(y=>1, MOI.GreaterThan(1.0)))
```
So
```julia
struct IndicatorSet{S}
    variable::Pair{MOI.VariableIndex, Bool}
    set::S
end
```
Then if `S` is a scalar set, it works for any scalar function, and if `S` is a  vector set, it works for any vector function.

matbesancon:
Ok, one design consideration I tried to keep is to represent the constraint as closely as possible to what solvers take as input

robertschwarz:
> Ok, one design consideration I tried to keep is to represent the constraint as closely as possible to what solvers take as input

In particular, solvers should be able to do a fine-grained `MOI.supports`, right?
For SCIP to declare that it can do only `y => 1` but not `y => 0`, this number would have to be type parameter of the set?

odow:
> Ok, one design consideration I tried to keep is to represent the constraint as closely as possible to what solvers take as input

Gurobi will just declare that it only supports `ScalarAffineFunction-in-IndicatorSet{S<:Union{GreaterThan, LessThan, EqualTo}}`. We can make the underlying design more general.

> For SCIP to declare that it can do only y => 1 but not y => 0, this number would have to be type parameter of the set?

It's probably fine for SCIP to throw and error at run-time when inspecting the value, rather than at compile time.

blegat:
> It's probably fine for SCIP to throw and error at run-time when inspecting the value, rather than at compile time.

If SCIP does not support `IndicatorSet{false, S}` then it can be bridged to `IndicatorSet{true, S}` by creating a new variable `y` and the constraint `x + y = 1` so we need the information in the set to be able to bridge it.

odow:
So
```julia
struct IndicatorSet{B, S}
    variable::VariableIndex
    set::S
    IndicatorSet(v::Pair{VariableIndex, Bool}, set::S) = new{last(v), S}(first(v), set)
end
```

mlubin:
Having a variable stored in the set means that we have to make nontrivial changes to any code that deals with copying between models (e.g., CachingOptimizer). Not worth it IMO.

blegat:
> Then use VAF to write down the constraints, e.g.,
> ```julia
> (1.0 * y, 2*x + 3*z) in IndicatorSet(LessThan(b))
> ```
> It's an error if the first component of the VAF isn't 1.0 times some variable.

If a solver supports the first component to be a `SingleVariable` and the second to be a `ScalarAffineFunction` then he should be able to say it in `supports_constraint` so that we can bridge the constraint if it the first component is an affine expression by adding a slack variable equal to this affine expression.

We could create a new function type
```julia
struct MixedTypesVectorFunction{FuncTypes} <: MOI.AbstractVectorFunction
    func::FuncTypes
end
```
and the solver would say, e.g.
```julia
MOI.supports(::Optimizer,
                      ::MOI.MixedTypesVectorFunction{Tuple{MOI.SingleVariable, MOI.ScalarAffineFunction{Float64}}},
                      ::MOI.IndicatorSet{true, <:Union{MOI.EqualTo{Float64}, MOI.LessThan{Float64}, MOI.GreaterThan{Float64}}})
```
Then
```julia
MOI.MixedTypesVectorFunction((f, g)) in MOI.IndicatorSet{true}(s)
```
would mean that `f => g in S` and
```julia
MOI.MixedTypesVectorFunction((f, g)) in MOI.IndicatorSet{false}(s)
```
would mean that `!f => g in S`.

mlubin:
> If a solver supports the first component to be a SingleVariable and the second to be a ScalarAffineFunction then he should be able to say it in supports_constraint so that we can bridge the constraint if it the first component is an affine expression by adding a slack variable equal to this affine expression.

I don't see this as especially useful for indicator constraints. It is standard and not a significant burden to model with an explicit indicator variable.

robertschwarz:
I like the not-so-general, but maybe still readable version (in JuMP):

    @constraint(model, [y, 2x + 3z] in IndicatorSet(1.0, LessThan(b)))

which also has some symmetry in that both the function and the set have two elements that correspond:

    y == 1.0
    2x + 3z in LessThan(b)

blegat:
> I don't see this as especially useful for indicator constraints. It is standard and not a significant burden to model with an explicit indicator variable.

As long as no solver support having the first component equal to an affine function and we create an MOI function to extract the `MOI.SingleVariable` from the VAF so have a consistent error across solvers then I agree that it is not worth creating this Tuple function.

@robertschwarz Why not use a bool ? It can only be 0.0 or 1.0 right ? What if the user puts 2.0 ?

robertschwarz:
I thought about using a bool, but then I remembered the `MOI.ZeroOrOne` set and wanted to be consistent with that. Of course, that set is not really related to `Float64` either...

mlubin:
> Why not use a bool ? It can only be 0.0 or 1.0 right ? What if the user puts 2.0 ?

We can also use an enum. `ONE_IMPLIES_SATISFIED` and `ZERO_IMPLIES_SATISFIED`.

mlubin:
> As long as no solver support having the first component equal to an affine function

There are no solvers that support having the first component equal to an affine function.

blegat:
> There are no solvers that support having the first component equal to an affine function.

Then I would be ok with either
```julia
@constraint(model, [y, 2x + 3z] in IndicatorSet{true}(LessThan(4.0)))
```
or
```julia
@constraint(model, [y, 2x + 3z] in IndicatorSet{ONE_IMPLIES_SATISFIED}(LessThan(4.0)))
```
to represent `y => (2x + 3z <= 4)`.

mlubin:
I'm on board with the type parameter. It will let us bridge for solvers that support only one of the two possibly values for the indicator variable.

src/sets.jl
```diff
@@ -439,22 +439,25 @@ Base.isapprox(a::SOS2{T}, b::SOS2{T}; kwargs...) where T = isapprox(a.weights, b
...
 `S` can be one of `LessThan`, `GreaterThan` or `EqualTo`.
+`ActiveCond` is either true or false.
 `IndicatorSet` is used with a `VectorAffineFunction` holding
 the indicator variable first.
 """
-struct IndicatorSet{S <:Union{LessThan,GreaterThan,EqualTo}} <: AbstractVectorSet
+struct IndicatorSet{ActiveCond, S <:Union{LessThan,GreaterThan,EqualTo}} <: AbstractVectorSet
     s::S
-    IndicatorSet(s::S) where {S <: Union{LessThan,GreaterThan,EqualTo}} = new{S}(s)
+    IndicatorSet(s::S, active_cond::Bool = true) where {S <: Union{LessThan,GreaterThan,EqualTo}} = new{active_cond, S}(s)
```
blegat: I would be type unstable to construct `IndicatorSet` where the boolean value is given as argument and not type argument

matbesancon: ah point yes

src/sets.jl
```diff
@@ -438,6 +438,25 @@ Base.isapprox(a::SOS2{T}, b::SOS2{T}; kwargs...) where T = isapprox(a.weights, b
...
 dimension(s::Union{SOS1, SOS2}) = length(s.weights)
 
+"""
+    IndicatorSet{ActiveCond, S <:Union{LessThan,GreaterThan,EqualTo}}
+
+Set of (x,y) that satisfy the indicator constraint:
+    y ∈ 𝔹, y = ActiveCond ==> a^T x ∈ S
+
+`S` can be one of `LessThan`, `GreaterThan` or `EqualTo`.
+`ActiveCond` is either true or false.
```
mlubin: For readability, I prefer the enum that I proposed over bools. `IndicatorSet{ZERO_IMPLIES_SATISFIED}` is super clear, while `IndicatorSet{false}` would send me looking to the definition of what `false` means.

src/sets.jl
```diff
@@ -438,6 +438,34 @@ Base.isapprox(a::SOS2{T}, b::SOS2{T}; kwargs...) where T = isapprox(a.weights, b
...
 dimension(s::Union{SOS1, SOS2}) = length(s.weights)
 
+"""
+	ActivationCond
+
+Activation condition for an indicator constraint.
+Used as first type parameter of `IndicatorSet{ActiveCond,S}`.
+"""
+@enum ActivationCond begin
+    ActivateOnFalse
```
odow: JuMP style guide would suggest `ACTIVATE_ON_FALSE`

src/sets.jl
```diff
@@ -438,6 +438,34 @@ Base.isapprox(a::SOS2{T}, b::SOS2{T}; kwargs...) where T = isapprox(a.weights, b
...
+
+Set of (x,y) that satisfy the indicator constraint:
+    y ∈ 𝔹, y = ActiveCond ==> a^T x ∈ S
+
+`S` can be one of `LessThan`, `GreaterThan` or `EqualTo`.
+`ActiveCond` is one of the value of the `ActivationCond` enum.
+`IndicatorSet` is used with a `VectorAffineFunction` holding
+the indicator variable first.
+"""
+struct IndicatorSet{ActiveCond, S <:Union{LessThan,GreaterThan,EqualTo}} <: AbstractVectorSet
```
blegat: Why restrict `S` and not let it be any set ?

matbesancon:
These are the sets supported by common solvers for indicator constraints.

blegat:
> These are the sets supported by common solvers for indicator constraints.

These can specify that they only support these tests by doing
```julia
MOI.supports(::Optimizer, ::Type{MOI.VectorAffineFunction{Float64}}, ::Type{MOI.IndicatorSet{ACTIVATE_ON_TRUE, <:Union{...}}}) = true
```

src/sets.jl
```diff
@@ -438,6 +438,34 @@ Base.isapprox(a::SOS2{T}, b::SOS2{T}; kwargs...) where T = isapprox(a.weights, b
...
+
+`S` has to be an `AbstractScalarSet`.
+`ActiveCond` is one of the value of the `ActivationCond` enum.
+`IndicatorSet` is used with a `VectorAffineFunction` holding
+the indicator variable first.
+"""
+struct IndicatorSet{ActiveCond, S <: AbstractScalarSet} <: AbstractVectorSet
+    s::S
+    IndicatorSet{ActiveCond}(s::S) where {ActiveCond, S <: AbstractScalarSet} = new{ActiveCond,S}(s)
+end
```
odow: I haven't tested, but you probably want something like the following
 - use meaningful names where possible, so `s` becomes `set`
 - `ActiveCond` should be restricted to the right type
 - use escaped latex in the docstring so that it renders properly

```julia
"""
	ActivationCondition

Activation condition for an indicator constraint.
Used as first type parameter of [`IndicatorSet`](@ref).
"""
@enum ActivationCondition begin
    ACTIVATE_ON_FALSE
    ACTIVATE_ON_TRUE
end

"""
    IndicatorSet{A, S <: AbstractScalarSet}

The set ``\\{(y, x) \\in \\{0, 1\\} \\times \\mathbb{R}: y = A \\implies x \\in set\\}``,
where `A` is an [`ActivationCondition`](@ref) enum.

`IndicatorSet` is used with an [`AbstractVectorFunction`](@ref) of dimension 2,
with the first element being the indicator variable `y`, and the second element
being a scalar function of `x`.
"""
struct IndicatorSet{A, S <: AbstractScalarSet} <: AbstractVectorSet
    set::S
    function IndicatorSet(A::ActivationCondition, set::AbstractScalarSet)
        return new{A, typeof(set)}(set)
    end
end
```

There is a debate about whether `set` should be an `AbstractScalarSet`. I don't see why it shouldn't be an `AbstractVectorSet` with one less dimension than the function.

mlubin: Is there any solver that natively supports the equivalent of indicator constraints for VAF? Why do we need this extra complexity when you could represent the same thing by adding multiple indicator constraints with a common indicator variable?

mlubin: > y = A ... where `A` is an [`ActivationCondition`](@ref) enum.
Doesn't make much sense. `y` doesn't take enum values. I'd explicitly list the cases, e.g.,
```
The set ``\\{(y, x) \\in \\{0, 1\\} \\times \\mathbb{R}: y = 0 \\implies x \\in set\\}``
when `A` is `ACTIVATE_WHEN_ZERO` and
``\\{(y, x) \\in \\{0, 1\\} \\times \\mathbb{R}: y = 1 \\implies x \\in set\\}``
when `A` is `ACTIVATE_WHEN_ONE`.

File: src/sets.jl [hunk 10]
```diff
@@ -438,6 +438,34 @@ Base.isapprox(a::SOS2{T}, b::SOS2{T}; kwargs...) where T = isapprox(a.weights, b
...
 dimension(s::Union{SOS1, SOS2}) = length(s.weights)
 
+"""
+	ActivationCond
+
+Activation condition for an indicator constraint.
+Used as first type parameter of `IndicatorSet{ActiveCond,S}`.
+"""
+@enum ActivationCond begin
+    ACTIVATE_ON_FALSE
```
mlubin: Use `ZERO` and `ONE` instead of `FALSE` and `TRUE`. We don't overload the meaning of 0/1 variables with bools anywhere else in the MOI.

src/sets.jl
blegat: > Is there any solver that natively supports the equivalent of indicator constraints for VAF? Why do we need this extra complexity when you could represent the same thing by adding multiple indicator constraints with a common indicator variable?

Adding multiple indicator variables does not work if the set is `SecondOrderCone` and modeling with such set may be useful even if it is not supported by any solver if it could be bridge. However I am not sure how to bridge a vector set.
For a scalar set you can add a variable `u` and add the constraints `u + x in set` and `[y, u]  in SOS1` but it does not work if `u` is a vector as you might need several components to be nonzero.

matbesancon: @odow 
> ActiveCond should be restricted to the right type

(replaced by `A` in the new version), it's not a type but an enum value, I don't think you can check type parameter type (just like `Array{T,N}` does not check that `N` is an integer

matbesancon: also, I'm waiting for a final design agreement to add tests

src/Test/intlinear.jl
```diff
@@ -358,9 +358,144 @@ function knapsacktest(model::MOI.ModelLike, config::TestConfig)
...
+    @test MOI.supports_constraint(model, MOI.SingleVariable, MOI.Interval{Float64})
+    @test MOI.supports_constraint(model, MOI.ScalarAffineFunction{Float64}, MOI.Interval{Float64})
+    @test MOI.supports_constraint(model, MOI.ScalarAffineFunction{Float64}, MOI.IndicatorSet{MOI.ACTIVATE_ON_ONE, MOI.LessThan{Float64}})
+    x1 = MOI.add_variable(model)
+    x2 = MOI.add_variable(model)
+    z1  = MOI.add_variable(model)
+    z2  = MOI.add_variable(model)
+    MOI.add_constraint(model, z1, MOI.ZeroOne())
+    MOI.add_constraint(model, z2, MOI.ZeroOne())
+    f1 = MOI.ScalarAffineFunction(
```
mlubin: This is the scalar-valued function z1 + x2. It should be the vector-valued function [z1, x2]. I think this affects all the tests here.

src/sets.jl
```diff
@@ -438,6 +438,38 @@ Base.isapprox(a::SOS2{T}, b::SOS2{T}; kwargs...) where T = isapprox(a.weights, b
...
+    ACTIVATE_ON_ZERO
+    ACTIVATE_ON_ONE
+end
+
+"""
+    IndicatorSet{A, S <: AbstractScalarSet}(set::S)
+
+Set ``\\{(y, x) \\in \\{0, 1\\} \\times \\mathbb{R}^n``
+Such that: ``y = 0 \\implies x \\in set\\}``
+when `A` is `ACTIVATE_WHEN_ZERO` and
```
mlubin: `ACTIVATE_ON_ZERO`

src/sets.jl
```diff
@@ -438,6 +438,38 @@ Base.isapprox(a::SOS2{T}, b::SOS2{T}; kwargs...) where T = isapprox(a.weights, b
...
+``\\{(y, x) \\in \\{0, 1\\} \\times \\mathbb{R}^n``
+Such that: ``y = 1 \\implies x \\in set\\}``
+when `A` is `ACTIVATE_WHEN_ONE`.
```
mlubin: `ACTIVATE_ON_ONE`

src/sets.jl
```diff
@@ -438,6 +438,38 @@ Base.isapprox(a::SOS2{T}, b::SOS2{T}; kwargs...) where T = isapprox(a.weights, b
...
+@enum ActivationCond begin
+    ACTIVATE_ON_ZERO
+    ACTIVATE_ON_ONE
+end
+
+"""
+    IndicatorSet{A, S <: AbstractScalarSet}(set::S)
+
+Set ``\\{(y, x) \\in \\{0, 1\\} \\times \\mathbb{R}^n``
+Such that: ``y = 0 \\implies x \\in set\\}``
```
mlubin: I'm having trouble parsing these two lines as a mathematical set. (There's text in the middle of the set definition?).
You can write this as ``\\{((y, x) \\in \\{0, 1\\} \\times \\mathbb{R}^n : y = 0 \\implies x \\in set\\}``
or also ``\\{0 \\} \\times \\mathbb{R} \cup \\{1\\} \\times set``. You should also compile this and check that the equations display reasonably in the doc.

src/sets.jl
mlubin: > Adding multiple indicator variables does not work if the set is SecondOrderCone and modeling with such set may be useful even if it is not supported by any solver if it could be bridge. 

Good point. But I propose we address the problem at hand and keep `S` as a scalar set. If, in the future, we want indicator constraints on vector sets, we can add them. We already have redundant ways to express scalar sets using 1-dimensional vector sets, so having a scalar indicator constraint and a vector indicator constraint would be consistent with MOI.

mlubin:
Also we can't merge with the CI failing. It looks like this introduces a syntax that doesn't work in 0.6. No objections to dropping 0.6 (although it means that we can't tag this code as MOI 0.8.x).

src/sets.jl
matbesancon: fixed in last commit yup. Doc built it renders well

src/sets.jl
matbesancon: fixed in last commit

blegat:
It might be a good time for dropping Julia v0.6 now that we need separate ways to tag v0.6 (attobot) and v1.0 (Registrator) release.

matbesancon:
> It might be a good time for dropping Julia v0.6 now that we need separate ways to tag v0.6 (attobot) and v1.0 (Registrator) release.

Indeed, adding `@compat` everywhere will just increase the number of things to watch when cleaning it out

matbesancon:
@blegat dropping 0.6 should probably be its own PR, including removing `Compat` stuff?

blegat:
> @blegat dropping 0.6 should probably be its own PR, including removing Compat stuff?

Yes, that would be easier to review as a separate PR

matbesancon:
waiting for #714

matbesancon:
good to go? ping @blegat @mlubin

src/Test/intlinear.jl
```diff
@@ -358,9 +358,147 @@ function knapsacktest(model::MOI.ModelLike, config::TestConfig)
     end
 end
 
+function indtest(model::MOI.ModelLike, config::TestConfig)
```
blegat: According to JuMP style guide, it should rather be `indicator_test`

src/Test/intlinear.jl
```diff
@@ -358,9 +358,147 @@ function knapsacktest(model::MOI.ModelLike, config::TestConfig)
...
+    atol = config.atol
+    rtol = config.rtol
+    # linear problem with indicator constraint
+    # max  2x1 + 3x2
+    # s.t. x1 + x2 <= 10
+    #      z1 ==> x2 <= 8
+    #      z2 ==> x2 + x1/5 <= 9
+    # z1 + z2 >= 1
```
blegat: Should be indented

src/Test/intlinear.jl
```diff
@@ -358,9 +358,147 @@ function knapsacktest(model::MOI.ModelLike, config::TestConfig)
...
+
+    @test MOI.supports(model, MOI.ObjectiveFunction{MOI.ScalarAffineFunction{Float64}}())
+    @test MOI.supports(model, MOI.ObjectiveSense())
+    @test MOI.supports_constraint(model, MOI.SingleVariable, MOI.ZeroOne)
+    @test MOI.supports_constraint(model, MOI.SingleVariable, MOI.Interval{Float64})
+    @test MOI.supports_constraint(model, MOI.ScalarAffineFunction{Float64}, MOI.Interval{Float64})
+    @test MOI.supports_constraint(model, MOI.ScalarAffineFunction{Float64}, MOI.IndicatorSet{MOI.ACTIVATE_ON_ONE, MOI.LessThan{Float64}})
+    x1 = MOI.add_variable(model)
+    x2 = MOI.add_variable(model)
+    z1  = MOI.add_variable(model)
```
blegat: Remove space before `=` (same on next line)

src/Test/intlinear.jl
```diff
@@ -358,9 +358,147 @@ function knapsacktest(model::MOI.ModelLike, config::TestConfig)
...
+    f1 = MOI.VectorAffineFunction(
+        MOI.VectorAffineTerm(1, MOI.ScalarAffineTerm(1.0, z1)),
+        MOI.VectorAffineTerm(2, MOI.ScalarAffineTerm(1.0, x2)),
+    )
+    iset1 = MOI.IndicatorSet{MOI.ACTIVATE_ON_ONE}(MOI.LessThan(8.))
+    MOI.add_constraint(model, f1, iset1)
+
+    f2 = MOI.VectorAffineFunction(
+        MOI.VectorAffineTerm(1, MOI.ScalarAffineTerm(1.0, z2)),
+        MOI.ScalarAffineTerm(1.0, z2),
```
blegat: What is this line doing ?

src/Test/intlinear.jl
```diff
@@ -358,9 +358,147 @@ function knapsacktest(model::MOI.ModelLike, config::TestConfig)
...
+    @test MOI.supports_constraint(model, MOI.ScalarAffineFunction{Float64}, MOI.Interval{Float64})
+    @test MOI.supports_constraint(model, MOI.ScalarAffineFunction{Float64}, MOI.IndicatorSet{MOI.ACTIVATE_ON_ONE, MOI.LessThan{Float64}})
+    x1 = MOI.add_variable(model)
+    x2 = MOI.add_variable(model)
+    z1  = MOI.add_variable(model)
+    z2  = MOI.add_variable(model)
+    MOI.add_constraint(model, z1, MOI.ZeroOne())
+    MOI.add_constraint(model, z2, MOI.ZeroOne())
+    f1 = MOI.VectorAffineFunction(
+        MOI.VectorAffineTerm(1, MOI.ScalarAffineTerm(1.0, z1)),
```
blegat: The terms should be in a vector

src/Test/intlinear.jl
```diff
@@ -358,9 +358,147 @@ function knapsacktest(model::MOI.ModelLike, config::TestConfig)
...
+        @test MOI.get(model, MOI.TerminationStatus()) == MOI.OPTIMAL
+        @test MOI.get(model, MOI.PrimalStatus()) == MOI.FEASIBLE_POINT
+        @test MOI.get(model, MOI.ObjectiveValue()) ≈ 28.75 atol=atol rtol=rtol
+        @test MOI.get(model, MOI.VariablePrimal(), x1) ≈ 1.25 atol=atol rtol=rtol
+        @test MOI.get(model, MOI.VariablePrimal(), x2) ≈ 8.75 atol=atol rtol=rtol
+        @test MOI.get(model, MOI.VariablePrimal(), z1) ≈ 0.0 atol=atol rtol=rtol
+        @test MOI.get(model, MOI.VariablePrimal(), z2) ≈ 1.0 atol=atol rtol=rtol
+    end
+
+    MOI.empty!(model)
```
blegat: Might be better to do a separate test so that a solver can exclude one of them and not the other one

src/Test/intlinear.jl
```diff
@@ -358,9 +358,147 @@ function knapsacktest(model::MOI.ModelLike, config::TestConfig)
...
+
+end
+
 const intlineartests = Dict("knapsack" => knapsacktest,
                             "int1"     => int1test,
                             "int2"     => int2test,
-                            "int3"     => int3test)
+                            "int3"     => int3test,
+                            "indcons"  => indtest,
+                            )
```
blegat: Remove one space before `)` to align with opening one

src/sets.jl
```diff
@@ -438,6 +438,36 @@ Base.isapprox(a::SOS2{T}, b::SOS2{T}; kwargs...) where T = isapprox(a.weights, b
...
+
+``\\{((y, x) \\in \\{0, 1\\} \\times \\mathbb{R}^n : y = 0 \\implies x \\in set\\}``
+when `A` is `ACTIVATE_ON_ZERO` and
+``\\{((y, x) \\in \\{0, 1\\} \\times \\mathbb{R}^n : y = 1 \\implies x \\in set\\}``
+when `A` is `ACTIVATE_ON_ONE`.
+
+`S` has to be a sub-type of `AbstractScalarSet`.
+`A` is one of the value of the `ActivationCond` enum.
+`IndicatorSet` is used with a `VectorAffineFunction` holding
+the indicator variable first.
```
blegat: Add an example

blegat:
The new test should be tested in https://github.com/JuliaOpt/MathOptInterface.jl/blob/master/test/Test/intlinear.jl

matbesancon:
Fixed most of the comments. @blegat I didn't find anything for testing new sets with MockOptimizer.
SOS constraints are not tested using it (just basic unit tests in `/test/sets.jl`

src/Test/intlinear.jl
```diff
@@ -485,21 +497,20 @@ function indicator_test(model::MOI.ModelLike, config::TestConfig)
...
-
-
 end
 
 const intlineartests = Dict("knapsack" => knapsacktest,
                             "int1"     => int1test,
                             "int2"     => int2test,
                             "int3"     => int3test,
-                            "indicator_cons"  => indicator_test,
+                            "indicator_cons1"  => indicator_test1,
```
blegat: Let's do `"indicator1" => indicator1_test" for consistency with the other function which will be renamed `knapsack_test`, `int1_test` , ... in the future

test/Test/intlinear.jl
```diff
@@ -24,9 +24,9 @@
     MOIU.set_mock_optimize!(mock,
         (mock::MOIU.MockOptimizer) -> MOIU.mock_optimize!(mock, [1.25, 8.75, 0., 1.])
     )
-    MOIT.indicator_test1(mock, config)
+    # MOIT.indicator1_test(mock, config)
```
blegat: Why is it commented ?

matbesancon: `MOIT.indicator1_test(mock, config)`
The `mock` optimizer does not support indicator constraints, I found nowhere to change this. Furthermore, `SOS` constraints are not tested here either

blegat: It should be added here:
https://github.com/JuliaOpt/MathOptInterface.jl/blob/589cd2252c5fb1aacbfb0852a931a8a703265746/test/runtests.jl#L28-L29
You should do something like
```julia
const LessThanIndicatorSet{T} = IndicatorSet{MOI.ACTIVATE_ON_ONE, MOI.LessThan{T}}
```
and add `LessThanIndicatorSet` in the macro because in this macro, the sets have either no type argument or only `T`, the coefficient type.

matbesancon:
tests passing cc @blegat @mlubin

src/sets.jl
```diff
@@ -435,6 +435,53 @@ Base.isapprox(a::T, b::T; kwargs...) where {T <: Union{SOS1, SOS2}} = isapprox(a
...
+```julia
+f = MOI.VectorAffineFunction(
+    [MOI.VectorAffineTerm(1, MOI.ScalarAffineTerm(1.0, z)),
+     MOI.VectorAffineTerm(2, MOI.ScalarAffineTerm(0.2, x1)),
+     MOI.VectorAffineTerm(2, MOI.ScalarAffineTerm(1.0, x2)),
+    ],
+    [0., 0.],
+)
+
+indicator_set = MOI.IndicatorSet{MOI.ACTIVATE_ON_ONE}(MOI.LessThan(9.))
```
blegat: 9. should be 9.0 according to the JuMP style guide

src/sets.jl
```diff
@@ -435,6 +435,53 @@ Base.isapprox(a::T, b::T; kwargs...) where {T <: Union{SOS1, SOS2}} = isapprox(a
 
 dimension(s::Union{SOS1, SOS2}) = length(s.weights)
 
+"""
+	ActivationCond
+
+Activation condition for an indicator constraint.
+Used as first type parameter of `IndicatorSet{ActiveCond,S}`.
+"""
+@enum ActivationCond begin
```
blegat: Why not call it `ActivationCondition` in full ?

matbesancon: changed

src/sets.jl
matbesancon: fixed, in tests also

matbesancon:
Thanks, added in last commit

test/sets.jl
```diff
@@ -6,5 +6,15 @@
             s_copy.weights[1] = 2.0
             @test s.weights[1] == 1.0
         end
+        @testset "IndicatorSet" begin
+            s1 = MOI.IndicatorSet{MOI.ACTIVATE_ON_ONE}(MOI.LessThan(4.0))
+            s2 = MOI.IndicatorSet{MOI.ACTIVATE_ON_ZERO}(MOI.GreaterThan(4.0))
+            s1_copy = copy(s1)
+            s2_copy = copy(s2)
+            @test s1_copy isa MOI.IndicatorSet{MOI.ACTIVATE_ON_ONE}
+            @test s1 == s1_copy
```
blegat: Check that modifying the set in `s1` does not modify the set in `s1_copy`. You will need a scalar set which is mutable

src/Test/intlinear.jl
```diff
@@ -356,9 +356,164 @@ function knapsacktest(model::MOI.ModelLike, config::TestConfig)
...
+         MOI.VectorAffineTerm(2, MOI.ScalarAffineTerm(0.2, x1)),
+         MOI.VectorAffineTerm(2, MOI.ScalarAffineTerm(1.0, x2)),
+        ],
+        [0.0, 0.0],
+    )
+    iset2 = MOI.IndicatorSet{MOI.ACTIVATE_ON_ONE}(MOI.LessThan(9.0))
+
+    MOI.add_constraint(model, f2, iset2)
+
+    # additional regular constraint
```
mlubin: Nit: Please appropriately capitalize and punctuate comments (http://www.juliaopt.org/JuMP.jl/dev/style/#Comments-1).

src/Test/intlinear.jl
```diff
@@ -356,9 +356,164 @@ function knapsacktest(model::MOI.ModelLike, config::TestConfig)
...
+    # s.t. x1 + x2 <= 10
+    #      z1 ==> x2 <= 8
+    #      z2 ==> x2 + x1/5 <= 9
+    #      z1 + z2 >= 1
+
+    MOI.empty!(model)
+    @test MOI.is_empty(model)
+
+    # same model as indicator_test1 with penalty
+    # on z2 switches active constraint to z1
```
mlubin: This is a run-on sentence. "This is the same model as indicator_test1, except that the penalty on z2 forces z1 to be 1."

test/runtests.jl
```diff
@@ -33,7 +35,7 @@ MOIU.@model(Model,
              MOI.PositiveSemidefiniteConeTriangle, MOI.PositiveSemidefiniteConeSquare,
              MOI.RootDetConeTriangle, MOI.RootDetConeSquare, MOI.LogDetConeTriangle,
              MOI.LogDetConeSquare),
-            (MOI.PowerCone, MOI.DualPowerCone, MOI.SOS1, MOI.SOS2),
+            (MOI.PowerCone, MOI.DualPowerCone, MOI.SOS1, MOI.SOS2, LessThanIndicatorSet),
```
mlubin: Is something lost by supporting `MOI.IndicatorSet` directly? Someone reading this code will immediately wonder why the choice of `LessThanIndicatorSet`.

src/Test/intlinear.jl
```diff
@@ -356,9 +356,164 @@ function knapsacktest(model::MOI.ModelLike, config::TestConfig)
...
+
+        @test MOI.get(model, MOI.TerminationStatus()) == MOI.OPTIMAL
+        @test MOI.get(model, MOI.PrimalStatus()) == MOI.FEASIBLE_POINT
+        @test MOI.get(model, MOI.ObjectiveValue()) ≈ 28.0 atol=atol rtol=rtol
+        @test MOI.get(model, MOI.VariablePrimal(), x1) ≈ 2.0  atol=atol rtol=rtol
+        @test MOI.get(model, MOI.VariablePrimal(), x2) ≈ 8.0  atol=atol rtol=rtol
+        @test MOI.get(model, MOI.VariablePrimal(), z1) ≈ 1.0 atol=atol rtol=rtol
+        @test MOI.get(model, MOI.VariablePrimal(), z2) ≈ 0.0 atol=atol rtol=rtol
+    end
+end
```
mlubin: `ACTIVATE_ON_ZERO` should be tested as well.

test/runtests.jl
blegat: When creating a `Model{T}`, sets `S` in the previous category will be typed `S` and sets in this category will be typed `S{T}`. Here `MOI.IndicatorSet` needs to be typed `MOI.IndicatorSet{MOI.ACTIVATE_ON_ONE, MOI.LessThan{T}}` so it does not fall into any category while `LessThanIndicatorSet` falls in the second category.
We could use `MOI.IndicatorSet{MOI.ACTIVATE_ON_ONE}` while could work (untested) but the container will have a non-concrete type.

src/Test/intlinear.jl
matbesancon: tested in last commit

test/runtests.jl
matbesancon: so this one is resolved?

src/Test/intlinear.jl
matbesancon: fixed in last commit

src/Test/intlinear.jl
matbesancon: fixed. Having a strict style guide without a linter is quite a pain

test/sets.jl
matbesancon: do we even have one? I understand the importance of testing it does not happen, but we are calling copy on the subtest, so if error there is, it's because `copy` is not properly implemented on it

blegat: @matbesancon `LessThan{BigInt}` is mutable because `BigInt` is mutable. You call also create a dummy mutable coefficient type or a dummy mutable set just for the test.

> we are calling copy on the subtest

That's what we want to test. If you remove this copy on the subtests, the tests still pass while they shouldn't

matbesancon: @blegat I played a bit with this idea. Fundamentally, this is breaking an assumption about numbers that will be everywhere in MOI, this PR doesn't change that. `BigInt` being mutable is just a constraint for their construction as far as I know, not supposed to be used.

The implementation from `src/sets`:
```julia
# isbits types, nothing to copy
function Base.copy(set::Union{Reals, Zeros, Nonnegatives, Nonpositives,
                              GreaterThan, LessThan, EqualTo, Interval,
                              SecondOrderCone, RotatedSecondOrderCone,
                              GeometricMeanCone, ExponentialCone,
                              DualExponentialCone, PowerCone, DualPowerCone,
                              PositiveSemidefiniteConeTriangle,
                              PositiveSemidefiniteConeSquare,
                              LogDetConeTriangle, LogDetConeSquare,
                              RootDetConeTriangle, RootDetConeSquare,
                              Integer, ZeroOne, Semicontinuous, Semiinteger})
    return set
end
```

Is then false because it should check `isbits` on the set

blegat: Good point. We could defer the question of whether we should copy the field of EqualTo... to a separate issue.
Still you might be able to create a mutable abstract set with Float64 to test the fact we ce copy the set for IndicatorSet

matbesancon: @blegat see issue #721, and tests added here for `IndicatorSet` copy

PR #711, opened 2019-04-18
Doubled methods removed

matbesancon:
Some `SOS{1/2}` methods could be merged.
Also, one was defined in Test for no apparent reason


matbesancon:
Should this be merged? (codecov seems broken on that)

PR #715, opened 2019-04-19
Refactor tests

blegat:
The idea is to be able to run subsets of the tests easily.
So if you just want to test utilities, you just do
```julia
$ julia test/Utilities/Utilities.jl
```
and if you just want to test the caching optimizer, you can do
```julia
$ julia test/Utilities/caching_optimizer.jl
```
The second important aspect is that we would like to have several testsets when we do that so that we have information about the progress of the tests.
If it runs for a few seconds and then just says
```julia
$ julia test/Utilities/caching_optimizer.jl
Test Summary:     | Pass  Total
Caching Optimizer |  109    109
```
then it's not helpful.
For this reason, inside `test/Utilities/caching_optimizer.jl`, there is a few test sets and in `test/Utilities/Utilities.jl`, we have
```julia
@testset "Caching Optimizer" begin
    include("caching_optimizer.jl")
end
```
so that if we run `test/Utilities/caching_optimizer.jl` we see each sub-testsets and if we run the whole `test/Utilities/Utilities.jl`, we just get `Caching Optimizer` and if we run the whole `runtests.jl`, we just have `Utilities`.


blegat:
Failing on Julia v0.7 when it redefines a struct:
https://github.com/JuliaOpt/MathOptInterface.jl/blob/208a9aa3d579d2811b1e7b6947fcb2b1ca890984/src/Utilities/model.jl#L582
I don't feel it's worth it adding a hack for Julia v0.7 given that it support was dropped:
https://discourse.julialang.org/t/julia-0-6-and-0-7-are-now-officially-unmaintained/20449/2

PR #718, opened 2019-04-27
Remove authors field

blegat:
We are deprecating JuliaOpt, see https://www.juliaopt.org/meetings/santiago2019/slides/miles_lubin.pdf


matbesancon:
It might be worth asking if the author field is not assumed by some tools (Registrator, Pkg or other) and if removing it is not gonna break some stuff

blegat:
That'a good question. I have not used yet any authors field in the packages I have registered with Registrator and it seems to work but there may be some tools requiring it

PR #723, opened 2019-05-02
common constant function

### LINKED ISSUE #618: Refactor MOI._constant and MOIU.getconstant
Create good names consistent with the [JuMP style guide](http://www.juliaopt.org/JuMP.jl/v0.19.0/style/)

matbesancon:
Merge `MOI._constant` and `MOIU.getconstant`. Fixes #618

### DIFF
--- a/docs/src/apireference.md
+++ b/docs/src/apireference.md
@@ -212,11 +212,14 @@ VectorQuadraticTerm
 VectorQuadraticFunction
 ```
 
-Functions for getting and setting properties of sets.
+Functions for getting and setting properties of function.
 
 ```@docs
 output_dimension
-constant
+constant(f::Union{ScalarAffineFunction, ScalarQuadraticFunction})
+constant(f::Union{VectorAffineFunction, VectorQuadraticFunction})
+constant(f::SingleVariable, ::DataType)
+constant(f::VectorOfVariables, T::DataType)
 ```
 
 ## Sets
@@ -259,6 +262,7 @@ Functions for getting and setting properties of sets.
 
 ```@docs
 dimension
+constant(s::EqualTo)
 ```
 
 ## Modifications
--- a/docs/src/apireference.md
+++ b/docs/src/apireference.md
@@ -216,6 +216,7 @@ Functions for getting and setting properties of sets.
 
 ```@docs
 output_dimension
+constant
 ```
 
 ## Sets
--- a/docs/src/apireference.md
+++ b/docs/src/apireference.md
@@ -212,7 +212,7 @@ VectorQuadraticTerm
 VectorQuadraticFunction
 ```
 
-Functions for getting and setting properties of function.
+Functions for getting and setting properties of functions.
 
 ```@docs
 output_dimension
--- a/src/Utilities/sets.jl
+++ b/src/Utilities/sets.jl
@@ -1,4 +1,3 @@
-
 const DimensionUpdatableSets = Union{MOI.Reals,
                                      MOI.Zeros,
                                      MOI.Nonnegatives,
--- a/src/Utilities/sets.jl
+++ b/src/Utilities/sets.jl
@@ -28,7 +28,7 @@ function shift_constant(set::Union{MOI.LessThan{T},
                                    MOI.GreaterThan{T},
                                    MOI.EqualTo{T}},
                         offset::T) where T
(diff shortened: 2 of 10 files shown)


blegat:
We might also add a way to get zero for the constant of `SingleVariable` and `VectorOfVariables` to solve the use case in:
https://github.com/JuliaOpt/MathOptInterface.jl/blob/c1d8f6a6949a39091c89d8c4053d51a7f6cc8d03/src/Bridges/scalarizebridge.jl#L7-L10
For that we need to know what is the coefficient type. We could allow a second argument which is the coefficient type but is optional for every set/function except `SingleVariable`/`VectorOfVariables`.

For the name, `getconstant` does not satisfy JuMP style guide, it should be `constant`, `get_constant`, or `set_constant`/function_constant`.
The second choice is inconsistent with the naming in the rest of JuMP/MOI but the first and third choices works (the first one is better IMO since the fact that it is a `set` or `function` is already in the signature so adding it in the name is redundant).

src/Utilities/sets.jl
```diff
@@ -3,9 +3,9 @@
 
 Returns the constant of the set.
 """
-getconstant(s::MOI.EqualTo) = s.value
-getconstant(s::MOI.GreaterThan) = s.lower
-getconstant(s::MOI.LessThan) = s.upper
+MOI.getconstant(s::MOI.EqualTo) = s.value
```
blegat: We can move the implementation to `MOI` if it's an MOI function.

src/functions.jl
```diff
@@ -352,14 +352,42 @@ function _dicts(f::Union{ScalarQuadraticFunction, VectorQuadraticFunction})
...
+    isapprox(constant(f), constant(g); kwargs...) && all(dict_isapprox.(_dicts(f), _dicts(g); kwargs...))
 end
 
+constant(f::Union{ScalarAffineFunction, ScalarQuadraticFunction}, T::DataType) = constant(f)
+constant(f::Union{VectorAffineFunction, VectorQuadraticFunction}, T::DataType) = constant(f)
+
+"""
+    constant(f::SingleVariable, T::DataType)
+
+The constant term of a `SingleVariable` function is zero.
```
blegat: `T` should appear in the docstring

blegat:
`constant` could appear somewhere in the doc, e.g. after http://www.juliaopt.org/MathOptInterface.jl/dev/apireference/#MathOptInterface.output_dimension

matbesancon:
I put it after `output_dimension`

docs/src/apireference.md
```diff
@@ -216,6 +216,7 @@ Functions for getting and setting properties of sets.
 
 ```@docs
 output_dimension
+constant
```
blegat: There is a way with Documenter to only make methods with functions appear here and methods with sets appear after
http://www.juliaopt.org/MathOptInterface.jl/dev/apireference/#MathOptInterface.dimension

matbesancon: I didn't find anything for this in Documenter's doc, do you have the syntax for that? The only solution I see is a really ugly hack using `Filters`

matbesancon: And didn't find anything like it in the rest of the docs

blegat: Normally, you can do something like `constant(::MathOptInterface.AbstractSet)`

blegat: See https://juliadocs.github.io/Documenter.jl/stable/man/guide/#Filtering-included-docstrings-1 and https://github.com/JuliaDocs/Documenter.jl/issues/569

matbesancon: ok I managed to fix it. It's a very tricky process I wouldn't advise using it whenever it's avoidable

matbesancon:
ok this looks good

src/Utilities/sets.jl
```diff
@@ -1,11 +1,3 @@
-"""
-    getconstant(s::Union{MOI.EqualTo, MOI.GreaterThan, MOI.LessThan})
-
-Returns the constant of the set.
-"""
-getconstant(s::MOI.EqualTo) = s.value
-getconstant(s::MOI.GreaterThan) = s.lower
-getconstant(s::MOI.LessThan) = s.upper
```
blegat: Remove this line

matbesancon:
@blegat seems good here

PR #728, opened 2019-05-05
Update testing guideline

blegat:


docs/src/apimanual.md
```diff
@@ -1012,38 +1012,53 @@ By convention, optimizers should not be exported and should be named
...
+const optimizer = FooBar.Optimizer()
+MOI.set(optimizer, MOI.Silent(), true)
+
+@testset "SolverName" begin
+    @test MOI.get(optimizer, MOI.SolverName()) == "FooBar"
+end
+
+@testset "supports_default_copy_to" begin
+    @test MOIU.supports_default_copy_to(optimizer, false)
+    # Use `@test !...` if names are not supported
```
Peiffap:
```suggestion
    # Use `@test !...` if names are not supported.
```

docs/src/apimanual.md
```diff
@@ -1012,38 +1012,53 @@ By convention, optimizers should not be exported and should be named
...
+    @test MOI.get(optimizer, MOI.SolverName()) == "FooBar"
+end
+
+@testset "supports_default_copy_to" begin
+    @test MOIU.supports_default_copy_to(optimizer, false)
+    # Use `@test !...` if names are not supported.
+    @test MOIU.supports_default_copy_to(optimizer, true)
+end
+
+const bridged = MOIB.full_bridge_optimizer(optimizer, Float64)
```
mlubin: Explain why `full_bridge_optimizer` is often needed to pass tests, or in what situations a solver could pass tests without a bridge.

docs/src/apimanual.md
```diff
@@ -1012,38 +1012,53 @@ By convention, optimizers should not be exported and should be named
...
 
-@testset "MOI Integer Conic" begin
-    MOIT.intconictest(optimizer, config)
+@testset "Integer Conic" begin
+    MOIT.intconictest(bridged, config)
 end
 ```
 
-If the wrapper does not support building the model incrementally (i.e. with `add_variable` and `add_constraint`), the line `const optimizer = FooBarOptimizer()` can be replaced with
+If the wrapper does not support building the model incrementally (i.e. with `add_variable` and `add_constraint`), then `supports_default_copy_to` can be replaced by `supports_allocate_load` if appropriate and the line `const bridged = ...` can be replaced with
```
mlubin: Explain why this code is needed, or point to a reference on the use of bridges with solvers that don't support `default_copy_to`.

docs/src/apimanual.md
```diff
@@ -1012,38 +1012,60 @@ By convention, optimizers should not be exported and should be named
...
+    MOIT.intconictest(bridged, config)
+end
+```
+The optimizer `bridged` constructed with [`full_bridge_optimizer`](@ref)
+automatically bridges constraints that are not supported by `optimizer`
+using the bridges listed in [Bridges](@ref). It is recommended for an
+implementation of MOI to only support constraints that are natively supported
+by the solver and let bridges transform the constraint to the appropriate form.
+For this reason it is expected that tests may not pass if `optimizer` is used
+instead of `bridged`.
```
mlubin: If a solver *does* natively support a large number of constraints, like SCIP, should we encourage the solver to test itself without the bridges? 

The issue with testing only with bridges is that you could be in the situation where you believe a constraint is natively supported but it's unintentionally covered by a bridge.

blegat: We don't really restrict the functions and sets that are going to be used in `@testset`s.
Even if the solver supports many things, a bridge can be as simple as replacing a `SingleVariable` objective by a `ScalarAffineFunction` objective. I don't expect many solvers to support `SingleVariable` objectives.

If a solver want to tests that something is supported natively, I would recommend to run a specific test with `optimizer` but not a `@testset`

mlubin: > I would recommend to run a specific test with optimizer but not a @testset

Could we add this recommendation to the docs?

docs/src/apimanual.md
```diff
@@ -1053,14 +1053,24 @@ end
...
 instead of `bridged`.
 
+To test that a specific problem can be solved without bridges, a specific test can
+be run with `optimizer` instead of `bridged`. For instance
+```julia
+@testset "Interval constraints" begin
+    MOIT.linear10test(optimizer, config)
+end
+```
+checks that `optimizer` implements supports for
```
mlubin: s/supports/support/

Issue #668, opened 2019-02-09
bridge_constraint function


blegat:
Currently, bridges are instantiated with the bridge constructor but this has the disadvantage of creating method ambiguities if the bridge has three fields which forces the constructor signature to also include all the type constraints of the bridge type.
We could add a `bridge_constraint` to avoid relying on the constructor.

PR #730, opened 2019-05-10
Add bridge_constraint

blegat:
Add the `bridge_constraint` function #668 and add a section in the doc about how to implement a bridge.

Closes https://github.com/JuliaOpt/MathOptInterface.jl/issues/668


blegat:
Merged as it conflicts with https://github.com/JuliaOpt/MathOptInterface.jl/pull/731

PR #731, opened 2019-05-10
Add abstract matrix sets

blegat:
Other matrix sets are defined in SumOfSquares:
https://github.com/JuliaOpt/SumOfSquares.jl/blob/master/src/diagonally_dominant.jl
https://github.com/JuliaOpt/SumOfSquares.jl/blob/master/src/psd2x2.jl
https://github.com/JuliaOpt/SumOfSquares.jl/blob/master/src/empty_bridge.jl
and having abstract sets would allow these sets to benefit from what is implemented for the PSDSquare and PSDTriangle sets such as the SquareBridge.


docs/src/apireference.md
```diff
@@ -225,47 +225,86 @@ constant(f::VectorOfVariables, T::DataType)
...
-Semiinteger
 SOS1
 SOS2
 IndicatorSet
 ```
 
-Functions for getting and setting properties of sets.
+### Matrix sets
 
+Matrix sets are represented in their vectorized form in order to be subtypes
```
mlubin: Define what vectorized form means, or provide a reference.

Issue #693, opened 2019-03-10
Modify ListOfVariableIndices


blegat:
`NumberOfVariables` is modified to remove variables added by bridges but not `ListOfVariableIndices`.

PR #732, opened 2019-05-11
SingleVariable constraint have same index as the variable

blegat:
- [x] Update universal fallback

Closes https://github.com/JuliaOpt/MathOptInterface.jl/issues/570

### DIFF
--- a/src/Bridges/bridgeoptimizer.jl
+++ b/src/Bridges/bridgeoptimizer.jl
@@ -375,8 +375,8 @@ end
 function store_bridge(b::AbstractBridgeOptimizer, func::MOI.AbstractFunction,
                       set::MOI.AbstractSet, bridge)
     push!(b.bridges, bridge)
-    push!(b.constraint_types, (typeof(func), typeof(func)))
-    return MOI.ConstraintIndex{typeof(func), typeof(func)}(length(b.bridges))
+    push!(b.constraint_types, (typeof(func), typeof(set)))
+    return MOI.ConstraintIndex{typeof(func), typeof(set)}(length(b.bridges))
 end
 function MOI.add_constraint(b::AbstractBridgeOptimizer, f::MOI.AbstractFunction,
                             s::MOI.AbstractSet)
--- a/src/Bridges/bridgeoptimizer.jl
+++ b/src/Bridges/bridgeoptimizer.jl
@@ -67,6 +67,10 @@ Return the `AbstractBridge` used to bridge the constraint with index `ci`.
 bridge(b::AbstractBridgeOptimizer, ci::CI) = b.bridges[ci.value]
 # By convention, they should be stored in a `bridges` field using a
 # dictionary-like object.
+function bridge(b::AbstractBridgeOptimizer,
+                ci::CI{MOI.SingleVariable, S}) where S
+    return b.single_variable_constraints[(ci.value, S)]
+end
 
 # Implementation of the MOI interface for AbstractBridgeOptimizer
 
@@ -74,12 +78,14 @@ MOI.optimize!(b::AbstractBridgeOptimizer) = MOI.optimize!(b.model)
 # By convention, the model should be stored in a `model` field
 
 function MOI.is_empty(b::AbstractBridgeOptimizer)
-    return isempty(b.bridges) && isempty(b.constraint_types) && MOI.is_empty(b.model)
+    return isempty(b.bridges) && isempty(b.constraint_types) &&
+           MOI.is_empty(b.model) && isempty(b.single_variable_constraints)
 end
 function MOI.empty!(b::AbstractBridgeOptimizer)
     MOI.empty!(b.model)
     empty!(b.bridges)
     empty!(b.constraint_types)
+    empty!(b.single_variable_constraints)
 end
 function MOI.supports(b::AbstractBridgeOptimizer,
                       attr::Union{MOI.AbstractModelAttribute,
@@ -361,6 +367,17 @@ function MOI.supports_constraint(b::AbstractBridgeOptimizer,
         return MOI.supports_constraint(b.model, F, S)
     end
 end
+function store_bridge(b::AbstractBridgeOptimizer, func::MOI.SingleVariable,
+                      set::MOI.AbstractSet, bridge)
+    b.single_variable_constraints[(func.variable.value, typeof(set))] = bridge
+    return MOI.ConstraintIndex{MOI.SingleVariable, typeof(set)}(func.variable.value)
+end
+function store_bridge(b::AbstractBridgeOptimizer, func::MOI.AbstractFunction,
+                      set::MOI.AbstractSet, bridge)
+    push!(b.bridges, bridge)
+    push!(b.constraint_types, (typeof(func), typeof(func)))
+    return MOI.ConstraintIndex{typeof(func), typeof(func)}(length(b.bridges))
+end
 function MOI.add_constraint(b::AbstractBridgeOptimizer, f::MOI.AbstractFunction,
                             s::MOI.AbstractSet)
     if is_bridged(b, typeof(f), typeof(s))
@@ -370,10 +387,7 @@ function MOI.add_constraint(b::AbstractBridgeOptimizer, f::MOI.AbstractFunction,
         BridgeType = concrete_bridge_type(b, typeof(f), typeof(s))
         # `add_constraint` might throw an `UnsupportedConstraint` but no
         # modification has been done in the previous line
-        push!(b.bridges, bridge_constraint(BridgeType, b, f, s))
-        push!(b.constraint_types, (typeof(f), typeof(s)))
-        ci = MOI.ConstraintIndex{typeof(f), typeof(s)}(length(b.bridges))
-        return ci
+        return store_bridge(b, f, s, bridge_constraint(BridgeType, b, f, s))
     else
         return MOI.add_constraint(b.model, f, s)
     end
--- a/src/Bridges/bridgeoptimizer.jl
+++ b/src/Bridges/bridgeoptimizer.jl
@@ -115,17 +115,29 @@ function MOI.is_valid(b::AbstractBridgeOptimizer, ci::CI{F, S}) where {F, S}
         return MOI.is_valid(b.model, ci)
     end
 end
+function MOI.is_valid(b::AbstractBridgeOptimizer,
+                      ci::CI{MOI.SingleVariable, S}) where S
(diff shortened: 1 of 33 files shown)


src/Utilities/model.jl
```diff
@@ -318,6 +327,62 @@ function MOI.get(model::AbstractModel, ::MOI.ListOfModelAttributesSet)::Vector{M
...
+single_variable_flag(::Type{<:MOI.EqualTo}) = 0x1
+single_variable_flag(::Type{<:MOI.GreaterThan}) = 0x2
+single_variable_flag(::Type{<:MOI.LessThan}) = 0x4
+single_variable_flag(::Type{<:MOI.Interval}) = 0x8
+single_variable_flag(::Type{MOI.Integer}) = 0x10
+single_variable_flag(::Type{MOI.ZeroOne}) = 0x20
+single_variable_flag(::Type{MOI.Semicontinuous}) = 0x40
+single_variable_flag(::Type{MOI.Semiinteger}) = 0x80
+
+# Sets setting lower bound:
```
mlubin: A "lower bound" of a set doesn't make much sense in isolation, consider another name. Maybe `extract_lower_bound`?

src/Utilities/model.jl
```diff
@@ -318,6 +327,62 @@ function MOI.get(model::AbstractModel, ::MOI.ListOfModelAttributesSet)::Vector{M
...
+           !iszero(mask & LOWER_BOUND_MASK)
+           error("Cannot add $f-in-$s constraint as another constraint on",
+                 " $(f.variable) already sets a lower bound.")
+        end
+        if !iszero(flag & UPPER_BOUND_MASK) &&
+           !iszero(mask & UPPER_BOUND_MASK)
+           error("Cannot add $f-in-$s constraint as another constraint on",
+                 " $(f.variable) already sets an upper bound.")
+        end
+        # No error should be thrown now, we can modify `model`
```
mlubin: Nit: Period at the end of sentences.

src/Utilities/model.jl
```diff
@@ -354,24 +426,98 @@ end
...
+                                  index)
+    return MOI.Integer()
+end
+function _get_single_variable_set(model::AbstractModel, ::Type{MOI.ZeroOne},
+                                  index)
+    return MOI.ZeroOne()
+end
+function MOI.get(model::AbstractModel, ::MOI.ConstraintSet,
+                 ci::CI{MOI.SingleVariable, S}) where S
+    return _get_single_variable_set(model, S, ci.value)
```
mlubin: Is this missing a check that the index is valid?

src/Utilities/model.jl
```diff
@@ -354,24 +426,98 @@ end
...
+                 ci::CI{MOI.SingleVariable})
+    return MOI.SingleVariable(MOI.VariableIndex(ci.value))
+end
 function MOI.get(model::AbstractModel, ::MOI.ConstraintFunction, ci::CI)
     _getfunction(model, ci, getconstrloc(model, ci))
 end
 
+function _get_single_variable_set(model::AbstractModel, ::Type{<:MOI.EqualTo},
+                                  index)
+    # lower and upper bounds are equal, we can take either of them
```
mlubin: Nit: Capitalization and punctuation in comments.

src/Utilities/model.jl
```diff
@@ -318,6 +327,62 @@ function MOI.get(model::AbstractModel, ::MOI.ListOfModelAttributesSet)::Vector{M
...
+single_variable_flag(::Type{MOI.Integer}) = 0x10
+single_variable_flag(::Type{MOI.ZeroOne}) = 0x20
+single_variable_flag(::Type{MOI.Semicontinuous}) = 0x40
+single_variable_flag(::Type{MOI.Semiinteger}) = 0x80
+
+# Sets setting lower bound:
+lower_bound(set::MOI.EqualTo) = set.value
+lower_bound(set::Union{MOI.GreaterThan, MOI.Interval}) = set.lower
+# 0xb = 0x8 | 0x2 | 0x1
+const LOWER_BOUND_MASK = 0xb
```
mlubin: I think this should include `Semicontinuous` and `Semiinteger`. These sets have bound parameters. Same for `UPPER_BOUND_MASK`.

Also please double check that `Semicontinuous` and `Semiinteger` are covered by tests.

test/Utilities/universalfallback.jl
```diff
@@ -113,17 +113,21 @@ y, z = MOI.add_variables(uf, 2)
...
+    _saf(vi) = convert(MOI.ScalarAffineFunction{Float64}, MOI.SingleVariable(vi))
+    function _add_constraint(vi, ub)
+        return MOI.add_constraint(uf, _saf(vi), MOI.LessThan(ub))
+    end
     attr = MOIT.UnknownConstraintAttribute()
     @testset "Supported constraint" begin
-        cx = MOI.add_constraint(uf, x, MOI.LessThan(0.))
-        cy = MOI.add_constraint(uf, y, MOI.LessThan(1.))
-        cz = MOI.add_constraint(uf, z, MOI.LessThan(2.))
+        cx = _add_constraint(x, 0.)
```
mlubin: An underscore isn't a license to stop using good names :).
What about `add_saf_constraint`? (I'm uncomfortable with the `saf` abbreviation, but this would at least be an improvement.)

src/Bridges/bridgeoptimizer.jl
```diff
@@ -109,76 +115,133 @@ function MOI.is_valid(b::AbstractBridgeOptimizer, ci::CI{F, S}) where {F, S}
...
     else
         MOI.delete(b.model, ci)
     end
 end
 
 # Attributes
-function MOI.get(b::AbstractBridgeOptimizer,
-                 attr::MOI.ListOfConstraintIndices{F, S}) where {F, S}
+# List of indices of all constraints, even those bridged
+function get_all_even_bridged(
```
mlubin: `get_all_including_bridged` maybe. I was confused and asking myself what's an even bridged or odd bridged constraint.

src/Bridges/bridgeoptimizer.jl
```diff
@@ -109,76 +115,133 @@ function MOI.is_valid(b::AbstractBridgeOptimizer, ci::CI{F, S}) where {F, S}
...
-            end
+            _remove_bridged(list, bridge, attr)
         end
     end
+    for bridge in values(b.single_variable_constraints)
+        _remove_bridged(list, bridge, attr)
+    end
     return list
 end
 function _numberof(b::AbstractBridgeOptimizer, model::MOI.ModelLike,
```
mlubin: nit: `_number_of`

src/Bridges/bridgeoptimizer.jl
mlubin: Why does `_remove_bridged` have an underscore and this function doesn't?

src/Test/UnitTests/modifications.jl
```diff
@@ -1,5 +1,28 @@
 const modificationtests = Dict{String, Function}()
 
+"""
+    set_function_singlevariable(model::MOI.ModelLike, config::TestConfig)
```
mlubin: `set_function_single_variable`

src/Test/contconic.jl
```diff
@@ -608,7 +608,9 @@ function _rotatedsoc1test(model::MOI.ModelLike, config::TestConfig, abvars::Bool
         a = MOI.add_variable(model)
         b = MOI.add_variable(model)
         vc1 = MOI.add_constraint(model, MOI.SingleVariable(a), MOI.EqualTo(0.5))
+        @test vc1.value == a.value
```
mlubin: I don't think we should litter the longer tests with this check after every `SingleVariable()` constraint. This distracts from the focus of the test which is to check that the solution is correct. Other tests check the value of the indices.

src/Bridges/bridgeoptimizer.jl
blegat: I don't want to user to call `_remove_bridged` as it could corrupt the integrity of the object.
`get_all_including_bridged` provides a well defined information which could be useful for inspecting and debugging.

src/Test/contconic.jl
blegat: Which other tests ? Some solvers do not call all the tests and many implementation would get singlevariable indices equal to the variables indices just by chances so I wanted to optimize the chance to catch it by checking it in as many corner cases as possible.

mlubin: It's good to catch it in corner cases, but this can be done by creating standalone unit tests. If solvers aren't running the right tests, that's a separate issue that should be addressed.

Doing the check here interferes with the readability of the tests. Tests should have a clear purpose and limited scope so that it's easy to understand what they're doing, how to modify them, and if they're correct. This is a particularly bad experience for someone who wants to add a new test and tries to copy the existing style. Are they responsible for adding these checks for every SingleVariable constraint in the test?

blegat: > It's good to catch it in corner cases, but this can be done by creating standalone unit tests
> If solvers aren't running the right tests, that's a separate issue that should be addressed

The right tests might be using constraints that are bridged so even if the optimizer runs these tests, it will work just because the bridgeoptimizer is correct. If the solver only supports `SingleVariable` for a specific constraint and has a bug when this constraint is added in some order, it needs to be tested on that test in that order. We cannot create every corner case in unit tests.

> This is a particularly bad experience for someone who wants to add a new test and tries to copy the existing style.

I understand this concern but I feel like having a good testing coverage of indices of SingleVariable constraints is more important.
In the long term, we might replace these tests with reading from a problem described in text like in unit tests and the check could be done automatically in a common function that calls the reading and then do common checks. The coverage for some attributes is currently not ideal, e.g. all the tests pass although we have this bug https://github.com/JuliaOpt/MathOptInterface.jl/issues/693 because this bug only happens in corner cases and we don't check `MOI.ListOfVariableIndices` very often.

> Are they responsible for adding these checks for every SingleVariable constraint in the test?

I think we should but it already has a decent coverage with existing tests so it might not be critical for new tests if they do not test new scalar sets.

mlubin: > In the long term, we might replace these tests with reading from a problem described in text like in unit tests and the check could be done automatically in a common function that calls the reading and then do common checks.

Sounds like a good idea.

I'll withdraw my objection if you add an explanatory comment on the first appearance in every file of `@test vc1.value == a.value` (or the equivalent) in a large solve-and-check-solution test (like `rotatedsoc1test` and `int1test`). I think this amount of documentation is proportional to the violation of readability that these tests incur.

PR #733, opened 2019-05-13
Add RawParameter

blegat:
See https://github.com/JuliaOpt/MathOptInterface.jl/pull/387#issuecomment-466648208

### DIFF
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -356,6 +356,15 @@ given.
 """
 struct Silent <: AbstractOptimizerAttribute end
 
+"""
+    RawParameter(name::String)
+
+An optimizer attribute for the solver-specific parameter identified by `name`.
+"""
+struct RawParameter <: AbstractOptimizerAttribute
+    name::String
+end
+
 ## Model attributes
 
 """
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -357,12 +357,13 @@ given.
 struct Silent <: AbstractOptimizerAttribute end
 
 """
-    RawParameter(name::String)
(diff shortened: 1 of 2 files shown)


src/attributes.jl
```diff
@@ -356,6 +356,15 @@ given.
 """
 struct Silent <: AbstractOptimizerAttribute end
 
+"""
+    RawParameter(name::String)
+
+An optimizer attribute for the solver-specific parameter identified by `name`.
+"""
+struct RawParameter <: AbstractOptimizerAttribute
+    name::String
```
mlubin: I'm wondering if `String` is sufficient or if some solvers might want to define enums for their parameters.

blegat: Would you suggest `RawParameter{T<:Any}` or `RawParameter{T<:Union{AbstractString, Enum}}` ?

blegat: In JuMP, we would have
```julia
set_parameter(model::Model, param, value) = MOI.set(model, MOI.RawParameter(param), value)
```
so `RawParameter{T<:Any}` would work.

mlubin: Or we could do
```julia
struct RawParameter <: AbstractOptimizerAttribute
    name::Any
end
```
A lot of code that deals with RawParameters won't know what solver its dealing with, so it won't know the value of `T` for collections of RawParameters. We can save the method errors and accept the type instability. It's hard to imagine a case where type instability here causes a performance bottleneck. Additionally, key-value collections of `RawParameter` won't know the type of the values, so there will be type instability in any case.

Issue #725, opened 2019-05-02
REDUCTION_CERTIFICATE


blegat:
We removed it as we thought it was dropped from Mosek (see https://github.com/JuliaOpt/MathOptInterface.jl/pull/575#discussion_r234392774) but I often still get it (with Mosek v8), i.e. these lines are executed:
https://github.com/JuliaOpt/MosekTools.jl/blob/d274a8381b6548f5ef25ad249e28884a6a557fed/src/attributes.jl#L494
https://github.com/JuliaOpt/MosekTools.jl/blob/d274a8381b6548f5ef25ad249e28884a6a557fed/src/attributes.jl#L519
and I get an `Undefined` error since it was removed from MOI.

@ulfworsoe @erling-d-andersen Was this feature removed in Mosek v9 ?

erling-d-andersen:
I am not sure who did

  elseif solsta == MSK_SOL_STA_PRIM_ILLPOSED_CER
        MOI.NO_SOLUTION
    elseif solsta == MSK_SOL_STA_DUAL_ILLPOSED_CER
        MOI.REDUCTION_CERTIFICATE

Why do say no solution in one case and certificate in the other. To me it makes zero sense.

Both solution statuses are of course still possible. 

It v9 removed all the statuses that something with NEAR in them.

blegat:
> Why do say no solution in one case and certificate in the other

For `MSK_SOL_STA_PRIM_ILLPOSED_CER`, the `MOI.PrimalStatus` is `MOI.NO_SOLUTION` and the `MOI.DualStatus` is `MOI.REDUCTION_CERTIFICATE` because the certificate is contained in the dual solution and there is no primal solution. This is similar to `MSK_SOL_STA_PRIM_INFEAS_CER` where there is no primal solution and the dual contains an infeasibility ray.

erling-d-andersen:
Got it logic now.

erling-d-andersen:
Btw if you see ill posed then you most likely have nasty model. Not that I cannot exclude a bug, but I would be worried about seeing illposed.

blegat:
> Btw if you see ill posed then you most likely have nasty model.

Indeed, it is nasty by design. This is a model parametrized by some parameter `γ`. For `γ < γ*`, it is infeasible and for `γ > γ*`, it is feasible. I am doing bisection to find `γ*`.
[...]

PR #734, opened 2019-05-14
Add REDUCTION_CERTIFICATE

blegat:
Corresponds to Mosek statuses:
* https://docs.mosek.com/8.1/capi/constants.html#mosek.solsta.prim_illposed_cer
* https://docs.mosek.com/8.1/capi/constants.html#mosek.solsta.dual_illposed_cer

Closes https://github.com/JuliaOpt/MathOptInterface.jl/issues/725

### DIFF
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -855,15 +855,23 @@ The values indicate how to interpret the result vector.
   infeasibility.
 * `NEARLY_INFEASIBILITY_CERTIFICATE`: the result satisfies a relaxed criterion for
   a certificate of infeasibility.
+* `REDUCTION_CERTIFICATE`: the result vector is an illposed certificate. If the
+  `PrimalStatus` is `REDUCTION_CERTIFICATE`, then the primal result vector is a
+  proof that the dual problem is illposed. If the `DualStatus` is
+  `REDUCTION_CERTIFICATE`, then the dual result vector is a proof that the primal
+  is illposed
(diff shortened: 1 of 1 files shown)


blegat:
> I think you're supposed to put a hyphen in "ill-posed".

Indeed, I used illposed as this is what is used in the Mosek doc but it seems ill-posed is the correct form.

src/attributes.jl
```diff
@@ -855,15 +855,23 @@ The values indicate how to interpret the result vector.
   infeasibility.
 * `NEARLY_INFEASIBILITY_CERTIFICATE`: the result satisfies a relaxed criterion for
   a certificate of infeasibility.
+* `REDUCTION_CERTIFICATE`: the result vector is an illposed certificate. If the
```
mlubin: Agreed on "ill-posed" over "illposed".

mlubin: This is a pretty esoteric status. Could you add a link to a page/paper that describes what it means?

src/attributes.jl
```diff
@@ -855,15 +855,24 @@ The values indicate how to interpret the result vector.
   infeasibility.
 * `NEARLY_INFEASIBILITY_CERTIFICATE`: the result satisfies a relaxed criterion for
   a certificate of infeasibility.
+* `REDUCTION_CERTIFICATE`: the result vector is an ill-posed certificate; see
+  [this article](https://doi.org/10.1007/s10107-017-1169-9) for details. If the
```
mlubin: It's better to link to the freely available version: https://arxiv.org/abs/1408.4685.

Issue #727, opened 2019-05-05
constant vector


blegat:
Rename to `constant_vector`:
https://github.com/JuliaOpt/MathOptInterface.jl/blob/7c2f9065dab776e171821394c4e581957d52a6e8/src/Utilities/functions.jl#L86-L88

blegat:
Closed by https://github.com/JuliaOpt/MathOptInterface.jl/pull/740 thanks to @brilhana

PR #740, opened 2019-05-19
Rename constant to constant_vector

brilhana:
Fix for #727.


blegat:
Thanks for the contribution. You should update the parts of the code that use it now, see the test failures:
https://travis-ci.org/JuliaOpt/MathOptInterface.jl/jobs/534460476

PR #741, opened 2019-05-20
Add test for MOI.Silent

blegat:
Also fixes `empty!` and `is_empty` with universal fallback and optimizer attributes.

### DIFF
--- a/src/Test/UnitTests/attributes.jl
+++ b/src/Test/UnitTests/attributes.jl
@@ -11,6 +11,27 @@ end
 unittests["solver_name"] = solver_name
 
 """
+    silent(model::MOI.ModelLike, config::TestConfig)
+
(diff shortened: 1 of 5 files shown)


rschwarz:
So I implemented `MOI.supports(::Optimizer, ::MOI.Silent) = true` for SCIP.jl, but the new test is failing for me, when using the `MOIB.full_bridge_optimizer` on top of a `CachingOptimizer`.

But it works fine when testing on direct `SCIP.Optimizer`. Should the call be forwarded to the underlying solver automatically?

blegat:
It should be fixed by https://github.com/JuliaOpt/MathOptInterface.jl/pull/745

rschwarz:
Yes, that does fix it, thanks!

Yesterday, I was still ignorant of the difference between `AbstractModelAttribute` and `AbstractOptimizerAttribute`, so it was a worthwhile learning experience ;-)

Issue #637, opened 2019-01-19
SOCtoRSOCBridge and disable SOCtoPSDBridge


blegat:
* Given a SOC constraint `norm(x) ≤ t` where x has length `n`, the SOCtoPSDBridge creates an `n+1`x`n+1` PSD constraint.
* Given a RSOC constraint `norm(x) ≤ 2tu` where `x` has length `n`, the RSOCtoPSDBridge creates an `n+1`x`n+1` PSD constraint
*  SOC `norm(x) ≤ t` is equivalent to a RSOC `y^2 ≤ 2uv` where the length of `y` is one lower that the length of `x`

In view of this, given a SOC, it would create a smaller PSD constraint to first convert it to RSOC and then to PSD using the RSOCtoPSDBridge hence removing the need for the SOCtoPSDBridge.
In fact, the 2x2 PSDConeTriangle on `[u y; y v]` is `y^2 ≤ uv`,  `u, y ≥ 0` which is almost the RSOC (just missing the `2` in front of `uv`), this illustrates why the PSD cone is closer to the rotated SOC than the SOC.

PR #751, opened 2019-05-27
Add SOCRBridge and disable SOCtoPSDBridge

blegat:
Closes https://github.com/JuliaOpt/MathOptInterface.jl/issues/637

### DIFF
--- a/docs/src/apireference.md
+++ b/docs/src/apireference.md
@@ -707,6 +707,7 @@ Bridges.Constraint.ScalarFunctionizeBridge
 Bridges.Constraint.VectorFunctionizeBridge
 Bridges.Constraint.SplitIntervalBridge
 Bridges.Constraint.RSOCBridge
+Bridges.Constraint.SOCRBridge
 Bridges.Constraint.QuadtoSOCBridge
 Bridges.Constraint.GeoMeanBridge
 Bridges.Constraint.SquareBridge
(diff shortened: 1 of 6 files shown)


src/Bridges/socr_bridge.jl
```diff
@@ -0,0 +1,73 @@
...
+                              G::Type{<:MOI.AbstractVectorFunction},
+                              ::Type{MOI.SecondOrderCone}) where T
+    S = MOIU.promote_operation(/, T, MOIU.scalar_type(G), T)
+    Y = MOIU.promote_operation(-, T, S, S)
+    Z = MOIU.promote_operation(+, T, S, S)
+    F = MOIU.promote_operation(vcat, T, Z, Y, G)
+    return SOCRBridge{T, F, G}
+end
+
+# Attributes, Bridge acting as an model
```
frapac: Is there a typo: `an` -> `a model`?

blegat:
Currently blocked by https://github.com/JuliaLang/julia/issues/32167

chriscoey:
So RSOC and SOCR are different bridges. It's not clear from the naming which one does which. Are there slightly more verbose names we could use? If @mlubin is ok with the naming then don't worry.

mlubin:
Neither RSOC nor SOCR are standard abbreviations so we should try to avoid using either. RotatedToSOC and SOCToRotated are alternatives.

blegat:
I moved the naming to a separate issue as it would be breaking:
https://github.com/JuliaOpt/MathOptInterface.jl/issues/860
Any other objection ? The failure is now resolved thanks to https://github.com/JuliaOpt/MathOptInterface.jl/pull/834

Issue #710, opened 2019-04-12
Constrained variables and variable bridges


blegat:
Currently, if a variable belongs to a set or cone, it should be entered as:
```julia
x = MOI.add_variable(model)
MOI.add_constraint(model, x in set)
```
or
```julia
x = MOI.add_variables(model, MOI.dimension(cone))
MOI.add_constraint(model, x in cone)
```
This means that the model first need to create free variables and is only informed afterwards that the variables are in fact not free and belong to a set or cone.

This approach has two major drawbacks:
1) For many solvers, this prevents supporting `default_copy_to` hence it should implement Allocate-Load or a custom `copy_to` and it prevents the solver from being available in direct mode without a cache. This is because variables are handled differently depending on the cone in which they belong. For instance:
    * In Mosek, SDP variables are treated differently, they are created with a different functions than scalar variables and their coefficients are stored in a different matrix for constraints. Previously, scalar variables were created and then when an SDP constraint was added on them, an SDP variable was created and set equal to the scalar variable but that gave a model very slow to solve (as evidenced by https://github.com/JuliaOpt/JuMP.jl/issues/1882) and it is currently resolved by a hack that was introduced in https://github.com/JuliaOpt/MosekTools.jl/pull/22 which is to not create the variable before it is used and if it was used before being set to SDP, we delete the created scalar variables to replace them by SDP variables and move their coefficients form the matrix of scalar variables to the matrices for psd variables. I discussed about this with @erling-d-andersen and eliminating the scalar variables in presolve that would be created in this manner is unlikely to happen soon. IMO relying on presolve instead of generating the correct problem isn't the cleanest solution anyway.
    * In SDPT3, when creating constraints, the coefficients are stored in different matrices depending on their order.
    * In SeDuMi, CDCS, the coefficients in the matrix are stored in a different order depending on the cone in which the variables belong.
[...]

PR #759, opened 2019-06-08
Add variable bridges

blegat:
This allows drastic simplifications to SDOI: https://github.com/JuliaOpt/SemidefiniteOptInterface.jl/pull/30 that even questions the reason of the need to keep this SDOI layer as it is rendered trivial by constraint + variable bridges.

- [x] Update `default_copy_to`
- [x] Update Allocate-Load API
- [x] Documentation
- [x] Tests

Closes https://github.com/JuliaOpt/MathOptInterface.jl/issues/710
Closes https://github.com/JuliaOpt/MathOptInterface.jl/issues/693

### DIFF
--- a/docs/src/apimanual.md
+++ b/docs/src/apimanual.md
@@ -954,7 +954,7 @@ true
 will additionally support constrained variables in [`GreaterThan`](@ref).
 Note that these [`Bridges.Variable.SingleBridgeOptimizer`](@ref) are mainly
 used for testing bridges. It is recommended to rather use
-[`Bridges.full_bridge_optimizer`](@ref) which automatically select the
+[`Bridges.full_bridge_optimizer`](@ref), which automatically selects the
 appropriate bridges for unsupported constrained variables.
 
 #### Constraint reformulation
--- a/docs/src/apimanual.md
+++ b/docs/src/apimanual.md
@@ -959,7 +959,7 @@ appropriate bridges for unsupported constrained variables.
 
 #### Constraint reformulation
 
-A constraint often possess different equivalent formulations, but a solver may only support one of them.
+A constraint often possesses different equivalent formulations, but a solver may only support one of them.
 It would be duplicate work to implement rewritting rules in every solver wrapper for every different formulation of the constraint to express it in the form supported by the solver.
 Constraint bridges provide a way to define a rewritting rule on top of the MOI interface which can be used by any optimizer.
 Some rules also implement constraint modifications and constraint primal and duals translations.
--- a/docs/src/apimanual.md
+++ b/docs/src/apimanual.md
@@ -960,7 +960,7 @@ appropriate bridges for unsupported constrained variables.
 #### Constraint reformulation
 
 A constraint often possesses different equivalent formulations, but a solver may only support one of them.
-It would be duplicate work to implement rewritting rules in every solver wrapper for every different formulation of the constraint to express it in the form supported by the solver.
+It would be duplicate work to implement rewriting rules in every solver wrapper for every different formulation of the constraint to express it in the form supported by the solver.
 Constraint bridges provide a way to define a rewritting rule on top of the MOI interface which can be used by any optimizer.
 Some rules also implement constraint modifications and constraint primal and duals translations.
 
--- a/docs/src/apimanual.md
+++ b/docs/src/apimanual.md
@@ -4,6 +4,7 @@ DocTestSetup = quote
     using MathOptInterface
     const MOI = MathOptInterface
 end
+DocTestFilters = [r"MathOptInterface|MOI"]
 ```
 
 # Manual
@@ -931,6 +932,31 @@ between `y_i` and the vector of scalar-valued quadratic functions.
 
 ### Automatic reformulation
 
+#### Variable reformulation
+
+A variable is often created constrained in a set unsupported by the solver while
+it could be parametrized by variables constrained in supported sets.
+For example, the [`Bridges.Variable.VectorizeBridge`](@ref) defines the
+reformulation of a constrained variable in [`GreaterThan`](@ref) into a
+constrained vector of one variable in [`Nonnegatives`](@ref).
+The `Bridges.Variable.Vectorize` is the bridge optimizer that applies the
+[`Bridges.Variable.VectorizeBridge`](@ref) rewritting rule. Given an optimizer
+`optimizer` implementing constrained variables in [`Nonnegatives`](@ref) and,
+the optimizer
+```jldoctest; setup=:(optimizer = MOI.Utilities.Model{Float64}())
+bridged_optimizer = MOI.Bridges.Variable.Vectorize{Float64}(optimizer)
+MOI.supports_constraint(bridged_optimizer, MOI.SingleVariable, MOI.GreaterThan{Float64})
+
+# output
+
+true
+```
+will additionally support constrained variables in [`GreaterThan`](@ref).
+Note that these [`Bridges.Variable.SingleBridgeOptimizer`](@ref) are mainly
+used for testing bridges. It is recommended to rather use
+[`Bridges.full_bridge_optimizer`](@ref) which automatically select the
+appropriate bridges for unsupported constrained variables.
+
 #### Constraint reformulation
 
 A constraint often possess different equivalent formulations, but a solver may only support one of them.
@@ -938,13 +964,28 @@ It would be duplicate work to implement rewritting rules in every solver wrapper
 Constraint bridges provide a way to define a rewritting rule on top of the MOI interface which can be used by any optimizer.
 Some rules also implement constraint modifications and constraint primal and duals translations.
 
-For example, the `SplitIntervalBridge` defines the reformulation of a `ScalarAffineFunction`-in-`Interval` constraint into a `ScalarAffineFunction`-in-`GreaterThan` and a `ScalarAffineFunction`-in-`LessThan` constraint.
-The `SplitInterval` is the bridge optimizer that applies the `SplitIntervalBridge` rewritting rule.
-Given an optimizer `optimizer` implementing `ScalarAffineFunction`-in-`GreaterThan` and `ScalarAffineFunction`-in-`LessThan`, the optimizer
-```
-bridgedoptimizer = SplitInterval(optimizer)
+For example, the [`Bridges.Constraint.SplitIntervalBridge`](@ref) defines the
+reformulation of a [`ScalarAffineFunction`](@ref)-in-[`Interval`](@ref)
+constraint into a [`ScalarAffineFunction`](@ref)-in-[`GreaterThan`](@ref) and a
+[`ScalarAffineFunction`](@ref)-in-[`LessThan`](@ref) constraint.
+The `Bridges.Constraint.SplitInterval` is the bridge optimizer that applies the
+[`Bridges.Constraint.SplitIntervalBridge`](@ref) rewritting rule. Given an
+optimizer `optimizer` implementing [`ScalarAffineFunction`](@ref)-in-[`GreaterThan`](@ref)
+and [`ScalarAffineFunction`](@ref)-in-[`LessThan`](@ref), the optimizer
+```jldoctest; setup=:(optimizer = MOI.Utilities.Model{Float64}())
+bridged_optimizer = MOI.Bridges.Constraint.SplitInterval{Float64}(optimizer)
+MOI.supports_constraint(bridged_optimizer, MOI.ScalarAffineFunction{Float64}, MOI.Interval{Float64})
+
+# output
+
+true
 ```
 will additionally support `ScalarAffineFunction`-in-`Interval`.
+Note that these [`Bridges.Constraint.SingleBridgeOptimizer`](@ref) are mainly
+used for testing bridges. It is recommended to rather use
+[`Bridges.full_bridge_optimizer`](@ref) which automatically select the
+appropriate constraint bridges for unsupported constraints.
+
 
 ## Implementing a solver interface
 
--- a/docs/src/apimanual.md
+++ b/docs/src/apimanual.md
@@ -940,7 +940,7 @@ For example, the [`Bridges.Variable.VectorizeBridge`](@ref) defines the
 reformulation of a constrained variable in [`GreaterThan`](@ref) into a
 constrained vector of one variable in [`Nonnegatives`](@ref).
 The `Bridges.Variable.Vectorize` is the bridge optimizer that applies the
-[`Bridges.Variable.VectorizeBridge`](@ref) rewritting rule. Given an optimizer
+[`Bridges.Variable.VectorizeBridge`](@ref) rewriting rule. Given an optimizer
 `optimizer` implementing constrained variables in [`Nonnegatives`](@ref) and,
 the optimizer
 ```jldoctest; setup=:(optimizer = MOI.Utilities.Model{Float64}())
--- a/docs/src/apimanual.md
+++ b/docs/src/apimanual.md
@@ -941,7 +941,7 @@ reformulation of a constrained variable in [`GreaterThan`](@ref) into a
 constrained vector of one variable in [`Nonnegatives`](@ref).
 The `Bridges.Variable.Vectorize` is the bridge optimizer that applies the
 [`Bridges.Variable.VectorizeBridge`](@ref) rewriting rule. Given an optimizer
-`optimizer` implementing constrained variables in [`Nonnegatives`](@ref) and,
+`optimizer` implementing constrained variables in [`Nonnegatives`](@ref),
 the optimizer
 ```jldoctest; setup=:(optimizer = MOI.Utilities.Model{Float64}())
 bridged_optimizer = MOI.Bridges.Variable.Vectorize{Float64}(optimizer)
@@ -961,7 +961,7 @@ appropriate bridges for unsupported constrained variables.
 
 A constraint often possesses different equivalent formulations, but a solver may only support one of them.
 It would be duplicate work to implement rewriting rules in every solver wrapper for every different formulation of the constraint to express it in the form supported by the solver.
-Constraint bridges provide a way to define a rewritting rule on top of the MOI interface which can be used by any optimizer.
+Constraint bridges provide a way to define a rewriting rule on top of the MOI interface which can be used by any optimizer.
 Some rules also implement constraint modifications and constraint primal and duals translations.
 
 For example, the [`Bridges.Constraint.SplitIntervalBridge`](@ref) defines the
@@ -969,7 +969,7 @@ reformulation of a [`ScalarAffineFunction`](@ref)-in-[`Interval`](@ref)
 constraint into a [`ScalarAffineFunction`](@ref)-in-[`GreaterThan`](@ref) and a
 [`ScalarAffineFunction`](@ref)-in-[`LessThan`](@ref) constraint.
 The `Bridges.Constraint.SplitInterval` is the bridge optimizer that applies the
-[`Bridges.Constraint.SplitIntervalBridge`](@ref) rewritting rule. Given an
+[`Bridges.Constraint.SplitIntervalBridge`](@ref) rewriting rule. Given an
 optimizer `optimizer` implementing [`ScalarAffineFunction`](@ref)-in-[`GreaterThan`](@ref)
 and [`ScalarAffineFunction`](@ref)-in-[`LessThan`](@ref), the optimizer
 ```jldoctest; setup=:(optimizer = MOI.Utilities.Model{Float64}())
@@ -983,7 +983,7 @@ true
 will additionally support `ScalarAffineFunction`-in-`Interval`.
 Note that these [`Bridges.Constraint.SingleBridgeOptimizer`](@ref) are mainly
 used for testing bridges. It is recommended to rather use
-[`Bridges.full_bridge_optimizer`](@ref) which automatically select the
+[`Bridges.full_bridge_optimizer`](@ref) which automatically selects the
 appropriate constraint bridges for unsupported constraints.
 
 
--- a/docs/src/apireference.md
+++ b/docs/src/apireference.md
@@ -480,8 +480,9 @@ Utilities.@model
 
 ## Bridges
 
-Bridges can be used for automatic reformulation of constrained variables or
-constraints into equivalent formulations using constrained variables and
+Bridges can be used for automatic reformulation of constrained variables (i.e.
+variables added with [`add_constrained_variable`](@ref)/[`add_constrained_variables`](@ref))
+or constraints into equivalent formulations using constrained variables and
 constraints of different types. There are two important concepts to distinguish:
 * [`Bridges.AbstractBridge`](@ref)s are recipes implementing a specific
   reformulation. Bridges are not directly subtypes of
@@ -496,8 +497,8 @@ constraints of different types. There are two important concepts to distinguish:
   [`Bridges.add_bridge`](@ref). [`Bridges.full_bridge_optimizer`](@ref) wraps a
   model in a [`Bridges.LazyBridgeOptimizer`](@ref) where all the bridges defined
   in MOI are added. This is the recommended way to use bridges in the
-  [Testing guideline](@ref), and JuMP automatically calls this function when
-  attaching an optimizer.
+  [Testing guideline](@ref), and JuMP automatically calls
+  [`Bridges.full_bridge_optimizer`](@ref) when attaching an optimizer.
 
 ```@docs
 Bridges.AbstractBridge
--- a/docs/src/apireference.md
+++ b/docs/src/apireference.md
@@ -74,7 +74,6 @@ AbstractSubmittable
 submit
 ```
 
-
 ## Model Interface
 
 ```@docs
@@ -192,8 +191,8 @@ delete(::ModelLike, ::Index)
 [`add_variables`](@ref) while *constrained variables* are
 the variables created with [`add_constrained_variable`](@ref) or
 [`add_constrained_variables`](@ref). Adding constrained variables instead of
-constraining free variables with [`add_constraint`](@ref) allows Variable
-bridges to be used.
+constraining free variables with [`add_constraint`](@ref) allows
+[Variable bridges](@ref) to be used.
 Note further that free variables that are constrained with
 [`add_constraint`](@ref) may be copied by [`copy_to`](@ref) with
 [`add_constrained_variable`](@ref) or [`add_constrained_variables`](@ref) by the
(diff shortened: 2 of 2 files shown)


src/Bridges/Constraint/single_bridge_optimizer.jl
```diff
@@ -25,13 +25,30 @@ function SingleBridgeOptimizer{BT}(model::OT) where {BT, OT <: MOI.ModelLike}
         Dict{Tuple{Int64, DataType}, AbstractBridge}())
 end
 
-function MOIB.supports_bridging_constraint(b::SingleBridgeOptimizer{BT},
-                                      F::Type{<:MOI.AbstractFunction},
-                                      S::Type{<:MOI.AbstractSet}) where BT
+function bridges(bridge::MOI.Bridges.AbstractBridgeOptimizer)
+    return MOI.Bridges.LazyFilter(bridge -> bridge !== nothing, Union{Nothing, AbstractBridge}[])
```
mlubin: Why filter an empty array?

src/Bridges/Constraint/single_bridge_optimizer.jl
```diff
@@ -25,13 +25,30 @@ function SingleBridgeOptimizer{BT}(model::OT) where {BT, OT <: MOI.ModelLike}
...
+end
+function bridges(bridge::SingleBridgeOptimizer)
+    return MOI.Bridges.LazyFilter(bridge -> bridge !== nothing, bridge.bridges)
+end
+
+function single_variable_constraints(bridge::MOI.Bridges.AbstractBridgeOptimizer)
```
mlubin: It's not clear if these are internal helper functions or part of the interface.

src/Bridges/Variable/Variable.jl
```diff
@@ -0,0 +1,18 @@
+module Variable
+
```
mlubin: Please add documentation on the design of variable bridges: what's the intention, how do they work, how do you add a new variable bridge. It's hard to review the code without this.

src/Bridges/Constraint/det.jl
```diff
@@ -98,7 +98,8 @@ function bridge_constraint(::Type{LogDetBridge{T}}, model,
 end
 
 MOI.supports_constraint(::Type{LogDetBridge{T}}, ::Type{<:Union{MOI.VectorOfVariables, MOI.VectorAffineFunction{T}}}, ::Type{MOI.LogDetConeTriangle}) where T = true
-added_constraint_types(::Type{LogDetBridge{T}}, ::Type{<:Union{MOI.VectorOfVariables, MOI.VectorAffineFunction{T}}}, ::Type{MOI.LogDetConeTriangle}) where T = [(MOI.VectorAffineFunction{T}, MOI.PositiveSemidefiniteConeTriangle), (MOI.VectorAffineFunction{T}, MOI.ExponentialCone), (MOI.ScalarAffineFunction{T}, MOI.LessThan{T})]
+MOIB.added_constrained_variable_types(::Type{<:LogDetBridge}) = Tuple{DataType}[]
```
odow: Is `Tuple{DataType}` really intended? That means just things like `(Int,)`? If so, why not just `DataType[]`? Maybe the return type should be specified in more detail in the doctoring.

odow: Even better, why not
```
Type{<:MOI.AbstractSet}[]
```

src/Bridges/Constraint/map.jl
```diff
@@ -0,0 +1,188 @@
+struct Map <: AbstractDict{MOI.VariableIndex, AbstractBridge}
+    # Constraint Index of bridged constraint -> Bridge.
+    # It is set to `nothing` when the constraint is deleted.
+    bridges::Vector{Union{Nothing, AbstractBridge}}
+    # Constraint Index of bridged constraint -> Constraint type.
+    constraint_types::Vector{Tuple{DataType, DataType}}
```
odow: Ditto here. I think we can do better than `DataType`, if only for readability sake.

```Julia
const TupleConstraintType = Tuple{<:MOI.AbstractFunction, <:MOI.AbstractSet}

constraint_types::Vector{TupleConstraintType}
```

src/Bridges/Constraint/map.jl
```diff
@@ -0,0 +1,188 @@
+struct Map <: AbstractDict{MOI.VariableIndex, AbstractBridge}
+    # Constraint Index of bridged constraint -> Bridge.
+    # It is set to `nothing` when the constraint is deleted.
+    bridges::Vector{Union{Nothing, AbstractBridge}}
```
odow: Potentially a use-case for `CleverDict`?

src/Bridges/Constraint/map.jl
```diff
@@ -0,0 +1,188 @@
...
+    map.single_variable_constraints[(func.variable.value, typeof(set))] = bridge
+    return MOI.ConstraintIndex{MOI.SingleVariable, typeof(set)}(func.variable.value)
+end
+function add_key_for_bridge(map::Map, bridge::AbstractBridge,
+                            func::MOI.VectorOfVariables, set::MOI.AbstractVectorSet)
+    index = first(func.variables).value
+    map.vector_of_variables_constraints[(index, typeof(set))] = bridge
+    return MOI.ConstraintIndex{MOI.VectorOfVariables, typeof(set)}(index)
+end
+function _iterate3(map::Map, elem_state=iterate(map.vector_of_variables_constraints))
```
odow: Documentation for what `1`, `2`, and `3` are? Or use more descriptive names?

src/Bridges/Constraint/quad_to_soc.jl
```diff
@@ -153,7 +153,8 @@ function MOI.supports_constraint(::Type{QuadtoSOCBridge{T}},
                                                 MOI.GreaterThan{T}}}) where T
     return true
 end
-function added_constraint_types(::Type{QuadtoSOCBridge{T}}) where T
+MOIB.added_constrained_variable_types(::Type{<:QuadtoSOCBridge}) = Tuple{DataType}[]
```
odow: Why is it necessary to add these to every bridge? Shouldn't there be a generic fallback?

src/Bridges/bridge.jl
```diff
@@ -0,0 +1,17 @@
+abstract type AbstractBridge end
+
+"""
+    added_constraint_types(BT::Type{<:Variable.AbstractBridge})::Bool
+
+Return a list of the types of constrained variables that bridges of concrete
+type `BT` add. This is used by the [`LazyBridgeOptimizer`](@ref).
+"""
+function added_constrained_variable_types end
```
odow: This should be a generic fallback
```Julia
function added_constrained_variable_Types(::Type{<:Variable.AbstractBridge})
    return Tuple{DataType}[]
end
```

src/Bridges/bridge.jl
```diff
@@ -0,0 +1,17 @@
+abstract type AbstractBridge end
+
+"""
+    added_constraint_types(BT::Type{<:Variable.AbstractBridge})::Bool
```
odow: Docstring is wrong. Doesn't return `Bool`.

src/Bridges/bridge.jl
```diff
@@ -0,0 +1,17 @@
...
+"""
+function added_constrained_variable_types end
+
+"""
+    added_constraint_types(BT::Type{<:Constraint.AbstractBridge})::Bool
+
+Return a list of the types of constraints that bridges of concrete type `BT`
+add. This is used by the [`LazyBridgeOptimizer`](@ref).
+"""
+function added_constraint_types end
```
odow: Ditto fallback

src/variables.jl
```diff
@@ -33,3 +33,26 @@ A [`AddVariableNotAllowed`](@ref) error is thrown if adding variables cannot be
 done in the current state of the model `model`.
 """
 add_variable(model::ModelLike) = throw(AddVariableNotAllowed())
+
+#function supports_constrained_variables(
+#    model::ModelLike, S::Type{<:AbstractVectorSet})
+#    return MOI.supports_constraint(model, MOI.VectorOfVariables, S)
+#end
+function add_constrained_variable(model::ModelLike, set::AbstractScalarSet)
```
odow: How does one add a variable with multiple sets? Like `LessThan`, `GreaterThan`, and `ZeroOne`? How does JuMP know when to call `add_constrained_variable`, or `add_variable` and then `add_constraint`?

src/Bridges/Constraint/det.jl
blegat: `Tuple{DataType}` for constrained variables is for consistency with `Tuple{DataType, DataType}` for constraints. I didn't know `Type{<:MOI.AbstractSet}[]` was possible.

src/Bridges/Constraint/map.jl
blegat: Yes, indeed, still not convinced that `Dict` is faster than `Vector{Union{Nothing, T}}` but early optimization is ...

src/Bridges/bridge.jl
blegat: It may be dangerous because if someone forget to implement the method, the only effect is that Bellman-Ford might work incorrectly.
That may produce nasty bugs

src/variables.jl
blegat: I would suggest the following:
```julia
@variable(model, x[1:n] in set)
```
does `MOI.add_constrained_variables(model, set)`.
If there are multiple set, you can use `add_constrained_variables` for one of the sets and `add_constraint` for the other ones.
There are cases where there is an impact on the formulation depending on which one you use in `add_constrained_variables` but the only cases impacted are cases which were worse before variable bridged and `add_constrained_variables`.
* If you use `add_constrained_variables` and it bridges the variable, then `add_constraint` on the bridged variable cannot work since it is an affine expression for the solver, not just a variable. It is then *force-bridged* by `FunctionizeBridge` into affine function-in-set. So it is better to do `add_constrained_variables` with a set `S` such that `supports_constraint(MOI.VectorOfVariables, S)` is true.
* For Mosek, SDP variables can only be added with `add_constrained_variables` so it should have priority.
For
```julia
@variable(model, -1 <= x[1:n] <= 1, Int)
```
it does not really matter for which set we use `add_constrained_variables` if the LP solver does not use variable bridges.

odow: > For Mosek, SDP variables can only be added with add_constrained_variables so it should have priority.

How does JuMP determine this priority? How can we add a PSD matrix that is constrained to be integer?

blegat: You cannot do that with Mosek. If the user tries to create VectorOfVariables-in-PSD with `add_constraint` it currently work because of a hack in the wrapper, i.e. I delete the variables created and add PSD ones but I am tempted to remove that once `add_constrained_variables` exists.
If the user tried to set PSD variables as binary, they will get a useful error from Mosek saying that PSD variables are handled differently and can't be binary.

odow: > You cannot do that with Mosek.

Sure, but what about a different solver in future.

blegat: > How does JuMP determine this priority?

I would say that in
```julia
@variable(model, x[1:3] in set, binary = true)
```
you do `MOI.add_constrained_variables(model, set)` and then you set the `binary` with `MOI.add_constraint` because the `set` variable-constraint is more likely to be bridged than the `MOI.ZeroOne` variable-constraint.
Another approach is to check with `MOI.supports_constraint` which variable-constraint is supported and use `MOI.add_constrained_variables` for one that is not supported so that it can be variable-bridged.

mlubin:
@blegat Do you have a timeline for getting this to a reviewable state? There are a lot of things blocked by the MOI 0.9 release (e.g., transition to new solver wrappers without LQOI, getting rid of `with_optimizer`, etc.).

blegat:
@mlubin I should be able to finish it in the next few days.

src/Bridges/Constraint/map.jl
blegat: This doesn't work because the elements are not functions, they are function *types*.
with your code, I get
```julia
julia> constraint_types[1] = (MOI.SingleVariable, MOI.EqualTo{Float64})
ERROR: MethodError: Cannot `convert` an object of type Type{MathOptInterface.SingleVariable} to an object of type MathOptInterface.AbstractFunction
Closest candidates are:
  convert(::Type{T}, ::T) where T at essentials.jl:154
Stacktrace:
 [1] convert(::Type{Tuple{#s2,#s1} where #s1<:MathOptInterface.AbstractSet where #s2<:MathOptInterface.AbstractFunction}, ::Tuple{DataType,DataType}) at ./essentials.jl:275
 [2] setindex!(::Array{Tuple{#s2,#s1} where #s1<:MathOptInterface.AbstractSet where #s2<:MathOptInterface.AbstractFunction,1}, ::Tuple{DataType,DataType}, ::Int64) at ./array.jl:767
 [3] top-level scope at none:0
```

src/Bridges/Constraint/det.jl
blegat: `Type{<:MOI.AbstractSet}[]` seems to work

blegat:
Ready from my side. I don't plan any change except trying @odow's suggestion: https://github.com/JuliaOpt/MathOptInterface.jl/pull/759#discussion_r295392751.
Last call for comments :)

src/Bridges/Constraint/det.jl
blegat: There seems to be issues implementing the solution:
```julia
julia> d = Dict{Tuple{Type{<:Integer}}, Int}()
Dict{Tuple{Type{#s712} where #s712<:Integer},Int64} with 0 entries

julia> d[(Int,)] = 1
ERROR: StackOverflowError:
Stacktrace:
 [1] convert(::Type{Tuple{Type{#s712} where #s712<:Integer}}, ::Tuple{DataType}) at ./essentials.jl:275
 [2] setindex!(::Dict{Tuple{Type{#s712} where #s712<:Integer},Int64}, ::Int64, ::Tuple{DataType}) at ./dict.jl:373
 [3] setindex!(::Dict{Tuple{Type{#s712} where #s712<:Integer},Int64}, ::Int64, ::Tuple{DataType}) at ./dict.jl:377 (repeats 9994 times)
 [4] top-level scope at none:0
```
It would be interesting this figure this out since we use a lot of `DataType` in MOI and it would be more readable to use `Tuple{<:...}` indeed.

blegat: Moving to a separate issue: https://github.com/JuliaOpt/MathOptInterface.jl/issues/783

mlubin:
I will start reviewing this soon.

blegat:
I have checked, the branch for MOI v0.9 of the following solvers passes with this branch: GLPK, Mosek, SCS, ECOS, CSDP, SDPA, SeDuMi and CDCS.

src/variables.jl
```diff
@@ -33,3 +33,98 @@ A [`AddVariableNotAllowed`](@ref) error is thrown if adding variables cannot be
...
+        model::ModelLike,
+        set::AbstractScalarSet
+    )::Tuple{MOI.VariableIndex,
+             MOI.ConstraintIndex{MOI.SingleVariable, typeof(set)}}
+
+Add to `model` a scalar variable constrained to belong to `set`, returning the
+index of the variable created and the index of the constraint constraining the
+variable to belong to `set`.
+
+By default, this function falls back to created a free variable with
```
mlubin: s/to created/to creating/

src/variables.jl
```diff
@@ -33,3 +33,98 @@ A [`AddVariableNotAllowed`](@ref) error is thrown if adding variables cannot be
...
+[`add_variable`](@ref) and then constrain it to belong to `set` with
```
mlubin: s/constrain/constraining/

src/variables.jl
```diff
@@ -33,3 +33,98 @@ A [`AddVariableNotAllowed`](@ref) error is thrown if adding variables cannot be
...
+[`add_constraint`](@ref). Therefore, there is no need to implement this function
+if `model` supports creating free variables and supports transforming free
+variables into variables in `set`. However,
+
+* if `model` does not support created free variables, it should only implement
```
mlubin: s/created/creating/

src/variables.jl
```diff
@@ -33,3 +33,98 @@ A [`AddVariableNotAllowed`](@ref) error is thrown if adding variables cannot be
...
+    )::Tuple{MOI.VariableIndex,
+             MOI.ConstraintIndex{MOI.SingleVariable, typeof(set)}}
+
+Add to `model` a scalar variable constrained to belong to `set`, returning the
+index of the variable created and the index of the constraint constraining the
+variable to belong to `set`.
+
+By default, this function falls back to created a free variable with
+[`add_variable`](@ref) and then constrain it to belong to `set` with
+[`add_constraint`](@ref). Therefore, there is no need to implement this function
```
mlubin: The discussion below seems like it should be moved to the "implementing a solver interface" section in the manual.

src/variables.jl
```diff
@@ -33,3 +33,98 @@ A [`AddVariableNotAllowed`](@ref) error is thrown if adding variables cannot be
...
+        model::ModelLike,
+        set::AbstractVectorSet
+    )::Tuple{Vector{MOI.VariableIndex},
+             MOI.ConstraintIndex{MOI.VectorOfVariables, typeof(set)}}
+
+Add to `model` a vector of variables constrained to belong to `set`, returning
+the indices of the variables created and the index of the constraint
+constraining the vector of variables to belong to `set`.
+
+By default, this function falls back to created free variables with
```
mlubin: s/created/creating/

src/variables.jl
```diff
@@ -33,3 +33,98 @@ A [`AddVariableNotAllowed`](@ref) error is thrown if adding variables cannot be
...
+[`add_variables`](@ref) and then constrain it to belong to `set` with
```
mlubin: s/constrain/constraining/

docs/src/apireference.md
```diff
@@ -437,16 +460,94 @@ Utilities.@model
...
+Bridges.Variable.add_all_bridges
+```
+
+### Constraint bridges
+
+When constraints are added with [`add_constraint`](@ref), constraint bridges
+allow to return *bridged constraints* that do not correspond to
+constraints of the underlying model. These constraints were enforced by an
+equivalent formulation that added constraints (and possibly also variables) in
+the underlying model.
```
mlubin: Consider adding an example here as well.

src/Bridges/Variable/single_bridge_optimizer.jl
```diff
@@ -0,0 +1,50 @@
+"""
+    SingleBridgeOptimizer{BT<:AbstractBridge, OT<:MOI.ModelLike} <: AbstractBridgeOptimizer
+
+The `SingleBridgeOptimizer` bridges any constrained variables supported by the
+bridge `BT`. This is in contrast with the [`MathOptInterface.Bridges.LazyBridgeOptimizer`](@ref)
+which only bridges the constrained variables that are unsupported by the internal model,
+even if they are supported by one of its bridges.
+
+!!! note
+    Two bridge optimizers using variable bridges cannot be used together as both
```
mlubin: What's the purpose of the `SingleBridgeOptimizer` given this restriction? Is it solely for testing?

src/constraints.jl
```diff
@@ -6,8 +6,29 @@
 Return a `Bool` indicating whether `model` supports `F`-in-`S` constraints, that is,
 `copy_to(model, src)` does not return `CopyUnsupportedConstraint` when `src` contains `F`-in-`S` constraints.
 If `F`-in-`S` constraints are only not supported in specific circumstances, e.g. `F`-in-`S` constraints cannot be combined with another type of constraint, it should still return `true`.
-"""
-supports_constraint(model::ModelLike, ::Type{<:AbstractFunction}, ::Type{<:AbstractSet}) = false
+
+Even if free variables are not explicitly set to be free by calling
+[`add_constraint`](@ref) with the set [`Reals`](@ref), `supports_constraint(::ModelLike`
```
mlubin: Is this `supports_constraint(::ModelLike` supposed to be here?

src/constraints.jl
```diff
@@ -6,8 +6,29 @@
...
+
+    supports_constraint(model::ModelLike, ::Type{VectorOfVariables}, ::Type{Reals})::Bool
+
+Return a `Bool` indicating whether `model` supports free variables.
+By default, this method returns `true` so it should only be implemented if
+`model` does not support free variables.
+
+Not that free variables are not explicitly set to be free by calling
```
mlubin: s/Not/Note/

src/constraints.jl
```diff
@@ -6,8 +6,29 @@
 Return a `Bool` indicating whether `model` supports `F`-in-`S` constraints, that is,
 `copy_to(model, src)` does not return `CopyUnsupportedConstraint` when `src` contains `F`-in-`S` constraints.
 If `F`-in-`S` constraints are only not supported in specific circumstances, e.g. `F`-in-`S` constraints cannot be combined with another type of constraint, it should still return `true`.
-"""
-supports_constraint(model::ModelLike, ::Type{<:AbstractFunction}, ::Type{<:AbstractSet}) = false
+
+Even if free variables are not explicitly set to be free by calling
```
mlubin: Where are free variables defined?

src/Bridges/bridge_optimizer.jl
```diff
@@ -463,26 +692,241 @@ function MOI.add_constraints(b::AbstractBridgeOptimizer, f::Vector{F},
...
+        # It could be solved by force-bridging the attribues (e.g. objective).
+        error("Using bridged variable in `SingleVariable` function.")
+    end
+    return func
+end
+bridged_function(bridge::AbstractBridgeOptimizer, value) = value
+
+"""
+    variable_unbridged_function(b::AbstractBridgeOptimizer,
+                              vi::MOI.VariableIndex)
```
mlubin: Missing space for alignment

src/Bridges/bridge_optimizer.jl
```diff
@@ -463,26 +692,241 @@ function MOI.add_constraints(b::AbstractBridgeOptimizer, f::Vector{F},
...
+        return unbridged_function(b, func)
+    end
+end
+
+"""
+    unbridged_function(b::AbstractBridgeOptimizer,
+                       func::MOI.AbstractFunction)::typeof(func)
+
+Return a function of variables of `b` that equals `func`. That is,
+internals variable of `b.model` created by a but not visible to the user, are
```
mlubin: Created by a what?

src/Bridges/bridge_optimizer.jl
```diff
@@ -15,31 +15,70 @@ abstract type AbstractBridgeOptimizer <: MOI.AbstractOptimizer end
...
 """
 function is_bridged end
-# Syntactic sugar
-function is_bridged(b::AbstractBridgeOptimizer, ::Type{CI{F, S}}) where {F, S}
+
+"""
+    is_bridged(b::AbstractBridgeOptimizer, vi::MOI.VariableIndex)
+
+Return a `Bool` indicating whether `vi` is bridged. The variable is said to be
+bridged if it a variable of `b` but not a variable of `b.model`.
```
mlubin: s/if it a/if it is a/

src/Bridges/bridge_optimizer.jl
```diff
@@ -15,31 +15,70 @@ abstract type AbstractBridgeOptimizer <: MOI.AbstractOptimizer end
...
+"""
+is_bridged(::AbstractBridgeOptimizer, vi::MOI.VariableIndex) = vi.value < 0
+
+"""
+    is_bridged(b::AbstractBridgeOptimizer, ci::MOI.ConstraintIndex)
+
+Return a `Bool` indicating whether `ci` is bridged. The constraint is said to be
+bridged if it a constraint of `b` but not a constraint of `b.model`.
```
mlubin: s/if it a/if it is a/

src/Bridges/bridge_optimizer.jl
```diff
@@ -54,16 +93,46 @@ This function should only be called if `is_bridged(b, F, S)`.
...
 
 """
-    bridge(b::AbstractBridgeOptimizer, ci::CI)
+    is_variable_bridged(b::AbstractBridgeOptimizer,
+                        ci::MOI.ConstraintIndex)
+
+Returns whether `ci` is the constraint of a bridged constrained variable. That
+is, if it was returned by `Variable.add_key_for_bridge` or
+`Variable.add_keys_for_bridge`.
+# If `ci.value < 0` is
```
mlubin: Why is there a comment inside the docstring?

src/Bridges/bridge_optimizer.jl
```diff
@@ -54,16 +93,46 @@ This function should only be called if `is_bridged(b, F, S)`.
...
+# but if `S` is `SingleVariable` or `VectorOfVariables`, it can also simply be a constraint on a
+# bridged variable.
+"""
+is_variable_bridged(::AbstractBridgeOptimizer, ::MOI.ConstraintIndex) = false
+function is_variable_bridged(
+    b::AbstractBridgeOptimizer,
+    ci::MOI.ConstraintIndex{<:Union{MOI.SingleVariable, MOI.VectorOfVariables}})
+    # It can be a constraint of bridged constrained variables so we `check` with
```
mlubin: s/of/corresponding to/

src/Bridges/bridge_optimizer.jl
```diff
@@ -98,79 +173,175 @@ function MOIU.supports_default_copy_to(b::AbstractBridgeOptimizer,
...
         return MOI.is_valid(b.model, ci)
     end
 end
-function MOI.is_valid(b::AbstractBridgeOptimizer,
-                      ci::CI{MOI.SingleVariable, S}) where S
-    if is_bridged(b, typeof(ci))
-        return haskey(b.single_variable_constraints, (ci.value, S))
+function MOI.delete(b::AbstractBridgeOptimizer, vis::Vector{MOI.VariableIndex})
+    if Constraint.has_bridges(Constraint.bridges(b))
+        # Delete all `MOI.VectorOfVariables` constraint of these variables
```
mlubin: s/constraint/constraints/

src/Bridges/bridge_optimizer.jl
```diff
@@ -98,79 +173,175 @@ function MOIU.supports_default_copy_to(b::AbstractBridgeOptimizer,
...
+        for ci in Constraint.variable_constraints(Constraint.bridges(b), vis)
+            if vis == MOI.get(b, MOI.ConstraintFunction(), ci).variables
+                MOI.delete(b, ci)
+            end
+        end
+        # Delete all `MOI.SingleVariable` constraint of these variables
```
mlubin: s/constraint/constraints/

src/Bridges/bridge_optimizer.jl
mlubin: Nit: periods at the end of sentences

src/Bridges/bridge_optimizer.jl
```diff
@@ -98,79 +173,175 @@ function MOIU.supports_default_copy_to(b::AbstractBridgeOptimizer,
...
-    if is_bridged(b, typeof(ci))
-        if !MOI.is_valid(b, ci)
-            throw(MOI.InvalidIndex(ci))
+function MOI.delete(b::AbstractBridgeOptimizer, ci::MOI.ConstraintIndex)
+    if is_bridged(b, ci)
+        MOI.throw_if_not_valid(b, ci)
+        br = bridge(b, ci)
+        if is_variable_bridged(b, ci)
+            error("Cannot delete constraint index of bridged constrained",
+                  " variables, delete the scalar variable or the vector of",
```
mlubin: This is a run-on sentence. s/variables, delete/variables. Delete/

src/Bridges/lazy_bridge_optimizer.jl
```diff
@@ -10,54 +10,105 @@ bridged into a constraint `F3`-in-`S3` (supported by the internal model) using b
...
-    # Constraint Index of bridged constraint -> Constraint type.
-    constraint_types::Vector{Tuple{DataType, DataType}}
-    # For `SingleVariable` constraints: (variable, set type) -> bridge
-    single_variable_constraints::Dict{Tuple{Int64, DataType}, Constraint.AbstractBridge}
-    bridgetypes::Vector{Any} # List of types of available bridges
-    dist::Dict{Tuple{DataType, DataType}, Int}      # (F, S) -> Number of bridges that need to be used for an `F`-in-`S` constraint
-    best::Dict{Tuple{DataType, DataType}, DataType} # (F, S) -> Bridge to be used for an `F`-in-`S` constraint
+    # Bellman-Ford
+    variable_bridge_types::Vector{Any} # List of types of available bridges
+    variable_dist::Dict{Tuple{DataType}, Int}      # (S,) -> Number of bridges that need to be used for constrained variables in `S`
```
mlubin: Nit: These comments break the line limit. They could be moved above the field names instead.

blegat:
> What do you think?

Sounds good, I will first address your comments in this PR to avoid conflicts and will then create smaller PRs.
Probably there will be
* One for fixing deleting of variables for Model and UniversalFallback
* One for defined constrained variables
* One for a variable bridges
* One for each variable bridge

There will probably not be a good code coverage for each separate PR but as they sum of to this one, it should be ok.

mlubin:
> There will probably not be a good code coverage for each separate PR but as they sum of to this one, it should be ok.

SGTM

docs/src/apireference.md
```diff
@@ -437,16 +460,94 @@ Utilities.@model
...
+[`Bridges.variable_bridged_function`](@ref). Similarly, the variables of the
+underlying model that were created by the bridge can be expressed in terms of
+the bridged variables and this expression can be obtained with
+[`Bridges.variable_unbridged_function`](@ref).
+
+!!! note
+    A notable exception is with [`Bridges.Variable.ZerosBridge`](@ref) where no
+    variable is created in the underlying model as the variables are simply
+    transformed to zeros. When this bridge is used, it is not possible to
+    recover functions with bridged variables from functions of the underlying
```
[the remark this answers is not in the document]
blegat: I renamed it `bridged_variable_function` and `unbridged_variable_function`

src/Bridges/Variable/single_bridge_optimizer.jl
blegat: Yes mostly for testing, I have added a remark in `apimanual`

chriscoey:
Maybe we can say bye to SDOI after this. Removing an extra layer simplifies things.

blegat:
> Maybe we can say bye to SDOI after this. Removing an extra layer simplifies things.

That's the plan, see https://github.com/JuliaOpt/SemidefiniteOptInterface.jl/pull/30#issue-291433670

blegat:
Just squashed everything into one commit to make rebase easier

blegat:
I don't plan to split this PR any further, reviews welcome on this one :)

docs/src/apimanual.md
```diff
@@ -931,20 +932,60 @@ between `y_i` and the vector of scalar-valued quadratic functions.
 
 ### Automatic reformulation
 
+#### Variable reformulation
+
+A variable is often created in a set unsupported by the solver while it could be
```
mlubin: Use the existing terminology around constrained variables. It's not obvious what "creating a variable in a set" means.

chriscoey:
> @chriscoey, since you're starting to write bridges, could you help review these docs?

Yes taking a look now.

docs/src/apimanual.md
```diff
@@ -931,20 +932,60 @@ between `y_i` and the vector of scalar-valued quadratic functions.
...
 
+#### Variable reformulation
+
+A variable is often created constrained in a set unsupported by the solver while
+it could be parametrized by variables constrained in supported sets.
+For example, the [`Bridges.Variable.VectorizeBridge`](@ref) defines the
+reformulation of a constrained variable in [`GreaterThan`](@ref) into a
+constrained vector of one variable in [`Nonnegatives`](@ref).
+The `Bridges.Variable.Vectorize` is the bridge optimizer that applies the
+[`Bridges.Variable.VectorizeBridge`](@ref) rewritting rule. Given an optimizer
```
chriscoey:
```suggestion
[`Bridges.Variable.VectorizeBridge`](@ref) rewriting rule. Given an optimizer
```

docs/src/apimanual.md
```diff
@@ -931,20 +932,60 @@ between `y_i` and the vector of scalar-valued quadratic functions.
...
+`optimizer` implementing constrained variables in [`Nonnegatives`](@ref) and,
```
chriscoey: sentence is grammatically odd to me with the comma after "and"

docs/src/apimanual.md
```diff
@@ -931,20 +932,60 @@ between `y_i` and the vector of scalar-valued quadratic functions.
...
+MOI.supports_constraint(bridged_optimizer, MOI.SingleVariable, MOI.GreaterThan{Float64})
+
+# output
+
+true
+```
+will additionally support constrained variables in [`GreaterThan`](@ref).
+Note that these [`Bridges.Variable.SingleBridgeOptimizer`](@ref) are mainly
+used for testing bridges. It is recommended to rather use
+[`Bridges.full_bridge_optimizer`](@ref) which automatically select the
```
chriscoey:
```suggestion
[`Bridges.full_bridge_optimizer`](@ref), which automatically selects the
```

docs/src/apimanual.md
```diff
@@ -931,20 +932,60 @@ between `y_i` and the vector of scalar-valued quadratic functions.
...
+appropriate bridges for unsupported constrained variables.
+
 #### Constraint reformulation
 
 A constraint often possess different equivalent formulations, but a solver may only support one of them.
```
chriscoey:
```suggestion
A constraint often possesses different equivalent formulations, but a solver may only support one of them.
```

docs/src/apimanual.md
```diff
@@ -931,20 +932,60 @@ between `y_i` and the vector of scalar-valued quadratic functions.
...
 It would be duplicate work to implement rewritting rules in every solver wrapper for every different formulation of the constraint to express it in the form supported by the solver.
```
chriscoey:
```suggestion
It would be duplicate work to implement rewriting rules in every solver wrapper for every different formulation of the constraint to express it in the form supported by the solver.
```

chriscoey: can you find and replace with "rewriting" lol

docs/src/apimanual.md
chriscoey: OK it seems you are mostly using these identifiers as plurals, so "select" is ok.

docs/src/apireference.md
```diff
@@ -192,8 +191,8 @@ delete(::ModelLike, ::Index)
 [`add_variables`](@ref) while *constrained variables* are
 the variables created with [`add_constrained_variable`](@ref) or
 [`add_constrained_variables`](@ref). Adding constrained variables instead of
-constraining free variables with [`add_constraint`](@ref) allows Variable
-bridges to be used.
+constraining free variables with [`add_constraint`](@ref) allows
+[Variable bridges](@ref) to be used.
```
chriscoey: If you capitalize variable here it probably would make sense to capitalize bridges too. I'm not sure I would be capitalizing either though.

docs/src/apireference.md
```diff
@@ -481,17 +480,219 @@ Utilities.@model
 
 ## Bridges
 
-Bridges can be used for automatic reformulation of a certain constraint type into equivalent constraints.
+Bridges can be used for automatic reformulation of constrained variables or
```
chriscoey: The distinction between "constrained variables" vs "constraints" isn't very clear. I think this should be defined first with examples. Is a constrained variable a SingleVariable or VectorOfVariables in a set, and a constraint a ScalarAffineFunction or VectorAffineFunction in a set?

docs/src/apireference.md
chriscoey: Ok perhaps this is needed for the ref to work? In that case I'd probably have both capitalized, but if that would be inconsistent with the rest of the repo and it's too much work to change everywhere then don't worry.

docs/src/apireference.md
blegat: The distinction was added in a separate PR, see http://www.juliaopt.org/MathOptInterface.jl/dev/apireference/#Variables-1

docs/src/apimanual.md
blegat: Isn't it weird if it's plural, we might need to add the 's' everywhere I made the mistake :-P

docs/src/apireference.md
blegat: Fixed

docs/src/apimanual.md
blegat: Fixed another occurence in https://github.com/JuliaOpt/MathOptInterface.jl/pull/759/commits/6d5df0054bb0af70b4ce6baa220dd7f765f420aa

docs/src/apireference.md
chriscoey: Oh boy. OK I think it's worth linking to that description here (and other places where constrained variables are discussed). The special meaning of "constrained variables" is not obvious from the name.

docs/src/apireference.md
```diff
@@ -481,17 +480,219 @@ Utilities.@model
...
+  [`Bridges.Constraint.AbstractBridge`](@ref).
+* [`Bridges.AbstractBridgeOptimizer`](@ref) is a layer that can be applied to
+  another [`ModelLike`](@ref) to apply the reformulation. The
+  [`Bridges.LazyBridgeOptimizer`](@ref) automatically chooses the appropriate
+  bridges to use when a constrained variable or constraint is not supported
+  by using the list of bridges that were added to it by
+  [`Bridges.add_bridge`](@ref). [`Bridges.full_bridge_optimizer`](@ref) wraps a
+  model in a [`Bridges.LazyBridgeOptimizer`](@ref) where all the bridges defined
+  in MOI are added. This is the recommended way to use bridges in the
+  [Testing guideline](@ref), and JuMP automatically calls this function when
```
chriscoey: `this function` - which function?

docs/src/apireference.md
```diff
@@ -481,17 +480,219 @@ Utilities.@model
...
 Bridges.add_bridge
+Bridges.full_bridge_optimizer
+```
+
+### Variable bridges
+
+When variables are added, either free with
+[`add_variable`](@ref)/[`add_variables`](@ref),
+or constrained with
+[`add_constrained_variable`](@ref)/[`add_constrained_variables`](@ref),
```
chriscoey: OK good that you mention `add_constrained_variable`

docs/src/apireference.md
chriscoey: Maybe mention in parens `[`add_constrained_variable`](@ref)/[`add_constrained_variables`](@ref)`

docs/src/apireference.md
```diff
@@ -481,17 +480,219 @@ Utilities.@model
...
+    ```
+    We can obtain the equivalent function in the variables of `model` as follows:
+    ```jldoctest cannot_unbridge_zero
+    inner_func = MOI.Bridges.bridged_function(bridged_model, func)
+
+    # output
+
+    MOI.ScalarAffineFunction{Float64}(MOI.ScalarAffineTerm{Float64}[], 0.0)
+    ```
+    However, it's not possible to invert this operations. Indeed, since the
```
chriscoey:
```suggestion
    However, it's not possible to invert this operation. Indeed, since the
```

docs/src/apireference.md
```diff
@@ -481,17 +480,219 @@ Utilities.@model
...
+```jldoctest split_interval
+MOI.get(model, MOI.ListOfConstraints())
+
+# output
+
+2-element Array{Tuple{DataType,DataType},1}:
+ (MOI.ScalarAffineFunction{Float64}, MOI.GreaterThan{Float64})
+ (MOI.ScalarAffineFunction{Float64}, MOI.LessThan{Float64})
+```
+However, `bridged_model` hides transparently hides these constraints and create the
```
chriscoey:
```suggestion
However, `bridged_model` transparently hides these constraints and create the
```

PR #762, opened 2019-06-09
Add UniversalFallback and @model in doc

blegat:


docs/src/apireference.md
```diff
@@ -404,6 +404,24 @@ ModifyObjectiveNotAllowed
 DeleteNotAllowed
 ```
 
+## Models
+
+The interface is generic and does not have a fixed list of functions and sets.
```
mlubin: Suggested text:

MOI is designed to be extensible, so there is no fixed list of possible functions and sets. This makes it challenging to define efficient storage representations for MOI models. For cases where the functions and sets of interest are known in advance (for example, solvers support a fixed list of functions and sets), we provide the [`Utilities.@model`](@ref) that macro defines a [`ModelLike`](@ref) given a list of functions and sets to support.

[`Utilities.UniversalFallback`](@ref) is a layer that sits on top of any `ModelLike` and provides non-specialized (slower) fallbacks for constraints and attributes that the underlying `ModeLike` does not support.

src/Utilities/model.jl
```diff
@@ -691,15 +691,37 @@ _callfield(f, s::SymbolFS) = :($f(model.$(_field(s))))
 _broadcastfield(b, s::SymbolFS) = :($b(f, model.$(_field(s))))
 
 """
-macro model(modelname, scalarsets, typedscalarsets, vectorsets, typedvectorsets, scalarfunctions, typedscalarfunctions, vectorfunctions, typedvectorfunctions)
+    macro model(modelname, scalarsets, typedscalarsets, vectorsets, typedvectorsets, scalarfunctions, typedscalarfunctions, vectorfunctions, typedvectorfunctions)
```
mlubin: These names are missing underscores.

src/Utilities/universalfallback.jl
```diff
@@ -1,14 +1,15 @@
...
-model `UniversalFallback(model)` supporting *any* constaints and attributes.
-This allows to have a specialized implementation in `model` for performance
-critical constraints and attributes while still supporting other attributes
-with a small performance penalty. Note that `model` is unaware of constraints
-and attributes stored by `UniversalFallback` so this is not appropriate if
-`model` is an optimizer (for this reason, `optimize!` has not been
-implemented). In that case, optimizer bridges should be used instead.
+The `UniversalFallback` can be applied on a [`MathOptInterface.ModelLike`](@ref)
+`model` to create the model `UniversalFallback(model)` supporting *any*
+constaints and attributes. This allows to have a specialized implementation in
```
mlubin: s/any constraints and attributes/any constraint and attribute/

src/Utilities/model.jl
mlubin: Can we provide a default instantiation of this type that supports *all* functions and sets that are defined in MOI? I think this would cover a lot of the existing use cases, and would save people from having to call the macro.

blegat: We already provide `JuMP._MOIModel` which I use most of the time

mlubin: That's not a public symbol, and it requires depending on JuMP.

docs/src/apireference.md
blegat: That's better :)

src/Utilities/model.jl
blegat: We could define a `@model(Model, ...)` which is the same as JuMP's `@model`. Then `JuMP` won't have to define `_MOIModel`, it can just use `MOI.Model`, which decrease the chance of breaking JuMP with MOI.
I would be ok with that.

mlubin: > We could define a `@model(Model, ...)` which is the same as JuMP's `@model`. Then JuMP won't have to define _MOIModel, it can just use MOI.Model, which decrease the chance of breaking JuMP with MOI.
I would be ok with that.

I support this. I would call it something like `DefaultModel` or `BasicModel` to make it clear that it's possible to have different implementations of ` ModeLike`. The `@model` macro then becomes a footnote, because most use cases are covered by `BasicModel`.

PR #767, opened 2019-06-14
Add CleverDicts as submodule to MOI.Utilities

odow:
Originally conceived and implemented here: https://github.com/JuliaOpt/Gurobi.jl/pull/216

### DIFF
--- a/src/Utilities/CleverDicts.jl
+++ b/src/Utilities/CleverDicts.jl
@@ -1,0 +1,193 @@
+module CleverDicts
+
+import OrderedCollections
+
+"""
+    CleverDict{K, V}
+
+A smart storage type for managing sequential objects with non-decreasing integer
+indices.
+
+Provided no keys are deleted, the backing storage is a `Vector{V}`. Once a key
+has been deleted, the backing storage switches to an `OrderedDict{K, V}`.
+
+The i'th ordered element can be obtained with `c[LinearIndex(i)]`.
+
+Note that querying a `LinearIndex` immediately after deleting a key via
+`delete!` is very slow. (It requires a rebuild of an ordered list of variables.)
+
+Use `new_key(c::CleverDict)` to obtain the next key in the sequential order.
+
+Overload the functions `index_to_key` and `key_to_index` to enable mappings
+between the integer index of the vector and the dictionary key.
+"""
+mutable struct CleverDict{K, V}
+    last_index::Int
+    vector::Union{Nothing, Vector{V}}
+    dict::Union{Nothing, OrderedCollections.OrderedDict{K, V}}
+    CleverDict{K, V}() where {K, V} = new{K, V}(0, V[], nothing)
+end
+
+"""
+    index_to_key(::Type{K}, index::Int)
+
+Create a new key associated with the integer value `index`.
+"""
+function index_to_key(::Type{K}, index::Int) where {K}
+    error("You need to define `index_to_key`.")
+end
+
+"""
+    key_to_index(key::K)
+
+Map `key` to an integer valued index, assuming that there have been no
+deletions.
+"""
+function key_to_index(key::K) where {K}
+    error("You need to define `key_to_index`.")
+end
+
+"""
+    new_item(c::CleverDict{K, V}, val::Val)::K where {K, V}
+
+Set `val` in the next available key, and return that key.
+"""
+function new_item(c::CleverDict{K, V}, val::V)::K where {K, V}
+    c.last_index += 1
+    key = index_to_key(K, c.last_index)
+    if c.dict === nothing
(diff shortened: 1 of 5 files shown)


src/Utilities/CleverDicts.jl
```diff
@@ -0,0 +1,198 @@
...
+
+Store an item `val` using `new_item(c::CleverDict, val)`. `new_item` returns a
+key corresponding to the stored item.
+
+Overload the functions `index_to_key` and `key_to_index` to enable mappings
+between the integer index of the vector and the dictionary key.
+
+## Example
+
+    struct MyKey
```
blegat: It is better to use triple backsticks to get color

src/Utilities/CleverDicts.jl
```diff
@@ -0,0 +1,198 @@
...
+        return 1 <= key_to_index(key) <= length(c.vector)
+    else
+        return haskey(c.dict, key)
+    end
+end
+
+function Base.iterate(
+    c::CleverDict{K, V}
+)::Union{Nothing, Tuple{Pair{K, V}, Int}} where {K, V}
+    if length(c) == 0
```
blegat: Seems more natural to call `isempty`

src/Utilities/CleverDicts.jl
```diff
@@ -0,0 +1,198 @@
...
+        @assert c.vector !== nothing
+        key = index_to_key(K, 1)
+        return key => c.vector[1], 2
+    end
+end
+
+function Base.iterate(
+    c::CleverDict{K, V}, s::Int
+)::Union{Nothing, Tuple{Pair{K, V}, Int}} where {K, V}
+    if s > length(c)
```
blegat: This could be used when the dict is not nothing which would not make sense

blegat: Or you rely on the fact that the state of the dict is always smaller than its length

src/Utilities/CleverDicts.jl
```diff
@@ -0,0 +1,198 @@
...
+        return iterate(c.dict)
+    else
+        @assert c.vector !== nothing
+        key = index_to_key(K, 1)
+        return key => c.vector[1], 2
+    end
+end
+
+function Base.iterate(
+    c::CleverDict{K, V}, s::Int
```
blegat: This would break if the state of the dict is no longer an Int

src/Utilities/CleverDicts.jl
```diff
@@ -0,0 +1,217 @@
...
+# `iterate(c, s::Int)`.
+
+function Base.iterate(
+    c::CleverDict{K, V}
+)::Union{Nothing, Tuple{Pair{K, V}, Int}} where {K, V}
+    if c.dict !== nothing
+        return iterate(c.dict)
+    else
+        @assert c.vector !== nothing
+        if length(c.vector) == 0
```
blegat: isempty

src/Utilities/CleverDicts.jl
```diff
@@ -0,0 +1,217 @@
...
+    c.vector = nothing
+    return
+end
+
+function Base.length(c::CleverDict)::Int
+    return c.dict == nothing ? length(c.vector) : length(c.dict)
+end
+
+function Base.isempty(c::CleverDict)
+    return c.dict !== nothing ? isempty(c.dict) : length(c.vector) == 0
```
blegat: isempty

src/Utilities/CleverDicts.jl
odow: Why is isempty preferred?

blegat: You are checking whether the vector is empty so it would be more readable to use this function.
We should always call the most specific function that does the task

src/Utilities/CleverDicts.jl
```diff
@@ -0,0 +1,231 @@
...
+# it is the original use-case for `CleverDict`, and it would be type-piracy for
+# solvers using `CleverDicts` to implement it themselves.
+
+import MathOptInterface
+
+function index_to_key(::Type{MathOptInterface.VariableIndex}, index::Int)
+    return MathOptInterface.VariableIndex(index)
+end
+
+key_to_index(key::MathOptInterface.VariableIndex) = key.value
```
odow: @blegat: I've implemented the methods to use `CleverDict` with `MOI.VariableIndex` since it would be type-piracy for solvers to implement the methods. Any ideas for better places to put them? Or is here okay?

blegat: No this seams appropriate. We it would make even more sense if we add `const CleverVariableDict{V} = CleverDict{MOI.VariableIndex, V}`

PR #769, opened 2019-06-17
Add MOI.Benchmarks submodule to facilitate solver benchmarking

odow:
Originally conceived and implemented in https://github.com/JuliaOpt/Gurobi.jl/pull/216 and https://github.com/JuliaOpt/GLPK.jl/pull/101.


blegat:
We could add an entry in the doc

PR #775, opened 2019-06-23
Add MOI.submit to add attribute

blegat:
Some operations do not fall into the category of `set` such as:
* Adding heuristic solutions (https://github.com/JuliaOpt/MathOptInterface.jl/pull/670)
* Adding Lazy constraints (https://github.com/JuliaOpt/MathOptInterface.jl/pull/670)

So when a solver want to expose a solver specific feature like this, we need to create a new `add_...` function in MOI and implement it in every layer + adding a function to JuMP, ...
This PR adds a `MOI.add` function which is similar to `MOI.set` except it returns an `AttributeIndex` which is consistent with `add_variable(s)` and `add_constraint(s)`.
With this PR, if a solver want to expose, say lazy constraints, it just have to define `LazyConstraint <: MOI.AbstractOptimizerAttribute`, implement `MOI.add(::Optimizer, ::LazyConstraint, ...)` and no change is needed either from JuMP or MOI.

This might be the only breaking change needed to add support for callbacks so it might be worth it to include it into v0.9

### DIFF
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -377,7 +377,7 @@ end
 
 """
     submit(optimizer::AbstractOptimizer, sub::AbstractSubmittable,
-           value)::Union{Nothing, SubmittedIndex{typeof(sub)}}
+           value)::Nothing
 
 Submit `value` to the submittable `sub` of the optimizer `optimizer`.
 
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -75,24 +75,53 @@ operation_name(err::SetAttributeNotAllowed) = "Setting attribute $(err.attr)"
 message(err::SetAttributeNotAllowed) = err.message
 
 """
-    struct AddAttributeNotAllowed{AttrType} <: NotAllowedError
-        attr::AttrType
+    AbstractSubmittable
+
+Abstract supertype for objects that can be submitted to the model.
+"""
+abstract type AbstractSubmittable end
+
+# This allows to use submittable in broadcast calls without the need to embed
+# it in a `Ref`.
+Base.broadcastable(sub::AbstractSubmittable) = Ref(sub)
+
+"""
+    struct UnsupportedSubmittable{SubmitType} <: UnsupportedError
+        sub::SubmitType
+        message::String
+    end
+
+An error indicating that the submittable `sub` is not supported by the model,
+i.e. that [`supports`](@ref) returns `false`.
+"""
+struct UnsupportedSubmittable{SubmitType<:AbstractSubmittable} <: UnsupportedError
+    sub::SubmitType
+    message::String
+end
+
+"""
+    struct SubmitNotAllowed{SubmitTyp<:AbstractSubmittable} <: NotAllowedError
+        sub::SubmitType
         message::String # Human-friendly explanation why the attribute cannot be set
     end
 
-An error indicating that the attribute `attr` is supported (see
+An error indicating that the submittable `sub` is supported (see
 [`supports`](@ref)) but cannot be added for some reason (see the error string).
 """
-struct AddAttributeNotAllowed{AttrType<:AnyAttribute} <: NotAllowedError
-    attr::AttrType
-	message::String # Human-friendly explanation why the attribute cannot be set
+struct SubmitNotAllowed{SubmitType<:AbstractSubmittable} <: NotAllowedError
+    sub::SubmitType
+    message::String # Human-friendly explanation why the attribute cannot be set
 end
-AddAttributeNotAllowed(attr::AnyAttribute) = AddAttributeNotAllowed(attr, "")
+SubmitNotAllowed(sub::AbstractSubmittable) = SubmitNotAllowed(sub, "")
 
-operation_name(err::AddAttributeNotAllowed) = "Adding attribute $(err.attr)"
-message(err::AddAttributeNotAllowed) = err.message
+operation_name(err::SubmitNotAllowed) = "Submitting $(err.sub)"
+message(err::SubmitNotAllowed) = err.message
 
 """
+    supports(model::ModelLike, sub::AbstractSubmittable)::Bool
+
+Return a `Bool` indicating whether `model` supports the submittable `sub`.
+
     supports(model::ModelLike, attr::AbstractOptimizerAttribute)::Bool
 
 Return a `Bool` indicating whether `model` supports the optimizer attribute
@@ -120,7 +149,7 @@ Return a `Bool` indicating whether `model` supports the constraint attribute
 `copy_to(model, src)` cannot be performed in case `attr` is in the
 [`ListOfConstraintAttributesSet`](@ref) of `src`.
 
-For all four methods, if the attribute is only not supported in specific
+For all five methods, if the attribute is only not supported in specific
 circumstances, it should still return `true`.
 
 Note that `supports` is only defined for attributes for which
@@ -128,6 +157,7 @@ Note that `supports` is only defined for attributes for which
 list of attributes set obtained by `ListOf...AttributesSet`.
 """
 function supports end
+supports(::ModelLike, ::AbstractSubmittable) = false
 function supports(::ModelLike, attr::Union{AbstractModelAttribute,
                                            AbstractOptimizerAttribute})
     if !is_copyable(attr)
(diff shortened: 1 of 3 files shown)


blegat:
> The assumption that you only need to set variable-wise values is a bit off. What if you want to add a second-order cone constraint as a lazy constraint?

You pass it as a tuple `(func, set)` in the `value` argument.

> I can only imagine this working in direct mode, because it would be hard to translate indices between the cache and the underlying solver given arbitrary payloads.

We could have a `map_indices` that should be implemented for payload containing indices but that can be left as future work. Having this work in direct mode is a good start.

> Also it wasn't clear to me if `LazyConstraint` in the proposal means a callback object or the actual constraint that would be added inside a callback.

The constraint object as `(func, set)` tuple.

> I'd propose something like: `submit...`

IMO `MOI.add` should return an index for consistency with `MOI.add_variable`, ... If we want something without indices, `MOI.submit` might indeed be a better name. The targeted application is callbacks so if they usually don't return indices, this PR should add `MOI.submit` and we'll see if we need to define a `MOI.add` in the future.

mlubin:
> The targeted application is callbacks so if they usually don't return indices

They don't usually return indices.

> The constraint object as (func, set) tuple.

Then what is the use case for the variable-wise and constraint-wise versions? Submitting a feasible solution?

```julia
  add(model::ModelLike, attr::AbstractVariableAttribute, v::VariableIndex,
         value)::AttributeIndex{typeof(attr)}
  add(model::ModelLike, attr::AbstractConstraintAttribute, c::ConstraintIndex,
          value)
```

blegat:
> Then what is the use case for the variable-wise and constraint-wise versions? Submitting a feasible solution?

Yes, submitting a heuristic solution.

mtanneau:
I'm only familiar with MILP callbacks so what follows is biased towards this use case; I have no experience (nor knowledge thereof) of callbacks for non-linear optimization solvers.

Some general background. MILP solvers allow you to do two types of things during with a callback:
  * Query some information from the solver, e.g. number of solutions, node statistics, time, etc... No information is _pushed_ to the solver.
   In terms of implementation, all this should be doable via MOI's `get` + the right attributes. I strongly suggest that any attribute being called within a callback gets the prefix `Cb`, e.g. `CbNodeIndex`. That would avoid some confusion.
  * Submit/push/post/provide (whatever word sounds best) information to the solver, e.g.: heuristic solution, lazy constraint, user cut, branching decision, etc... This _does_ affect the solver's behaviour.
   Which information can be passed, is which form, and from where in the resolution process, is solver-specific (although MILP solvers share _some_ common ground). For instance, you typically can't query dual variables if your callback is called after a heuristic found an integer solution.

mtanneau:
Now the specifics:

### Heuristic solution
This allows you to _suggest_ a heuristic solution to the solver.
Some will require you to provide a full solution, some allow to provide partial solutions (e.g. integer assignment on a subset of the variables).

This does not modify the model. It does not create a new variable. Your heuristic solution may not even be _considered_ by the solver. Therefore, the `submit` proposal of Miles seems more appropriate than tweaking the `add` function.

IMO, an intuitive syntax could look like
```julia
MOI.submit(model, ::MOI.CbHeuristicSolution, ::Any)
```

### Lazy constraint and user cut

Provide an additional constraint to the solver, which were not part of the original formulation.
The semantic difference between user cuts and lazy constraints is the following:
* A lazy constraint may cut an integer solution that was feasible w.r.t. the original formulation.
* A user cut may not cut any integer feasible solution. If it does, things may break.

Similar to heuristic solution, user cuts and lazy constraints may never enter the formulation (if the solver decides so).
* Lazy constraints / user cuts are not indexed
* They many not be enforced: [Gurobi's docs](https://www.gurobi.com/documentation/8.1/refman/c_grbcblazy.html) states that
> Node solutions will usually respect previously added lazy constraints, but not always.
* You cannot query the associated dual variables (because they may not even exist)
* I do not know whether lazy constraints / user cuts are kept when copying a model

The syntax proposed by Miles also fits this nicely, e.g.:
```julia
MOI.submit(model, ::MOI.CbUserCut, ::Any)
MOI.submit(model, ::MOI.CbLazyConstraint, ::Any)
```
the third argument will most likely encapsulate the constraint itself (`func`, `set`), and some callback-related data.

_Note:_ as mentioned by Miles, some of the initial constraints may be marked as lazy: http://www.gurobi.com/documentation/8.1/refman/lazy.html. This has nothing to do with callbacks.

### Adding a new variable
AFAIK, no MILP solver supports it. The only exception may be SCIP, but I'm not familiar enough.

mtanneau:
Now that I think about it, maybe we should be returning an index when adding constraints (and possibly variables) through a callback. Just because existing MILP solvers don't do it doesn't mean MOI should not, and it would restrict the future capabilities, if someone wants to implement lazy constraints a bit differently.

My idea would like the following.
There could be an `InvalidConstraintIndex`, which -as its name suggests- is always invalid. By default, this is the index that would be returned when adding a lazy constraint / user cut via a callback. If a solver wishes to return a valid index because they support it, they can.

The same thing could be done with variables. It would actually be quite useful in a branch-and-price context: new variables are added at each node, but one may want to remove some when taking certain branching decisions.

mlubin:
@odow thoughts?

src/attributes.jl
```diff
@@ -74,6 +74,24 @@ SetAttributeNotAllowed(attr::AnyAttribute) = SetAttributeNotAllowed(attr, "")
...
+        attr::AttrType
+        message::String # Human-friendly explanation why the attribute cannot be set
+    end
+
+An error indicating that the attribute `attr` is supported (see
+[`supports`](@ref)) but cannot be added for some reason (see the error string).
+"""
+struct AddAttributeNotAllowed{AttrType<:AnyAttribute} <: NotAllowedError
+    attr::AttrType
+	message::String # Human-friendly explanation why the attribute cannot be set
```
odow: Weird indent.

src/attributes.jl
```diff
@@ -74,6 +74,24 @@ SetAttributeNotAllowed(attr::AnyAttribute) = SetAttributeNotAllowed(attr, "")
...
+end
+AddAttributeNotAllowed(attr::AnyAttribute) = AddAttributeNotAllowed(attr, "")
```
odow: Maybe a more descriptive default message?

blegat: There is already a more descriptive one printed by `NotAllowedError`, the message field is for custom messages.

blegat:
@mlubin @mtanneau @odow Just iterated based on your comments. Let me know what you think of the new design.

src/indextypes.jl
```diff
@@ -35,10 +35,21 @@ end
...
 Base.hash(v::VariableIndex, h::UInt) = hash(v.value, h)
 
+"""
+    SubmittedIndex{SubmitType}
+
+A type-safe wrapper for `Int64` for use in referencing elements submitted for
+submittable of type `SubmitType`.
+"""
+struct SubmittedIndex{SubmitType}
+    attr::SubmitType
```
odow: Is this necessary? What is it needed for?

src/attributes.jl
```diff
@@ -328,7 +376,26 @@ function throw_set_error_fallback(model::ModelLike,
...
 """
-    ListOfOptimizerAttributesSet()
+    submit(optimizer::AbstractOptimizer, sub::AbstractSubmittable,
+           value)::Union{Nothing, SubmittedIndex{typeof(sub)}}
+
+Submit `value` to the submittable `sub` of the optimizer `optimizer`.
+
+An [`UnsupportedSubmittable`](@ref) error is thrown if `model` does not support
+the attribute `attr` (see [`supports`](@ref)) and a [`SubmitNotAllowed`](@ref)
+error is thrown if it supports the submittable `sub` but it cannot be submitted.
```
odow: When should it return nothing and when should it return an index? Only if supported by the solver? Or always?

mlubin:
I'm skeptical of the need for `SubmittedIndex` given that the use cases for it are only hypothetical and could possibly never materialize. It's clunky to carry around pieces of code that won't be used. https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it.

It would be a different story if we had a solver developer who said, "I want to use this for X."

src/indextypes.jl
blegat: Indices of different submittable should be different to avoid mixing indices of different ones. The hole point of these indices types is to ensure that they are not used in incorrect places and adding the attribute type might help in this respect

blegat:
It might also not be too hard to add later when it's needed so I am ok to not have it for now.

mtanneau:
> I'm skeptical of the need for SubmittedIndex given that the use cases for it are only hypothetical and could possibly never materialize.

Miles has a strong point. 
In this direction, I can think of only two hypothetical cases/solvers where one could want to support indices for submitted stuff: Pajarito and Coluna. And that's assuming that `add_constraint` and `add_variable` were not satisfactory.

In the mean time, it indeed seems preferable not to carry `SubmittedIndex` around, but add it later if needed.

blegat:
@mlubin @mtanneau @odow Just removed the `SubmittedIndex`. Any more comments ?

mtanneau:
> @mlubin @mtanneau @odow Just removed the `SubmittedIndex`. Any more comments ?

I think what we need now are a couple of examples.

I'd be happy to port the examples that were in JuMP v18 ([heuristic](https://github.com/JuliaOpt/JuMP.jl/blob/v0.18.6/examples/simpleheur.jl), [lazy](https://github.com/JuliaOpt/JuMP.jl/blob/v0.18.6/examples/simplelazy.jl) & [user](https://github.com/JuliaOpt/JuMP.jl/blob/v0.18.6/examples/simpleusercut.jl)) to MOI, but I don't have time to commit to this until the end of the month.

mlubin:
> I think what we need now are a couple of examples to check whether something's missing/odd.

This PR doesn't have anything specific for callbacks, so I don't think it needs to be blocked until we have a full callback example.

PR #782, opened 2019-07-10
Add Lazy/Heuristic MIP callbacks

blegat:
This PR defines the necessary attributes and submittables for lazy and heuristic callbacks.
The docstrings are based on https://github.com/JuliaOpt/MathOptInterface.jl/pull/670.

Closes https://github.com/JuliaOpt/MathOptInterface.jl/pull/670

cc @mtanneau

### DIFF
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -378,6 +378,14 @@ function throw_set_error_fallback(model::ModelLike,
 end
 
 """
+    SettingSingleVariableFunctionNotAllowed()
+
+Error type that should be thrown when the user [`set`](@ref) the
+[`ConstraintFunction`](@ref) of a [`SingleVariable`](@ref) constraint.
+"""
+struct SettingSingleVariableFunctionNotAllowed <: Exception end
+
+"""
     submit(optimizer::AbstractOptimizer, sub::AbstractSubmittable,
            values...)::Nothing
 
@@ -397,13 +405,39 @@ function submit(model::ModelLike, sub::AbstractSubmittable, args...)
     end
 end
 
-"""
-    SettingSingleVariableFunctionNotAllowed()
+## Submittables
 
-Error type that should be thrown when the user [`set`](@ref) the
-[`ConstraintFunction`](@ref) of a [`SingleVariable`](@ref) constraint.
 """
-struct SettingSingleVariableFunctionNotAllowed <: Exception end
+    LazyConstraint(callback_data)
+
+Lazy constraint `func`-in-`set` submitted as a tuple `(func, set)`. The optimal
+solution returned by [`VariablePrimal`](@ref) will satisfy all lazy
+constraints that have been submitted.
+
+This can be submitted only from the lazy callback (that is, the function set to
+[`LazyCallback`](@ref)). The field `callback_data` is a solver-specific
+callback type that is passed as the argument to the lazy callback.
+"""
+struct LazyConstraint <: AbstractSubmittable
+    callback_data::Any
+end
+
+"""
+    HeuristicSolution(callback_data)
+
+Heuristically obtained integer-feasible solutions. The solution is given by a
+`Dict{VariableIndex, Float64}` mapping the variables to their solution value.
+
+This can be submitted only from the heuristic callback (that is, the function
+set to [`HeuristicCallback`](@ref)). The field `callback_data` is a
+solver-specific callback type that is passed as the argument to the heuristic
+callback.
+
+Note that the solver may silently reject the provided solution.
+"""
+struct HeuristicSolution <: AbstractSubmittable
+    callback_data::Any
+end
 
 ## Optimizer attributes
 
@@ -460,6 +494,74 @@ which is typically an `Enum` or a `String`.
 struct RawParameter <: AbstractOptimizerAttribute
     name::Any
 end
+
+### Callbacks
+
+"""
+    struct OptimizeInProgress{AttrType<:AnyAttribute} <: Exception
+        attr::AttrType
+    end
+
+Error thrown from optimizer when `MOI.get(optimizer, attr)` is called inside an
+[`AbstractCallback`](@ref) while it is only defined once [`optimize!`](@ref) has
+completed. This is only defined if `is_set_by_optimize(attr)` is `true`.
+"""
+struct OptimizeInProgress{AttrType<:AnyAttribute} <: Exception
+    attr::AttrType
+end
+
+function Base.showerror(io::IO, err::OptimizeInProgress)
+    print(io, typeof(err), ": Cannot get result as the `MOI.optimize!` has not",
+          " finished.")
+end
+
+"""
+    abstract type AbstractCallback <: AbstractOptimizerAttribute end
+
+Abstract type for optimizer attribute representing a callback functions. The
+value set to subtypes of `AbstractCallback` is a function that may be called
+during [`optimize!`](@ref). As [`optimize!`](@ref) is in progress, the result
+attributes (i.e, the attributes `attr` such that `is_set_by_optimize(attr)`)
+may not be accessible from the callback hence trying to get result attributes
+might throw a [`OptimizeInProgress`](@ref) error.
+
+The value of the attribute should be a function taking only one argument, that
+is commonly called `callback_data`, that can be used for instance in
+[`LazyCallback`](@ref), [`HeuristicSolution`](@ref).
+"""
+abstract type AbstractCallback <: AbstractOptimizerAttribute end
+
+"""
+    LazyCallback() <: AbstractCallback
+
+The *lazy* callback can be used to submit [`LazyConstraint`](@ref) which are
+added on-demand at integer nodes in the branch and bound tree. Note that there
+is no guarantee that the callback is called at *every* integer primal solution.
+
+The solution at the integer node is accessed through
+[`VariablePrimalAtIntegerNode`](@ref). Trying to access other result
+attributes will throw [`OptimizeInProgress`](@ref) as discussed in
+[`AbstractCallback`](@ref).
+"""
+struct LazyCallback <: AbstractCallback end
+
+"""
+    HeuristicCallback() <: AbstractCallback
+
+The *heuristic* callback can be used to submit [`HeuristicSolution`](@ref) at
+fractional (i.e., non-integer) nodes in the branch and bound tree. Note that
+there is not guarantee that the callback is called *everytime* the solver has a
+fractional solution.
+
+The solution at the fractional node is accessed through
+[`VariablePrimalAtFractionalNode`](@ref). Trying to access other result
+attributes will throw [`OptimizeInProgress`](@ref) as discussed in
+[`AbstractCallback`](@ref).
+
+Some solvers require a complete solution, others only partial solutions. It's up
+to you to provide the appropriate one. If in doubt, give a complete solution.
+"""
+struct HeuristicCallback <: AbstractCallback end
 
 ## Model attributes
 
@@ -680,6 +782,28 @@ struct VariablePrimal <: AbstractVariableAttribute
     N::Int
 end
 VariablePrimal() = VariablePrimal(1)
+
+"""
+    VariablePrimalAtIntegerNode(callback_data)
+
+A variable attribute for the assignment to some primal variable's value at the
+integer node from which the [`LazyCallback`](@ref) corresponding to
+`callback_data` was called.
+"""
+struct VariablePrimalAtIntegerNode <: AbstractVariableAttribute
+    callback_data::Any
+end
+
+"""
+    VariablePrimalAtFractionalNode(callback_data)
+
+A variable attribute for the assignment to some primal variable's value at the
+fractional node from which the [`HeuristicCallback`](@ref) corresponding to
+`callback_data` was called.
+"""
+struct VariablePrimalAtFractionalNode <: AbstractVariableAttribute
+    callback_data::Any
+end
 
 """
     BasisStatusCode
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -408,35 +408,70 @@ end
 ## Submittables
 
 """
+    RejectSolution(callback_data)
+
+Reject feasible solution provided by [`CallbackVariablePrimal`](@ref).
+
+This can be submitted only from the [`FeasibleSolutionCallback`](@ref). The
+field `callback_data` is a solver-specific callback type that is passed as the
+argument to the feasible solution callback.
+"""
+struct RejectSolution{CBDT} <: AbstractSubmittable
+    callback_data::CBDT
+end
+
+"""
     LazyConstraint(callback_data)
 
-Lazy constraint `func`-in-`set` submitted as a tuple `(func, set)`. The optimal
+Lazy constraint `func`-in-`set` submitted as `func, set`. The optimal
 solution returned by [`VariablePrimal`](@ref) will satisfy all lazy
 constraints that have been submitted.
 
-This can be submitted only from the lazy callback (that is, the function set to
-[`LazyCallback`](@ref)). The field `callback_data` is a solver-specific
-callback type that is passed as the argument to the lazy callback.
+This can be submitted only from the [`FeasibleSolutionCallback`](@ref). The
+field `callback_data` is a solver-specific callback type that is passed as the
+argument to the feasible solution callback.
 """
-struct LazyConstraint <: AbstractSubmittable
-    callback_data::Any
+struct LazyConstraint{CBDT} <: AbstractSubmittable
+    callback_data::CBDT
 end
 
 """
     HeuristicSolution(callback_data)
 
-Heuristically obtained integer-feasible solutions. The solution is given by a
-`Dict{VariableIndex, Float64}` mapping the variables to their solution value.
+Heuristically obtained feasible solution. The solution is submitted as
+`variables, values` where `values[i]` gives the value of `variables[i]`,
+similarly to [`set`](@ref).
 
-This can be submitted only from the heuristic callback (that is, the function
-set to [`HeuristicCallback`](@ref)). The field `callback_data` is a
-solver-specific callback type that is passed as the argument to the heuristic
-callback.
+This can be submitted only from the [`InfeasibleSolutionCallback`](@ref). The
+field `callback_data` is a solver-specific callback type that is passed as the
+argument to the infeasible solution callback.
 
 Note that the solver may silently reject the provided solution.
 """
-struct HeuristicSolution <: AbstractSubmittable
-    callback_data::Any
+struct HeuristicSolution{CBDT} <: AbstractSubmittable
+    callback_data::CBDT
+end
+
+"""
+    UserCut(callback_data)
+
+Constraint `func`-to-`set` suggested to render the solution given by
+[`CallbackVariablePrimal`](@ref) trivially infeasible. The solution is submitted
+as `func, set`. For instance, for a solution satisfying all constraints except
+[`SingleVariable`](@ref)-in-[`Integer`](@ref) constraints, a
+[`ScalarAffineFunction`](@ref)-in-[`LessThan`](@ref) may be submitted cut it
+off. Note that, as opposed to [`LazyConstraint`](@ref), the provided constraint
+cannot modify the feasible set, the constraint should be redundant, e.g., it
+may be a consequence of affine and integrality constraints.
+
+This can be submitted only from the [`InfeasibleSolutionCallback`](@ref). The
+field `callback_data` is a solver-specific callback type that is passed as the
+argument to the infeasible solution callback.
+
+Note that the solver may silently ignore the provided constraint.
+"""
+struct UserCut{CBDT} <: AbstractSubmittable
+    callback_data::CBDT
 end
 
 ## Optimizer attributes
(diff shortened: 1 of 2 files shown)


src/attributes.jl
```diff
@@ -449,6 +483,74 @@ struct RawParameter <: AbstractOptimizerAttribute
...
+attributes will throw [`OptimizeInProgress`](@ref) as discussed in
+[`AbstractCallback`](@ref).
+"""
+struct LazyCallback <: AbstractCallback end
+
+"""
+    HeuristicCallback() <: AbstractCallback
+
+The *heuristic* callback can be used to submit [`HeuristicSolution`](@ref) at
+fractional (i.e., non-integer) nodes in the branch and bound tree. Note that
```
odow: Are HeuristicCallbacks ever called at integer nodes asking for a better solution?

src/attributes.jl
```diff
@@ -449,6 +483,74 @@ struct RawParameter <: AbstractOptimizerAttribute
...
+is commonly called `callback_data`, that can be used for instance in
+[`LazyCallback`](@ref), [`HeuristicSolution`](@ref).
+"""
+abstract type AbstractCallback <: AbstractOptimizerAttribute end
+
+"""
+    LazyCallback() <: AbstractCallback
+
+The *lazy* callback can be used to submit [`LazyConstraint`](@ref) which are
+added on-demand at integer nodes in the branch and bound tree. Note that there
```
odow: Are lazy constraints only ever added at integer nodes?

src/attributes.jl
mtanneau: In general, no.
See for instance the [CPLEX docs](https://www.ibm.com/support/knowledgecenter/SSSA5P_12.9.0/ilog.odms.cplex.help/refcallablelibrary/cpxapi/callbackpostheursoln.html)

mtanneau:
In terms of interface, Gurobi and CPLEX are good examples that we can follow.
Pointers to the docs: 
* CPLEX
  * [Overview](https://www.ibm.com/support/knowledgecenter/en/SSSA5P_12.9.0/ilog.odms.cplex.help/CPLEX/UsrMan/topics/progr_adv/callbacks/introCallbacks.html)
  * [What can be queried from where](https://www.ibm.com/support/knowledgecenter/SSSA5P_12.9.0/ilog.odms.cplex.help/refcallablelibrary/cpxapi/callbacksetfunc.html)
* Gurobi
  * [Callback codes](https://www.gurobi.com/documentation/8.1/refman/callback_codes.html#sec:CallbackCodes)
  * [Callback example](https://www.gurobi.com/documentation/8.1/examples/callback_c_c.html) (in C)
  * [`GRBcbget` function](https://www.gurobi.com/documentation/8.1/refman/c_grbcbget.html#routine:GRBcbget)

src/attributes.jl
mtanneau: Whether you _can_ is solver-specific.

Whether you _want_ to depends on the application; in general you may want to separate lazy constraints from a fractional point (e.g. Benders cut)

blegat:
One approach would be to have only one callback `MIPCallback` and you would decide what to do based on whether you can get a fractional and/or an integer solution and whether you can submit a lazy constraint and/or a heuristic solution. That would require adding `can_get` and `can_submit`. `supports` does not do it because `supports` cannot depend on the state of the solver.

A second approach would be to have `FractionalNodeCallback` where you are at a fractional node and `IntegerNodeCallback` where you are at an integer node. What you can get is fixed so no need for `can_get` but you can check what you can submit with `can_submit`.

The third approach is having `LazyCallback` and `HeuristicCallback` where both what you can get and what you can submit is fixed (this is the current approach of this PR).

So to summarize in a table:

|                    | Integer Solution | Fractional Solution |
|--------------------|------------------|---------------------|
| Lazy Constraint    | LazyCallback     |                     |
| Heuristic Solution |                  | HeuristicCallback   |

* The first solution is having a single callback for the 4 cells.
* The second solution is having a callback for each column.
* The third solution is having a callback for each diagonal cell.

mtanneau:
IMO, `MIPCallback` is much preferable. (this approach is what I refer to as "generic callback", not to be confused with solver-independent callback).
* It is lighter: you only ever use `MOI.get`, `MOI.submit` and possibly `MOI.add` and `MOI.set`.
This is also more in line with MOI's general philosophy.
* It is more flexible than specific callbacks. For instance, all informational callbacks would be automatically  covered.
Importantly, one can define a new kind of callback by just defining their own attributes.
* It's the direction that the main solvers are taking, and thus would make the wrapping more intuitive.


FYI, here is a table of what MIP solvers currently support. `Y` is yes, `N` is no, `?` is I don't know. Docs are linked in the names.

| Solver     | Cbc| [CPLEX](https://www.ibm.com/support/knowledgecenter/en/SSSA5P_12.9.0/ilog.odms.cplex.help/CPLEX/UsrMan/topics/progr_adv/callbacks/whatToDoCallbacks.html) | [GLPK](http://most.ccib.rutgers.edu/glpk.pdf#a8) | [Gurobi](https://www.gurobi.com/documentation/8.1/refman/callback_codes.html#sec:CallbackCodes) | SCIP | [Xpress](https://www.fico.com/fico-xpress-optimization/docs/latest/solver/optimizer/HTML/chapter5_sec_section5002.html) |
|-------------|:-----:|:-----:|:-------:|:------:|:------:|:-------:|
| Generic callback 		| ?  	| Y 	| Y 	| Y 	| N/? 	| N 	
| Heuristics	 	| ?  	| Y 	| Y 	| Y 	| ? 	| Y 	
| Lazy constraint	| ? 	| Y	| Y 	| Y 	| ? 	| Y 	
| User cuts	 	| ?  	| Y 	| ? 	| Y 	| ? 	| Y 

With respect to `can_get` and `can_submit`, there are two cases that should be distinguished:
1. I'm using a MIP solver that does not support lazy constraints, and I'm trying to add one.
This should raise an "MOI-level" error, because I'm trying to do something the solver doesn't support. In particular, the compiler can infer it.
2. I'm using Gurobi, and I want to add a lazy constraint at the root.
This should raise a "Gurobi-level" error, and typically cannot be inferred at compile time (unless you type every single callback code).

This distinction is the same as, say querying the solution value after calling `optimize!`. For instance, if Gurobi ran into some error, you get a Gurobi error when asking for the optimal solution.

Back to `MIPCallback`. Solvers with generic callbacks will provide a `where` argument to the callback function, indicating what information can and cannot be queried/submitted (the correspondence would be indicated in the solver's docs).

One question I don't have an answer to is: what about solvers that don't do generic callbacks (e.g. XPress)? How do you break a generic callback into usercut callback / heuristic callback, etc?
This approach also puts a toll on solvers' wrappers. But callbacks are very solver-specific anyway, so it makes sense that most of the heavy lifting is done by solvers themselves.

blegat:
> It is more flexible than specific callbacks. For instance, all informational callbacks would be automatically  covered.
> Importantly, one can define a new kind of callback by just defining their own attributes.

Not sure to follow. Solvers can already define new callbacks by defining new attributes, this PR will not change the situation in that respect.
For informational callbacks, this is what they should do. How would informational callbacks fit in `MIPCallback` ?

> here is a table of what MIP solvers currently support

Thanks a lot !

> With respect to `can_get` and `can_submit`, there are two cases that should be distinguished:

The case 1. is already covered by `supports`. For case 2., if there is enough info in the `callback_data` (let's say `where` is also in `callback_data`), then we might be able to implement `can_get(::Optimizer, SomeAttribute(callback_data)` and `can_submit(::Optimizer, SomeSubmittable(callback_data)`. This is still run-time but it allows to avoid having solvers throwing errors.

> Back to `MIPCallback`. Solvers with generic callbacks will provide a `where` argument to the callback function, indicating what information can and cannot be queried/submitted (the correspondence would be indicated in the solver's docs).

I agree, this could be included in `callback_data`.

> One question I don't have an answer to is: what about solvers that don't do generic callbacks (e.g. XPress)? How do you break a generic callback into usercut callback / heuristic callback, etc?

You could put in the `callback_data` that this is a `usercut`, etc... so that the wrapper can redirect appropriately and have enough info to answer `can_submit`, ...

src/attributes.jl
```diff
@@ -394,13 +402,39 @@ function submit(model::ModelLike, sub::AbstractSubmittable, args...)
...
-struct SettingSingleVariableFunctionNotAllowed <: Exception end
+struct LazyConstraint <: AbstractSubmittable
+    callback_data::Any
+end
+
+"""
+    HeuristicSolution(callback_data)
+
+Heuristically obtained integer-feasible solutions. The solution is given by a
+`Dict{VariableIndex, Float64}` mapping the variables to their solution value.
```
mtanneau: Why a `Dict{VariableIndex, Float64}`?

Wouldn't it be easier to pass a `Vector{VariableIndex}` containing the list of variables, and a `Vector{<:Number}` containing the associated values?

Then, to post a heuristic solution, one does something like
```julia
MOI.submit(model, ::MOI.CbHeuristicSolution, vars::Vector{VariableIndex}, vals::Vector{<:Number})
```

(I wrote `<:Number` to allow more multiple-precision and possibly complex numbers)

mtanneau:
> How would informational callbacks fit in MIPCallback ?

To implement an informational callback, just define a `MIPCallback` that only queries the information you want (as long as the solver you're using supports querying it).
I guess my point was: if MOI has a generic callback, then we don't need to change anything the day someone wants to support some fancy feature. I think it would make the codebase easier to maintain.

> if there is enough info in the callback_data (let's say where is also in callback_data), then we might be able to implement can_get(::Optimizer, SomeAttribute(callback_data) and can_submit(::Optimizer, SomeSubmittable(callback_data). This is still run-time but it allows to avoid having solvers throwing errors.

Wouldn't this be duplicating code? Solvers already perform this check, and raise an error if needed.
Implementing `can_get(::Optimizer, ...)` and `can_submit(::Optimizer, ...)` essentially does the same, but at the solver's MOI wrapper level.
Why not pass on the call to the solver, and intercept the error instead?

blegat:
Suppose you have some algo to generate a heuristic solution but it takes some time. With `can_submit` you can check before running the algo if it's worth it

mtanneau:
> You could put in the callback_data that this is a usercut, etc... so that the wrapper can redirect appropriately and have enough info to answer can_submit, ...

The problematic part is more about registering the callback in the first place.
Say you define (at MOI level) a generic callback function `my_callback`. That callback function is then passed on to the Xpress.jl wrapper, which passes it to the Xpress object.
What callback do you register with Xpress? You'd have to pretty much cover all possible cases...

One possibility (but quite cumbersome) would be to have Xpress.jl register a callback function for every possible specific callback in Xpress, each of them calling `my_callback` with the right argument.

blegat:
To correct something I said in earlier posts: In fact, we don't need `can_submit`, `supports` is already appropriate. When you write `MOI.supports(optimizer, MOI.LazyConstraint(callback_data))`, the optimizer can see in `callback_data` whether lazy constraints can be submitted inside the callback (since it sees which type of callback it is) hence it can return `false` or `true` depending on this. This does not depend on the state of `optimizer` but only on the value of the submittable so it satisfy the definition of `supports` and we don't need to add `can_submit`.

Thinking more about it, I prefer solution 2 of https://github.com/JuliaOpt/MathOptInterface.jl/pull/782#issuecomment-510100413. That is, similarly to `MOI.TerminationStatus` which tells *why* the solvers *stopped*. The type of the callback would say *why* the solver *called*, e.g.
* It found a feasible solution and asks whether you have any last word (e.g. by submitting lazy constraints you were hidding from him).
* It found a infeasible solution and asks whether you want to help (e.g. by submitting heuristic solutions or user cuts).
This type of fallbacks kind of defines what the user can `MOI.get` but one can do `try`-`catch` to be sure to be solver independent.
For submitting, the user can do `supports` to see if it's worth doing all the computation to get something to submit, e.g.
```julia
if MOI.supports(optimizer, MOI.UserCut(callback_data))
    func, set = # Long computation
    MOI.submit(optimizer, MOI.UserCut(callback_data), (func, set))
end
```

@mtanneau You suggest using rather solution 3. So a single callback and you have a way to query *why* it was *called* (or *from where* it was *called* but it seems kind of the same), let's say by `MOI.get(optimizer, MOI.WhyDidYouCall())`.
It seems rather similar in practice:
* In solution 2, for each *why*, you define a different a different callback.
* In solution 3, for each *why*, you define a different value of `MOI.WhyDidYouCall`.

However, there are several reasons for which IMO solution 2 is more appropriate for solver independent callbacks in MOI:
* In solution 3, it's impossible to have a solver-independent `enum` that covers all possible *why* (or *where*) to return for `MOI.WhyDidYouCall` so we would need to return different types which is slightly inconsistent with the rest of MOI where `MOI.get` usually returns enums, functions, sets or numbers. Some types will be defined in MOI and the solver specific ones will be defined in the solver wrappers.
* In solution 3, having to deal with everything in the same callback would give a spagetti code difficult to maintain. If you have a code for, say a `Gurobi` specific callback in addition to your code for solver-indepedent ones, you have to add it to the mix. And since the type returned by `MOI.WhyDidYouCall` is defined in the Gurobi wrapper, you have to use a trick to only execute this `ifelse` condition if the Gurobi package is loaded. While in solution 2, you can `MOI.set` the solver-specific callback next to `using Gurobi; solver = with_optimizer(Gurobi.Optimizer)` in a `gurobi.jl` file that you load when you use Gurobi.
* In solution 3, we want to define one callback to rule them all. But to be consistent, that means NLP callbacks should also be part of it, that means that  `MOI.WhyDidYouCall` could return `NeedObjectiveValueAt`, `NeedGradientAt`, `NeedHessianAt`. Here we clearly see that having everything in the same function is limiting since we need to mix the NLP autodiff code with the code for generating lazy constraints etc... all in the same functions. While it is feasible in theory to create a function grouping all different callbacks in the wrappers, it is inconvenient as it would need to recreate this function everytime a callback is added.

I am not saying that solution 3 was a bad design decision for Gurobi. The argument above only apply for solver-independent callbacks for MOI. Not for solver-dependent ones in C.
For instance, I suppose that it might be convenient to have a single function in C so that you can share values with `static` variables, this clearly does not apply to Julia where you can easily defined closures, etc...

> One possibility (but quite cumbersome) would be to have Xpress.jl register a callback function for every possible specific callback in Xpress, each of them calling `my_callback` with the right argument.

Yes, that would be the way if we go for solution 3.

src/attributes.jl
blegat: Indeed, it would be more consistent with the vectorized version of `set`.

rschwarz:
> FYI, here is a table of what MIP solvers currently support. `Y` is yes, `N` is no, `?` is I don't know. Docs are linked in the names.

It's not clear to me what you mean by "generic callback" (but it might be covered by SCIP's event handlers). Other than that, I'm confident that SCIP provides *functionality* that is equivalent (and goes beyond) what other solvers offer in terms of callbacks. However, the plugin system in SCIP typically requires that the user implement multiple callback methods and only imperfect mappings are possible (as we did with the MPB-wrapper of SCIP).
See [SCIP docs](https://scip.zib.de/doc-6.0.2/html/HOWTOADD.php) and related SCIP.jl issue: [#94](https://github.com/SCIP-Interfaces/SCIP.jl/issues/94). Because SCIP's plugins are so idiosyncratic, I think it's best not to try and make solver-independent callbacks work now.

mtanneau:
I still want to believe in generic callbacks, but... I don't see any of it being practical (or consensual) anytime soon.

So @blegat I'm overturning my previous comments, and I support having a small set of specific callbacks that expose a limited set of functionalities. That should be enough to satisfy most users and hopefully be merged soon.

Coming soon: a table with what you can get from where for various solvers. I hope it will help in deciding what deserves to be in MOI.

> It's not clear to me what you mean by "generic callback"

@rschwarz same terminology as [CPLEX' generic callback](https://www.ibm.com/support/knowledgecenter/SSSA5P_12.9.0/ilog.odms.cplex.help/CPLEX/UsrMan/topics/progr_adv/callbacks/defnCallbacks.html).
Basically, a "generic callback" wraps-up all possible callbacks you may think of.

blegat:
PR updated based on the discussion above as well as discussion with @ccoffrin, @harshangrjn and @kaarthiksundar.

mtanneau:
Continuing the discussion (@blegat this is mostly aligned with what you're proposing).

I would suggest to go with the following minimal set of functionalities which _should_ be supported by most (if not all) MIP solvers:
* **Heuristic callback**: submit a heuristic solution. The user can query the current value of the primal solution, which is (generally) a fractional point.
* **User cut callback**: submit a cut that's violated by the current _fractional_ point. The user can query the current value of the solution, which is in general fractional. The current solution may be integer (e.g., if the current LP relaxation yields an integer point).
* **Lazy constraint callback**: submit a constraint that is violated by the current point, and may cut some integer point. The user can query the current value of the primal solution, which may be fractional (e.g., at a node) or integer (e.g., at an incumbent).

To motivate this, here's a compilation of what (some) MIP solvers supports in terms of callback functionality.
`H` is heuristic (submit a heuristic solution), `U` is user cut (cut a fractional point), `L` is lazy constraint (cut an integer or fractional point).

 Solver     | Node relaxation | Incumbent |
|-----------|:-----:|:-----:|
| CPLEX	 	| `H`, `U`           | `H`, `L`
| GLPK      | `H`, `L`, `U` | X
| Gurobi	| `H`, `L`, `U` | `L`
| Xpress 	| `H`, `U`  	| ?

For instance: Gurobi allows to submit heuristic solutions, user cuts and lazy constraints after solving a node relaxation, and only lazy constraints when an incumbent is found.

Notes & caveats.
* CPLEX:
    * Docs do not specify in which context one can post a heuristic solution, so I put it in both cases.
    * No explicit mention of "lazy constraint" in the docs (none that I could find). For rejecting and incumbent, the function is `CPXcallbackrejectcandidate` which can submit constraints that cut off the integer point. No guarantee these "lazy constraints" will be enforced though
* GLPK:
    * Incumbent callback (`GLP_IBINGO`) is for information only.
    I don't know whether a node callback is called every time an incumbent is found by one of GLPK's internal heuristics (and this is not mentioned in the docs).
* Gurobi:
    * Heuristic solution can only be added when at a node relaxation.
    * Lazy constraints may be added at node relaxation, or when an incumbent is found. In both cases, the user has access to the value of the current primal solution (either fractional or integer).
* Xpress:
    * I could not find any mention of "lazy constraints" in Xpress docs. Either lazy constraints can be added through `XPRSaddcuts`, or there are no "lazy constraint" equivalent in Xpress, or there's a function I missed.
    * [Docs](https://www.fico.com/fico-xpress-optimization/docs/latest/solver/optimizer/HTML/XPRSaddcuts.html) (Further info #3) say cuts can only be added at a node relaxation
    * Incumbent solution can be rejected (see [here](https://www.fico.com/fico-xpress-optimization/docs/latest/solver/optimizer/HTML/XPRSaddcbpreintsol.html)), but it's not clear whether there's a procedure to include a violated cut from an incumbent.

rschwarz:
A question for understanding: Is the idea that the user can only add a single callback function for each type (lazy constraint, heuristic) via the new attribute?

So, if I have a model where I want to use represent two separate constraints via lazy constraints, I would have to create a wrapper callback that checks violation and adds cuts for either of them?

blegat:
> A question for understanding: Is the idea that the user can only add a single callback function for each type (lazy constraint, heuristic) via the new attribute?

Yes, in the current design there is only one callback. If you set another callback it overwrites the previous one. The wrapper approach would be a solution.

blegat:
@mtanneau I agree with your latest comment, let's define `HeuristicCallback`, `UserCutCallback` and `LazyConstraintCallback`. That corresponds to solution 3 of https://github.com/JuliaOpt/MathOptInterface.jl/pull/782#issuecomment-510100413
I have removed `RejectSolution`, it can be added later.

I have updated the PR, last call for objections.

src/attributes.jl
```diff
@@ -461,6 +517,136 @@ struct RawParameter <: AbstractOptimizerAttribute
...
+The feasible primal solution is accessed through
+[`CallbackVariablePrimal`](@ref). Trying to access other result
+attributes will throw [`OptimizeInProgress`](@ref) as discussed in
+[`AbstractCallback`](@ref).
+
+## Examples
+
+```julia
+x = MOI.add_variables(optimizer, 8)
+MOI.submit(optimizer, MOI.LazyConstraintCallback(), callback_data -> begin
```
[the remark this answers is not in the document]
mtanneau: The docs should state whether one can register multiple callbacks of the same type (with the same optimizer). For instance, if the user has two types of lazy constraints, can she register two `LazyConstraintsCallback`, or should she merge them into a single one?
Similarly, state what happens if two callbacks of the same type are register (are the two active? does the last one replace the previous one? is an error thrown?)

Convention-wise, the simpler is probably to go with:
* at most one callback of each type can be registered
* If an optimizer already has a callback `cb1::CBType`, and the user registers a callback `cb2::CBType`, then `cb1` is erased, and `cb2` is registered.

Users with multiple callbacks (of the same type, e.g. different heuristics, or different types of lazy constraints) can still wrap them in a single callback (e.g., merge all heuristics within one `HeuristicCallback`)

src/attributes.jl
```diff
@@ -461,6 +517,136 @@ struct RawParameter <: AbstractOptimizerAttribute
...
+    abstract type AbstractCallback <: AbstractOptimizerAttribute end
+
+Abstract type for optimizer attribute representing a callback functions. The
+value set to subtypes of `AbstractCallback` is a function that may be called
+during [`optimize!`](@ref). As [`optimize!`](@ref) is in progress, the result
+attributes (i.e, the attributes `attr` such that `is_set_by_optimize(attr)`)
+may not be accessible from the callback hence trying to get result attributes
+might throw a [`OptimizeInProgress`](@ref) error.
+
+Some solvers require a complete solution, others only partial solutions. It's up
```
mtanneau: > Some solvers require a complete solution, others only partial solutions...

Shouldn't this go into `HeuristicSolution`?

src/attributes.jl
```diff
@@ -461,6 +517,136 @@ struct RawParameter <: AbstractOptimizerAttribute
...
+"""
+    UserCutCallback() <: AbstractCallback
+
+The callback can be used to submit [`UserCut`](@ref) given the current primal
+solution. For instance, it may be called at fractional (i.e., non-integer) nodes
+in the branch and bound tree of a mixed-integer problem. Note that there is not
+guarantee that the callback is called *everytime* the solver has an infeasible
+solution.
+
+The infeasible solution is accessed through
```
mtanneau: > infeasible solution

This terminology might be confusing in MILP context, since. "Fractional solution" would be less ambiguous.

src/attributes.jl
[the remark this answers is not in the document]
blegat: Yes, let's start with only one callback, it should be `MOI.set`, not `MOI.submit`.

src/attributes.jl
blegat: Let's stay general and say "current solution" to cover the case of heuristic callback from incumbent.

src/attributes.jl
```diff
@@ -397,13 +405,64 @@ function submit(model::ModelLike, sub::AbstractSubmittable, args...)
...
+similarly to [`set`](@ref).
+
+This can be submitted only from the [`HeuristicCallback`](@ref). The
+field `callback_data` is a solver-specific callback type that is passed as the
+argument to the heuristic callback.
+
+Some solvers require a complete solution, others only partial solutions. It's up
+to you to provide the appropriate one. If in doubt, give a complete solution.
+
+Note that the solver may silently reject the provided solution.
```
mlubin: Gurobi at least provides a mechanism for finding out if the solution was accepted: https://www.gurobi.com/documentation/8.1/refman/c_grbcbsolution.html (see the `objP` argument). This is *very* useful for unit testing, so we may want to think of a way to expose it. Maybe the `HeuristicSolution` can hold a `Ref{SomeEnumType}` where `SomeEnumType` tells you if the solution was accepted, rejected, or unknown?

src/attributes.jl
```diff
@@ -461,6 +520,137 @@ struct RawParameter <: AbstractOptimizerAttribute
...
+Abstract type for optimizer attribute representing a callback functions. The
+value set to subtypes of `AbstractCallback` is a function that may be called
+during [`optimize!`](@ref). As [`optimize!`](@ref) is in progress, the result
+attributes (i.e, the attributes `attr` such that `is_set_by_optimize(attr)`)
+may not be accessible from the callback hence trying to get result attributes
+might throw a [`OptimizeInProgress`](@ref) error.
+
+At most one callback of each type can be registered. If an optimizer already
+has a function for a callback type, and the user registers a new function,
+then the old one is erased, and the new one is registered.
```
mlubin: "the old one is erased, and the new one is registered" can be shortened to "the old one is replaced"

src/attributes.jl
rschwarz: Would it be better to use "solution candidate", then, because it's not known whether the values are feasible, and can be a solution? In terms of B&B tree nodes, one could also talk of current relaxation, or current relaxation's solution.

src/attributes.jl
blegat: We could also say that it returns the enum. We never said that `submit` always returns nothing, it could something that depends on the `Submittable`

PR #785, opened 2019-07-11
 Add AbstractCallback and OptimizeInProgress

blegat:
This is a subset of https://github.com/JuliaOpt/MathOptInterface.jl/pull/782 that should be less subject to bikeshedding and that would allow solver specific callbacks to start using `AbstractCallback` and `OptimizeInProgress`.

cc @mtanneau

### DIFF
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -470,6 +470,7 @@ end
 
 """
     abstract type AbstractCallback <: AbstractOptimizerAttribute end
+
 Abstract type for optimizer attribute representing a callback functions. The
 value set to subtypes of `AbstractCallback` is a function that may be called
 during [`optimize!`](@ref). As [`optimize!`](@ref) is in progress, the result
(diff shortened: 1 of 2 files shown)


chriscoey:
Bump now that moi 0.9 is out

mlubin:
IMO it should be a higher priority to fully complete the release of MOI 0.9 by releasing JuMP 0.20 and making sure the relevant solvers are available so that fall semester courses don't have serious installation issues. After that callbacks are at the top of the list.

src/attributes.jl
```diff
@@ -449,6 +449,39 @@ struct RawParameter <: AbstractOptimizerAttribute
...
+    attr::AttrType
+end
+
+function Base.showerror(io::IO, err::OptimizeInProgress)
+    print(io, typeof(err), ": Cannot get result as the `MOI.optimize!` has not",
+          " finished.")
+end
+
+"""
+    abstract type AbstractCallback <: AbstractOptimizerAttribute end
```
rschwarz: Wouldn't `AbstractCallback` be a `Submittable` instead? That is, one can usually more than one callback to a solver?

blegat: That could work. That would allow callbacks to be added independently from different places that may not know about each other (as we discussed with @juan-pablo-vielma ).
One drawback is that since `submit` does not return any index, we would not be allowed to remove a callback.
With the attribute, setting it to another callback removes the previous one.

rschwarz: Thanks, I have now learned that most solvers only work with a single callback (different from SCIP), so this choice makes sense for the solver-independent infrastructure.

blegat:
Closed in favor of https://github.com/JuliaOpt/MathOptInterface.jl/pull/782

Issue #780, opened 2019-07-05
Address TODOs in apimanual.md


mlubin:
`[Discuss how models are constructed, optimizer attributes.]`
`[Define notation more precisely. ``a`` vector; ``A`` matrix; don't reuse ``u,l,b`` as scalar and vector]`
`[ needs formatting help, doc tests ]`

PR #786, opened 2019-07-11
Address TODOs

blegat:
Notably transition the knapsack example to a doctest with `MockOptimizer`. More doctests are added in https://github.com/JuliaOpt/MathOptInterface.jl/pull/759.

Closes https://github.com/JuliaOpt/MathOptInterface.jl/issues/780

### DIFF
--- a/docs/src/apimanual.md
+++ b/docs/src/apimanual.md
@@ -153,7 +153,7 @@ Models are constructed by
 * adding variables using [`add_variables`](@ref) (or [`add_variables`](@ref)),
   see [Adding variables](@ref);
 * setting an objective sense and function using [`set`](@ref),
-  see [Setting objective](@ref).
+  see [Setting an objective](@ref).
 * and adding constraints using [`add_constraint`](@ref) (or
   [`add_constraints`](@ref)), see [Sets and Constraints](@ref).
 
@@ -224,7 +224,7 @@ the function ``5x_1 - 2.3x_2 + 1``.
     `[ScalarAffineTerm(5.0, x[1]), ScalarAffineTerm(-2.3, x[2])]`. This is
     Julia's broadcast syntax and is used quite often.
 
-### Setting objective
+### Setting an objective
 
 Objective functions are assigned to a model by setting the
 [`ObjectiveFunction`](@ref) attribute. The [`ObjectiveSense`](@ref) attribute is
@@ -484,7 +484,6 @@ non-global tree search solvers like
 
 ## A complete example: solving a knapsack problem
 
-[ needs formatting help, doc tests ]
 We first need to select a solver supporting the given problem (see
 [`supports`](@ref) and [`supports_constraint`](@ref)). In this example, we
 want to solve a binary-constrained knapsack problem:
@@ -493,37 +492,13 @@ want to solve a binary-constrained knapsack problem:
 using GLPK
 optimizer = GLPK.Optimizer()
 ```
-we can check that it supports the objective as follows:
-```jldoctest knapsack; setup = :(optimizer = MOI.Utilities.MockOptimizer(MOI.Utilities.Model{Float64}()); MOI.Utilities.set_mock_optimize!(optimizer, mock -> MOI.Utilities.mock_optimize!(mock, ones(3))))
-MOI.supports(optimizer, MOI.ObjectiveFunction{MOI.ScalarAffineFunction{Float64}}())
-
-# output
-
-true
-```
-we can check that it supports the knapsack constraint as follows:
-```jldoctest knapsack
-MOI.supports_constraint(optimizer, MOI.ScalarAffineFunction{Float64}, MOI.LessThan{Float64})
-
-# output
-
-true
-```
-and we can check that it supports binary variables as follows:
-```jldoctest knapsack
-MOI.supports_constraint(optimizer, MOI.SingleVariable, MOI.ZeroOne)
-
-# output
-
-true
-```
 We first define the constants of the problem:
-```jldoctest knapsack
+```jldoctest knapsack; setup = :(optimizer = MOI.Utilities.MockOptimizer(MOI.Utilities.Model{Float64}()); MOI.Utilities.set_mock_optimize!(optimizer, mock -> MOI.Utilities.mock_optimize!(mock, ones(3))))
 c = [1.0, 2.0, 3.0]
 w = [0.3, 0.5, 1.0]
 C = 3.2
 
-num_variables = length(c)
+num_variables_to_create = length(c)
 
 # output
 
@@ -531,7 +506,7 @@ num_variables = length(c)
 ```
 We create the variables of the problem and set the objective function:
 ```jldoctest knapsack
-x = MOI.add_variables(optimizer, num_variables)
+x = MOI.add_variables(optimizer, num_variables_to_create)
 objective_function = MOI.ScalarAffineFunction(MOI.ScalarAffineTerm.(c, x), 0.0)
 MOI.set(optimizer, MOI.ObjectiveFunction{MOI.ScalarAffineFunction{Float64}}(),
         objective_function)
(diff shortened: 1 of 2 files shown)


docs/src/apimanual.md
```diff
@@ -210,6 +224,8 @@ the function ``5x_1 - 2.3x_2 + 1``.
     `[ScalarAffineTerm(5.0, x[1]), ScalarAffineTerm(-2.3, x[2])]`. This is
     Julia's broadcast syntax and is used quite often.
 
+### Setting objective
```
mlubin: "Setting an objective"

docs/src/apimanual.md
```diff
@@ -470,58 +485,128 @@ non-global tree search solvers like
 ## A complete example: solving a knapsack problem
 
 [ needs formatting help, doc tests ]
```
mlubin: This text can be removed :)

docs/src/apimanual.md
```diff
@@ -470,58 +485,128 @@ non-global tree search solvers like
...
+```jldoctest knapsack; setup = :(optimizer = MOI.Utilities.MockOptimizer(MOI.Utilities.Model{Float64}()); MOI.Utilities.set_mock_optimize!(optimizer, mock -> MOI.Utilities.mock_optimize!(mock, ones(3))))
+MOI.supports(optimizer, MOI.ObjectiveFunction{MOI.ScalarAffineFunction{Float64}}())
 
-# Solves the binary-constrained knapsack problem:
-# max c'x: w'x <= C, x binary using GLPK.
+# output
 
+true
+```
+we can check that it supports the knapsack constraint as follows:
```
mlubin: I don't know if showing how to check if the solver supports the constraints is helpful at the start of the example. This isn't part of the normal workflow.

If you know that GLPK supports the problem type, then you just use it. If you don't know, then this approach of calling `supports_constraint` isn't especially helpful. Are you supposed to go through this exercise with each possible solver until you find one that works?

docs/src/apimanual.md
```diff
@@ -470,58 +485,128 @@ non-global tree search solvers like
...
+
+true
+```
+We first define the constants of the problem:
+```jldoctest knapsack
 c = [1.0, 2.0, 3.0]
 w = [0.3, 0.5, 1.0]
 C = 3.2
 
 num_variables = length(c)
```
mlubin: Perhaps `num_variables_to_create`?

docs/src/apimanual.md
```diff
@@ -470,58 +485,128 @@ non-global tree search solvers like
...
+It found the optimal solution! Now let's see what is that solution.
+But first, let's check if it has more than one solution to share:
+```jldoctest knapsack
+MOI.get(optimizer, MOI.ResultCount())
+
+# output
+
+1
+```
+Only one. As the termination status is `MOI.OPTIMAL` and there is only one
```
mlubin: This merits a note on why a solver might return more than one solution.

docs/src/apimanual.md
blegat: Now, it was just to illustrate the use of `supports_constraint`. It might not be the best place for a first example

mlubin: Readers will be looking at this section as a first example of how to use MOI. It's good to have an illustration of `supports_constraint`, but it could be in a more advanced example.

blegat: Yes, I kind of forgot why I included it. Removed :)

docs/src/apimanual.md
```diff
@@ -290,9 +306,8 @@ add_constraint(model, VectorOfVariables([x,y,z]), SecondOrderCone(3))
 
 Below is a list of common constraint types and how they are represented
 as function-set pairs in MOI. In the notation below, ``x`` is a vector of decision variables,
-``x_i`` is a scalar decision variable, and all other terms are fixed constants.
-
-[Define notation more precisely. ``a`` vector; ``A`` matrix; don't reuse ``u,l,b`` as scalar and vector]
+``x_i`` is a scalar decision variable, ``\alpha, \beta`` are scalar constants,
+``a, b`` are a constant vectors and `A` is a constant matrix.
```
mlubin: There's another TODO in the section below that I missed in #780: [Define ``\mathbb{R}_+, \mathbb{R}_-``]. You don't have to address it in this PR, but we should leave that issue open if not.

mlubin: s/are a constant vectors/are constant vectors/

docs/src/apimanual.md
```diff
@@ -469,59 +484,115 @@ non-global tree search solvers like
...
-obj_value = MOI.get(optimizer, MOI.ObjectiveValue())
-if termination_status != MOI.OPTIMAL
-    error("Solver terminated with status $termination_status")
-end
+# output
+
+```
+The first thing to check after optimization is why the solver stopped, e.g.,
+did it stop because of a time limit or did it stop because it found the optimal
+solution ?
```
mlubin: We usually don't put spaces before question marks.

docs/src/apimanual.md
```diff
@@ -469,59 +484,115 @@ non-global tree search solvers like
...
+# output
+
+1
+```
+Only one.
 
-@assert MOI.get(optimizer, MOI.ResultCount()) > 0
+!!! note
+    While the value of `MOI.get(optimizer, MOI.ResultCount())` is often one, it
+    is important to check its value in order to write a robust code. For
```
mlubin: Not everyone wants to write robust code. For example, it really doesn't matter in a classroom. You could say that robust code should check it.

docs/src/apimanual.md
```diff
@@ -469,59 +484,115 @@ non-global tree search solvers like
...
+    instance, when the problem is unbounded, the solver might return two
```
mlubin: Which solvers, if any, actually do this?

blegat: [VRepOptimizer](https://github.com/JuliaPolyhedra/Polyhedra.jl/blob/master/src/vrep_optimizer.jl) does not currently do it but it should. Apart from that, not sure. I chose the example that makes most sense in theory. Note that the example that `ResultCount` could be zero if it is infeasible is already discussed in `apimanual`.

mlubin: The theory doesn't really matter if there are zero solvers that do this. Another example in theory is MIP solvers that could return multiple suboptimal solutions, but AFAIK this isn't implemented anywhere yet. Maybe we could remove the discussion about result counts from the intro example.

docs/src/apimanual.md
```diff
@@ -469,59 +486,94 @@ non-global tree search solvers like
...
+MAX_SENSE::OptimizationSense = 1
+```
+We add the knapsack constraint and integrality constraints:
+```jldoctest knapsack
 knapsack_function = MOI.ScalarAffineFunction(MOI.ScalarAffineTerm.(w, x), 0.0)
 MOI.add_constraint(optimizer, knapsack_function, MOI.LessThan(C))
-
-# Add integrality constraints.
-for i in 1:num_variables
+for i in 1:num_variables_to_create
```
mlubin: Looping over `num_variables_to_create` after the variables are created doesn't read well. What about looping over `1:length(x)`? Or remove `num_variables_to_create` and do `MOI.add_variables(optimizer, length(c))`.

docs/src/apimanual.md
```diff
@@ -138,15 +142,25 @@ from the [`ModelLike`](@ref) abstract type.
...
 
 Through the rest of the manual, `model` is used as a generic `ModelLike`, and
 `optimizer` is used as a generic `AbstractOptimizer`.
 
-[Discuss how models are constructed, optimizer attributes.]
+Models are constructed by
+* adding variables using [`add_variables`](@ref) (or [`add_variables`](@ref)),
+  see [Adding variables](@ref);
+* setting an objective sense and function using [`set`](@ref),
+  see [Setting an objective](@ref).
```
mlubin: Nit: semicolon instead of period at the end for consistency with above. (Or you could use periods everywhere.)

PR #789, opened 2019-07-13
Add Objective bridges

blegat:
Note that the base branch set is `bl/variable_bridge` but I don't plan to merge it into it, it's only set to have a small diff. This PR will only be merged once `bl/variable_bridge` is merged into master.

Closes https://github.com/JuliaOpt/MathOptInterface.jl/issues/529

### DIFF
--- a/src/Bridges/Objective/functionize.jl
+++ b/src/Bridges/Objective/functionize.jl
@@ -24,10 +24,14 @@ function MOIB.set_objective_function_type(::Type{FunctionizeBridge{T}}) where T
     return MOI.ScalarAffineFunction{T}
 end
 
+# No variables or constraints are created in this bridge so there is nothing to
+# delete.
 function MOI.delete(model::MOI.ModelLike, bridge::FunctionizeBridge) end
 
 function MOI.set(::MOI.ModelLike, ::MOI.ObjectiveSense,
                  ::FunctionizeBridge, ::MOI.OptimizationSense)
+    # `FunctionizeBridge` is sense agnostic, therefore, we don't need to change
+    # anything.
 end
 function MOI.get(model::MOI.ModelLike,
                  attr::MOIB.ObjectiveFunctionValue{MOI.SingleVariable},
--- a/src/Bridges/Objective/functionize.jl
+++ b/src/Bridges/Objective/functionize.jl
@@ -24,6 +24,14 @@ function MOIB.set_objective_function_type(::Type{FunctionizeBridge{T}}) where T
     return MOI.ScalarAffineFunction{T}
 end
 
+# Attributes, Bridge acting as a model
+function MOI.get(bridge::FunctionizeBridge, ::MOI.NumberOfVariables)
+    return 0
+end
+function MOI.get(bridge::FunctionizeBridge, ::MOI.ListOfVariableIndices)
+    return MOI.VariableIndex[]
+end
+
 # No variables or constraints are created in this bridge so there is nothing to
 # delete.
 function MOI.delete(model::MOI.ModelLike, bridge::FunctionizeBridge) end
--- a/src/Bridges/Objective/functionize.jl
+++ b/src/Bridges/Objective/functionize.jl
@@ -1,0 +1,43 @@
+"""
+    FunctionizeBridge{T}
+
+The `FunctionizeBridge` converts a `SingleVariable` objective into a
+`ScalarAffineFunction{T}` objective.
+"""
+struct FunctionizeBridge{T} <: AbstractBridge end
+function bridge_objective(::Type{FunctionizeBridge{T}}, model::MOI.ModelLike,
+                           func::MOI.SingleVariable) where T
+    F = MOI.ScalarAffineFunction{T}
+    MOI.set(model, MOI.ObjectiveFunction{F}(), convert(F, func))
+    return FunctionizeBridge{T}()
+end
+
+function supports_objective_function(
+    ::Type{<:FunctionizeBridge}, ::Type{MOI.SingleVariable})
+    return true
+end
+MOIB.added_constrained_variable_types(::Type{<:FunctionizeBridge}) = Tuple{DataType}[]
+function MOIB.added_constraint_types(::Type{<:FunctionizeBridge})
+    return Tuple{DataType, DataType}[]
+end
+function MOIB.set_objective_function_type(::Type{FunctionizeBridge{T}}) where T
+    return MOI.ScalarAffineFunction{T}
+end
+
+function MOI.delete(model::MOI.ModelLike, bridge::FunctionizeBridge) end
+
+function MOI.set(::MOI.ModelLike, ::MOI.ObjectiveSense,
+                 ::FunctionizeBridge, ::MOI.OptimizationSense)
+end
+function MOI.get(model::MOI.ModelLike,
+                 attr::MOIB.ObjectiveFunctionValue{MOI.SingleVariable},
+                 bridge::FunctionizeBridge{T}) where T
+    F = MOI.ScalarAffineFunction{T}
+    return MOI.get(model, MOIB.ObjectiveFunctionValue{F}())
+end
+function MOI.get(model::MOI.ModelLike, attr::MOI.ObjectiveFunction{MOI.SingleVariable},
+                 bridge::FunctionizeBridge{T}) where T
+    F = MOI.ScalarAffineFunction{T}
+    func = MOI.get(model, MOI.ObjectiveFunction{F}())
+    return convert(MOI.SingleVariable, func)
+end
--- a/src/Bridges/Objective/slack.jl
+++ b/src/Bridges/Objective/slack.jl
@@ -1,0 +1,76 @@
+"""
+    SlackBridge{T, F, G}
+
+The `SlackBridge` converts an objective function of type `G` into a
+[`SingleVariable`](@ref) objective by creating a slack variable and a
+`F`-in-[`LessThan`](@ref) constraint for minimization or
+`F`-in-[`LessThan`](@ref) constraint for maximization where `F` is
+`MOI.Utilities.promote_operation(-, T, G, MOI.SingleVariable}`.
+Note that when using this bridge, changing the optimization sense
+is not supported. Set the sense to `MOI.FEASIBILITY_SENSE` first
+in order to delete the bridge in order to change the sense.
+"""
+struct SlackBridge{T, F<:MOI.AbstractScalarFunction, G<:MOI.AbstractScalarFunction} <: AbstractBridge
+    slack::MOI.VariableIndex
+    constraint::Union{MOI.ConstraintIndex{F, MOI.LessThan{T}},
+                      MOI.ConstraintIndex{F, MOI.GreaterThan{T}}}
+end
+function bridge_objective(::Type{SlackBridge{T, F, G}}, model::MOI.ModelLike,
+                          func::G) where {T, F, G<:MOI.AbstractScalarFunction}
+    slack = MOI.add_variable(model)
+    fslack = MOI.SingleVariable(slack)
+    f = MOIU.operate(-, T, func, fslack)
+    if MOI.get(model, MOI.ObjectiveSense()) == MOI.MIN_SENSE
+        set = MOI.LessThan(zero(T))
+    elseif MOI.get(model, MOI.ObjectiveSense()) == MOI.MAX_SENSE
+        set = MOI.GreaterThan(zero(T))
+    else
+        error("Set `MOI.ObjectiveSense` before `MOI.ObjectiveFunction` when",
+              " using `MOI.Bridges.Objective.SlackBridge`.")
+    end
+    constraint = MOIU.add_scalar_constraint(model, f, set)
+    MOI.set(model, MOI.ObjectiveFunction{MOI.SingleVariable}(), fslack)
+    return SlackBridge{T, F, G}(slack, constraint)
+end
+
+function supports_objective_function(
+    ::Type{<:SlackBridge}, ::Type{MOI.SingleVariable})
+    return false
+end
+function supports_objective_function(
+    ::Type{<:SlackBridge}, ::Type{<:MOI.AbstractScalarFunction})
+    return true
+end
+MOIB.added_constrained_variable_types(::Type{<:SlackBridge}) = Tuple{DataType}[]
+function MOIB.added_constraint_types(::Type{<:SlackBridge{T, F}}) where {T, F}
+    return [(F, MOI.GreaterThan{T}), (F, MOI.LessThan{T})]
+end
+function MOIB.set_objective_function_type(::Type{<:SlackBridge})
+    return MOI.SingleVariable
+end
+function concrete_bridge_type(::Type{<:SlackBridge{T}},
+                              G::Type{<:MOI.AbstractScalarFunction}) where T
+    F = MOIU.promote_operation(-, T, G, MOI.SingleVariable)
+    return SlackBridge{T, F, G}
+end
+
+
+function MOI.delete(model::MOI.ModelLike, bridge::SlackBridge)
+    MOI.delete(model, bridge.constraint)
+    MOI.delete(model, bridge.slack)
+end
+
+function MOI.get(model::MOI.ModelLike,
+                 attr::MOIB.ObjectiveFunctionValue{G},
+                 bridge::SlackBridge{T, F, G}) where {T, F, G}
+    slack = MOI.get(model, MOIB.ObjectiveFunctionValue{MOI.SingleVariable}())
+    obj_slack_constant = MOI.get(model, MOI.ConstraintPrimal(), bridge.constraint)
+    # The constant was moved to the set as it is a scalar constraint.
+    constant = MOI.constant(MOI.get(model, MOI.ConstraintSet(), bridge.constraint))
+    return obj_slack_constant + slack + constant
+end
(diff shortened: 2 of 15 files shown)


chriscoey:
So this is for converting unsupported objective types to supported constraints? like quadratic obj to quadratic constraint? max logdet to max of a new variable and a logdet cone constraint? any weird use cases planned?

blegat:
> So this is for converting unsupported objective types to supported constraints? like quadratic obj to quadratic constraint? max logdet to max of a new variable and a logdet cone constraint? any weird use cases planned?

Yes, exactly log-det (that is supported by SDPT3 and should be bridged for others) and quadratic objectives which is suppoted by some solvers and should be transformed to RSOC are the main use cases.
The diff is abnormally large since it is based on `bl/variable_bridges` which has changed quite a lot since when it was written

chriscoey:
cool. so the objective bridges are aware of the objective sense? like can it check whether the sense is compatible with reformulating using a convex epi/hypograph constraint?

blegat:
> so the objective bridges are aware of the objective sense?

That's the most tricky part with objective bridges, it would have been easier if both sense and objective are set at the same times.
So yes, they are aware of objective sense but there are tricky cases that do not work now, e.g.
```julia
set sense to Min
set function to convex quad
set sense to Max # error
set function to concave quad
```

chriscoey:
seems like it would make sense to set an objective sense and function at same time. how breaking would it be to change the interface?

blegat:
It might be quite breaking. We might start with an error as this use case it not that common and evaluate if it's worth the breakage.

chriscoey:
I'm thinking of cases like https://github.com/JuliaOpt/Pajarito.jl/blob/master/examples/expdesign.jl where you have different convex objectives some of which are "min convex function" and some that are "max concave function" and you want to switch between the objectives easily to see how they affect the solution. Or cases where you try different convex objectives for regression problems, some that are max and some min.

mlubin:
> seems like it would make sense to set an objective sense and function at same time. how breaking would it be to change the interface?

This would be a usability loss for the API without a clear justification. Why can't we look at the combination of objective sense and objective function and then decide which transformation to apply?

blegat:
> Why can't we look at the combination of objective sense and objective function and then decide which transformation to apply?

That's another approach, cache the objective function, in case the bridge is sensitive to min vs max, cache the objective function until `optimize!` is called hence you are sure to have the sense set

rschwarz:
From the user's point of view, when switching from "min convex" to "max concave", couldn't we clear the objective by switching to feasibility mode as an intermediate step?

rschwarz:
Also (without looking at the diff), does this PR allow the bridging of nonlinear objectives to nonlinear constraints (In the `NLPBlock` sense)?

blegat:
> From the user's point of view, when switching from "min convex" to "max concave", couldn't we clear the objective by switching to feasibility mode as an intermediate step?

Yes, that's a valid workaround

blegat:
Adding to the v0.9.2 milestone as currently, bridged variables cannot be used in SingleVariable objective which is a regression as [SDOI](https://github.com/JuliaOpt/SemidefiniteOptInterface.jl) could handle that. This PR fixes this issues by force-bridging the SingleVariable to a ScalarAffineFunction with Bridges.Objective.FunctionizeBridge

ericphanson:
> Adding to the v0.9.2 milestone as currently, bridged variables cannot be used in SingleVariable objective which is a regression as [SDOI](https://github.com/JuliaOpt/SemidefiniteOptInterface.jl) could handle that. This PR fixes this issues by force-bridging the SingleVariable to a ScalarAffineFunction with Bridges.Objective.FunctionizeBridge

Just to say @jiazhengzhu ran into this when trying to test SDPA-GMP with some JuMP examples.

blegat:
This PR should be ready now. The `SlackBridge` needs the objective sense to be set before the objective function and does not support the objective sense to be changed. A workaround is to set it to `FeasibilitySense` first to remove the bridge in order to change it from max to min or min to max. At a later stage, we could cache the objective and sense to and only set it before optimize but that's out of scope for this PR.

I have created a PR with an independent subset of the changes for easier review: https://github.com/JuliaOpt/MathOptInterface.jl/pull/867

src/Bridges/Constraint/det.jl
```diff
@@ -99,7 +99,7 @@ end
 
 MOI.supports_constraint(::Type{LogDetBridge{T}}, ::Type{<:Union{MOI.VectorOfVariables, MOI.VectorAffineFunction{T}}}, ::Type{MOI.LogDetConeTriangle}) where T = true
 MOIB.added_constrained_variable_types(::Type{<:LogDetBridge}) = Tuple{DataType}[]
-MOIB.added_constraint_types(::Type{LogDetBridge{T}}, ::Type{<:Union{MOI.VectorOfVariables, MOI.VectorAffineFunction{T}}}, ::Type{MOI.LogDetConeTriangle}) where T = [(MOI.VectorAffineFunction{T}, MOI.PositiveSemidefiniteConeTriangle), (MOI.VectorAffineFunction{T}, MOI.ExponentialCone), (MOI.ScalarAffineFunction{T}, MOI.LessThan{T})]
+MOIB.added_constraint_types(::Type{LogDetBridge{T}}) where T = [(MOI.VectorAffineFunction{T}, MOI.PositiveSemidefiniteConeTriangle), (MOI.VectorAffineFunction{T}, MOI.ExponentialCone), (MOI.ScalarAffineFunction{T}, MOI.LessThan{T})]
```
odow: Is this an un-related bug fix? Do we not have tests?

src/Bridges/Constraint/det.jl
```diff
@@ -197,7 +197,7 @@ end
 
 MOI.supports_constraint(::Type{RootDetBridge{T}}, ::Type{<:Union{MOI.VectorOfVariables, MOI.VectorAffineFunction{T}}}, ::Type{MOI.RootDetConeTriangle}) where T = true
 MOIB.added_constrained_variable_types(::Type{<:RootDetBridge}) = Tuple{DataType}[]
-MOIB.added_constraint_types(::Type{RootDetBridge{T}}, ::Type{<:Union{MOI.VectorOfVariables, MOI.VectorAffineFunction{T}}}, ::Type{MOI.RootDetConeTriangle}) where T = [(MOI.VectorAffineFunction{T}, MOI.PositiveSemidefiniteConeTriangle), (MOI.VectorAffineFunction{T}, MOI.GeometricMeanCone)]
+MOIB.added_constraint_types(::Type{RootDetBridge{T}}) where T = [(MOI.VectorAffineFunction{T}, MOI.PositiveSemidefiniteConeTriangle), (MOI.VectorAffineFunction{T}, MOI.GeometricMeanCone)]
```
odow: Ditto re bug-fix/tests.

src/Bridges/Objective/bridge.jl
```diff
@@ -0,0 +1,118 @@
...
+Return the value of the attribute `attr` of the model `model` for the
+variable bridged by `bridge`.
+"""
+function MOI.set(::MOI.ModelLike, ::MOI.ObjectiveSense,
+                 bridge::AbstractBridge, ::MOI.OptimizationSense)
+    throw(ArgumentError(
+        "Objective bridge of type `$(typeof(bridge))` does not support" *
+        " modifying the objective sense. As a workaround, set the sense to" *
+        " `MOI.FEASIBILITY_SENSE` to clear the objective function and" *
+        " bridges."))
```
odow: So what is JuMP going to do? Is it possible to easily set `FEASIBILITY_SENSE` from JuMP?

src/Bridges/Objective/bridge.jl
```diff
@@ -0,0 +1,118 @@
...
+end
+
+"""
+    function MOI.get(model::MOI.ModelLike, attr::MOI.ObjectiveFunction,
+                     bridge::AbstractBridge)
+
+Return the value of the objective function bridged by `bridge` for the model
```
odow: Not the value, the function object.

src/Bridges/Objective/bridge.jl
```diff
@@ -0,0 +1,118 @@
...
+add for bridging objective functions of type `F`. This fallbacks to
+`added_constraint_types(concrete_bridge_type(BT, S))`
+so bridges should not implement this method.
+"""
+function MOIB.added_constraint_types(
+    BT::Type{<:AbstractBridge}, F::Type{<:MOI.AbstractScalarFunction})
+    return MOIB.added_constraint_types(concrete_bridge_type(BT, F))
+end
+
+function MOIB.set_objective_function_type(
```
odow: Missing a docstring similar to the ones above?

src/Bridges/Objective/functionize.jl
```diff
@@ -0,0 +1,43 @@
...
+end
+MOIB.added_constrained_variable_types(::Type{<:FunctionizeBridge}) = Tuple{DataType}[]
+function MOIB.added_constraint_types(::Type{<:FunctionizeBridge})
+    return Tuple{DataType, DataType}[]
+end
+function MOIB.set_objective_function_type(::Type{FunctionizeBridge{T}}) where T
+    return MOI.ScalarAffineFunction{T}
+end
+
+function MOI.delete(model::MOI.ModelLike, bridge::FunctionizeBridge) end
```
odow: What does it mean to delete an objective bridge?

src/Bridges/Objective/functionize.jl
```diff
@@ -0,0 +1,43 @@
...
+
+function MOI.set(::MOI.ModelLike, ::MOI.ObjectiveSense,
+                 ::FunctionizeBridge, ::MOI.OptimizationSense)
+end
```
odow: When is this function called?

src/Bridges/Objective/map.jl
```diff
@@ -0,0 +1,81 @@
+"""
+    Map <: AbstractDict{MOI.ObjectiveFunction, AbstractBridge}
+
+Mapping between bridged variables and the bridge that bridged the variable.
+"""
+mutable struct Map <: AbstractDict{MOI.ObjectiveFunction, AbstractBridge}
```
odow: Does using `AbstractBridge` instead of `Any` hurt/help performance? (There were some recent discussions on Discourse about bad compile times from abstract types in dicts, e.g., https://discourse.julialang.org/t/compiler-performance-of-dict/27986/12?u=odow)

src/Bridges/Objective/map.jl
```diff
@@ -0,0 +1,81 @@
+"""
+    Map <: AbstractDict{MOI.ObjectiveFunction, AbstractBridge}
+
+Mapping between bridged variables and the bridge that bridged the variable.
```
odow: I don't think this is the correct sentence. Maybe:

A mapping from a bridged objective function to the bridge responsible for creating that objective function.

src/Bridges/Objective/map.jl
```diff
@@ -0,0 +1,81 @@
...
+    map.function_type = typeof(func)
+    map.bridges[attr] = bridge
+end
+
+"""
+    EmptyMap <: AbstractDict{MOI.ObjectiveFunction, AbstractBridge}
+
+Empty version of [`Map`](@ref). It is used by
+[`MathOptInterface.Bridges.Constraint.SingleBridgeOptimizer`](@ref) as it does
+not bridge any variable.
```
odow: Doc-string hasn't been updated.

src/Bridges/Objective/slack.jl
```diff
@@ -0,0 +1,76 @@
...
+    SlackBridge{T, F, G}
+
+The `SlackBridge` converts an objective function of type `G` into a
+[`SingleVariable`](@ref) objective by creating a slack variable and a
+`F`-in-[`LessThan`](@ref) constraint for minimization or
+`F`-in-[`LessThan`](@ref) constraint for maximization where `F` is
+`MOI.Utilities.promote_operation(-, T, G, MOI.SingleVariable}`.
+Note that when using this bridge, changing the optimization sense
+is not supported. Set the sense to `MOI.FEASIBILITY_SENSE` first
+in order to delete the bridge in order to change the sense.
```
odow: Set the sense to `MOI.FEASIBILITY_SENSE` first to delete the bridge in order to change the sense, then re-add the objective.

src/Bridges/bridge.jl
```diff
@@ -76,3 +76,11 @@ Return a list of the types of constraints that bridges of concrete type `BT`
 add. This is used by the [`LazyBridgeOptimizer`](@ref).
 """
 function added_constraint_types end
+
+"""
+    set_objective_function_type(BT::Type{<:Objective.AbstractBridge})::Bool
```
odow: `Bool` is incorrect as a return type.

src/Bridges/bridge_optimizer.jl
```diff
@@ -28,6 +28,11 @@ instead of passing it as is to its internal model.
 
 Return a `Bool` indicating whether `b` tries to bridge constrained variables in
 `S` instead of passing it as is to its internal model.
+
+    is_bridged(b::AbstractBridgeOptimizer, F::Type{<:MOI.AbstractFunction})::Bool
+
+Return a `Bool` indicating whether `b` tries to bridge objective functions of
```
odow: `tries to bridge` or `bridges`?

src/Bridges/bridge_optimizer.jl
```diff
@@ -450,6 +495,123 @@ function MOI.set(b::AbstractBridgeOptimizer,
...
+
+function MOI.supports(b::AbstractBridgeOptimizer,
+                      attr::MOI.ObjectiveFunction{F}) where F
+    if is_bridged(b, F)
+        return supports_bridging_objective_function(b, F)
+    else
+        return MOI.supports(b.model, attr)
+    end
+end
+struct ObjectiveFunctionValue{F<:MOI.AbstractScalarFunction} end
```
odow: Why is it necessary to have a different attribute? What's the difference between `ObjectiveFunctionValue` and `ObjectiveValue`?

src/Bridges/Objective/functionize.jl
blegat: When the sense of the model is changed, i.e. in `MOI.get(::AbstractBridgeOptimizer, ::ObjectiveSense, ::OptimizationSense)`.
The docstring in https://github.com/JuliaOpt/MathOptInterface.jl/pull/867/ should make that clear

src/Bridges/Objective/functionize.jl
blegat: I am adding a docstring in https://github.com/JuliaOpt/MathOptInterface.jl/pull/867/

src/Bridges/Objective/bridge.jl
blegat: It was added in https://github.com/JuliaOpt/MathOptInterface.jl/pull/867/

src/Bridges/Objective/bridge.jl
blegat: The user can do `set_objective_sense(model, MOI.FEASIBILITY_SENSE)`. This is not ideal but this is a rare use case. We can implement https://github.com/JuliaOpt/MathOptInterface.jl/pull/789#issuecomment-520178883 if this is an issue

src/Bridges/Objective/map.jl
blegat: I didn't know that it increase compile time, it's there mostly for type-safety, to make sure to get an error if anything else is stored there.
We can check if changing this improves things but I'd say it's out of the scope of this PR.

src/Bridges/bridge_optimizer.jl
blegat: For `LazyBridgeOptimizer`, `is_bridged(b, ...) = !supports_constraint(b.model, ...)` so it does not mean we can bridge it, it just mean that we should try since the inner model does not support it.

src/Bridges/bridge_optimizer.jl
blegat: Added a docstring

src/Bridges/Constraint/det.jl
blegat: It's now required for `fill_...` to work for LazyBridgeOptimizer. I have made a independent PR with additional tests: https://github.com/JuliaOpt/MathOptInterface.jl/pull/870

src/Bridges/Objective/functionize.jl
odow: See the comment for `ObjectiveSense`. Either add a descriptive comment (`No variables to delete`), or call it `_delete_variables_and_constraints_from_bridge`

src/Bridges/Objective/functionize.jl
odow: It would be useful to add a comment inside the function
```Julia
# FunctionizeBridge is sense agnostic, therefore, we don't need to tell it if the sense changes.
```
Alternatively, do these need to overload `MOI.set`? Why not just call it `_update_objectivesense_of_bridge`?

src/Bridges/Objective/functionize.jl
blegat: `MOI.delete` is consistent with the `MOI.delete` of variable and constraint bridges: "you are going to be deleted so first undo anything you have done." so I have kept the name `MOI.delete` and added the comment.

src/Bridges/Objective/functionize.jl
blegat: The reuse of `MOI.set` and `MOI.get` for bridges where bridges are used in place of the model of indices is common in bridges so I find it consistent, even if the meaning is not directly obvious without reading the docstring. I have added a comment to make it clearer.

PR #796, opened 2019-07-19
Define add_constrained_variable

blegat:

### DIFF
--- a/docs/src/apimanual.md
+++ b/docs/src/apimanual.md
@@ -971,7 +971,7 @@ MOI.set(model, MyPackage.PrintLevel(), 0)
 ### Supported constrained variables and constraints
 
 The solver interface should only implement support for constrained variables
-(see [`add_constrained_variable`](@ref)/[`add_constrained_variable`](@ref))
+(see [`add_constrained_variable`](@ref)/[`add_constrained_variables`](@ref))
 or constraints that directly map to a structure exploited by the solver
 algorithm. There is no need to add support for additional types, this is
 handled by the [Automatic reformulation](@ref). Furthermore, this allows
--- a/docs/src/apimanual.md
+++ b/docs/src/apimanual.md
@@ -929,7 +929,9 @@ If ``\mathcal{C}_i`` is a vector set, the discussion remains valid with
 ``y_i(\frac{1}{2}x^TQ_ix + a_i^T x + b_i)`` replaced with the scalar product
 between `y_i` and the vector of scalar-valued quadratic functions.
 
-### Constraint bridges
+### Automatic reformulation
+
+#### Constraint reformulation
 
 A constraint often possess different equivalent formulations, but a solver may only support one of them.
 It would be duplicate work to implement rewritting rules in every solver wrapper for every different formulation of the constraint to express it in the form supported by the solver.
@@ -965,6 +967,34 @@ Then, the user can write:
 model = MyPackage.Optimizer()
 MOI.set(model, MyPackage.PrintLevel(), 0)
 ```
+
+### Supported constrained variables and constraints
+
+The solver interface should only implement support for constrained variables
+or constraints that directly map to a structure exploited by the solver
+algorithm. There is no need to add support for additional types, this is
+handled by the [Automatic reformulation](@ref). Furthermore, this allows
+[`supports_constraint`](@ref) to indicate which types are exploited by the
+solver and hence allow layers such as [`Bridges.LazyBridgeOptimizer`](@ref)
+to accurately select the most appropriate transformations.
+
+As [`add_constrained_variable`](@ref) (resp. [`add_constrained_variables`](@ref))
+falls back to [`add_variable`](@ref) (resp. [`add_variables`](@ref)) followed by
+[`add_constraint`](@ref), there is no need to implement this function
+if `model` supports creating free variables and supports transforming free
+variables into variables in `set`. However,
+
+* if `model` does not support creating free variables, it should only implement
+  `add_constrained_variable` and not [`add_variable`](@ref) nor
+  [`add_constraint`](@ref) for [`SingleVariable`](@ref)-in-`typeof(set)`.
+  In addition, it should implement `supports_constraint(::Optimizer,
+  ::Type{VectorOfVariables}, ::Type{Reals})` and return `false` so that free
+  variables are bridged, see [`supports_constraint`](@ref).
+* if `model` supports free variables but does not support transforming free
+  variables into variables in `set`, then it should implement both
+  [`add_variable`](@ref) and `add_constraint_variable` but should not implement
+  any method for [`add_constraint`](@ref) for
+  [`SingleVariable`](@ref)-in-`typeof(set)`.
 
 ### Implementing copy
 
--- a/docs/src/apimanual.md
+++ b/docs/src/apimanual.md
@@ -971,30 +971,24 @@ MOI.set(model, MyPackage.PrintLevel(), 0)
 ### Supported constrained variables and constraints
 
 The solver interface should only implement support for constrained variables
+(see [`add_constrained_variable`](@ref)/[`add_constrained_variable`](@ref))
 or constraints that directly map to a structure exploited by the solver
 algorithm. There is no need to add support for additional types, this is
 handled by the [Automatic reformulation](@ref). Furthermore, this allows
 [`supports_constraint`](@ref) to indicate which types are exploited by the
-solver and hence allow layers such as [`Bridges.LazyBridgeOptimizer`](@ref)
+solver and hence allows layers such as [`Bridges.LazyBridgeOptimizer`](@ref)
 to accurately select the most appropriate transformations.
 
 As [`add_constrained_variable`](@ref) (resp. [`add_constrained_variables`](@ref))
 falls back to [`add_variable`](@ref) (resp. [`add_variables`](@ref)) followed by
 [`add_constraint`](@ref), there is no need to implement this function
-if `model` supports creating free variables and supports transforming free
-variables into variables in `set`. However,
-
-* if `model` does not support creating free variables, it should only implement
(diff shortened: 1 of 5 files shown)


src/variables.jl
```diff
@@ -33,3 +33,84 @@ A [`AddVariableNotAllowed`](@ref) error is thrown if adding variables cannot be
...
+the indices of the variables created and the index of the constraint
+constraining the vector of variables to belong to `set`.
+
+By default, this function falls back to creating free variables with
+[`add_variables`](@ref) and then constraining it to belong to `set` with
+[`add_constraint`](@ref). Therefore, there is no need to implement this function
+if `model` supports creating free variables and supports transforming free
+variables into variables in `set`. However,
+
+* if `model` does not support creating variables, it should only implement
```
mlubin: if `model` does not support creating free variables

src/variables.jl
```diff
@@ -33,3 +33,84 @@ A [`AddVariableNotAllowed`](@ref) error is thrown if adding variables cannot be
...
+  `add_constrained_variables` and not [`add_variable`](@ref) nor
+  [`add_constraint`](@ref) for [`VectorOfVariables`](@ref)-in-`typeof(set)`.
+  In addition, it should implement `supports_constraint(::Optimizer,
+  ::Type{VectorOfVariables}, ::Type{Reals})` and return `false` so that free
+  variables are bridged, see [`supports_constraint`](@ref).
+* if `model` supports free variables but does not support transforming free
```
mlubin: Another unclear use of "supports transforming".

docs/src/apimanual.md
```diff
@@ -966,6 +968,34 @@ model = MyPackage.Optimizer()
 MOI.set(model, MyPackage.PrintLevel(), 0)
 ```
 
+### Supported constrained variables and constraints
+
+The solver interface should only implement support for constrained variables
```
mlubin: This should link to the definition of constrained variables.

docs/src/apimanual.md
```diff
@@ -966,6 +968,34 @@ model = MyPackage.Optimizer()
...
+or constraints that directly map to a structure exploited by the solver
+algorithm. There is no need to add support for additional types, this is
+handled by the [Automatic reformulation](@ref). Furthermore, this allows
+[`supports_constraint`](@ref) to indicate which types are exploited by the
+solver and hence allow layers such as [`Bridges.LazyBridgeOptimizer`](@ref)
```
mlubin: Nit: allows (the subject is presumably "this")

docs/src/apimanual.md
```diff
@@ -966,6 +968,34 @@ model = MyPackage.Optimizer()
...
+to accurately select the most appropriate transformations.
+
+As [`add_constrained_variable`](@ref) (resp. [`add_constrained_variables`](@ref))
+falls back to [`add_variable`](@ref) (resp. [`add_variables`](@ref)) followed by
+[`add_constraint`](@ref), there is no need to implement this function
+if `model` supports creating free variables and supports transforming free
```
mlubin: What does "transforming free variables" mean? This doesn't appear to refer to an MOI concept. I wouldn't say that adding a lower or upper bound on a variable involves a transformation of the variable.

docs/src/apimanual.md
```diff
@@ -966,6 +968,28 @@ model = MyPackage.Optimizer()
 MOI.set(model, MyPackage.PrintLevel(), 0)
 ```
 
+### Supported constrained variables and constraints
+
+The solver interface should only implement support for constrained variables
+(see [`add_constrained_variable`](@ref)/[`add_constrained_variable`](@ref))
```
mlubin: [`add_constrained_variable`] is repeated

docs/src/apireference.md
```diff
@@ -182,11 +182,25 @@ delete(::ModelLike, ::Index)
 
 ### Variables
 
-Functions for adding variables. For deleting, see index types section.
+*Free variables* are created with with [`add_variable`](@ref) or
+[`add_variables`](@ref) and *constrained variables* are created with
```
mlubin: The discussion of constrained variables should be consistent with the MOI [standard form problem](http://www.juliaopt.org/MathOptInterface.jl/stable/apimanual/#Standard-form-problem-1). To do this, you can say that a constrained variable is equivalent to a `SingleVariable` or `VectorOfVariables` constrant.

src/constraints.jl
```diff
@@ -8,10 +8,26 @@ is, `copy_to(model, src)` does not throw [`UnsupportedConstraint`](@ref) when
 `src` contains `F`-in-`S` constraints. If `F`-in-`S` constraints are only not
 supported in specific circumstances, e.g. `F`-in-`S` constraints cannot be
 combined with another type of constraint, it should still return `true`.
-"""
+
+    supports_constraint(model::ModelLike, ::Type{VectorOfVariables}, ::Type{Reals})::Bool
+
+Return a `Bool` indicating whether `model` supports free variables. That is,
+`copy_to(model, src)` does error when `src` contains variables that are not
```
mlubin: s/does error/does not error/

mlubin: The double negative "does not error when ... contains variables that are not ..." is pretty confusing. Could you provide a concrete example? E.g., if a solver requires all variables to be nonnegative.

src/variables.jl
```diff
@@ -33,3 +33,70 @@ A [`AddVariableNotAllowed`](@ref) error is thrown if adding variables cannot be
...
+    )::Tuple{MOI.VariableIndex,
+             MOI.ConstraintIndex{MOI.SingleVariable, typeof(set)}}
+
+Add to `model` a scalar variable constrained to belong to `set`, returning the
+index of the variable created and the index of the constraint constraining the
+variable to belong to `set`.
+
+By default, this function falls back to creating a free variable with
+[`add_variable`](@ref) and then constraining it to belong to `set` with
+[`add_constraint`](@ref).
```
mlubin: By the definition above, calling `add_constraint` turns a free variable into a constrained variable, right? Maybe add, ", which turns the free variable into a constrained variable."

docs/src/apireference.md
```diff
@@ -182,11 +182,26 @@ delete(::ModelLike, ::Index)
 
 ### Variables
 
-Functions for adding variables. For deleting, see index types section.
+*Free variables* are the variables created with [`add_variable`](@ref) or
+[`add_variables`](@ref) and never constrained while *constrained variables* are
+the variables either created with [`add_constrained_variable`](@ref) or
+[`add_constrained_variables`](@ref) or created with [`add_variable`](@ref) or
+[`add_variables`](@ref) and constrained after being created using
```
mlubin: For a constrained variable, is it valid to ask what set is constraining the variable? This is nonsymmetric because you can add multiple `SingleVariable` constraints on the same variable using `add_constraint`, but only one when using `add_constrained_variable`.

docs/src/apireference.md
```diff
@@ -182,11 +182,26 @@ delete(::ModelLike, ::Index)
...
+[`add_constraint`](@ref) with the [`SingleVariable`](@ref) or
+[`VectorOfVariables`](@ref). However, they need to be added with
+[`add_constrained_variable`](@ref) or [`add_constrained_variables`](@ref) to
+allow Variable bridges to be used.
+Note further that variables that are constrained after being created using
+[`add_constraint`](@ref) may be copied by [`copy_to`](@ref) with
+[`add_constrained_variable`](@ref) or [`add_constrained_variables`](@ref) by the
```
mlubin: If a variable is constrained to multiple sets, which one is used in the `copy_to` call with `add_constrained_variable`?

docs/src/apireference.md
blegat: No you cannot it may be a relevant next step. That would allow `copy_to` to copy variables with the same set they were added. Currently on bl/variable_bridges, if a variable is constrained twice then it's not determined by the order of creation but rather by the type of set

mlubin: What if the definition of "constrained variable" is changed to one that is created with `add_constrained_variable`? This would make it simpler in the future to implement this change.
Then the current `copy_to` behavior would be a bit quirky but not a big deal.

src/constraints.jl
```diff
@@ -12,10 +12,12 @@ combined with another type of constraint, it should still return `true`.
...
 
 Return a `Bool` indicating whether `model` supports free variables. That is,
-`copy_to(model, src)` does error when `src` contains variables that are not
+`copy_to(model, src)` does not error when `src` contains variables that are not
 constrained by any [`SingleVariable`](@ref) or [`VectorOfVariables`](@ref)
 constraint. By default, this method returns `true` so it should only be
-implemented if `model` does not support free variables.
+implemented if `model` does not support free variables. For instance, if a
+solver requires all variables to be nonnegative, it should implement this
+method and return `false` free variables cannot be copied to the solver.
```
mlubin: and return `false` *because* free variables cannot be copied to the solver

docs/src/apireference.md
mlubin: I don't consider this comment resolved.

blegat: Oops, didn't see your comment. Good idea, I have updated the PR

src/variables.jl
```diff
@@ -33,3 +33,72 @@ A [`AddVariableNotAllowed`](@ref) error is thrown if adding variables cannot be
...
+    )::Tuple{MOI.VariableIndex,
+             MOI.ConstraintIndex{MOI.SingleVariable, typeof(set)}}
+
+Add to `model` a scalar variable constrained to belong to `set`, returning the
+index of the variable created and the index of the constraint constraining the
+variable to belong to `set`.
+
+By default, this function falls back to creating a free variable with
+[`add_variable`](@ref) and then constraining it to belong to `set` with
+[`add_constraint`](@ref), which turns the free variable into a constrained
```
mlubin: Is this description still accurate given the new definition of constrained variables?

PR #797, opened 2019-07-19
Allow arbitrary number of arguments to submit

blegat:
It would have zero argument for `RejectSolution`, see https://github.com/JuliaOpt/MathOptInterface.jl/pull/782/, it would be weird to require the user to provide a `tuple()`

### DIFF
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -388,7 +388,8 @@ error is thrown if it supports the submittable `sub` but it cannot be submitted.
 function submit end
 function submit(model::ModelLike, sub::AbstractSubmittable, args...)
     if supports(model, sub)
-        throw(MethodError(submit, (model, sub, args...)))
+        throw(ArgumentError(
+            "Submitting $(typeof.(args)) for `$(typeof(sub))` is not valid."))
     else
         throw(UnsupportedSubmittable(sub))
     end
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -377,9 +377,9 @@ end
 
 """
     submit(optimizer::AbstractOptimizer, sub::AbstractSubmittable,
-           value)::Nothing
+           values...)::Nothing
 
-Submit `value` to the submittable `sub` of the optimizer `optimizer`.
+Submit `values` to the submittable `sub` of the optimizer `optimizer`.
 
 An [`UnsupportedSubmittable`](@ref) error is thrown if `model` does not support
 the attribute `attr` (see [`supports`](@ref)) and a [`SubmitNotAllowed`](@ref)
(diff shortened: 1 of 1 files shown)


mlubin:
Wait, hold on. Aren't there `MethodError` concerns with this? How can JuMP expose this function and know that the caller is providing the correct number of arguments?

blegat:
They will get an error like it's not supported:
https://github.com/JuliaOpt/MathOptInterface.jl/blob/e458dfd55b581d2c0470a69095ab0001df823d82/src/attributes.jl#L389-L395
We have the same issue for get:
https://github.com/JuliaOpt/MathOptInterface.jl/blob/e458dfd55b581d2c0470a69095ab0001df823d82/src/attributes.jl#L248-L250
But not for set, which throws a `MethodError` (but still violating the `MethodError` principle since it's not the same one but avoiding this causes ambiguity as mentioned in the comment):
https://github.com/JuliaOpt/MathOptInterface.jl/blob/e458dfd55b581d2c0470a69095ab0001df823d82/src/attributes.jl#L348-L376

mlubin:
> They will get an error like it's not supported:

If you try to submit using a submittable that's supported, then you'll hit `throw(SubmitNotAllowed(sub))`. This seems like the wrong error message, because submit may be allowed, you just submitted with the wrong number of arguments.

I don't see an easy workaround if we're following the `MethodError` principle. 

> We have the same issue for get

Is this an issue with the fallback or with the definition of `get`? Why do we need `get` to allow an arbitrary number of arguments?

blegat:
> Is this an issue with the fallback or with the definition of get? Why do we need get to allow an arbitrary number of arguments?

Not, just with the fallback.

blegat:
In fact, it does not make sense to throw `SubmitNotAllowed` in the fallback since for submittable you support iff you implement `submit`.
For `set`, it's different since you may support an attribute but only implement `allocate`/`load`.

mlubin:
> Not, just with the fallback.

What is preventing us from defining the fallback without using varargs?

blegat:
> What is preventing us from defining the fallback without using varargs?

Nothing, let's open an issue https://github.com/JuliaOpt/MathOptInterface.jl/issues/799

src/attributes.jl
```diff
@@ -388,7 +388,7 @@ error is thrown if it supports the submittable `sub` but it cannot be submitted.
 function submit end
 function submit(model::ModelLike, sub::AbstractSubmittable, args...)
     if supports(model, sub)
-        throw(SubmitNotAllowed(sub))
+        throw(MethodError(submit, (model, sub, args...)))
```
mlubin: This change doesn't address the concern that a JuMP user will get an MOI MethodError if they pass the wrong number of arguments (and there's no obvious way for JuMP to fix this).

mlubin:
LGTM. Merge when ready.

PR #798, opened 2019-07-19
Fix broadcast for scalar sets

blegat:
Defining scalar sets as 0-dimensional for broadcast seems reasonable.

cc @tweisser

### DIFF
--- a/test/sets.jl
+++ b/test/sets.jl
@@ -1,3 +1,10 @@
+using Test
+
+using MathOptInterface
+const MOI = MathOptInterface
+
+include("dummy.jl")
+
 """
     MutLessThan{T<:Real} <: MOI.AbstractScalarSet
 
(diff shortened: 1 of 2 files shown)


test/sets.jl
```diff
@@ -32,7 +39,17 @@ Base.copy(mlt::MutLessThan) = MutLessThan(Base.copy(mlt.upper))
...
+        end
+    end
+    @testset "Broadcast" begin
+        model = DummyModelWithAdd()
+        x = MOI.add_variables(model, 3)
+        cis = MOI.add_constraint.(model, x, MOI.EqualTo(0.0))
+        @test length(cis) == 3
+        for i in 1:3
+            @test cis[i] == MOI.ConstraintIndex{MOI.SingleVariable,
+                                                MOI.EqualTo{Float64}}(0)
```
mlubin: Should this be `i` instead of `0`? Is the test passing as is?

blegat: It's with the `DummyModel`, not `MOIU.Model`

mlubin: That's confusing, probably shouldn't be part of the test.

mlubin: Especially since MOI defines what the constraint index for a `SingleVariable` constraint should be.

PR #801, opened 2019-07-22
Add normalize_constant

blegat:

### DIFF
--- a/docs/src/apireference.md
+++ b/docs/src/apireference.md
@@ -677,11 +677,14 @@ Utilities.operate!
 Utilities.vectorize
 ```
 
-## Set utilities
+## Constraint utilities
 
-The following utilities are available for sets:
+The following utilities are available for moving the function constant to the
+set for scalar constraints:
 ```@docs
 Utilities.shift_constant
+Utilities.add_scalar_constraint
+Utilities.normalize_constant
 ```
 
 ## Benchmarks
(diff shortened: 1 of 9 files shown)


src/Utilities/constraints.jl
```diff
@@ -10,20 +10,26 @@ the set in `model`. If `allow_modify_function` is `true` then the function
...
-                                           MOI.ScalarQuadraticFunction{T}},
+                               func::MOI.AbstractScalarFunction,
                                set::MOI.AbstractScalarSet;
                                allow_modify_function::Bool=false) where T
+    return MOI.add_constraint(
+        model, normalize_constant(
+            func, set; allow_modify_function=allow_modify_function)...)
+end
+
+function normalize_constant(func::MOI.AbstractFunction, set::MOI.AbstractSet;
```
mlubin: This name doesn't make it clear what's being normalized to what. Maybe `move_function_constant_to_set`?

blegat: It's only moved in the scalar case. It's also consistent with: https://github.com/JuliaOpt/JuMP.jl/pull/1960
Maybe a docstring would make it clear ?

docs/src/apireference.md
```diff
@@ -677,11 +677,14 @@ Utilities.operate!
...
 
-## Set utilities
+## Constraint utilities
 
-The following utilities are available for sets:
+The following utilities are available for moving the function constant to the
+set for scalar constraints:
 ```@docs
 Utilities.shift_constant
+Utilities.add_scalar_constraint
```
mlubin: Sorry for the bikeshedding, but could `add_scalar_constraint` be named `normalize_and_add_constraint`?  The name `add_scalar_constraint` doesn't suggest that any normalization is happening; it sounds like a specialization of `add_constraint`.

PR #804, opened 2019-07-23
Dual sets

guilhermebodin:
Moved functions from https://github.com/guilhermebodin/Dualization.jl to MOI


src/sets.jl
```diff
@@ -30,6 +30,28 @@ julia> dimension(PositiveSemidefiniteConeTriangle(2))
...
 function dimension end
 
+"""
+    dual_set(s::AbstractSet)
+
+Return the dual set of `s`, if there is no dual set returns an error.
+
+### Examples
+
+```julia-repl
```
blegat: Could this be a `jldoctest` ?

test/Utilities/sets.jl
```diff
@@ -23,3 +23,65 @@ end
...
+    pow03 = MOI.PowerCone(0.3)
+    pow04 = MOI.PowerCone(0.4)
+    dual_pow03 = MOI.DualPowerCone(0.3)
+    @test MOI.dual_set(pow03) == dual_pow03
+    @test MOI.dual_set(pow03) != pow03
+    @test MOI.dual_set(dual_pow03) == pow03
+    @test MOI.dual_set(dual_pow03) != pow04
+    @test MOI.dual_set(dual_pow03) != dual_pow03
+    # Not implemented Set
+    @test_throws ErrorException MOI.dual_set(MOI.PositiveSemidefiniteConeSquare(4))
```
blegat: You can be more specific with `err = ErrorException("...")` and `@test_throws err ...`

src/sets.jl
guilhermebodin: Maybe like this
```jldoctest
julia> MOI.dual_set(Reals(4))
MOI.Zeros(4)

julia> MOI.dual_set(SecondOrderCone(5))
MOI.SecondOrderCone(5)

julia> MOI.dual_set(ExponentialCone())
MOI.DualExponentialCone()
```

blegat: Yes :)

src/sets.jl
```diff
@@ -30,6 +30,28 @@ julia> dimension(PositiveSemidefiniteConeTriangle(2))
 """
 function dimension end
 
+"""
+    dual_set(s::AbstractSet)
+
+Return the dual set of `s`, if there is no dual set returns an error.
```
mlubin: Please provide or reference a definition of a dual set.

test/Utilities/sets.jl
```diff
@@ -23,3 +23,66 @@ end
...
+    pow03 = MOI.PowerCone(0.3)
+    pow04 = MOI.PowerCone(0.4)
+    dual_pow03 = MOI.DualPowerCone(0.3)
+    @test MOI.dual_set(pow03) == dual_pow03
+    @test MOI.dual_set(pow03) != pow03
+    @test MOI.dual_set(dual_pow03) == pow03
+    @test MOI.dual_set(dual_pow03) != pow04
+    @test MOI.dual_set(dual_pow03) != dual_pow03
+    # Not implemented Set
+    err = ErrorException("Dual of $(MOI.PositiveSemidefiniteConeSquare(4)) is not implemented.")
```
mlubin: Please add a note about why this dual is not defined. It's a very subtle point.

blegat: You can reference the comment in `src/Bridges/Constraint/square.jl`

guilhermebodin: Do you think I should create some specific dual_set functions that provide custom error messages?
This should also be useful for GreaterThan and LessThan

src/sets.jl
guilhermebodin: In MOI duality is defined in terms of conic duality so the dual set is the dual cone. Is this a good reference https://en.wikipedia.org/wiki/Dual_cone_and_polar_cone to put in documentation?

src/sets.jl
```diff
@@ -30,6 +30,29 @@ julia> dimension(PositiveSemidefiniteConeTriangle(2))
 """
 function dimension end
 
+"""
+    dual_set(s::AbstractSet)
+
+Return the dual set of `s`, in terms of the dual cone of the set.
```
mlubin: "in terms of", should be "that is,"

src/sets.jl
```diff
@@ -30,6 +30,29 @@ julia> dimension(PositiveSemidefiniteConeTriangle(2))
 """
 function dimension end
 
+"""
+    dual_set(s::AbstractSet)
+
+Return the dual set of `s`, in terms of the dual cone of the set. 
+See https://en.wikipedia.org/wiki/Dual_cone_and_polar_cone for more information.
+
```
mlubin: The discussion about what the function returns if the dual set is not defined got lost somehow.

src/sets.jl
```diff
@@ -30,6 +30,29 @@ julia> dimension(PositiveSemidefiniteConeTriangle(2))
 """
 function dimension end
 
+"""
+    dual_set(s::AbstractSet)
+
+Return the dual set of `s`, in terms of the dual cone of the set. 
+See https://en.wikipedia.org/wiki/Dual_cone_and_polar_cone for more information.
```
mlubin: Also refer to the discussion of conic duality in the MOI manual.

src/sets.jl
```diff
@@ -411,6 +456,8 @@ struct PositiveSemidefiniteConeSquare <: AbstractSymmetricMatrixSetSquare
     side_dimension::Int
 end
 
+dual_set(s::PositiveSemidefiniteConeSquare) = error("Dual of $s is not defined, $s can be a non-covex set.")
```
mlubin: How can `PositiveSemidefiniteConeSquare` be a non-convex set?

guilhermebodin: I misunderstood it, I thought it could be non-symmetric but in fact, in MOI documentation it explains that the matrix is considered to be symmetric and positive semidefinite.

src/sets.jl
```diff
@@ -30,6 +30,31 @@ julia> dimension(PositiveSemidefiniteConeTriangle(2))
 """
 function dimension end
 
+"""
+    dual_set(s::AbstractSet)
+
+Return the dual set of `s`, that is the dual cone of the set. This follows the 
+definition of `MathOptInterface` discussed in http://www.juliaopt.org/MathOptInterface.jl/stable/apimanual/#Advanced-1
```
mlubin: For version consistency, please reference this section using an internal link instead of a URL. See https://juliadocs.github.io/Documenter.jl/stable/man/syntax/#@ref-link-1. It might take a couple tries to work as I'm not sure of the exact syntax. You can double check that the link works by building the documentation locally.

src/sets.jl
```diff
@@ -411,6 +458,11 @@ struct PositiveSemidefiniteConeSquare <: AbstractSymmetricMatrixSetSquare
     side_dimension::Int
 end
 
+function dual_set(s::PositiveSemidefiniteConeSquare)
+    return error("""Dual of $s is not defined because $s is not a proper cone.
```
mlubin: I don't think this is the correct explanation. The discussion at square.jl says that "(P₊)* = D₊" where " D₊ [is] the cone of matrices A such that A+Aᵀ ∈ P₊" There's a well-defined dual of this cone, but it's not defined as an MOI set.

src/sets.jl
```diff
@@ -411,6 +458,11 @@ struct PositiveSemidefiniteConeSquare <: AbstractSymmetricMatrixSetSquare
     side_dimension::Int
 end
 
+function dual_set(s::PositiveSemidefiniteConeSquare)
+    return error("""Dual of $s is not defined because $s is not a proper cone. 
+                    For more details please see comments on https://github.com/JuliaOpt/MathOptInterface.jl/blob/master/src/Bridges/Constraint/square.jl""")
```
mlubin: This is an awkward link to include in an error message. I would just say, "see the comments in src/Bridges/Constraint/square.jl".

src/sets.jl
guilhermebodin: image

blegat:
@guilhermebodin Could you make a PR to update the `NEWS.md` ? (use [ci skip]  in your commit to avoid unnecessary travis checks)

PR #807, opened 2019-07-31
Add array utilities

blegat:


src/Utilities/lazy_iterators.jl
```diff
@@ -0,0 +1,65 @@
...
+    if elem_state === nothing
+        return nothing
+    else
+        return it.f(elem_state[1]), elem_state[2]
+    end
+end
+Base.IteratorSize(it::LazyMap) = Base.IteratorSize(it.data)
+Base.eltype(::LazyMap{T}) where {T} = T
+
+struct LazyFilter{T, VT}
```
mlubin: Is this the same as `Base.Iterators.Filter`?

src/Utilities/lazy_iterators.jl
```diff
@@ -0,0 +1,65 @@
+struct EmptyVector{T} <: AbstractVector{T} end
+Base.size(::EmptyVector) = (0,)
+Base.isempty(::EmptyVector) = true
+Base.eltype(::EmptyVector{T}) where {T} = T
+Base.iterate(::EmptyVector) = nothing
+
+struct LazyMap{T, VT}
```
mlubin: This could be replaced by `Base.Generator(f, data)`. See https://discourse.julialang.org/t/why-no-base-iterators-map/14280/7.

src/Utilities/lazy_iterators.jl
```diff
@@ -0,0 +1,65 @@
...
+    if elem_state === nothing
+        return nothing
+    else
+        return elem_state
+    end
+end
+Base.IteratorSize(::LazyFilter) = Base.SizeUnknown()
+Base.eltype(::LazyFilter{T}) where {T} = T
+
+struct LazyCat{VT}
```
mlubin: `Base.Iterators.flatten`?

src/Utilities/lazy_iterators.jl
blegat: The difference is that `eltype(Base.Generator(x -> x^2, 1:2))` while for `LazyMap`, you need to set the eltype

mlubin: Got it. Please add a comment explaining.

blegat: In fact, this is a simplified version of https://github.com/JuliaArrays/MappedArrays.jl

src/Utilities/lazy_iterators.jl
blegat: Good point, `LazyCat` is useless

src/Utilities/lazy_iterators.jl
blegat: Indeed, `filter` was not lazy but I didn't see `Base.Iterators.Filter`

PR #808, opened 2019-08-01
Implement TimeLimit

remi-garcia:
Implement TimeLimit using the example of Silent #695 

~~Note: for the tests I used Integers TimeLimit because I didn't know the right way to `@test` a float value~~

### DIFF
--- a/docs/src/apimanual.md
+++ b/docs/src/apimanual.md
@@ -1132,6 +1132,7 @@ const MOIB = MOI.Bridges
 import FooBar
 const optimizer = FooBar.Optimizer()
 MOI.set(optimizer, MOI.Silent(), true)
+MOI.set(optimizer, MOI.TimeLimit(), 0)
 
 @testset "SolverName" begin
     @test MOI.get(optimizer, MOI.SolverName()) == "FooBar"
--- a/docs/src/apimanual.md
+++ b/docs/src/apimanual.md
@@ -1132,7 +1132,6 @@ const MOIB = MOI.Bridges
 import FooBar
 const optimizer = FooBar.Optimizer()
 MOI.set(optimizer, MOI.Silent(), true)
-MOI.set(optimizer, MOI.TimeLimit(), 0)
 
 @testset "SolverName" begin
     @test MOI.get(optimizer, MOI.SolverName()) == "FooBar"
--- a/src/Test/UnitTests/attributes.jl
+++ b/src/Test/UnitTests/attributes.jl
@@ -32,25 +32,25 @@ end
 unittests["silent"] = silent
 
 """
-    timelimit(model::MOI.ModelLike, config::TestConfig)
+    timelimitsec(model::MOI.ModelLike, config::TestConfig)
 
-Test that the [`MOI.TimeLimit`](@ref) attribute is implemented for `model`.
+Test that the [`MOI.TimeLimitSec`](@ref) attribute is implemented for `model`.
 """
-function timelimit(model::MOI.ModelLike, config::TestConfig)
+function timelimitsec(model::MOI.ModelLike, config::TestConfig)
     if config.solve
-        @test MOI.supports(model, MOI.TimeLimit())
+        @test MOI.supports(model, MOI.TimeLimitSec())
         # Get the current value to restore it at the end of the test
-        value = MOI.get(model, MOI.TimeLimit())
-        MOI.set(model, MOI.TimeLimit(), value + 1)
-        @test value + 1 == MOI.get(model, MOI.TimeLimit())
-        # Check that `set` does not just increment the current value
-        MOI.set(model, MOI.TimeLimit(), value + 1)
-        @test value + 1 == MOI.get(model, MOI.TimeLimit())
-        MOI.set(model, MOI.TimeLimit(), value)
-        @test value == MOI.get(model, MOI.TimeLimit())
+        value = MOI.get(model, MOI.TimeLimitSec())
+        MOI.set(model, MOI.TimeLimitSec(), 0.0)
+        @test isapprox(MOI.get(model, MOI.TimeLimitSec()), 0.0, atol = eps(0.0))
+        # Check that `set` does not just set TimeLimitSec to `0.0`
+        MOI.set(model, MOI.TimeLimitSec(), 1.0)
+        @test isapprox(MOI.get(model, MOI.TimeLimitSec()), 1.0, atol = eps(0.0))
+        MOI.set(model, MOI.TimeLimitSec(), value)
+        @test value == MOI.get(model, MOI.TimeLimitSec()) # Equality should hold
     end
 end
-unittests["timelimit"] = timelimit
+unittests["timelimitsec"] = timelimitsec
 
 """
     raw_status_string(model::MOI.ModelLike, config::TestConfig)
--- a/src/Test/UnitTests/attributes.jl
+++ b/src/Test/UnitTests/attributes.jl
@@ -32,7 +32,7 @@ end
 unittests["silent"] = silent
 
 """
-    timelimitsec(model::MOI.ModelLike, config::TestConfig)
+    time_limit_sec(model::MOI.ModelLike, config::TestConfig)
 
 Test that the [`MOI.TimeLimitSec`](@ref) attribute is implemented for `model`.
 """
--- a/src/Test/UnitTests/attributes.jl
+++ b/src/Test/UnitTests/attributes.jl
@@ -36,21 +36,20 @@ unittests["silent"] = silent
 
 Test that the [`MOI.TimeLimitSec`](@ref) attribute is implemented for `model`.
 """
-function timelimitsec(model::MOI.ModelLike, config::TestConfig)
+function time_limit_sec(model::MOI.ModelLike, config::TestConfig)
     if config.solve
         @test MOI.supports(model, MOI.TimeLimitSec())
         # Get the current value to restore it at the end of the test
         value = MOI.get(model, MOI.TimeLimitSec())
         MOI.set(model, MOI.TimeLimitSec(), 0.0)
-        @test isapprox(MOI.get(model, MOI.TimeLimitSec()), 0.0, atol = eps(0.0))
-        # Check that `set` does not just set TimeLimitSec to `0.0`
+        @test MOI.get(model, MOI.TimeLimitSec()) == 0.0
         MOI.set(model, MOI.TimeLimitSec(), 1.0)
-        @test isapprox(MOI.get(model, MOI.TimeLimitSec()), 1.0, atol = eps(0.0))
+        @test MOI.get(model, MOI.TimeLimitSec()) == 1.0
         MOI.set(model, MOI.TimeLimitSec(), value)
         @test value == MOI.get(model, MOI.TimeLimitSec()) # Equality should hold
     end
 end
-unittests["timelimitsec"] = timelimitsec
+unittests["time_limit_sec"] = time_limit_sec
 
 """
     raw_status_string(model::MOI.ModelLike, config::TestConfig)
--- a/src/Test/UnitTests/attributes.jl
+++ b/src/Test/UnitTests/attributes.jl
@@ -32,6 +32,27 @@ end
 unittests["silent"] = silent
 
 """
+    timelimit(model::MOI.ModelLike, config::TestConfig)
+
+Test that the [`MOI.TimeLimit`](@ref) attribute is implemented for `model`.
+"""
+function timelimit(model::MOI.ModelLike, config::TestConfig)
+    if config.solve
+        @test MOI.supports(model, MOI.TimeLimit())
+        # Get the current value to restore it at the end of the test
+        value = MOI.get(model, MOI.TimeLimit())
+        MOI.set(model, MOI.TimeLimit(), value + 1)
+        @test value + 1 == MOI.get(model, MOI.TimeLimit())
+        # Check that `set` does not just increment the current value
+        MOI.set(model, MOI.TimeLimit(), value + 1)
+        @test value + 1 == MOI.get(model, MOI.TimeLimit())
+        MOI.set(model, MOI.TimeLimit(), value)
+        @test value == MOI.get(model, MOI.TimeLimit())
+    end
+end
+unittests["timelimit"] = timelimit
+
+"""
     raw_status_string(model::MOI.ModelLike, config::TestConfig)
 
 Test that the [`MOI.RawStatusString`](@ref) attribute is implemented for
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -446,8 +446,7 @@ struct Silent <: AbstractOptimizerAttribute end
 An optimizer attribute for setting a time limit for an optimization. When `set`
 to `nothing`, it deactivates the solver time limit. The default value is
 `nothing`. The time limit is in seconds.
-""" # TODO add a test checking if the solver returns TIME_LIMIT status when
-    # the time limit is hit
+""" # TODO add a test checking if the solver returns TIME_LIMIT status when the time limit is hit
 struct TimeLimitSec <: AbstractOptimizerAttribute end
 
 """
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -420,7 +420,7 @@ An optimizer attribute for the string identifying the solver/optimizer.
 struct SolverName <: AbstractOptimizerAttribute end
 
 """
-    Silent
+    Silent()
 
 An optimizer attribute for silencing the output of an optimizer. When `set`
 to `true`, it takes precedence over any other attribute controlling verbosity
@@ -441,11 +441,11 @@ given.
 struct Silent <: AbstractOptimizerAttribute end
 
 """
-    TimeLimitSec
+    TimeLimitSec()
 
-An optimizer attribute for setting a time limit to an optimization. When `set`
-to `nothing`, it deactivate the solver time limit. The default value is
-`nothing`. The time limit is `set` in seconds.
+An optimizer attribute for setting a time limit for an optimization. When `set`
+to `nothing`, it deactivates the solver time limit. The default value is
+`nothing`. The time limit is in seconds.
 """
 struct TimeLimitSec <: AbstractOptimizerAttribute end
 
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -446,7 +446,8 @@ struct Silent <: AbstractOptimizerAttribute end
 An optimizer attribute for setting a time limit for an optimization. When `set`
 to `nothing`, it deactivates the solver time limit. The default value is
 `nothing`. The time limit is in seconds.
-"""
+""" # TODO add a test checking if the solver returns TIME_LIMIT status when
+    # the time limit is hit
 struct TimeLimitSec <: AbstractOptimizerAttribute end
 
 """
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -446,8 +446,7 @@ struct Silent <: AbstractOptimizerAttribute end
 An optimizer attribute for setting a time limit for an optimization. When `set`
 to `nothing`, it deactivates the solver time limit. The default value is
 `nothing`. The time limit is in seconds.
-""" # TODO add a test checking if the solver returns TIME_LIMIT status when
-    # the time limit is hit
+"""
 struct TimeLimitSec <: AbstractOptimizerAttribute end
 
 """
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -441,6 +441,15 @@ given.
 struct Silent <: AbstractOptimizerAttribute end
 
 """
+    TimeLimit
+
+An optimizer attribute for setting a time limit to an optimization. When `set`
+to `0`, it deactivate the solver time limit. The default value is `0`. The time
+limit is `set` in seconds.
+"""
+struct TimeLimit <: AbstractOptimizerAttribute end
+
+"""
     RawParameter(name)
 
 An optimizer attribute for the solver-specific parameter identified by `name`
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -441,13 +441,13 @@ given.
 struct Silent <: AbstractOptimizerAttribute end
 
 """
-    TimeLimit
+    TimeLimitSec
 
 An optimizer attribute for setting a time limit to an optimization. When `set`
-to `0`, it deactivate the solver time limit. The default value is `0`. The time
-limit is `set` in seconds.
+to `nothing`, it deactivate the solver time limit. The default value is
+`nothing`. The time limit is `set` in seconds.
 """
-struct TimeLimit <: AbstractOptimizerAttribute end
+struct TimeLimitSec <: AbstractOptimizerAttribute end
 
 """
     RawParameter(name)
--- a/test/Test/unit.jl
+++ b/test/Test/unit.jl
@@ -14,7 +14,6 @@ end
     # `UniversalFallback` needed for `MOI.Silent`
     mock = MOIU.MockOptimizer(MOIU.UniversalFallback(MOIU.Model{Float64}()))
     MOI.set(mock, MOI.Silent(), true)
-    MOI.set(mock, MOI.TimeLimitSec(), nothing)
     config = MOIT.TestConfig()
     for model in [mock,
                   MOIU.CachingOptimizer(MOIU.UniversalFallback(
--- a/test/Test/unit.jl
+++ b/test/Test/unit.jl
@@ -14,6 +14,7 @@ end
     # `UniversalFallback` needed for `MOI.Silent`
     mock = MOIU.MockOptimizer(MOIU.UniversalFallback(MOIU.Model{Float64}()))
     MOI.set(mock, MOI.Silent(), true)
+    MOI.set(mock, MOI.TimeLimit(), 0)
     config = MOIT.TestConfig()
     for model in [mock,
                   MOIU.CachingOptimizer(MOIU.UniversalFallback(
--- a/test/Test/unit.jl
+++ b/test/Test/unit.jl
@@ -14,6 +14,9 @@ end
     # `UniversalFallback` needed for `MOI.Silent`
     mock = MOIU.MockOptimizer(MOIU.UniversalFallback(MOIU.Model{Float64}()))
     MOI.set(mock, MOI.Silent(), true)
+    # TimeLimitSec() has to be set to default value since the mock optimizer
+    # doesn't handle this
+    MOI.set(mock, MOI.TimeLimitSec(), nothing)
     config = MOIT.TestConfig()
     for model in [mock,
                   MOIU.CachingOptimizer(MOIU.UniversalFallback(
--- a/test/Test/unit.jl
+++ b/test/Test/unit.jl
@@ -14,6 +14,7 @@ end
     # `UniversalFallback` needed for `MOI.Silent`
     mock = MOIU.MockOptimizer(MOIU.UniversalFallback(MOIU.Model{Float64}()))
     MOI.set(mock, MOI.Silent(), true)
+    MOI.set(mock, MOI.TimeLimitSec(), nothing)
     config = MOIT.TestConfig()
     for model in [mock,
                   MOIU.CachingOptimizer(MOIU.UniversalFallback(
--- a/test/Test/unit.jl
+++ b/test/Test/unit.jl
@@ -14,7 +14,7 @@ end
     # `UniversalFallback` needed for `MOI.Silent`
     mock = MOIU.MockOptimizer(MOIU.UniversalFallback(MOIU.Model{Float64}()))
     MOI.set(mock, MOI.Silent(), true)
-    MOI.set(mock, MOI.TimeLimit(), 0)
+    MOI.set(mock, MOI.TimeLimitSec(), nothing)
     config = MOIT.TestConfig()
     for model in [mock,
                   MOIU.CachingOptimizer(MOIU.UniversalFallback(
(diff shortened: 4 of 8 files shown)


docs/src/apimanual.md
```diff
@@ -1132,6 +1132,7 @@ const MOIB = MOI.Bridges
 import FooBar
 const optimizer = FooBar.Optimizer()
 MOI.set(optimizer, MOI.Silent(), true)
+MOI.set(optimizer, MOI.TimeLimit(), 0)
```
blegat: Adding this line here means that all the tests will be executed with a `TimeLimit` of 0 hence they will all fail

src/Test/UnitTests/attributes.jl
```diff
@@ -31,6 +31,27 @@ function silent(model::MOI.ModelLike, config::TestConfig)
...
+
+Test that the [`MOI.TimeLimit`](@ref) attribute is implemented for `model`.
+"""
+function timelimit(model::MOI.ModelLike, config::TestConfig)
+    if config.solve
+        @test MOI.supports(model, MOI.TimeLimit())
+        # Get the current value to restore it at the end of the test
+        value = MOI.get(model, MOI.TimeLimit())
+        MOI.set(model, MOI.TimeLimit(), value + 1)
+        @test value + 1 == MOI.get(model, MOI.TimeLimit())
```
blegat: Use `isapprox` instead of `==`. The solver might store the time in a `Float32` for instance and won't be able to return exactly the same value

docs/src/apimanual.md
blegat: Just read the doc. Disabling the time limit for the tests may be annoying as you might have your tests looping infinitely

test/Test/unit.jl
```diff
@@ -14,6 +14,7 @@ end
     # `UniversalFallback` needed for `MOI.Silent`
     mock = MOIU.MockOptimizer(MOIU.UniversalFallback(MOIU.Model{Float64}()))
     MOI.set(mock, MOI.Silent(), true)
+    MOI.set(mock, MOI.TimeLimit(), 0)
```
blegat: `0` should be `0.0`

test/Utilities/cachingoptimizer.jl
```diff
@@ -35,6 +35,13 @@ const MOIU = MOI.Utilities
         " is attached.")
     @test_throws exception MOI.get(model, attr)
 
+    @test_throws exception MOI.get(model, attr)
```
blegat: This is a duplicate of the previous line

test/Utilities/cachingoptimizer.jl
```diff
@@ -46,24 +53,37 @@ end
     cache = MOIU.UniversalFallback(MOIU.Model{Float64}())
     cached = MOIU.CachingOptimizer(cache, MOIU.MANUAL)
     MOI.set(cached, MOI.Silent(), true)
+    MOI.set(cached, MOI.TimeLimit(), 0)
```
blegat: `0` should be `0.0`

src/Test/UnitTests/attributes.jl
remi-garcia: I wasn't sure because I had issues with isapprox(. , 0). I'm changing that.

remi-garcia:
I'll set it to `nothing`, I don't know why but I though 0 was the usual default value while in CPLEX it is 1e+75.

src/Test/UnitTests/attributes.jl
```diff
@@ -31,6 +31,27 @@ function silent(model::MOI.ModelLike, config::TestConfig)
...
+    timelimit(model::MOI.ModelLike, config::TestConfig)
+
+Test that the [`MOI.TimeLimit`](@ref) attribute is implemented for `model`.
+"""
+function timelimit(model::MOI.ModelLike, config::TestConfig)
+    if config.solve
+        @test MOI.supports(model, MOI.TimeLimit())
+        # Get the current value to restore it at the end of the test
+        value = MOI.get(model, MOI.TimeLimit())
+        MOI.set(model, MOI.TimeLimit(), value + 1)
```
mlubin: These tests with `value + 1` seem overly complex. Why not do:
```julia
value = MOI.get(model, MOI.TimeLimit())

MOI.set(model, MOI.TimeLimit(), 15.0)
@test 15.0 ≈ MOI.get(model, MOI.TimeLimit())

MOI.set(model, MOI.TimeLimit() value)
@test value == MOI.get(model, MOI.TimeLimit()) # Equality should hold.
```

src/attributes.jl
```diff
@@ -440,6 +440,15 @@ given.
 """
 struct Silent <: AbstractOptimizerAttribute end
 
+"""
+    TimeLimit
+
+An optimizer attribute for setting a time limit to an optimization. When `set`
+to `0`, it deactivate the solver time limit. The default value is `0`. The time
```
mlubin: +1 for `nothing` instead of `0` as a default.

src/attributes.jl
```diff
@@ -440,6 +440,15 @@ given.
 """
 struct Silent <: AbstractOptimizerAttribute end
 
+"""
+    TimeLimit
```
mlubin: Consider naming this `TimeLimitSec` for clarity.

src/Test/UnitTests/attributes.jl
blegat: You need to set `atol` to a nonzero value for `isapprox` to work when one of the argument is zero

blegat: See https://github.com/JuliaLang/julia/issues/23376

src/attributes.jl
blegat: Why not set the default to whatever default the solver has. Some solver have a default time limit

src/Test/UnitTests/attributes.jl
remi-garcia: I'm changing that part to check 2 values:
```Julia
MOI.set(model, MOI.TimeLimit(), 0.0)
@test isapprox(MOI.get(model, MOI.TimeLimit()), 0.0)
# Check that `set` does not just set TimeLimit to `0.0`
MOI.set(model, MOI.TimeLimit(), 15.0)
@test isapprox(MOI.get(model, MOI.TimeLimit()), 15.0)
```

Then for `value` I was thinking of:
```Julia
MOI.set(model, MOI.TimeLimit(), value)
if value !== nothing
    @test isapprox(value, MOI.get(model, MOI.TimeLimit()))
else
    @test value === MOI.get(model, MOI.TimeLimit())
end
```
but if equality should hold is enough I can do that instead.

mlubin: > equality should hold is enough I can do that instead

I believe it's safe to assume that equality will hold for a time limit that the solver already had set. There shouldn't be loss in the process.

src/attributes.jl
mlubin: No solvers should have a default finite time limit. They might represent "no limit" it as either a large number or as 0.0, but we should abstract over that.

src/Test/UnitTests/attributes.jl
remi-garcia: That what I did for me but in a test I have no clue how to choose the right `atol`

remi-garcia: I choosed `atol = eps(0.0)`. Testing locally now.

src/attributes.jl
blegat: Ok so we should say that if a solver has a default time limit, it should be disabled

remi-garcia:
I made the following changes:
`TimeLimit` -> `TimeLimitSec`
Default value: `0` -> `nothing`
Tests: `==` -> `isapprox`
Tests: `#` -> `#.0`

TODO:
> The tests do not check that the solver returns the `TIME_LIMIT` status when the time limit is hit. This is actually pretty challenging to do correctly because some solvers could solve a tiny problem very fast and return `OPTIMAL` before checking the time limit. However, it's worth a try or a TODO.

src/Test/UnitTests/attributes.jl
```diff
@@ -31,6 +31,27 @@ function silent(model::MOI.ModelLike, config::TestConfig)
 end
 unittests["silent"] = silent
 
+"""
+    timelimitsec(model::MOI.ModelLike, config::TestConfig)
```
mlubin: The [style guide](http://www.juliaopt.org/JuMP.jl/v0.19.2/style/#Naming-1) says this should be named `time_limit_sec`.

src/Test/UnitTests/attributes.jl
```diff
@@ -31,6 +31,27 @@ function silent(model::MOI.ModelLike, config::TestConfig)
...
+
+Test that the [`MOI.TimeLimitSec`](@ref) attribute is implemented for `model`.
+"""
+function timelimitsec(model::MOI.ModelLike, config::TestConfig)
+    if config.solve
+        @test MOI.supports(model, MOI.TimeLimitSec())
+        # Get the current value to restore it at the end of the test
+        value = MOI.get(model, MOI.TimeLimitSec())
+        MOI.set(model, MOI.TimeLimitSec(), 0.0)
+        @test isapprox(MOI.get(model, MOI.TimeLimitSec()), 0.0, atol = eps(0.0))
```
mlubin: If you set the time limit to `0.0`, `get` should return `0.0`. There's no need for `isapprox`.

src/Test/UnitTests/attributes.jl
```diff
@@ -31,6 +31,27 @@ function silent(model::MOI.ModelLike, config::TestConfig)
...
+        # Check that `set` does not just set TimeLimitSec to `0.0`
```
mlubin: This comment isn't needed. It's obvious from the code.

src/Test/UnitTests/attributes.jl
```diff
@@ -31,6 +31,27 @@ function silent(model::MOI.ModelLike, config::TestConfig)
...
+        MOI.set(model, MOI.TimeLimitSec(), 1.0)
+        @test isapprox(MOI.get(model, MOI.TimeLimitSec()), 1.0, atol = eps(0.0))
```
mlubin: `atol` isn't needed when comparing with 1.0. Just write `@test MOI.get(model, MOI.TimeLimitSec()) ≈ 1.0`.

src/attributes.jl
```diff
@@ -440,6 +440,15 @@ given.
 """
 struct Silent <: AbstractOptimizerAttribute end
 
+"""
+    TimeLimitSec
```
mlubin: `TimeLimitSec()` for consistency with the other docstrings.

src/attributes.jl
```diff
@@ -440,6 +440,15 @@ given.
...
+
+An optimizer attribute for setting a time limit to an optimization. When `set`
+to `nothing`, it deactivate the solver time limit. The default value is
```
mlubin: s/deactivate/deactivates/

mlubin: I'm confused about which solvers have a default finite time limit. Which are they?

src/attributes.jl
```diff
@@ -440,6 +440,15 @@ given.
...
+`nothing`. The time limit is `set` in seconds.
```
mlubin: "The time limit is in seconds."

src/attributes.jl
```diff
@@ -440,6 +440,15 @@ given.
 """
 struct Silent <: AbstractOptimizerAttribute end
 
+"""
+    TimeLimitSec
+
+An optimizer attribute for setting a time limit to an optimization. When `set`
```
mlubin: s/to/for/

test/Utilities/cachingoptimizer.jl
```diff
@@ -46,24 +50,37 @@ end
     cache = MOIU.UniversalFallback(MOIU.Model{Float64}())
     cached = MOIU.CachingOptimizer(cache, MOIU.MANUAL)
     MOI.set(cached, MOI.Silent(), true)
+    MOI.set(cached, MOI.TimeLimitSec(), 0.0)
     mock = MOIU.MockOptimizer(MOIU.UniversalFallback(MOIU.Model{Float64}()))
     MOIU.reset_optimizer(cached, mock)
     @test MOI.get(mock, MOI.Silent())
     @test MOI.get(cached, MOI.Silent())
+    @test isapprox(MOI.get(mock, MOI.TimeLimitSec()), 0.0, atol = eps(0.0))
```
mlubin: Use equality tests here. `0.0 == 0.0` is true.

src/attributes.jl
remi-garcia: Should I do the same change to `Silent` too?

src/Test/UnitTests/attributes.jl
remi-garcia: I'll make that change but I was wondering why JuMP style guide differ from [Julia style guide](https://docs.julialang.org/en/v1/manual/style-guide/index.html)?

src/attributes.jl
mlubin: Yes, thanks.

src/Test/UnitTests/attributes.jl
mlubin: The julia style guide has an ambiguous guideline for underscores: "functions are lowercase (maximum, convert) and, when readable, with multiple words squashed together (isequal, haskey)." According to this rule, I could argue that `timelimitsec` is not readable. Someone else could argue that it is. As the preamble to the JuMP style guide states, a goal is "reducing the amount of time spent on bike-shedding by establishing basic naming and formatting conventions". By using underscores to separate words *always*, we don't have to spend time debating this.

src/attributes.jl
remi-garcia: By "it deactivate the solver time limit" I meant it set it back to default. For CPLEX default value is 1e+75.

src/Test/UnitTests/attributes.jl
remi-garcia: Ok, thank you very much for taking the time to answer

remi-garcia:
I made the following changes:
- `timelimitsec()` -> `time_limit_sec()`
- `isapprox(., 0.0)` -> `. == 0.0`
- `isapprox(., 1.0)` -> `. ≈ 1.0`
- English corrections and small changes

Still TODO:
> The tests do not check that the solver returns the `TIME_LIMIT` status when the time limit is hit. This is actually pretty challenging to do correctly because some solvers could solve a tiny problem very fast and return `OPTIMAL` before checking the time limit. However, it's worth a try or a TODO.

test/Test/unit.jl
```diff
@@ -14,6 +14,7 @@ end
     # `UniversalFallback` needed for `MOI.Silent`
     mock = MOIU.MockOptimizer(MOIU.UniversalFallback(MOIU.Model{Float64}()))
     MOI.set(mock, MOI.Silent(), true)
+    MOI.set(mock, MOI.TimeLimitSec(), nothing)
```
mlubin: This is the default value, so why is the `set` needed?

remi-garcia: I think it should be remove. My bad.

remi-garcia:
I'll add a TODO in the code for now but I'm not sure where is the right place to add it since I didn't find any test for Termination status.

test/Test/unit.jl
remi-garcia: Well, removing that line breaks. `nothing` is the default value but I didn't set it in first place. I put that line back for now.
I don't know how to set a default value, and I think `Silent()` default value isn't set too. I'll be looking into it soon.

mlubin: What breaks? This needs to be resolved before we can merge.

src/Test/UnitTests/attributes.jl
```diff
@@ -31,6 +31,26 @@ function silent(model::MOI.ModelLike, config::TestConfig)
...
+"""
+    time_limit_sec(model::MOI.ModelLike, config::TestConfig)
+
+Test that the [`MOI.TimeLimitSec`](@ref) attribute is implemented for `model`.
+"""
+function time_limit_sec(model::MOI.ModelLike, config::TestConfig)
+    if config.solve
+        @test MOI.supports(model, MOI.TimeLimitSec())
+        # Get the current value to restore it at the end of the test
+        value = MOI.get(model, MOI.TimeLimitSec())
```
remi-garcia: This `get()` throws an error if `TimeLimitSec` is not `set` to `nothing` before. It's done in `test/Test/unit.jl` for now.

test/Test/unit.jl
remi-garcia: When `get()` is used on `TimeLimitSec()` instead of returning `nothing` as default value it throws an error. That happened in `src/Test/UnitTests/attributes.jl`:
`KeyError: key MathOptInterface.TimeLimitSec() not found`

remi-garcia:
```Julia
(MathOptInterface) pkg> test MathOptInterface
   Testing MathOptInterface
 Resolving package versions...
Test Summary: | Pass  Total
MOI           |  116    116
Test Summary: | Pass  Total
MOI.Utilities | 5220   5220
time_limit_sec: Error During Test at /.../MathOptInterface.jl/src/Test/config.jl:45
  KeyError: key MathOptInterface.TimeLimitSec() not found
  Stacktrace:
   ...
time_limit_sec: Error During Test at /.../MathOptInterface.jl/src/Test/config.jl:45
  KeyError: key MathOptInterface.TimeLimitSec() not found
  Stacktrace:
   ...
Test Summary:                             | Pass  Error  Total
MOI.Test                                  | 3197      2   3199
  Config                                  |    6             6
  Unit                                    | 1062      2   1064
    Basic Constraint Tests                |  582           582
    Unit Tests                            |  343      2    345
      max_sense                           |    3             3
      min_sense                           |    3             3
      get_objective_function              |    5             5
      silent                              |    4             4
      time_limit_sec                      |    1      1      2
      getvariable                         |    2             2
      feasibility_sense                   |    3             3
      solver_name                         |    1             1
      delete_variables                    |    4             4
      getconstraint                       |    7             7
      variablenames                       |    5             5
      add_variables                       |    3             3
      delete_variable                     |    4             4
      add_variable                        |    3             3
      max_sense                           |    3             3
      min_sense                           |    3             3
      get_objective_function              |    5             5
      silent                              |    4             4
      time_limit_sec                      |    1      1      2
      getvariable                         |    2             2
      feasibility_sense                   |    3             3
      solver_name                         |    1             1
      delete_variables                    |    4             4
      getconstraint                       |    7             7
      variablenames                       |    5             5
      add_variables                       |    3             3
      delete_variable                     |    4             4
      add_variable                        |    3             3
      solve_blank_obj                     |   18            18
      solve_constant_obj                  |   10            10
      solve_singlevariable_obj            |   10            10
      solve_with_lowerbound               |   13            13
      solve_with_upperbound               |   13            13
      solve_affine_lessthan               |    8             8
      solve_affine_greaterthan            |    8             8
      solve_affine_equalto                |    8             8
      solve_affine_interval               |    8             8
      solve_duplicate_terms_scalar_affine |    8             8
      solve_duplicate_terms_vector_affine |    8             8
      solve_qcp_edge_cases                |   14            14
      solve_qp_edge_cases                 |   21            21
      solve_duplicate_terms_obj           |   10            10
      solve_affine_deletion_edge_cases    |   24            24
      solve_integer_edge_cases            |   20            20
      solve_objbound_edge_cases           |   24            24
      raw_status_string                   |    6             6
      solve_time                          |    6             6
      solve_zero_one_with_bounds          |    9             9
      solve_unbounded_model               |    1             1
    modifications                         |  137           137
  Continuous Linear                       |  737           737
  Continuous Conic                        |  965           965
  Continuous Quadratic                    |  227           227
  Integer Linear                          |  180           180
  Integer Conic                           |   20            20
ERROR: LoadError: Some tests did not pass: 3197 passed, 0 failed, 2 errored, 0 broken.
in expression starting at /.../MathOptInterface.jl/test/runtests.jl:31
ERROR: Package MathOptInterface errored during testing
```

remi-garcia: Same error occurs if I remove `MOI.set(mock, MOI.Silent(), true)`

mlubin: @blegat, any ideas?

odow: It's just because the mock optimizer doesn't implement the default values, so they need to be set here. I think it's okay to set to `nothing` here for now.

remi-garcia: Then why I don't add Silent and TimeLimitSec here:
https://github.com/JuliaOpt/MathOptInterface.jl/blob/0923d1e3a5fc9fdabc9c322bee84f29abe3805c3/src/Utilities/mockoptimizer.jl#L89

That way I can set the default value to what they should be. Then we can add test checking for that values to be sure every solver implement them as default.

odow: Sure. That seems like the right thing to do.

blegat: Yes, we could set it directly in the constructor for `Silent` and `TimeLimitSec`. You should also update `is_empty` so that it does not check `isempty(mock.optimizer_attribute)`

remi-garcia: I'm working on it, I changed `empty!()` too. Still `set` and `get` aren't working. They change `mock.inner_model` because `Silent` and `TimeLimitSec` aren't set by optimize:
https://github.com/JuliaOpt/MathOptInterface.jl/blob/0923d1e3a5fc9fdabc9c322bee84f29abe3805c3/src/Utilities/mockoptimizer.jl#L197-L203

remi-garcia: In `function MockOptimizer()` should I leave `Dict{MOI.AbstractOptimizerAttribute, Any}(),` and modify `inner_model`?

remi-garcia: I did try to change `inner_model` when `MockOptimizer()` is called but since `Silent()` only works with `UniversalFallback` that fails.
I could modify the way `set` and `get` work so we can store `Silent()` and `TimeLimitSec()` in `mock` instead of `mock.inner_model`. I'll try that and if I can get anything working I'll push so you can review.

remi-garcia: I have a temporary workaround. I use a `try catch` in `set` and `get`.
The problem is that I didn't find a way to know if `inner_model` supports an `AbstractOptimizerAttribute`. If it does and I stores it in `optimizer_attributes` then `is_empty` fails:
```Julia
AssertionError: MOI.is_empty(optimizer)
  Stacktrace:
   [1] MathOptInterface.Utilities.CachingOptimizer(::MathOptInterface.Utilities.UniversalFallback{MathOptInterface.Utilities.Model{Float64}}, ::MathOptInterface.Utilities.MockOptimizer{MathOptInterface.Utilities.UniversalFallback{MathOptInterface.Utilities.Model{Float64}}}) at /.../MathOptInterface.jl/src/Utilities/cachingoptimizer.jl:61
...```
I tried to use `supports` but it's not defined since `Silent()` and `TimeLimitSec()` aren't copyable. Maybe I didn't understand what `supports` is used for.

Comment (blegat) [reply test/Test/unit.jl]: > I did try to change `inner_model` when `MockOptimizer()` is called but since `Silent()` only works with `UniversalFallback` that fails.

Why don't you use a `UniversalFallback` ?

Comment (remi-garcia) [reply test/Test/unit.jl]: I think I have no control on this. If there is a test that don't use `UniversalFallback` `inner_model` can't accept `Silent()` and `TimeLimitSec()`. I think I need to `set` those attributes with default values without knowing which `ModelLike` is given to `MockOptimizer`.

Comment (blegat) [reply test/Test/unit.jl]: In `src/Test/unit.jl`, you cannot assume what the `ModelLike` is but in `test/Test/unit.Jl` you can choose to use a `UniversalFallback` like here:
https://github.com/JuliaOpt/MathOptInterface.jl/blob/d3106f434293a2aae7c664a22a30a7bd4069111a/test/Test/unit.jl#L14-L15

Comment (remi-garcia) [reply test/Test/unit.jl]: I will change that then but `MockOptimizer` will always be using `UniversalFallback`. It feels that it shouldn't be that way.

Comment (remi-garcia) [reply test/Test/unit.jl]: https://github.com/JuliaOpt/MathOptInterface.jl/blob/d3106f434293a2aae7c664a22a30a7bd4069111a/test/Utilities/copy.jl#L30-L39
That breaks if I try to change the `::ModelLike` for UniversalFallback.

Comment (blegat) [reply test/Test/unit.jl]: You only need to change it in `test/Test/unit.jl`, don't modify anything in `src`

Comment (remi-garcia) [reply test/Test/unit.jl]: I still have to make changes in `src/Utilities/mockoptimizer.jl` to set default values. With that done if I don't change `set` and `get` this happen - `MOI.set(mock.inner_model, attr, value)` - on MockOptimizer call.
That's fine if the call is made with UniversalFallback so I'm replacing `MOIU.Model{Float64}()` with `MOIU.UniversalFallback(MOIU.Model{Float64}())` in `test/`.

When I did that in `test/Utilities/copy.jl` I got new errors:
```Julia
@testset "Allocate-Load" begin
    @test !MOIU.supports_allocate_load(DummyModel(), false)
    @test !MOIU.supports_allocate_load(DummyModel(), true)
    mock = MOIU.MockOptimizer(MOIU.UniversalFallback(MOIU.Model{Float64}()), needs_allocate_load=true)
    MOIT.failcopytestc(mock)
    MOIT.failcopytestia(mock)
    MOIT.failcopytestva(mock)
    MOIT.failcopytestca(mock)
    MOIT.copytest(mock, MOIU.UniversalFallback(MOIU.Model{Float64}()))
end
```
```Julia
Test Summary:        | Pass  Fail  Error  Total
MOI.Utilities        | 5182     8      1   5191
  Functions          |  388                 388
  Sets               |   63                  63
  Constraints        |    8                   8
  Model              |  948                 948
  Universal Fallback |  661                 661
  Parser             |   35                  35
  Mock Optimizer     |  203                 203
  Caching Optimizer  | 2756                2756
  Copy               |   58     8      1     67
    AUTOMATIC        |    2                   2
    Default          |   47                  47
    Allocate-Load    |    9     8      1     18
  CleverDicts        |   62                  62
```

blegat: > I still have to make changes in `src/Utilities/mockoptimizer.jl` to set default values

Indeed, this requires the internal model to always support `TimeLimitSec`. You can either not set the default (like it is currently done for `Silent`) or only set it in case the internal model supports it (you can check that with `supports`).
I would rather do the first option (not setting the default) as the other behavior is somewhat weird, setting the optimizer attributes of ModelLike passed to MockOptimizer is a rather suprising behavior

remi-garcia: I think that would be strange to put `default value is` in docs but not testing for it because it's not implemented in MOI side.
I had issue with the use of `supports` since it returns an error instead of false for `Silent()` and `TimeLimitSec()`:
https://github.com/JuliaOpt/MathOptInterface.jl/blob/ab2d0f05205a81c3848f9f1d870b75332b1bddd2/src/attributes.jl#L155-L169

To be honest I don't know how to solve that without some important changes as this bad `try catch` I did in last push.

blegat: > it's not implemented in MOI side

`MockOptimizer` is just a utility for testing so I don't think it is an issul if it does not have the default values for `Silent` and `TimeLimitSec`.
If new optimizer attributes are defined outside of MOI, `MockOptimizer` won't be able to have the right default values. It's case is a bit specific since it supports **any** optimizer attributes (when used with UniversalFallback), even some that may not defined in MOI. But it actually completely ignores their value.
That's not the case for solvers that only support specific attributes, take their value into account and should have the correct default values.

remi-garcia: My understanding is: if I remove default from `MockOptimizer` then MOI can't have a test to check if Optimizers implement defaults correctly. If I didn't understand that well then I don't see why `MockOptimizer` should have this feature but if I'm right then I believe this feature is quite important.

I'll push without the default values.

blegat: > if I remove default from `MockOptimizer` then MOI can't have a test to check if Optimizers implement defaults correctly.

You can set the attribute's value to its default before running the tests as it is done for `Silent` here:
https://github.com/JuliaOpt/MathOptInterface.jl/blob/42cdc3454cdb88345da0adb9cce2a57497c2fad4/test/Test/unit.jl#L15-L16
Of course solvers shouldn't do that but for the `MockOptimizer` it's ok to mock the test

odow:
Looking good @remi-garcia. Sorry for the back and forth on the default options.

remi-garcia:
I'm actually learning so much doing this so I really don't mind. Thank you for taking so much time helping me with this PR.

frapac:
@remi-garcia Thanks for implementing this! Just a quick question: does the time limit concerns the CPU time passed in the solver, or the real time?

blegat:
I think it would be better to use the CPU time passed in the solver

remi-garcia:
I should have precise that in docs, sorry.
From my perspective it was **real time** since I wanted it to replace [CPX_PARAM_TILIM](https://www.ibm.com/support/knowledgecenter/SSSA5P_12.9.0/ilog.odms.cplex.help/CPLEX/Parameters/topics/TiLim.html) for CPLEX.

Maybe we should precise that in `/src/attributes.jl` and duplicate it for CPU time. Or [as CPLEX](https://www.ibm.com/support/knowledgecenter/SSSA5P_12.9.0/ilog.odms.cplex.help/CPLEX/Parameters/topics/ClockType.html) we could have a second parameter with 2 values `CPUTime` and `RealTime`.

blegat:
For Mosek, it's unclear if it's CPUTime or REALTime (see https://docs.mosek.com/8.1/rmosek/parameters.html#mosek.dparam.optimizer_max_time).
cc @ulfworsoe

mlubin:
I understood the limit as wall clock time (real time). We need to clarify this in the docs.

I'm not sure how essential it is to have a second solver-independent option with a CPU time limit, that's a bit esoteric.

PR #815, opened 2019-08-04
Change ObjectiveSense() doc

### LINKED ISSUE #814: ObjectiveSense or OptimizationSense
https://github.com/JuliaOpt/MathOptInterface.jl/blob/0923d1e3a5fc9fdabc9c322bee84f29abe3805c3/src/attributes.jl#L473-L481
I think there is an inconsistency in the comment.
If I understand that well in the doc only `ObjectiveSense()` should appear, `OptimizationSense` is only here so that works on MOI side and MOI users don't need it. Is that right?
I also have noticed that `FeasiblitySense` should be `FEASIBILITY_SENSE`

Sorry if posting an issue wasn't what I should have done. I've still a lot to learn with MOI.

remi-garcia:
Closes #814 
This change makes it more clear for me. Is that so for native English speaker?

### DIFF
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -472,9 +472,9 @@ struct Name <: AbstractModelAttribute end
 """
     ObjectiveSense()
 
-A model attribute for the `OptimizationSense` of the objective function, which
-can be `MIN_SENSE`, `MAX_SENSE`, or `FeasiblitySense`. The default is
-`FEASIBILITY_SENSE`.
+A model attribute for the objective sense of the objective function, which
+must be an `OptimizationSense`: `MIN_SENSE`, `MAX_SENSE`, or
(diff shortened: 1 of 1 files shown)


src/attributes.jl
```diff
@@ -472,9 +472,9 @@ struct Name <: AbstractModelAttribute end
 """
     ObjectiveSense()
 
-A model attribute for the `OptimizationSense` of the objective function, which
-can be `MIN_SENSE`, `MAX_SENSE`, or `FeasiblitySense`. The default is
-`FEASIBILITY_SENSE`.
+A model attribute for the objective sense of the objective function, which
+must be an `OptimizationSense`: `MIN_SENSE`, `MAX_SENSE`, or
```
blegat: Use `[OptimizationSense](@ref)`

mlubin: `OptimizationSense` has no docstring, so this won't be useful.

Issue #750, opened 2019-05-27
Run basic_constraint_tests for all bridges


blegat:
Now that bridges need to implement `MOI.get` for `MOI.ConstraintFunction` and `MOI.ConstraintSet`, it is good practice to run `basic_constraint_tests` with the constraints that can be bridged.

- [x] flip_sign.jl https://github.com/JuliaOpt/MathOptInterface.jl/pull/755
- [x] geomean.jl https://github.com/JuliaOpt/MathOptInterface.jl/pull/888
- [x] interval.jl
- [x] quadtosoc.jl
- [x] scalarize.jl
- [x] slack.jl
- [x] square.jl https://github.com/JuliaOpt/MathOptInterface.jl/pull/899
- [x] det.jl
- [x] functionize.jl
- [ ] indicator.jl
- [x] rsoc.jl
- [x] soc_to_psd.jl
- [x] vectorize.jl

odow:
Closing in favor of https://github.com/jump-dev/MathOptInterface.jl/issues/820

PR #818, opened 2019-08-07
add L_1-norm and L_infinity-norm cones and tests

chriscoey:
The names can be bikeshed. 
I plan to add bridges from these sets to LP constraints (edit: in a PR after this is merged) (edit 2: may as well add bridges here).
Hypatia supports these cones, and has significantly better performance with these cones than with the LP extended formulations.


chriscoey:
Note: the second test instance for both cones is the same as the SOC3 test, but with the 2-dim SOC replaced with equivalent 2-dim norm1 or norminf cones. The contconic file can do with quite a bit of refactoring and cleanup but that should be done separately.

chriscoey:
Failures for LazyBridgeOptimizer. I don't understand that part.

blegat:
> Failures for LazyBridgeOptimizer. I don't understand that part.

Just exclude the tests and add a comment to add them back once bridges are implemented

src/Bridges/Constraint/norm_to_lp.jl
```diff
@@ -0,0 +1,52 @@
...
+function bridge_constraint(::Type{NormOneBridge{T}}, model, f::MOI.VectorOfVariables, s::MOI.NormOneCone) where T
+    return bridge_constraint(NormOneBridge{T}, model, MOI.VectorAffineFunction{T}(f), s)
+end
+function bridge_constraint(::Type{NormOneBridge{T}}, model, f::MOI.VectorAffineFunction{T}, s::MOI.NormOneCone) where T
+    f_scalars = MOIU.eachscalar(f)
+    t = f_scalars[1]
+    x = f_scalars[2:end]
+    d = MOI.dimension(s) - 1
+    y = MOI.add_variables(model, d)
+    eq_index = MOI.add_constraint(model, t - sum(y), MOI.EqualTo(zero(T))) # TODO fix
```
chriscoey: @blegat what is the best way of doing the `t - sum(y)` here?

src/Bridges/Constraint/norm_to_lp.jl
```diff
@@ -0,0 +1,52 @@
...
+    ge_index = MOI.add_constraint(model, MOIU.operate(vcat, T, y - x, y + x), MOI.Nonnegatives(2d)) # TODO fix
```
chriscoey: @blegat how should I do the `MOIU.operate(vcat, T, y - x, y + x)` part?

src/Bridges/Constraint/norm_to_lp.jl
```diff
@@ -1,3 +1,45 @@
...
+end
+function bridge_constraint(::Type{NormInfinityBridge{T}}, model, f::MOI.VectorOfVariables, s::MOI.NormInfinityCone) where T
+    return bridge_constraint(NormInfinityBridge{T}, model, MOI.VectorAffineFunction{T}(f), s)
+end
+function bridge_constraint(::Type{NormInfinityBridge{T}}, model, f::MOI.VectorAffineFunction{T}, s::MOI.NormInfinityCone) where T
+    f_scalars = MOIU.eachscalar(f)
+    t = f_scalars[1]
+    x = f_scalars[2:end]
+    d = MOI.dimension(s) - 1
+    ge_index = MOI.add_constraint(model, MOIU.operate(vcat, T, t - x, t + x), MOI.Nonnegatives(2d)) # TODO fix
```
chriscoey: @blegat how should I do the `t - x` here?

src/Bridges/Constraint/norm_to_lp.jl
```diff
@@ -18,7 +60,7 @@ function bridge_constraint(::Type{NormOneBridge{T}}, model, f::MOI.VectorAffineF
     d = MOI.dimension(s) - 1
     y = MOI.add_variables(model, d)
     eq_index = MOI.add_constraint(model, t - sum(y), MOI.EqualTo(zero(T))) # TODO fix
-    ge_index = MOI.add_constraint(model, MOIU.operate(vcat, T, y - x, y + x), MOI.Nonnegatives(2d)) # TODO fix 
+    ge_index = MOI.add_constraint(model, MOIU.operate(vcat, T, y - x, y + x), MOI.Nonnegatives(2d)) # TODO fix
```
chriscoey: @blegat how should I do the `y - x` here?

src/Bridges/Constraint/norm_to_lp.jl
blegat: Currently `sum` will be inefficient as it won't exploit the mutability of the temporary affine expressions created (cc @Peiffap ). You an build the sum manually with `MOI.ScalarAffineFunction(...)` or use a `for` loop and `operate!(+, ...)`

chriscoey: If sum will be faster in future then wouldn't it make more sense to use sum now, rather than the more complicated way?

src/Bridges/Constraint/norm_to_lp.jl
chriscoey: This is different from my previous question that you answered, because this is a vector.

chriscoey:
Bridges added, NEWS updated, tests passing. Please review

src/Bridges/Constraint/norm_to_lp.jl
```diff
@@ -0,0 +1,92 @@
...
+    ge_index::CI{MOI.VectorAffineFunction{T}, MOI.Nonnegatives}
+end
+function bridge_constraint(::Type{NormInfinityBridge{T}}, model, f::MOI.VectorOfVariables, s::MOI.NormInfinityCone) where T
+    return bridge_constraint(NormInfinityBridge{T}, model, MOI.VectorAffineFunction{T}(f), s)
+end
+function bridge_constraint(::Type{NormInfinityBridge{T}}, model, f::MOI.VectorAffineFunction{T}, s::MOI.NormInfinityCone) where T
+    f_scalars = MOIU.eachscalar(f)
+    t = f_scalars[1]
+    d = MOI.dimension(s)
+    new_f = MOIU.vectorize(vcat([MOIU.operate(-, T, t, f_scalars[i]) for i in 2:d], [MOIU.operate(+, T, t, f_scalars[i]) for i in 2:d]))
```
blegat: You can do `ub = MOIU.operate!(-, T, f_scalars[2:d])` and then `MOIU.operate_output_index!(+, i, T, ub, t)`

src/Bridges/Constraint/norm_to_lp.jl
```diff
@@ -0,0 +1,92 @@
+"""
+    NormInfinityBridge{T}
+
+The `NormInfinityCone` is representable with LP constraints, since ``t \\ge \\max_i |x_i|`` if and only if ``t \\ge x_i`` and ``t \\ge -x_i`` for all ``i``.
+"""
+struct NormInfinityBridge{T} <: AbstractBridge
+    ge_index::CI{MOI.VectorAffineFunction{T}, MOI.Nonnegatives}
+end
+function bridge_constraint(::Type{NormInfinityBridge{T}}, model, f::MOI.VectorOfVariables, s::MOI.NormInfinityCone) where T
+    return bridge_constraint(NormInfinityBridge{T}, model, MOI.VectorAffineFunction{T}(f), s)
```
blegat: You shouldn't need this if you use `MOIU.operate`, it should convert when needed. The bridges that uses this were written before `MOIU.operate`

src/Bridges/Constraint/norm_to_lp.jl
```diff
@@ -0,0 +1,92 @@
+"""
+    NormInfinityBridge{T}
+
+The `NormInfinityCone` is representable with LP constraints, since ``t \\ge \\max_i |x_i|`` if and only if ``t \\ge x_i`` and ``t \\ge -x_i`` for all ``i``.
+"""
+struct NormInfinityBridge{T} <: AbstractBridge
+    ge_index::CI{MOI.VectorAffineFunction{T}, MOI.Nonnegatives}
```
blegat: The function should be determined in `concrete_bridge_type` using `promote_operation`

chriscoey: OK this means nothing to me. Can you point to an example in another cone that I can copy-paste?

chriscoey:
@blegat I addressed your comments except for https://github.com/JuliaOpt/MathOptInterface.jl/pull/818#discussion_r311935747

src/Bridges/Constraint/norm_to_lp.jl
```diff
@@ -6,21 +6,26 @@ The `NormInfinityCone` is representable with LP constraints, since ``t \\ge \\ma
...
     ge_index::CI{MOI.VectorAffineFunction{T}, MOI.Nonnegatives}
 end
-function bridge_constraint(::Type{NormInfinityBridge{T}}, model, f::MOI.VectorOfVariables, s::MOI.NormInfinityCone) where T
-    return bridge_constraint(NormInfinityBridge{T}, model, MOI.VectorAffineFunction{T}(f), s)
-end
-function bridge_constraint(::Type{NormInfinityBridge{T}}, model, f::MOI.VectorAffineFunction{T}, s::MOI.NormInfinityCone) where T
+function bridge_constraint(::Type{NormInfinityBridge{T}}, model, f::MOI.AbstractVectorFunction, s::MOI.NormInfinityCone) where T
     f_scalars = MOIU.eachscalar(f)
     t = f_scalars[1]
+    if t isa MOI.SingleVariable # TODO currently needed for operate_output_index!
```
chriscoey: @blegat this was needed due to a missing method for operate_output_index!

src/Bridges/Constraint/norm_to_lp.jl
```diff
@@ -6,21 +6,26 @@ The `NormInfinityCone` is representable with LP constraints, since ``t \\ge \\ma
...
+        t = MOI.ScalarAffineFunction{T}(t)
+    end
     d = MOI.dimension(s)
-    new_f = MOIU.vectorize(vcat([MOIU.operate(-, T, t, f_scalars[i]) for i in 2:d], [MOIU.operate(+, T, t, f_scalars[i]) for i in 2:d]))
-    ge_index = MOI.add_constraint(model, new_f, MOI.Nonnegatives(2d - 2))
+    lb = f_scalars[2:d]
+    ub = MOIU.operate(-, T, lb)
+    f_new = MOIU.operate(vcat, T, ub, lb)
+    for i in 1:MOI.output_dimension(f_new)
+        MOIU.operate_output_index!(+, T, i, f_new, t)
```
chriscoey: @blegat is this how you envisioned it?

src/Bridges/Constraint/norm_to_lp.jl
```diff
@@ -50,21 +55,26 @@ struct NormOneBridge{T} <: AbstractBridge
     eq_index::CI{MOI.ScalarAffineFunction{T}, MOI.EqualTo{T}}
     ge_index::CI{MOI.VectorAffineFunction{T}, MOI.Nonnegatives}
 end
-function bridge_constraint(::Type{NormOneBridge{T}}, model, f::MOI.VectorOfVariables, s::MOI.NormOneCone) where T
-    return bridge_constraint(NormOneBridge{T}, model, MOI.VectorAffineFunction{T}(f), s)
-end
-function bridge_constraint(::Type{NormOneBridge{T}}, model, f::MOI.VectorAffineFunction{T}, s::MOI.NormOneCone) where T
+function bridge_constraint(::Type{NormOneBridge{T}}, model, f::MOI.AbstractVectorFunction, s::MOI.NormOneCone) where T
+    if f isa MOI.VectorOfVariables # TODO currently needed for test to pass
```
chriscoey: @blegat this may have exposed a bug? without these 3 lines, normone1v test fails (though normone1f test passes).

src/Bridges/Constraint/norm_to_lp.jl
blegat: See below:
https://github.com/JuliaOpt/MathOptInterface.jl/blob/351269c98ebfeb5edeea05e0c5f393812c9d185f/src/Bridges/Constraint/scalarize.jl#L34-L38
https://github.com/JuliaOpt/MathOptInterface.jl/blob/351269c98ebfeb5edeea05e0c5f393812c9d185f/src/Bridges/Constraint/flip_sign.jl#L97-L102

src/Bridges/Constraint/norm_to_lp.jl
blegat: Yes, I like that! Maybe you are missing a second line `MOIU.operate_output_index!(+, T, d + i, f_new, t)` for the `lb` part ?

test/Bridges/Constraint/norm_to_lp.jl
```diff
@@ -0,0 +1,32 @@
...
+config = MOIT.TestConfig(duals = false)
+
+@testset "NormInfinity" begin
+    mock.optimize! = (mock::MOIU.MockOptimizer) -> MOIU.mock_optimize!(mock, [1.0, 0.5, 1.0])
+    bridged_mock = MOIB.Constraint.NormInfinity{Float64}(mock)
+    MOIT.norminf1vtest(bridged_mock, config)
+    MOIT.norminf1ftest(bridged_mock, config)
+    # Dual is not yet implemented for NormInfinity bridge
+    ci = first(MOI.get(bridged_mock, MOI.ListOfConstraintIndices{MOI.VectorAffineFunction{Float64}, MOI.NormInfinityCone}()))
+    test_delete_bridge(bridged_mock, ci, 3, ((MOI.VectorAffineFunction{Float64}, MOI.Nonnegatives, 0),))
```
blegat: Needs to test with `test_model_equals`, see https://github.com/JuliaOpt/MathOptInterface.jl/issues/820

src/Bridges/Constraint/norm_to_lp.jl
chriscoey: on line 18 (before this) I vcat ub and lb

test/Bridges/Constraint/norm_to_lp.jl
chriscoey: I'm going to need to wait to see more examples of how that is supposed to look

src/Bridges/Constraint/norm_to_lp.jl
chriscoey: I do not understand what additional type parameters these bridges need, or the purpose of it. The geomean bridge code was my basis for implementing these new cones. Perhaps we can update these new bridges in future when the geomean bridge code is also updated?

blegat: > I do not understand what additional type parameters these bridges need, or the purpose of it

This would allow the bridge to be used with any `AbstractVectorFunction`, e.g. `VectorQuadraticFunction`.

test/Bridges/Constraint/norm_to_lp.jl
blegat: See https://github.com/JuliaOpt/MathOptInterface.jl/pull/821/commits/ae71c8b6e3a5f189f63513dd11cbc145bc21b6c7

src/Bridges/Constraint/norm_to_lp.jl
blegat: Indeed, MWE welcome :)

src/Bridges/Constraint/norm_to_lp.jl
blegat: Yes but then you only add `t` to the indices corresponding to `ub`, so `t` is not constrained to be higher than the lower bound, you simply have `-lb <= 0`

chriscoey: I did it differently to how you suggested - on line 18, I vcat to get `f_new`, then I add `t` to every element of `f_new`, which is double the length of ub

chriscoey: the loop is `for i in 1:MOI.output_dimension(f_new)`

blegat: > the loop is `for i in 1:MOI.output_dimension(f_new)`

Aaaaah, that's what I was missing, I thought it was `for i in 1:d`

src/Bridges/Constraint/norm_to_lp.jl
chriscoey: So do I need `NormInfinityBridge{T, F, G}` and `NormOneBridge{T, F, G}`?

chriscoey:
@blegat should I have a go at implementing constraint duals thru the bridge? any expected gotchas there?

src/Bridges/Constraint/norm_to_lp.jl
chriscoey: I'm now hesitant about this because it VQF complicate duals-through-bridge: https://github.com/JuliaOpt/MathOptInterface.jl/pull/818#issuecomment-519782772

blegat:
> I'm not sure whether allowing VectorQuadraticFunction would complicate that

I don't think so, just implement duals thinking about the VectorAffineFunction and it would automatically work also for VQF because what you do is basically multiplying by the adjoint of the linear map between the two sets and you can proof that it's correct in the lagrangian function so it should work with any function so using `AbstractVectorFunction` is correct. See slide 6 of https://www.researchgate.net/publication/334574258_Automatic_reformulation_using_optimization_bridges

> that the bridge could convert a convex model to a nonconvex model

Why is that ? If the quadratic function `x_i` have a PSD Q then `t >= -x_i` is indeed nonconvex but wasn't the initial model also nonconvex ?

chriscoey:
OK I allow VQF and I changed the equality constraint in L1 cone LP formulation to an inequality, so now it won't convert convex constraints to nonconvex ones. 
New tests are passing and also the contconic new tests pass with Hypatia's L1 and Linf cones (incl duals).
But I haven't yet added the model equality tests. I would still like more examples to see how to do this, because the example at ae71c8b is for a variable bridge.

chriscoey:
@blegat @mlubin if a bridge adds vector constraints and scalar constraints, is it preferable to implement the scalar constraints as vector constraints (eg VAF in Nonnegatives(1))? Since some conic solver wrappers may decide to vectorize.

src/Bridges/Constraint/norm_to_lp.jl
```diff
@@ -0,0 +1,124 @@
+"""
+    NormInfinityBridge{T}
+
+The `NormInfinityCone` is representable with LP constraints, since ``t \\ge \\max_i |x_i|`` if and only if ``t \\ge x_i`` and ``t \\ge -x_i`` for all ``i``.
+"""
+struct NormInfinityBridge{T} <: AbstractBridge
+    nn_index::CI{<:Union{MOI.VectorAffineFunction{T}, MOI.VectorQuadraticFunction{T}}, MOI.Nonnegatives}
```
blegat: You should infer the type of the function in `concrete_bridge_type`, see
https://github.com/JuliaOpt/MathOptInterface.jl/blob/af426f0f96142609ab18ee1ba91c6b4ca425a50c/src/Bridges/Constraint/geomean.jl#L108-L116

chriscoey: OK I'll have a go at this though I'm still a little confused. 
How in the GeoMeanBridge code can we sometimes have `GeoMeanBridge{T}` and sometimes `GeoMeanBridge{T, F}` and sometimes `GeoMeanBridge{T, F, G}`?

chriscoey: I mean does Julia just allow us to lop off the last type parameters?

blegat: When you do `::GeoMeanBridge{T}` for an argument, you allow the second and third type parameters to be anything and you won't reference them in the function.
It is equivalent to doing `::GeoMeanBridge{T, F, G} ... where {F, G}` but when you don't use `F` and `G` in the function, I prefer just writing `::GeoMeanBridge{T}`.
For `::Type{<:GeoMeanBridge{T}}`, it's the same.

src/Bridges/Constraint/norm_to_lp.jl
```diff
@@ -0,0 +1,124 @@
...
+    nn_index::CI{<:Union{MOI.VectorOfVariables, MOI.VectorAffineFunction{T}}, MOI.Nonnegatives}
+end
+function bridge_constraint(::Type{NormOneBridge{T}}, model::MOI.ModelLike, f::MOI.AbstractVectorFunction, s::MOI.NormOneCone) where T
+    if f isa MOI.VectorOfVariables # TODO currently needed for test to pass
+        f = MOI.VectorAffineFunction{T}(f)
+    end
+    f_scalars = MOIU.eachscalar(f)
+    d = MOI.dimension(s)
+    y = MOI.add_variables(model, d - 1)
+    ge_index = MOI.add_constraint(model, MOIU.operate(-, T, f_scalars[1], MOIU.operate(sum, T, y)), MOI.GreaterThan(zero(T)))
```
blegat: Use MOIU.normalise_and_add_constraint` to move the constant to the set in case it is nonzero

blegat:
> is it preferable to implement the scalar constraints as vector constraints (eg VAF in Nonnegatives(1))

Conic solvers are usually: SAF-in-EqualTo/VOV-in-AbstractVectorSet in the primal and VAF-in-AbstractVectorSet in the dual or the opposite.
For bridges, I prefer entering the constraint in the closest form from the bridge so look for the constraint that requires the less effort. Otherwise, you end up hardcoding VectorizeBridge or ScalarizeBridge inside your transformation.
In this case, you have it in the form of a SAF-in-GreaterThan and vectorizing it might end up hardcoding the VectorizeBridge in your bridge which is a waste if the solver need to scalarize it afterwards.

chriscoey:
@blegat I tried adding `concrete_bridge_type` but now tests are failing, eg for `norminf1ftest` (but not `norminf1vtest`)
```
No mock dual is set for constraint `MathOptInterface.ConstraintIndex{MathOptInterface.VectorAffineFunction{Float64},MathOptInterface.NormInfinityCone}(12345677)`
```
I can't figure out why.

src/Bridges/Constraint/norm_to_lp.jl
```diff
@@ -23,19 +25,23 @@ function bridge_constraint(::Type{NormInfinityBridge{T}}, model::MOI.ModelLike,
     return NormInfinityBridge(nn_index)
 end
 
-MOI.supports_constraint(::Type{NormInfinityBridge{T}}, ::Type{<:MOI.AbstractVectorFunction}, ::Type{MOI.NormInfinityCone}) where T = true
+MOI.supports_constraint(::Type{NormInfinityBridge}, ::Type{<:MOI.AbstractVectorFunction}, ::Type{MOI.NormInfinityCone}) = true
```
blegat: This should be `Type{NormInfinityBridge{T}}`, it should fix your issue

chriscoey:
@blegat tests passing, ready for review

src/sets.jl
```diff
@@ -184,6 +184,28 @@ constant(s::EqualTo) = s.value
 constant(s::GreaterThan) = s.lower
 constant(s::LessThan) = s.upper
 
+"""
+    NormInfinityCone(dimension)
+
+The epigraph of the L∞ norm function ``\\{ (t,x) \\in \\mathbb{R}^{dimension} : t \\ge || x ||_∞ = \\max_i |x_i| \\}`` of dimension `dimension`.
```
blegat: Shouldn't it be \infty in LaTeX ?

src/Bridges/Constraint/norm_to_lp.jl
```diff
@@ -0,0 +1,134 @@
...
+    if f isa MOI.VectorOfVariables # TODO currently needed for test to pass
+        f = MOI.VectorAffineFunction{T}(f)
+    end
+    f_scalars = MOIU.eachscalar(f)
+    d = MOI.dimension(s)
+    y = MOI.add_variables(model, d - 1)
+    ge_index = MOIU.normalize_and_add_constraint(model, MOIU.operate(-, T, f_scalars[1], MOIU.operate(sum, T, y)), MOI.GreaterThan(zero(T)), allow_modify_function=true)
+    lb = f_scalars[2:d]
+    ub = MOIU.operate(-, T, lb)
+    MOIU.operate!(+, T, lb, MOI.VectorOfVariables(y))
```
blegat: `operate!` is not guarnteed to store the result in the first arg. For instance, it is not the case for VectorOfVariables hence the bug you have (I guess it's the issue).
You need to catch the output of `operate!` in another variable and not use `lb` anymore (as it might or might not have been modified, you don't know).

src/sets.jl
chriscoey: sure. is there a way to preview the latex output?

src/Bridges/Constraint/norm_to_lp.jl
chriscoey: ok that's a little confusing. will try it.

chriscoey: indeed that fixed it

src/sets.jl
```diff
@@ -184,6 +184,28 @@ constant(s::EqualTo) = s.value
 constant(s::GreaterThan) = s.lower
 constant(s::LessThan) = s.upper
 
+"""
+    NormInfinityCone(dimension)
+
+The epigraph of the ``\\ell_\\infty``-norm function ``\\{ (t,x) \\in \\mathbb{R}^{dimension} : t \\ge || x ||_∞ = \\max_i |x_i| \\}`` of dimension `dimension`.
```
blegat: I meant the infinity symbol in the subscript of the norm. By the way, the norm should be `\|` or `\lVert`/`\rVert` instead of `||`

chriscoey:
FYI I'll probably have a go at spectral norm and its dual nuclear norm after this PR, and I'll use the code here as my template.

test/Bridges/Constraint/norm_to_lp.jl
mlubin: I would like to see tests with `test_models_equal`. I find these really helpful to see concretely what the bridge is doing.

mlubin:
This is non-breaking (except in that it adds new tests that solvers could fail), so we'll let it into 0.9.x after the finishing touches.

test/Bridges/Constraint/norm_to_lp.jl
chriscoey: OK. does that mean tests in the style of both https://github.com/JuliaOpt/MathOptInterface.jl/blob/d9673e8c04f23519436a747cbc8aeb881235ceae/test/Bridges/Variable/rsoc_to_psd.jl#L26 and https://github.com/JuliaOpt/MathOptInterface.jl/blob/d9673e8c04f23519436a747cbc8aeb881235ceae/test/Bridges/Variable/rsoc_to_psd.jl#L61? what are the high-level purposes of those two testsets?

@blegat those are for a variable bridge instead of a constraint bridge like in this PR. how different do they need to look for constraint bridge? do we have any closer examples I can use as a template here?

blegat:
Then we should add support for `ConstraintSet` and `ConstraintFunction` to the bridges, otherwise, e.g. printing a JuMP model using the bridge in direct mode will fail.
You can test that it works with `basic_constraint_test`: https://github.com/JuliaOpt/MathOptInterface.jl/issues/750

chriscoey:
> Then we should add support for `ConstraintSet` and `ConstraintFunction` to the bridges, otherwise, e.g. printing a JuMP model using the bridge in direct mode will fail.
> You can test that it works with `basic_constraint_test`: #750

Is get ConstraintFunction going to be quite complicated because we need to reconstruct it? in #750 it's easy because that is a simple bridge. can we just store the original function in the bridge and return that?

blegat:
> can we just store the original function in the bridge and return that?

It's ok of it's costly, no need to have cache. If the user wants efficient access to the function he should use CachingOptimizer.

chriscoey:
But it's also a matter of unnecessarily complex code to reconstruct it, when it could just be stored.

Similarly for get function primal value... It is complex to reconstruct that when we could just return the evaluation of the stored function at the primal variable values. This was somewhat of a burden to implement. 

Complex code is also more prone to errors and harder to maintain.

chriscoey:
I could see it becoming even more burdensome to reconstruct the original function when we want to add more complicated bridges eg for piecewise linear functions and MIP formulations. I don't even know how I'd do it for the cones in this PR in a way that's fairly compact and readable.

blegat:
The philosophy in MOI is more: support what you can, the user will use appropriate utility in case it wants to do something you don't support.
It's not the role of bridges to cache and use get_fallback for ConstraintPrimal, it's the role of CachingOptimizer and it's the role of the user to use `get_fallback` if it needs to query ConstraintPrimal for a model that does not support it.
The advantage of this is that it gives more flexibility to the user and MOI utilities are more atomic.
If it's not possible to get the ConstraintFunction for a bridge, it's ok if it's not implemented.
Here, the code you have for ConstraintPrimal can be used to get the `ConstraintFunction` by, e.g. replacing `+` by `operate(+, ...)`

chriscoey:
Ok. It's not a hill I'm prepared to die on. Is there a nice way we can avoid the almost-duplicated code needed for get constraint function and get constraint primal? Not just for this bridge but for others too

blegat:
> Is there a nice way we can avoid the almost-duplicated code needed for get constraint function and get constraint primal?

Good point. If the transformation for the constraint primal is written with `MOIU.operate`, it should work for `ConstraintPrimal` too, there might just be a few `MOIU.operate` methods to add. That might be the direction to go.

src/Bridges/Constraint/norm_to_lp.jl
```diff
@@ -41,6 +41,23 @@ MOI.get(b::NormInfinityBridge{T, F}, ::MOI.ListOfConstraintIndices{F, MOI.Nonneg
 MOI.delete(model::MOI.ModelLike, c::NormInfinityBridge) = MOI.delete(model, c.nn_index)
 
 # Attributes, Bridge acting as a constraint
+function MOI.get(model::MOI.ModelLike, ::MOI.ConstraintFunction, c::NormInfinityBridge{T, F}) where {T, F}
+    ge = MOIU.eachscalar(MOI.get(model, MOI.ConstraintFunction(), c.nn_index))
+    t = MOIU.operate!(/, T, sum(ge), T(length(ge)))
+    d = div(length(ge), 2)
+    x = MOIU.operate!(-, T, ge[(d + 1):end], ge[1:d])
+    # x = x / T(2)
+    # x = MOIU.operate!(/, T, x, T(2))
```
chriscoey: @blegat I need to divide a VAF by 2 but I can't find a way to do this. If you could add the necessary method for operate in this PR (or in master and I will merge in) that would be a great help.

src/Bridges/Constraint/norm_to_lp.jl
```diff
@@ -42,37 +42,31 @@ MOI.delete(model::MOI.ModelLike, c::NormInfinityBridge) = MOI.delete(model, c.nn
...
-    # x = x / T(2)
-    # x = MOIU.operate!(/, T, x, T(2))
-    # for i in 1:d
-    #     MOIU.operate_output_index!(/, T, i, x, T(2))
-    # end
-    @show MOIU.canonicalize!(MOIU.operate(vcat, T, t, x))
+    nn_func = MOIU.eachscalar(MOI.get(model, MOI.ConstraintFunction(), c.nn_index))
+    t = MOIU.operate!(/, T, sum(nn_func), T(length(nn_func)))
+    d = div(length(nn_func), 2)
+    x = MOIU.operate!(/, T, MOIU.operate!(-, T, nn_func[(d + 1):end], nn_func[1:d]), T(2))
```
chriscoey: @blegat This line fails because I need to divide a VAF by 2. If you could add the necessary method for `operate` in this PR (or in master and I will merge in) that would be a great help.

src/Bridges/Constraint/norm_to_lp.jl
```diff
@@ -128,12 +122,24 @@ function MOI.delete(model::MOI.ModelLike, c::NormOneBridge)
 end
 
 # Attributes, Bridge acting as a constraint
+function MOI.get(model::MOI.ModelLike, ::MOI.ConstraintFunction, c::NormOneBridge{T, F, G}) where {T, F, G}
+    ge_func = MOI.get(model, MOI.ConstraintFunction(), c.ge_index)
+    nn_func = MOIU.eachscalar(MOI.get(model, MOI.ConstraintFunction(), c.nn_index))
+    t = MOIU.operate!(+, T, ge_func, MOIU.operate!(/, T, sum(nn_func), T(2)))
+    d = div(length(nn_func), 2)
+    x = MOIU.operate!(/, T, MOIU.operate!(-, T, nn_func[(d + 1):end], nn_func[1:d]), T(2))
```
chriscoey: @blegat see https://github.com/JuliaOpt/MathOptInterface.jl/pull/818/commits/913e3db3cd4abf4e57ccba2336e0f16327fc6b88#r314057544

chriscoey:
@blegat I need your help to finish this. See my comment https://github.com/JuliaOpt/MathOptInterface.jl/pull/818/commits/913e3db3cd4abf4e57ccba2336e0f16327fc6b88#r314057544.
Beyond that, there are a couple of failing "basic constraint tests" that I can't figure out.

test/Bridges/Constraint/norm_to_lp.jl
chriscoey: done

chriscoey:
@blegat thanks for the fix in #837.
Basic constraint tests are now failing later, and I'll need your help again. 
There seem to be 3 types of failure/error (same for NormOne and NormInfinity):
```
julia> @testset begin include("test/Bridges/Constraint/norm_to_lp.jl") end

ConstraintFunction: Error During Test at /home/coey/.julia/dev/MathOptInterface/src/Test/UnitTests/basic_constraint_tests.jl:210
  Test threw exception
  Expression: func ≈ constraint_function
  MethodError: no method matching isapprox(::MathOptInterface.VectorAffineFunction{Float64}, ::MathOptInterface.VectorOfVariables)
  Closest candidates are:
    isapprox(!Matched::Missing, ::Any; kwargs...) at missing.jl:69
    isapprox(::Any, !Matched::Missing; kwargs...) at missing.jl:70
    isapprox(!Matched::Union{MathOptInterface.SingleVariable, MathOptInterface.VectorOfVariables}, ::Union{MathOptInterface.SingleVariable, MathOptInterface.VectorOfVariables}; kwargs...) at /home/coey/.julia/dev/MathOptInterface/src/functions.jl:281
    ...
  Stacktrace:
   [1] eval_test(::Expr, ::Expr, ::LineNumberNode, ::Bool) at /home/coey/julia/usr/share/julia/stdlib/v1.3/Test/src/Test.jl:244
   [2] macro expansion at /home/coey/.julia/dev/MathOptInterface/src/Test/UnitTests/basic_constraint_tests.jl:210 [inlined]
   [3] macro expansion at /home/coey/julia/usr/share/julia/stdlib/v1.3/Test/src/Test.jl:1114 [inlined]
   [4] macro expansion at /home/coey/.julia/dev/MathOptInterface/src/Test/UnitTests/basic_constraint_tests.jl:209 [inlined]
   [5] macro expansion at /home/coey/julia/usr/share/julia/stdlib/v1.3/Test/src/Test.jl:1114 [inlined]
   [6] #basic_constraint_test_helper#39(::Bool, ::Bool, ::Bool, ::typeof(MathOptInterface.Test.basic_constraint_test_helper), ::MathOptInterface.Bridges.Constraint.SingleBridgeOptimizer{MathOptInterface.Bridges.Constraint.NormInfinityBridge{Float64,F} where F,MathOptInterface.Utilities.MockOptimizer{MathOptInterface.Utilities.Model{Float64}}}, ::MathOptInterface.Test.TestConfig, ::getfield(MathOptInterface.Test, Symbol("##10#11")), ::MathOptInterface.NormInfinityCone, ::Int64) at /home/coey/.julia/dev/MathOptInterface/src/Test/UnitTests/basic_constraint_tests.jl:196
  
ConstraintFunction: Test Failed at /home/coey/.julia/dev/MathOptInterface/src/Test/UnitTests/basic_constraint_tests.jl:211
  Expression: variables(func) == variables(constraint_function)
   Evaluated: Set(MathOptInterface.VariableIndex[MathOptInterface.VariableIndex(12345677), MathOptInterface.VariableIndex(12345676), MathOptInterface.VariableIndex(12345679)]) == MathOptInterface.VariableIndex[MathOptInterface.VariableIndex(12345679), MathOptInterface.VariableIndex(12345676), MathOptInterface.VariableIndex(12345677)]
Stacktrace:
 [1] macro expansion at /home/coey/.julia/dev/MathOptInterface/src/Test/UnitTests/basic_constraint_tests.jl:211 [inlined]
 [2] macro expansion at /home/coey/julia/usr/share/julia/stdlib/v1.3/Test/src/Test.jl:1114 [inlined]
 [3] macro expansion at /home/coey/.julia/dev/MathOptInterface/src/Test/UnitTests/basic_constraint_tests.jl:209 [inlined]
 [4] macro expansion at /home/coey/julia/usr/share/julia/stdlib/v1.3/Test/src/Test.jl:1114 [inlined]
 [5] #basic_constraint_test_helper#39(::Bool, ::Bool, ::Bool, ::typeof(MathOptInterface.Test.basic_constraint_test_helper), ::MathOptInterface.Bridges.Constraint.SingleBridgeOptimizer{MathOptInterface.Bridges.Constraint.NormInfinityBridge{Float64,F} where F,MathOptInterface.Utilities.MockOptimizer{MathOptInterface.Utilities.Model{Float64}}}, ::MathOptInterface.Test.TestConfig, ::getfield(MathOptInterface.Test, Symbol("##10#11")), ::MathOptInterface.NormInfinityCone, ::Int64) at /home/coey/.julia/dev/MathOptInterface/src/Test/UnitTests/basic_constraint_tests.jl:196

ConstraintFunction: Error During Test at /home/coey/.julia/dev/MathOptInterface/src/Test/UnitTests/basic_constraint_tests.jl:208
  MethodError: no method matching MathOptInterface.VectorAffineTerm(::Int64, ::MathOptInterface.ScalarQuadraticTerm{Float64})
  Closest candidates are:
    MathOptInterface.VectorAffineTerm(::Int64, !Matched::MathOptInterface.ScalarAffineTerm{T}) where T at /home/coey/.julia/dev/MathOptInterface/src/functions.jl:103
    MathOptInterface.VectorAffineTerm(::Integer, !Matched::MathOptInterface.ScalarAffineTerm) at /home/coey/.julia/dev/MathOptInterface/src/functions.jl:108
  Stacktrace:
   [1] unsafe_add(::MathOptInterface.VectorQuadraticTerm{Float64}, ::MathOptInterface.VectorQuadraticTerm{Float64}) at /home/coey/.julia/dev/MathOptInterface/src/Utilities/functions.jl:331
   [2] sort_and_compress!(::Array{MathOptInterface.VectorQuadraticTerm{Float64},1}, ::typeof(MathOptInterface.term_indices), ::getfield(MathOptInterface.Utilities, Symbol("##24#26")), ::typeof(MathOptInterface.Utilities.unsafe_add)) at /home/coey/.julia/dev/MathOptInterface/src/Utilities/functions.jl:427
   [3] canonicalize! at /home/coey/.julia/dev/MathOptInterface/src/Utilities/functions.jl:408 [inlined]
   [4] get(::MathOptInterface.Bridges.Constraint.SingleBridgeOptimizer{MathOptInterface.Bridges.Constraint.NormInfinityBridge{Float64,F} where F,MathOptInterface.Utilities.MockOptimizer{MathOptInterface.Utilities.Model{Float64}}}, ::MathOptInterface.ConstraintFunction, ::MathOptInterface.Bridges.Constraint.NormInfinityBridge{Float64,MathOptInterface.VectorQuadraticFunction{Float64}}) at /home/coey/.julia/dev/MathOptInterface/src/Bridges/Constraint/norm_to_lp.jl:49
   [5] get(::MathOptInterface.Bridges.Constraint.SingleBridgeOptimizer{MathOptInterface.Bridges.Constraint.NormInfinityBridge{Float64,F} where F,MathOptInterface.Utilities.MockOptimizer{MathOptInterface.Utilities.Model{Float64}}}, ::MathOptInterface.ConstraintFunction, ::MathOptInterface.ConstraintIndex{MathOptInterface.VectorQuadraticFunction{Float64},MathOptInterface.NormInfinityCone}) at /home/coey/.julia/dev/MathOptInterface/src/Bridges/bridge_optimizer.jl:515
   [6] macro expansion at /home/coey/.julia/dev/MathOptInterface/src/Test/UnitTests/basic_constraint_tests.jl:209 [inlined] (repeats 2 times)
   [7] macro expansion at /home/coey/julia/usr/share/julia/stdlib/v1.3/Test/src/Test.jl:1114 [inlined]
   [8] #basic_constraint_test_helper#39(::Bool, ::Bool, ::Bool, ::typeof(MathOptInterface.Test.basic_constraint_test_helper), ::MathOptInterface.Bridges.Constraint.SingleBridgeOptimizer{MathOptInterface.Bridges.Constraint.NormInfinityBridge{Float64,F} where F,MathOptInterface.Utilities.MockOptimizer{MathOptInterface.Utilities.Model{Float64}}}, ::MathOptInterface.Test.TestConfig, ::getfield(MathOptInterface.Test, Symbol("##18#19")), ::MathOptInterface.NormInfinityCone, ::Int64) at /home/coey/.julia/dev/MathOptInterface/src/Test/UnitTests/basic_constraint_tests.jl:196
   [9] (::getfield(MathOptInterface.Test, Symbol("#kw##basic_constraint_test_helper")))(::NamedTuple{(:delete, :get_constraint_function, :get_constraint_set),Tuple{Bool,Bool,Bool}}, ::typeof(MathOptInterface.Test.basic_constraint_test_helper), ::MathOptInterface.Bridges.Constraint.SingleBridgeOptimizer{MathOptInterface.Bridges.Constraint.NormInfinityBridge{Float64,F} where F,MathOptInterface.Utilities.MockOptimizer{MathOptInterface.Utilities.Model{Float64}}}, ::MathOptInterface.Test.TestConfig, ::Function, ::MathOptInterface.NormInfinityCone, ::Int64) at ./none:0
   [10] macro expansion at /home/coey/.julia/dev/MathOptInterface/src/Test/UnitTests/basic_constraint_tests.jl:131 [inlined]
   [11] macro expansion at /home/coey/julia/usr/share/julia/stdlib/v1.3/Test/src/Test.jl:1114 [inlined]
   [12] #basic_constraint_tests#20(::Bool, ::Bool, ::Bool, ::Array{Tuple{DataType,DataType},1}, ::Array{Any,1}, ::typeof(MathOptInterface.Test.basic_constraint_tests), ::MathOptInterface.Bridges.Constraint.SingleBridgeOptimizer{MathOptInterface.Bridges.Constraint.NormInfinityBridge{Float64,F} where F,MathOptInterface.Utilities.MockOptimizer{MathOptInterface.Utilities.Model{Float64}}}, ::MathOptInterface.Test.TestConfig) at /home/coey/.julia/dev/MathOptInterface/src/Test/UnitTests/basic_constraint_tests.jl:130
   [13] (::getfield(MathOptInterface.Test, Symbol("#kw##basic_constraint_tests")))(::NamedTuple{(:include,),Tuple{Array{Tuple{DataType,DataType},1}}}, ::typeof(MathOptInterface.Test.basic_constraint_tests), ::MathOptInterface.Bridges.Constraint.SingleBridgeOptimizer{MathOptInterface.Bridges.Constraint.NormInfinityBridge{Float64,F} where F,MathOptInterface.Utilities.MockOptimizer{MathOptInterface.Utilities.Model{Float64}}}, ::MathOptInterface.Test.TestConfig) at ./none:0
   [14] top-level scope at /home/coey/.julia/dev/MathOptInterface/test/Bridges/Constraint/norm_to_lp.jl:17
   [15] top-level scope at /home/coey/julia/usr/share/julia/stdlib/v1.3/Test/src/Test.jl:1114
   [16] top-level scope at /home/coey/.julia/dev/MathOptInterface/test/Bridges/Constraint/norm_to_lp.jl:15
   [17] include at ./boot.jl:328 [inlined]
   [18] include_relative(::Module, ::String) at ./loading.jl:1094
   [19] include(::Module, ::String) at ./Base.jl:31
   [20] include(::String) at ./client.jl:432
   [21] top-level scope at REPL[3]:1
   [22] top-level scope at /home/coey/julia/usr/share/julia/stdlib/v1.3/Test/src/Test.jl:1114
   [23] top-level scope at REPL[3]:1
   [24] eval(::Module, ::Any) at ./boot.jl:330
   [25] eval_user_input(::Any, ::REPL.REPLBackend) at /home/coey/julia/usr/share/julia/stdlib/v1.3/REPL/src/REPL.jl:86
   [26] macro expansion at /home/coey/julia/usr/share/julia/stdlib/v1.3/REPL/src/REPL.jl:118 [inlined]
   [27] (::getfield(REPL, Symbol("##26#27")){REPL.REPLBackend})() at ./task.jl:333
```

blegat:
You need to convert the VAF back into a VOV as done in:
https://github.com/JuliaOpt/MathOptInterface.jl/blob/0ed068ff8342cf2501305963c22708b0b18a4b80/src/Bridges/Constraint/rsoc.jl#L84

chriscoey:
That fixed one of the test failures. Then I made two small modifications to MOIU.functions.jl to get the new tests passing.

chriscoey:
@mlubin @blegat please review

src/Utilities/functions.jl
```diff
@@ -1821,9 +1821,9 @@ function count_terms(dimension::I, terms::Vector{T}) where {I,T}
     return counting
 end
 
-convert_approx(::Type{T}, func::T; kws...) where {T} = func
+convert_approx(::Type{T}, func::T; kws...) where {T} = MOIU.canonical(func)
```
blegat: Why is this needed ?

chriscoey: It's needed to eliminate variables. Without it, MOI thinks it has more variables in the VAF then there are following elimination.

chriscoey: Without this change, we get the following errors for NormOne bridge tests:
```
julia> include("test/Bridges/Constraint/norm_to_lp.jl")
[ Info: Recompiling stale cache file /home/coey/.julia/compiled/v1.3/MathOptInterface/tyub8.ji for MathOptInterface [b8f27783-ece8-5eb3-8dc8-9495eed66fee]
Test Summary: | Pass  Total
NormInfinity  |  139    139
ConstraintFunction: Test Failed at /home/coey/.julia/dev/MathOptInterface/src/Test/UnitTests/basic_constraint_tests.jl:211
  Expression: variables(func) == variables(constraint_function)
   Evaluated: Set(MathOptInterface.VariableIndex[MathOptInterface.VariableIndex(12345677), MathOptInterface.VariableIndex(12345674), MathOptInterface.VariableIndex(12345676), MathOptInterface.VariableIndex(12345675), MathOptInterface.VariableIndex(12345679)]) == Set(MathOptInterface.VariableIndex[MathOptInterface.VariableIndex(12345677), MathOptInterface.VariableIndex(12345676), MathOptInterface.VariableIndex(12345679)])
Stacktrace:
 [1] macro expansion at /home/coey/.julia/dev/MathOptInterface/src/Test/UnitTests/basic_constraint_tests.jl:211 [inlined]
 [2] macro expansion at /home/coey/julia/usr/share/julia/stdlib/v1.3/Test/src/Test.jl:1114 [inlined]
 [3] macro expansion at /home/coey/.julia/dev/MathOptInterface/src/Test/UnitTests/basic_constraint_tests.jl:209 [inlined]
 [4] macro expansion at /home/coey/julia/usr/share/julia/stdlib/v1.3/Test/src/Test.jl:1114 [inlined]
 [5] #basic_constraint_test_helper#39(::Bool, ::Bool, ::Bool, ::typeof(MathOptInterface.Test.basic_constraint_test_helper), ::MathOptInterface.Bridges.Constraint.SingleBridgeOptimizer{MathOptInterface.Bridges.Constraint.NormOneBridge{Float64,F,G,H} where H where G where F,MathOptInterface.Utilities.MockOptimizer{MathOptInterface.Utilities.Model{Float64}}}, ::MathOptInterface.Test.TestConfig, ::getfield(MathOptInterface.Test, Symbol("##16#17")), ::MathOptInterface.NormOneCone, ::Int64) at /home/coey/.julia/dev/MathOptInterface/src/Test/UnitTests/basic_constraint_tests.jl:196
ConstraintFunction: Test Failed at /home/coey/.julia/dev/MathOptInterface/src/Test/UnitTests/basic_constraint_tests.jl:211
  Expression: variables(func) == variables(constraint_function)
   Evaluated: Set(MathOptInterface.VariableIndex[MathOptInterface.VariableIndex(12345677), MathOptInterface.VariableIndex(12345674), MathOptInterface.VariableIndex(12345676), MathOptInterface.VariableIndex(12345675), MathOptInterface.VariableIndex(12345679)]) == Set(MathOptInterface.VariableIndex[MathOptInterface.VariableIndex(12345677), MathOptInterface.VariableIndex(12345676), MathOptInterface.VariableIndex(12345679)])
Stacktrace:
 [1] macro expansion at /home/coey/.julia/dev/MathOptInterface/src/Test/UnitTests/basic_constraint_tests.jl:211 [inlined]
 [2] macro expansion at /home/coey/julia/usr/share/julia/stdlib/v1.3/Test/src/Test.jl:1114 [inlined]
 [3] macro expansion at /home/coey/.julia/dev/MathOptInterface/src/Test/UnitTests/basic_constraint_tests.jl:209 [inlined]
 [4] macro expansion at /home/coey/julia/usr/share/julia/stdlib/v1.3/Test/src/Test.jl:1114 [inlined]
 [5] #basic_constraint_test_helper#39(::Bool, ::Bool, ::Bool, ::typeof(MathOptInterface.Test.basic_constraint_test_helper), ::MathOptInterface.Bridges.Constraint.SingleBridgeOptimizer{MathOptInterface.Bridges.Constraint.NormOneBridge{Float64,F,G,H} where H where G where F,MathOptInterface.Utilities.MockOptimizer{MathOptInterface.Utilities.Model{Float64}}}, ::MathOptInterface.Test.TestConfig, ::getfield(MathOptInterface.Test, Symbol("##18#19")), ::MathOptInterface.NormOneCone, ::Int64) at /home/coey/.julia/dev/MathOptInterface/src/Test/UnitTests/basic_constraint_tests.jl:196
Test Summary:                                                                    | Pass  Fail  Total
NormOne                                                                          |  146     2    148
  MathOptInterface.VectorOfVariables-MathOptInterface.NormOneCone                |   13           13
  MathOptInterface.VectorAffineFunction{Float64}-MathOptInterface.NormOneCone    |   14     1     15
    add_constraint                                                               |    7     1      8
      ConstraintName                                                             |    3            3
      ConstraintFunction                                                         |    1     1      2
      ConstraintSet                                                              |    1            1
    ListOfConstraintIndices                                                      |    1            1
    add_constraints                                                              |    2            2
    is_valid                                                                     |    1            1
    delete                                                                       |    2            2
  MathOptInterface.VectorQuadraticFunction{Float64}-MathOptInterface.NormOneCone |   14     1     15
    add_constraint                                                               |    7     1      8
      ConstraintName                                                             |    3            3
      ConstraintFunction                                                         |    1     1      2
      ConstraintSet                                                              |    1            1
    ListOfConstraintIndices                                                      |    1            1
    add_constraints                                                              |    2            2
    is_valid                                                                     |    1            1
    delete                                                                       |    2            2
  Test mock model                                                                |   15           15
  Test bridged model                                                             |   12           12
ERROR: LoadError: Some tests did not pass: 146 passed, 2 failed, 0 errored, 0 broken.
in expression starting at /home/coey/.julia/dev/MathOptInterface/test/Bridges/Constraint/norm_to_lp.jl:79
```
there are others ways to fix the issue probably, like canonicalizing in the basic constraint tests before getting the variables. @blegat what is your preferred fix?

blegat: You can do `removevariable` in the bridge like in:
https://github.com/JuliaOpt/MathOptInterface.jl/blob/0ed068ff8342cf2501305963c22708b0b18a4b80/src/Bridges/Constraint/slack.jl#L98

chriscoey: OK fixed

chriscoey:
@blegat please merge if you're satisfied

src/Bridges/Constraint/norm_to_lp.jl
blegat: Why are the tests passing if the line fails ?

chriscoey: my comment was 2 days ago and I've merged in changes from master. works now.

chriscoey:
tests passing. good to merge?

PR #823, opened 2019-08-09
Add Variable.VectorizeBridge

blegat:
Also fix a bug in `bridge_optimizer.jl` caught while adding the model tests.

### DIFF
--- a/src/Bridges/Variable/vectorize.jl
+++ b/src/Bridges/Variable/vectorize.jl
@@ -1,0 +1,110 @@
+"""
+    VectorizeBridge{T, S}
+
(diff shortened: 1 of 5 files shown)


src/Bridges/Variable/vectorize.jl
```diff
@@ -0,0 +1,110 @@
+"""
+    VectorizeBridge{T, S}
+
+Transforms a constrained variable in `scalar_set_type(S, T)` where
+`S <: VectorLinearSet` constrained variables in `S`.
```
mlubin: This isn't clear, could you give a concrete example?

PR #833, opened 2019-08-13
Add extra tests to nametest

odow:
Current tests weren't testing the edge-case where we rename variables/constraints.

ref https://github.com/JuliaOpt/Gurobi.jl/pull/216, https://github.com/JuliaOpt/GLPK.jl/pull/101

<s>Also, it doesn't look like the `modellike.jl` tests are tested?</s>


odow:
Bump.  Barring objections, I'll merge this tomorrow.

PR #837, opened 2019-08-15
Implement operate: Vector + and / constant

blegat:
See https://github.com/JuliaOpt/MathOptInterface.jl/pull/818/files#r314057544


chriscoey:
Thanks!

Issue #836, opened 2019-08-14
Fix SOC3 test


odow:
The termination status should accept `INFEASIBLE_OR_UNBOUNDED`, and instead of `if config.duals`, it should be `if config.infeas_certificates`.

https://github.com/JuliaOpt/MathOptInterface.jl/blob/bb800e909389ec62d00f1ecbfd8c4a549797b310/src/Test/contconic.jl#L499-L505

PR #839, opened 2019-08-15
Fix SOC3

guilhermebodin:
#836


src/Test/contconic.jl
```diff
@@ -499,8 +499,9 @@ function soc3test(model::MOI.ModelLike, config::TestConfig)
         @test MOI.get(model, MOI.TerminationStatus()) == MOI.INFEASIBLE
 
         @test MOI.get(model, MOI.PrimalStatus()) in (MOI.NO_SOLUTION,
-                                                     MOI.INFEASIBLE_POINT)
-        if config.duals
+                                                     MOI.INFEASIBLE_POINT,
+                                                     MOI.INFEASIBLE_OR_UNBOUNDED)
```
blegat: This should be added for `TerminationStatus`, not `PrimalStatus`

guilhermebodin: oops, hehehe

src/Test/contconic.jl
```diff
@@ -496,11 +496,12 @@ function soc3test(model::MOI.ModelLike, config::TestConfig)
 
         MOI.optimize!(model)
 
-        @test MOI.get(model, MOI.TerminationStatus()) == MOI.INFEASIBLE
+        @test MOI.get(model, MOI.TerminationStatus()) in (MOI.INFEASIBLE, 
+                                                          MOI.INFEASIBLE_OR_UNBOUNDED)
```
mlubin: @odow Why do we need to support `INFEASIBLE_OR_UNBOUNDED`? We can set the following parameters so that Gurobi doesn't return this status:
https://www.gurobi.com/documentation/8.0/refman/infunbdinfo.html
https://www.gurobi.com/documentation/8.0/refman/dualreductions.html#parameter:DualReductions

odow: Oooo. I didn't realize there was this parameter.

odow: I've updated Gurobi to use the parameter, and reverted the change here.

Issue #838, opened 2019-08-15
Bug in SOCtoPSDBridge


JiazhengZhu:
There is a bug in `Bridges/Constraints/soc_to_psd.jl` at line 19: [https://github.com/JuliaOpt/MathOptInterface.jl/blob/f5a10367c70aad3ff26adc062462498d0da41cec/src/Bridges/Constraint/soc_to_psd.jl#L19](url)

Here the constant vector for the resultant VAF was not filled in correctly. It seems that the constants from input VAF f was directly written into the beginning of this vector, rather than going into the entries corresponding to the vectorised matrix using `trimap`. For this part the code perhaps should be

```julia
constant = zeros(T, n)
    for i in 1:dim
        constant[trimap(i,1)] = f.constants[i]
    end
```
A non-working example would be

```julia
using SCS
using MathOptInterface
MOI = MathOptInterface
MOIU = MathOptInterface.Utilities
MOIU.@model(ModelData, (), (),
                         (MOI.Zeros, MOI.Nonnegatives, MOI.SecondOrderCone,
                          MOI.ExponentialCone, MOI.PositiveSemidefiniteConeTriangle),
                         (), (), (), (), (MOI.VectorAffineFunction,));
x = MOI.add_variables(MOIU.CachingOptimizer(MOIU.UniversalFallback(ModelData{Float64}()), SCS.Optimizer()), 3);
Fterms = MOI.ScalarAffineTerm.([0., 0., 0.], x);
F = MOI.VectorAffineFunction(MOI.VectorAffineTerm.([1, 2, 3], Fterms), [1., 2., 3.]);
g = MOI.ScalarAffineFunction(MOI.ScalarAffineTerm.([0., 0., 0.], x), 4.);
psd_constr = MOI.Bridges._SOCtoPSDaff(F, g);
psd_constr.constants
```
So the constants in the output VAF should be a vector representation of the upper triangular part of 
![](http://latex.codecogs.com/gif.latex?%5Cdpi%7B110%7D%20%5Cbegin%7Bpmatrix%7D%201%20%26%202%20%26%203%5C%5C%202%20%26%204%20%26%200%5C%5C%203%26%200%26%204%20%5Cend%7Bpmatrix%7D)
[...]

PR #840, opened 2019-08-16
Fix SOC to PSD bridge

blegat:
Closes https://github.com/JuliaOpt/MathOptInterface.jl/issues/838
@chriscoey Also adds `operate_output_index!` methods you need for https://github.com/JuliaOpt/MathOptInterface.jl/pull/818


chriscoey:
Thanks. After you merge this, I'll update #818.

Issue #849, opened 2019-08-23
map_indices(, ::MOI.NLPBlockData) is wrong


mlubin:
The issue in https://github.com/JuliaOpt/JuMP.jl/issues/2037 is that the variable indices are permuted 2 <-> 1 between JuMP and Ipopt.jl. The starting points therefore get swapped. The NLPBlock is unaware of the permutation of indices. Wrapping the `NLPBlock` to permute all the indices in the evaluation callbacks would be wasteful work. IMO we should instead guarantee that `default_copy_to` maintains the original ordering of the variables.

This is a blocker for JuMP 0.20.

Issue #846, opened 2019-08-22
Next-generation nonlinear


blegat:
We discussed with @ccoffrin, @harshangrjn and @kaarthiksundar about what the next-generation NLP could look like.
Here is a gist of the idea:
We create two new function types. First `NonlinearExpressionFunction` which will be used by JuMP to give the objective function and the constraints (as `NonlinearExpressionFunction`-in-`EqualTo`/`GreaterThan`/`LessThan`) to the MOI backend.
```julia
struct NonlinearExpressionFunction <: AbstractScalarFunction
    expression
end
```
Then `SecondOrderBlackBoxFunction` will be what NLP solvers such as Ipopt supports:
```julia
struct SecondOrderBlackBoxFunction <: AbstractScalarFunction
    evaluation_oracle::Function
    gradient_oracle::Function # May need to replace it by `append_to_jacobian_sparsity!` and `fill_constraint_jacobian!`
    hessian_oracle::Function
end
```
But NLP solvers such as [Alpine](https://github.com/lanl-ansi/Alpine.jl) can directly support `NonlinearExpressionFunction` as it exploits the expression structure.

We can then define the bridges `AffinetoSecondOrderBlackBox` and `QuadratictoSecondOrderBlackBoxBridge` which transforms respectively `ScalarAffineFunction` and `ScalarQuadraticFunction` into `SecondOrderBlackBoxFunctionBridge` (as is currently done in the Ipopt MOI wrapper for instance).

The transformation from `NonlinearExpressionFunction` to `SecondOrderBlackBoxFunction` can be achieved by a `AutomaticDifferentiationBridge{ADBackend}` which computes the callbacks using `ADBackend`.
We can move the current AD implementation form JuMP to `MOI.Utilities` into an `DefaultADBackend` and `full_bridge_optimizer` adds `AutomaticDifferentiationBridge{DefaultADBackend}`.
If the user wants to change the AD backend to `OtherADBackend`, there will be a JuMP function that internally removes `AutomaticDifferentiationBridge{DefaultADBackend}` and adds `AutomaticDifferentiationBridge{OtherADBackend}`.

@kaarthiksundar would be interested to work on this and we thought a good first step is to define `SecondOrderBlackBoxFunction` and create the bridges `AffinetoSecondOrderBlackBox` and `QuadratictoSecondOrderBlackBoxBridge`.

Let us know what you think !

ccoffrin:
[...]

PR #850, opened 2019-08-23
Don't change order of variables when NLPBlock is used

blegat:
Closes https://github.com/JuliaOpt/MathOptInterface.jl/issues/849

### DIFF
--- a/src/Utilities/copy.jl
+++ b/src/Utilities/copy.jl
@@ -267,6 +267,21 @@ function pass_constraints(
     end
 end
 
+function copy_free_variables(dest::MOI.ModelLike, idxmap::IndexMap, vis_src, copy_variables::Function)
+    if length(vis_src) != length(keys(idxmap.varmap))
+        vars = copy_variables(dest, length(vis_src) - length(idxmap.varmap))
+        i = 1
+        for vi in vis_src
+            if !haskey(idxmap.varmap, vi)
+                idxmap.varmap[vi] = vars[i]
+                i += 1
+            end
+        end
+        @assert i == length(vars) + 1
+        @assert length(vis_src) == length(idxmap.varmap)
+    end
+end
+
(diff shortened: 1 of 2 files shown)


blegat:
I have verified that it does fix the failure on `mle.jl` in JuMP examples:
https://travis-ci.org/JuliaOpt/JuMP.jl/jobs/575715644#L504

src/Utilities/copy.jl
```diff
@@ -292,24 +307,29 @@ function default_copy_to(dest::MOI.ModelLike, src::MOI.ModelLike, copy_names::Bo
...
-    if length(vis_src) != length(keys(idxmap.varmap))
-        # Copy free variables
-        variables_not_added = setdiff(Set(vis_src), keys(idxmap.varmap))
-        vars = MOI.add_variables(dest, length(variables_not_added))
-        for (vi, var) in zip(variables_not_added, vars)
-            idxmap.varmap[vi] = var
-        end
+    # The `NLPBlock` assumes that the order of variables does not change (#849)
+    if MOI.NLPBlock() in MOI.get(src, MOI.ListOfModelAttributesSet())
+        vector_of_variables_not_added = [
```
mlubin: Why do this only for `NLPBlock` and not always?

blegat: You mean first determine how each variable is going to be added (either free or constrained) and then add them in the same order (hence splitting the `add_variables` call in separate one for each contiguous block) so that they are added in the same order even if some variables are contrained ?

Note that with https://github.com/JuliaOpt/MathOptInterface.jl/issues/846, the need to not change the order of the variables might go away. The assumption that variables have indices 1 to n that is used by NLP in MOI does not match the rest of the MOI ecosystem and it will go away once `NonlinearExpressionFunction` implements `map_variables`.

Issue #841, opened 2019-08-16
Update MOI.Test to support any coefficient that is a subtype of Real


chriscoey:
Looking for an easy way to help contribute to MOI? 

Help us by updating the tests so that we can run them with a generic coefficient type, instead of just `Float64`.

## Problem

I am trying to test a solver that works on a `Model{T}` for any `T <: Real`, but I can only run the vast majority of tests when `T = Float64`.

## Solution 

Things have changed quite a lot since the discussion below. Here's the latest (August 2021):

Step 1: pick a file from https://github.com/jump-dev/MathOptInterface.jl/tree/master/src/Test

Step 2: pick a test which contains explicit `Float64`s. Here's one
https://github.com/jump-dev/MathOptInterface.jl/blob/f08c78cc776e36e92270d42912128543d131c025/src/Test/test_variable.jl#L180-L201

Step 3: rewrite it to use a type parameter `T` instead
```Julia
function test_variable_solve_with_upperbound(
    model::MOI.ModelLike,
    config::Config{T},
) where {T}
    x = MOI.add_variable(model)
    c1 = MOI.add_constraint(model, MOI.SingleVariable(x), MOI.LessThan(one(T)))
    @test x.value == c1.value
    c2 = MOI.add_constraint(model, MOI.SingleVariable(x), MOI.GreaterThan(zero(T)))
    @test x.value == c2.value
    f = MOI.ScalarAffineFunction([MOI.ScalarAffineTerm(T(2), x)], zero(T))
    MOI.set(model, MOI.ObjectiveFunction{typeof(f)}(), f)
    MOI.set(model, MOI.ObjectiveSense(), MOI.MAX_SENSE)
    _test_model_solution(
        model,
        config;
        objective_value = T(2),
        variable_primal = [(x, one(T))],
        constraint_primal = [(c1, one(T)), (c2, one(T))],
        constraint_dual = [(c1, -T(2)), (c2, zero(T))],
    )
    return
end
```

ericphanson:
@jiazhengzhu and I would be interested in this too! Could someone point us to how to go about generalizing the tests?

blegat:
It might also be a type parameter of `TestConfig` so that the types local variables of the test functions are known at compile time

ericphanson:
In #855 some decisions were made about the details of doing this, and I think the following was more-or-less settled on for how to update a set of tests:

* Change one test file per PR
[...]

PR #855, opened 2019-08-28
Generalize tests for `contlinear`

ericphanson:
I tried starting to address #841. I haven't incorporated @blegat's suggestion to use a type parameter for `TestConfig` but that should be easy to change. I run into a problem, however, when trying to run the `contlinear` tests with `BigFloat`s for SDPA-GMP, which is that `Float64`'s are used internally in a few places in MOI. For example,

https://github.com/JuliaOpt/MathOptInterface.jl/blob/6ac4e6d6ee47486062ab9a14816aca6d51328760/src/Bridges/Constraint/slack.jl#L20

causes
```julia
  MethodError: no method matching shift_constant(::MathOptInterface.EqualTo{Float64}, ::BigFloat)
```
in the test on line `1539` of `contlinear.jl`. I imagine this is why it's important to test with other numeric types if we want to use them with MOI! I think therefore something beyond changing the tests here is needed. Perhaps auditing the codebase for `Float64` usage and replacing it with an appropriate type? It looks easy to fix in the slack.jl#L20 case. I saw also `Float64`'s here:

https://github.com/JuliaOpt/MathOptInterface.jl/blob/6ac4e6d6ee47486062ab9a14816aca6d51328760/src/Bridges/bridge_optimizer.jl#L724

and I'm not sure what the replacement would be, because `SingleVariable`'s don't carry a type parameter.

Regarding the changes in this PR:

My technique for updating the file was lots of find-and-replace. There are four types of constructors which need to be parametrized (VAFs, SAFs, SATs, and VATs), and integers written as floating point numbers (`3.0`) can be regex replaced by finding `(\d*?_?\d+?)\.0` and replacing by `T($1)` (assuming `T` is the numeric type). I also first replaced `0.0`'s by `zero(T)` and `1.0` by `one(T)`, although I'm not sure if that's better than `T(0)` and `T(1)`. Then I searched for remaining decimal points and replaced them with fractions by hand (e.g. `T(1)/T(2)`).

I'm not sure if replacing the constant literals was desired; @odow suggested just checking that they promote correctly. But because I was running into lots of promotion-related errors when running the `contlinear` test with SDPA-GMP, I decided to replace them. I think that helped isolate which problems were coming from MOI's internal use of Float64 numbers, versus ones introduced in `contlinear.jl`.

### DIFF
--- a/src/Test/config.jl
+++ b/src/Test/config.jl
@@ -16,7 +16,7 @@ struct TestConfig{T <: Number}
         infeas_certificates::Bool = true, optimal_status = MOI.OPTIMAL,
         basis::Bool = false) where {T <: Number}
         new(atol, rtol, solve, query, modify_lhs, duals, infeas_certificates,
-            optimal_status, basis) 
+            optimal_status, basis)
     end
     TestConfig(;kwargs...) = TestConfig{Float64}(; kwargs...)
 end
--- a/src/Test/config.jl
+++ b/src/Test/config.jl
@@ -10,13 +10,14 @@ struct TestConfig
     # MOI.OPTIMAL or MOI.LOCALLY_SOLVED.
     optimal_status::MOI.TerminationStatusCode
     basis::Bool # can get variable and constraint basis status
+    number_type::Type{<:Number}
     function TestConfig(;
         atol::Float64 = 1e-8, rtol::Float64 = 1e-8, solve::Bool = true,
         query::Bool = true, modify_lhs::Bool = true, duals::Bool = true,
         infeas_certificates::Bool = true, optimal_status = MOI.OPTIMAL,
-        basis::Bool = false)
+        basis::Bool = false, number_type::Type{<:Number} = Float64)
         new(atol, rtol, solve, query, modify_lhs, duals, infeas_certificates,
-            optimal_status, basis)
+            optimal_status, basis, number_type)
     end
 end
 
--- a/src/Test/config.jl
+++ b/src/Test/config.jl
@@ -1,4 +1,4 @@
-struct TestConfig
+struct TestConfig{T <: Number}
     atol::Float64 # absolute tolerance for ...
     rtol::Float64 # relative tolerance for ...
     solve::Bool # optimize and test result
@@ -10,15 +10,15 @@ struct TestConfig
     # MOI.OPTIMAL or MOI.LOCALLY_SOLVED.
     optimal_status::MOI.TerminationStatusCode
     basis::Bool # can get variable and constraint basis status
-    number_type::Type{<:Number}
-    function TestConfig(;
+    function TestConfig{T}(;
         atol::Float64 = 1e-8, rtol::Float64 = 1e-8, solve::Bool = true,
         query::Bool = true, modify_lhs::Bool = true, duals::Bool = true,
         infeas_certificates::Bool = true, optimal_status = MOI.OPTIMAL,
-        basis::Bool = false, number_type::Type{<:Number} = Float64)
+        basis::Bool = false) where {T <: Number}
         new(atol, rtol, solve, query, modify_lhs, duals, infeas_certificates,
-            optimal_status, basis, number_type)
+            optimal_status, basis) 
     end
+    TestConfig(;kwargs...) = TestConfig{Float64}(; kwargs...)
 end
 
 """
--- a/src/Test/contlinear.jl
+++ b/src/Test/contlinear.jl
@@ -28,7 +28,7 @@ function linear1test(model::MOI.ModelLike, config::TestConfig{T}) where {T <: Nu
     v = MOI.add_variables(model, 2)
     @test MOI.get(model, MOI.NumberOfVariables()) == 2
 
-    cf = MOI.ScalarAffineFunction{T}(MOI.ScalarAffineTerm{T}.([one(T),one(T)], v), zero(T))
+    cf = MOI.ScalarAffineFunction{T}(MOI.ScalarAffineTerm{T}.([one(T), one(T)], v), zero(T))
     c = MOI.add_constraint(model, cf, MOI.LessThan(one(T)))
     @test MOI.get(model, MOI.NumberOfConstraints{MOI.ScalarAffineFunction{T},MOI.LessThan{T}}()) == 1
 
@@ -42,7 +42,7 @@ function linear1test(model::MOI.ModelLike, config::TestConfig{T}) where {T <: Nu
     @test MOI.get(model, MOI.NumberOfConstraints{MOI.SingleVariable,MOI.GreaterThan{T}}()) == 2
 
     # note: adding some redundant zero coefficients to catch solvers that don't handle duplicate coefficients correctly:
-    objf = MOI.ScalarAffineFunction{T}(MOI.ScalarAffineTerm{T}.([zero(T),zero(T),-one(T),zero(T),zero(T),zero(T)], [v; v; v]), zero(T))
+    objf = MOI.ScalarAffineFunction{T}(MOI.ScalarAffineTerm{T}.([zero(T), zero(T), -one(T), zero(T), zero(T), zero(T)], [v; v; v]), zero(T))
     MOI.set(model, MOI.ObjectiveFunction{MOI.ScalarAffineFunction{T}}(), objf)
     MOI.set(model, MOI.ObjectiveSense(), MOI.MIN_SENSE)
 
@@ -94,7 +94,7 @@ function linear1test(model::MOI.ModelLike, config::TestConfig{T}) where {T <: Nu
 
     # change objective to Max +x
 
-    objf = MOI.ScalarAffineFunction{T}(MOI.ScalarAffineTerm{T}.([one(T),zero(T)], v), zero(T))
+    objf = MOI.ScalarAffineFunction{T}(MOI.ScalarAffineTerm{T}.([one(T), zero(T)], v), zero(T))
     MOI.set(model, MOI.ObjectiveFunction{MOI.ScalarAffineFunction{T}}(), objf)
     MOI.set(model, MOI.ObjectiveSense(), MOI.MAX_SENSE)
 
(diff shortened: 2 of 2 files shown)


src/Test/config.jl
```diff
@@ -10,13 +10,14 @@ struct TestConfig
     # MOI.OPTIMAL or MOI.LOCALLY_SOLVED.
     optimal_status::MOI.TerminationStatusCode
     basis::Bool # can get variable and constraint basis status
+    number_type::Type{<:Number}
```
ericphanson: Parametrizing TestConfig would get rid of the abstract type here, which is probably desirable

ericphanson:
I parametrized `TestConfig`. I have two questions:

1. Should we use `one(T)` and `zero(T)` or `T(1)` and `T(0)`? If `one(T)` is really better, what about `100*one(T)` vs `T(100)`? I think in the `one(T)` version, we are allowing a custom numeric type to have some kind of strange multiplicative identity which is not `==1.0`, whereas in the `T(1)` picture we are assuming that we want something essentially `==1.0` and that we can construct our type from integers without loss of precision.
2. How do we test this? It seems like maybe we need to add tests that use this code with various numeric types. And do we need to add tests for the `TestConfig` object itself?

src/Test/contlinear.jl
```diff
@@ -333,12 +333,12 @@ function linear1test(model::MOI.ModelLike, config::TestConfig)
             @test MOI.get(model, MOI.DualStatus(1)) == MOI.FEASIBLE_POINT
             @test MOI.get(model, MOI.DualObjectiveValue()) ≈ 3 atol=atol rtol=rtol
 
-            @test MOI.get(model, MOI.ConstraintDual(), c) ≈ -1.5 atol=atol rtol=rtol
-            @test MOI.get(model, MOI.ConstraintDual(), c2) ≈ 0.5 atol=atol rtol=rtol
+            @test MOI.get(model, MOI.ConstraintDual(), c) ≈ -T(3)/T(2) atol=atol rtol=rtol
+            @test MOI.get(model, MOI.ConstraintDual(), c2) ≈ T(1)/T(2) atol=atol rtol=rtol
```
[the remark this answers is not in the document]
ericphanson: Ah that does sound better, I’ll change those

src/Test/config.jl
```diff
@@ -10,14 +10,15 @@ struct TestConfig
...
-    function TestConfig(;
+    function TestConfig{T}(;
         atol::Float64 = 1e-8, rtol::Float64 = 1e-8, solve::Bool = true,
         query::Bool = true, modify_lhs::Bool = true, duals::Bool = true,
         infeas_certificates::Bool = true, optimal_status = MOI.OPTIMAL,
-        basis::Bool = false)
+        basis::Bool = false) where {T <: Number}
         new(atol, rtol, solve, query, modify_lhs, duals, infeas_certificates,
-            optimal_status, basis)
+            optimal_status, basis)
```
blegat: Remove trailing space

blegat:
@odow `Complex{Rational}` is not a concrete type hence the inference failure. With `Complex{Rational{Int}}`, both give the same thing:
```julia
julia> f(x) = x * one(Complex{Rational{Int}})
f (generic function with 2 methods)

julia> g(x) = Complex{Rational{Int}}(x)
g (generic function with 2 methods)

julia> @code_warntype g(1)
Variables
  #self#::Core.Compiler.Const(g, false)
  x::Int64

Body::Complex{Rational{Int64}}
1 ─ %1 = Core.apply_type(Main.Rational, Main.Int)::Core.Compiler.Const(Rational{Int64}, false)
│   %2 = Core.apply_type(Main.Complex, %1)::Core.Compiler.Const(Complex{Rational{Int64}}, false)
│   %3 = (%2)(x)::Complex{Rational{Int64}}
└──      return %3

julia> @code_warntype g(1)
Variables
  #self#::Core.Compiler.Const(g, false)
  x::Int64

Body::Complex{Rational{Int64}}
1 ─ %1 = Core.apply_type(Main.Rational, Main.Int)::Core.Compiler.Const(Rational{Int64}, false)
│   %2 = Core.apply_type(Main.Complex, %1)::Core.Compiler.Const(Complex{Rational{Int64}}, false)
│   %3 = (%2)(x)::Complex{Rational{Int64}}
└──      return %3
```
I prefer `zero(T)` and `one(T)` over `T(0)` and `T(1)` since `one` and `zero` are more specific than `T(::Int)`. However, for other number `x`, `T(x)` seems better than `x * one(T)` or `x + zero(T)`.

blegat:
https://github.com/JuliaOpt/MathOptInterface.jl/blob/6ac4e6d6ee47486062ab9a14816aca6d51328760/src/Bridges/Constraint/slack.jl#L20
Should be `zero(T)`.
https://github.com/JuliaOpt/MathOptInterface.jl/blob/6ac4e6d6ee47486062ab9a14816aca6d51328760/src/Bridges/bridge_optimizer.jl#L724
That's trickier.
One solution would be to create a new function in the MOI.AbstractBridgeOptimizer API to get the MOI.Bridges.Constraint.ScalarFunctionizeBridge.
For `LazyBridgeOptimizer`, it will loop through the list of of bridges and return the first one with this type.

ericphanson:
> I prefer `zero(T)` and `one(T)` over `T(0)` and `T(1)` since `one` and `zero` are more specific than `T(::Int)`. However, for other numbers, say `x`, I agree that `T(x)` seems better than `x * one(T)` or `x + zero(T)`.

Ok, I reverted that to go back to `zero(T)` and `one(T)`.

So in summary (and to make it easy to do these changes for other tests):
* `one(T)` and `zero(T)` to replace `0.0` and `1.0`
* `T(n)` to replace `n.0` (integer `n`)
* `T(n//m)` to replace floats like `1.5`

Besides that, I added type parameters to `MOI.EqualTo`, `MOI.GreaterThan`, `MOI.LessThan`, and `MOI.ScalarAffineTerm`, `MOI.ScalarAffineFunction`, `MOI.VectorAffineTerm`, and `MOI.`VectorAffineFunction`.

To test the changes (which helped me find that `MOI.EqualTo` and the other similar constraints needed to be parametrized), I've been using the script
```
using Test

using MathOptInterface
const MOI = MathOptInterface
const MOIT = MOI.Test
const MOIU = MOI.Utilities
const MOIB = MOI.Bridges

import SDPA_GMP

function test_with_type(T)
    optimizer = SDPA_GMP.Optimizer{T}()
    MOI.set(optimizer, MOI.Silent(), true)

    @testset "SolverName" begin
        @test MOI.get(optimizer, MOI.SolverName()) == "SDPA-GMP"
    end

    @testset "supports_default_copy_to" begin
        @test MOIU.supports_allocate_load(optimizer, false)
        @test !MOIU.supports_allocate_load(optimizer, true)
    end

    # UniversalFallback is needed for starting values, even if they are ignored by SDPA
    cache = MOIU.UniversalFallback(MOIU.Model{T}())
    cached = MOIU.CachingOptimizer(cache, optimizer)
    bridged = MOIB.full_bridge_optimizer(cached, T)
    # test 1e-3 because of rsoc3 test, otherwise, 1e-5 is enough
    config = MOIT.TestConfig{T}(atol=1e-3, rtol=1e-3)

    @testset "MOIT.contlineartest" begin
        MOIB.remove_bridge(bridged, MOIB.Constraint.ScalarSlackBridge{T})
        MOIT.contlineartest(bridged, config, [
            "linear8a",
            "linear12"
        ])
    end

end

@testset "Tests with type $T" for T in (Float64, BigFloat)
    test_with_type(T)
end
```
It still has some errors on `BigFloat` runs, but I think those are due to the use of `Float64` in MOI itself.

src/Test/contlinear.jl
```diff
@@ -577,42 +577,42 @@ function linear4test(model::MOI.ModelLike, config::TestConfig)
...
+        @test MOI.get(model, MOI.VariablePrimal(), y) ≈ zero(T) atol=atol rtol=rtol
     end
 
     # Min  x - y
-    # s.t. 100.0 <= x
-    #               y <= -100.0
-    MOI.set(model, MOI.ConstraintSet(), c2, MOI.LessThan(-100.0))
+    # s.t. T(100) <= x
+    #               y <= -T(100)
+    MOI.set(model, MOI.ConstraintSet(), c2, MOI.LessThan{T}(-T(100)))
```
blegat: `MOI.LessThan(T(-100))` should be sufficient, no need to do `MOI.LessThan{T}`

blegat:
> which helped me find that MOI.EqualTo and the other similar constraints needed to be parametrized

Why is that ? It automatically picks the type of the argument:
```julia
julia> MOI.EqualTo(1)
MathOptInterface.EqualTo{Int64}(1)
```

ericphanson:
> > which helped me find that MOI.EqualTo and the other similar constraints needed to be parametrized
> 
> Why is that ? It automatically picks the type of the argument:
> 
> ```julia
> julia> MOI.EqualTo(1)
> MathOptInterface.EqualTo{Int64}(1)
> ```

Hmm, well it might be simply because after I parametrized the functions, things like:

https://github.com/JuliaOpt/MathOptInterface.jl/blob/111fba5622b1e8ce1055dce2e9f01b2ad14ad971/src/Test/contlinear.jl#L15

won't work unless I also change the constraint there. I didn't realize that I only needed to add the type parameter in some places though, and just find-and-replaced it everywhere.

chriscoey:
We may as well make obvious style fixes while making these changes. There should be spaces between elements of vectors, like `[one(T), zero(T)]` rather than `[one(T),zero(T)]`

blegat:
@ericphanson Indeed, it's needed when you just write the type but not when it's a constructor

ericphanson:
@blegat Ok, I changed it to only put type parameters when it is not used as a constructor. I'll update my little summary of changes.

ericphanson:
> We may as well make obvious style fixes while making these changes. There should be spaces between elements of vectors, like `[one(T), zero(T)]` rather than `[one(T),zero(T)]`

Ok, I made that specific fix (regex-find `(\[.+?),([^ ])`, replace `$1, $2`)

src/Test/contlinear.jl
```diff
@@ -1,7 +1,7 @@
 # Continuous linear problems
 
 # Basic solver, query, resolve
-function linear1test(model::MOI.ModelLike, config::TestConfig)
+function linear1test(model::MOI.ModelLike, config::TestConfig{T}) where {T <: Number}
```
blegat: Why adding `<: Number` here ? As `T` is a type param of `TestConfig`, it is already constrained to be `<: Number`

ericphanson: No big reason; I thought maybe it adds clarity to the reader about what T is for, but I agree it’s not needed. Should I take these off?

blegat: Yes, it's more consistent with the rest of the MOI code

ericphanson: Ok, will do

Issue #862, opened 2019-08-29
Optimizer printing is too verbose


ericphanson:
Sorry if this has already been discussed somewhere; I didn't see an issue. If I accidentally let some optimizer object print to the REPL (maybe with nested layers of caches and bridges), it prints a huge amount. E.g. running
```julia
using SDPA_GMP, Convex, Test
E12, E21 = ComplexVariable(2, 2), ComplexVariable(2, 2)
s1, s2 = [big"0.25" big"-0.25"*im; big"0.25"*im big"0.25"], Complex{BigFloat}[big"0.5" big"0.0"; big"0.0" big"0.0"]
p = Problem{BigFloat}(:minimize, real(tr(E12 * (s1 + 2 * s2) + E21 * (s2 + 2 * s1))), [E12 ⪰ 0, E21 ⪰ 0, E12 + E21 == Diagonal(ones(2)) ])
solve!(p, SDPA_GMP.Optimizer())
p.MOI_model
```
means that last line yields 162_723 characters dumped in the REPL (as measured by `show`ing to a buffer, converting to a string, and asking for the `length`). (That code requires `https://github.com/ericphanson/Convex.jl#MathOptInterface` and `https://github.com/ericphanson/SDPA_GMP.jl`). 

Could MOI overload `show` to provide some kind of nice output? Maybe we could even somehow print a nice description of the various layers. Or even just redirect to `Base.summary` which gives e.g.
```julia
julia> summary(p.MOI_model)
"MathOptInterface.Utilities.CachingOptimizer{MathOptInterface.AbstractOptimizer,MathOptInterface.Utilities.Model{BigFloat}}"
```
That doesn't seem too bad, and seems way more helpful than a massive wall of text.

odow:
Any PR to improve printing would be great! Printing the summary seems helpful.

In the solvers, I have overloaded `show` (e.g., [Gurobi](https://github.com/JuliaOpt/Gurobi.jl/blob/db91e1e9fabc3f62f30265780e6a577016ebba4e/src/MOI_wrapper.jl#L139)), but this doesn't work when Gurobi get's wrapped in a bridge.

ericphanson:
Ok great, glad you're open to it! I'll try a few things locally and post a PR (probably soon, but possibly not).

blegat:
This also annoyed me several times, showing `summary` by default seems reasonable.

PR #864, opened 2019-08-31
Make `ModelLike` printing less verbose

ericphanson:
Fixes https://github.com/JuliaOpt/MathOptInterface.jl/issues/862

This adds a default show method for `ModelLike` to just print the summary (by default the type of a struct, without printing any fields).  Instead of printing the summary directly though, it substitutes MOI, MOIU, MOIB, and MOIT for the relevant modules and submodules. I found this made the summaries easier to read since otherwise lots of the space is used for writing out the module names.

Then for most of the `ModelLike`s defined in MOI itself, I added a `show` method which prints (what I hope is) a relevant summary of the fields as well.

The same code which generated 160k characters in #862 now yields
```julia
julia> p.MOI_model
MOIU.CachingOptimizer{MOI.AbstractOptimizer,MOIU.Model{BigFloat}}
in state ATTACHED_OPTIMIZER
in mode MANUAL
with model cache MOIU.Model{BigFloat}
with optimizer MOIB.LazyBridgeOptimizer{MOIU.CachingOptimizer{SDPA_GMP.Optimizer{BigFloat},MOIU.UniversalFallback{MOIU.Model{BigFloat}}}}
  with 1 variable bridge
  with 8 constraint bridges
  with inner model MOIU.CachingOptimizer{SDPA_GMP.Optimizer{BigFloat},MOIU.UniversalFallback{MOIU.Model{BigFloat}}}
    in state ATTACHED_OPTIMIZER
    in mode AUTOMATIC
    with model cache MOIU.UniversalFallback{MOIU.Model{BigFloat}}
      fallback for MOIU.Model{BigFloat}
    with optimizer SDPA_GMP.Optimizer{BigFloat}
```

The indentation works by storing an `indent` parameter in the `IOContext`. That means if a solver implements a custom show, it should also use this parameter to make sure it respects indentation too. For example, if I add

```julia
function Base.show(io::IO, O::Optimizer)
    indent = " "^get(io, :indent, 0)
    summary(io, O)
    print(io, "(presolve=$(O.presolve), verbose=$(!O.silent))")
    print(io, "\n$(indent)using binary located at $(O.binary_path)")
end
```
to SDPA_GMP.jl, then the example from #862 becomes
```julia
julia> p.MOI_model
MOIU.CachingOptimizer{MOI.AbstractOptimizer,MOIU.Model{BigFloat}}
in state ATTACHED_OPTIMIZER
in mode MANUAL
with model cache MOIU.Model{BigFloat}
with optimizer MOIB.LazyBridgeOptimizer{MOIU.CachingOptimizer{SDPA_GMP.Optimizer{BigFloat},MOIU.UniversalFallback{MOIU.Model{BigFloat}}}}
  with 1 variable bridge
  with 8 constraint bridges
  with inner model MOIU.CachingOptimizer{SDPA_GMP.Optimizer{BigFloat},MOIU.UniversalFallback{MOIU.Model{BigFloat}}}
    in state ATTACHED_OPTIMIZER
    in mode AUTOMATIC
    with model cache MOIU.UniversalFallback{MOIU.Model{BigFloat}}
      fallback for MOIU.Model{BigFloat}
    with optimizer SDPA_GMP.Optimizer{BigFloat}(presolve=true, verbose=true)
      using binary located at /Users/eh540/.julia/dev/SDPA_GMP/deps/usr/bin/sdpa_gmp
```
which adds an extra line of text (about the binary path) at the right indentation level.


src/Bridges/lazy_bridge_optimizer.jl
```diff
@@ -45,6 +45,18 @@ function LazyBridgeOptimizer(model::MOI.ModelLike)
...
 end
 
+function Base.show(io::IO, B::LazyBridgeOptimizer)
+    s(n) = n == 1 ? "" : "s"
+    indent = " "^get(io, :indent, 0)
+    n_var = length(Variable.bridges(B))
+    n_con = length(Constraint.bridges(B))
+    MOIU._print(io, summary(B))
+    print(io, "\n$(indent)with $(n_var) variable bridge$(s(n_var))")
+    print(io, "\n$(indent)with $(n_con) constraint bridge$(s(n_con))")
```
blegat: You may be able to create a show for AbstractBridgeOptimizer by creating a function like `print_bridges_summary` that prints nothing for `Constraint.EmptyMap` and prints this line for `Constraint.Map`.

ericphanson: Sounds like a great idea :). Let me know what you think of the implementation in the commit I just pushed.

ericphanson:
> Very cool. Can you add some tests?

Thanks! Yes, that sounds like a great idea. I'm not 100% sure how to test this. I just pushed one test for a bridge optimizer, where I ran it once to get the output, then hardcoded that and test that I actually get that output. Is this the kind of thing you're thinking of? If so, I can add a couple more like that. Or if you have any other testing ideas I'm happy to hear them.

blegat:
> Is this the kind of thing you're thinking of?

Yes, except that we usually just use `sprint`. Is there a reason you are doing something more elaborate ?

ericphanson:
> we usually just use `sprint`. Is there a reason you are doing something more elaborate ?

I just didn’t know about `sprint`. I’ll switch to that. Are there similar tests already?

blegat:
Yes, in `test/errors.jl` and `test/Utilities/copy.jl`

src/Utilities/Utilities.jl
```diff
@@ -17,6 +17,14 @@ const VQF{T} = MOI.VectorQuadraticFunction{T}
 const VI = MOI.VariableIndex
 const CI{F,S} = MOI.ConstraintIndex{F,S}
 
+function _print(io::IO, s::AbstractString)
```
blegat: Give a more meaningful name like `print_with_acronym`

ericphanson: Sounds good, done

ericphanson:
🎉

PR #870, opened 2019-09-04
Test that added_... is implemented for concrete types and warn if typ…

blegat:
…es are missing in test_delete_...

### DIFF
--- a/test/Bridges/utilities.jl
+++ b/test/Bridges/utilities.jl
@@ -8,12 +8,12 @@ function warn_incomplete_list_num_constraints(BT, list_num_constraints)
     for (S,) in MOIB.added_constrained_variable_types(BT)
         F = MOIU.variable_function_type(S)
         if !any(c -> c[1] == F && c[2] == S, list_num_constraints)
-            @warn("Bridges of type $BT add constrained variable in $S but their number is not tested.")
+            error("Bridges of type $BT add constrained variable in $S but their number is not tested.")
         end
     end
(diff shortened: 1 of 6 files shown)


test/Bridges/utilities.jl
```diff
@@ -4,10 +4,25 @@ function test_num_constraints(bridged_mock, F, S, n)
     @test ((F, S) in MOI.get(bridged_mock, MOI.ListOfConstraints())) == !iszero(n)
 end
 
+function warn_incomplete_list_num_constraints(BT, list_num_constraints)
+    for (S,) in MOIB.added_constrained_variable_types(BT)
+        F = MOIU.variable_function_type(S)
+        if !any(c -> c[1] == F && c[2] == S, list_num_constraints)
+            @warn("Bridges of type $BT add constrained variable in $S but their number is not tested.")
```
odow: Warn or make this an error?

blegat: This is used only by us in tests, it's useful to have it as a warning so that we can see the rest of the test. The advantage of an error though is that it will fail in CI so an error might be better.

odow: > The advantage of an error though is that it will fail in CI so an error might be better.

This is what I was thinking.

blegat: It was indeed the right choice, it just had its first use case:
https://travis-ci.org/JuliaOpt/MathOptInterface.jl/builds/580800484

Issue #868, opened 2019-09-03
Add MOI tests for DualExponentialCone and DualPowerCone Problems


guilhermebodin:
This would help to tests the cone support for SCS and COSMO, maybe others too.

guilhermebodin:
https://github.com/JuliaOpt/SCS.jl/pull/154

blegat:
@guilhermebodin Is it resolved by https://github.com/JuliaOpt/MathOptInterface.jl/pull/873 ?

guilhermebodin:
Yes :)

PR #873, opened 2019-09-06
DualExponentialCone and DualPowerCone

guilhermebodin:
This is a start to add new tests for DualExponentialCone and DualPowerCone as in #868.

I still need to figure out problem solution and add to the test
- [x] Solution for Dual Exponential
- [x] Solution for Dual Power
- [x] Add Tests for DualExponential
- [x] Add Tests for DualPower


test/Test/contconic.jl
```diff
@@ -122,6 +122,24 @@ end
                           (MOI.VectorAffineFunction{Float64}, MOI.ExponentialCone)   => [[-1., log(5)-1, 1/5]])
     MOIT.exp3test(mock, config)
 end
+@testset "Dual Exponential" begin
+    mock.optimize! = (mock::MOIU.MockOptimizer) -> MOIU.mock_optimize!(mock, [-exp(1/2), -exp(1/2)/2, 1., 1 + exp(1/2), 1 + exp(1/2)/2],
+                          (MOI.ScalarAffineFunction{Float64}, MOI.EqualTo{Float64})  => [-1., -2., -2exp(1/2)])
```
blegat: `1.` should be `1.0`, see JuMP style guide

test/Test/contconic.jl
```diff
@@ -122,6 +122,24 @@ end
                           (MOI.VectorAffineFunction{Float64}, MOI.ExponentialCone)   => [[-1., log(5)-1, 1/5]])
     MOIT.exp3test(mock, config)
 end
+@testset "Dual Exponential" begin
+    mock.optimize! = (mock::MOIU.MockOptimizer) -> MOIU.mock_optimize!(mock, [-exp(1/2), -exp(1/2)/2, 1., 1 + exp(1/2), 1 + exp(1/2)/2],
```
blegat: Need spaces around `/` according to JuMP style guide

src/Test/contconic.jl
```diff
@@ -1580,6 +1669,94 @@ powtests = Dict("pow1v" => pow1vtest,
...
+
+    vov = MOI.VectorOfVariables(v)
+
+    if vecofvars
+        vc = MOI.add_constraint(model, vov, MOI.DualPowerCone(a));
+    else
+        vc = MOI.add_constraint(model, MOI.VectorAffineFunction{Float64}(vov), MOI.DualPowerCone(a))
+    end
+
+    cu = MOI.add_constraint(model, MOI.ScalarAffineFunction(MOI.ScalarAffineTerm.([1.0, 1.0], [x[1], v[1]]), 0.), MOI.EqualTo(0.))
```
blegat: `0.` should be `0.0` according to JuMP style guide

src/Test/contconic.jl
```diff
@@ -1580,6 +1669,94 @@ powtests = Dict("pow1v" => pow1vtest,
...
 @moitestset pow
 
+function _dualpow1test(model::MOI.ModelLike, config::TestConfig, vecofvars::Bool)
+    atol = config.atol
+    rtol = config.rtol
+    # Problem dual POW1
+    # min -2x_1 - x_2
+    #  st  x_1 + u == 0
+    #      x_2 + v == 0
+    #      w == -1
```
blegat: The equation seems to be `w == 1`

src/Test/contconic.jl
```diff
@@ -1580,6 +1669,94 @@ powtests = Dict("pow1v" => pow1vtest,
 
 @moitestset pow
 
+function _dualpow1test(model::MOI.ModelLike, config::TestConfig, vecofvars::Bool)
+    atol = config.atol
+    rtol = config.rtol
+    # Problem dual POW1
+    # min -2x_1 - x_2
```
blegat: If the objective here is `-x_1 - x_2` the problem is easy to solve by hand so it might be easier to debug and we have the analytical solution.
By the [Weighted AM–GM inequality](https://en.wikipedia.org/wiki/Inequality_of_arithmetic_and_geometric_means), you have
0.9a + 0.1b >= a^0.9 b^0.1
with equality iff a == b
here taking a = u/0.9 and b = v/0.1, we have
u + v >= (u/0.9)^0.9 (v/0.1)^0.1
with equality iff u/0.9 == v/0.1.
Here the best you can do is `u + v == 1` and for that inequality must hold so u = 9v hence you get `v = 0.1` and `u = 0.9`.
You can move `a` as argument of the function with default value `0.9` and use `u = a` and `v = 1 - a` in the solution adding the explanation of why this is the optimal solution.

Issue #820, opened 2019-08-07
Use Bridges.runtests in bridges tests


blegat:
See https://github.com/JuliaOpt/MathOptInterface.jl/pull/819#discussion_r311582930

Variables:
- [x] flip_sign_bridge.jl https://github.com/jump-dev/MathOptInterface.jl/pull/1896
- [x] free_bridge.jl
- [x] rsoc_to_soc.jl
- [x] rsoc_to_psd.jl
- [x] soc_to_rsoc.jl
- [x] vectorize.jl
- [x] zeros_bridge.jl

Constraint
 - [x] det.jl
 - [x] flip_sign.jl
 - [x] functionize.jl
 - [x] geomean.jl
 - [x] geomean_to_relentr.jl
 - [x] indicator_activate_on_zero.jl https://github.com/jump-dev/MathOptInterface.jl/pull/1897
 - [x] indicator_sos.jl https://github.com/jump-dev/MathOptInterface.jl/pull/1895
 - [x] interval.jl https://github.com/jump-dev/MathOptInterface.jl/pull/1895
 - [x] ltgt_to_interval.jl https://github.com/jump-dev/MathOptInterface.jl/pull/1895
 - [x] norm_infinity.jl https://github.com/jump-dev/MathOptInterface.jl/pull/1895
 - [x] norm_one.jl https://github.com/jump-dev/MathOptInterface.jl/pull/1895
 - [x] norm_spec_nuc_to_psd.jl
 - [x] quad_to_soc.jl
 - [x] relentr_to_exp.jl
 - [x] rsoc_soc.jl
 - [x] scalarize.jl https://github.com/jump-dev/MathOptInterface.jl/pull/1895
 - [x] semi_to_binary.jl
 - [x] slack.jl
 - [x] soc_rsoc.jl
 - [x] soc_to_nonconvex_quad.jl
 - [x] soc_to_psd.jl
 - [x] square.jl
 - [x] vectorize.jl
 - [x] zero_one.jl 

Objective
 - [x] functionize.jl
 - [x] slack.jl

odow:
This is greatly improved by https://github.com/jump-dev/MathOptInterface.jl/blob/8d4d9988498d377d379e9244513c2ba418d49419/test/Bridges/Constraint/soc_to_nonconvex_quad.jl#L140-L166. 

But we still need a solution for variable bridges.

odow:
[...]

PR #877, opened 2019-09-11
indicator to SOS bridge

matbesancon:
Based on the comment on indicator constraints in the [SCIP documentation](https://scip.zib.de/doc/html/cons__indicator_8h.php):
> This constraint is equivalent to a linear constraint `ax−s ≤ b` and an SOS1 constraint on z and s (at most one should be nonzero). In the indicator context we can, however, separate more inequalities.


matbesancon:
this should be good for review

matbesancon:
There seems to be a problem, but I haven't figured out what exactly. I still haven't figured how to check exactly the newly formulated problem

odow:
> I still haven't figured how to check exactly the newly formulated problem

Try saving to MathOptFormat?

matbesancon:
Ah yes thanks, I was looking for an excuse to try it :)

src/Bridges/Constraint/indicator.jl
```diff
@@ -51,3 +51,89 @@ function concrete_bridge_type(::Type{<:IndicatorActiveOnFalseBridge},
                               ::Type{MOI.IndicatorSet{MOI.ACTIVATE_ON_ZERO, S}}) where {F<:MOI.VectorAffineFunction, S<:MOI.AbstractScalarSet}
     return IndicatorActiveOnFalseBridge{Float64, F, S}
 end
+
+## Indicator constraint by SOS1 bridge
+
+"""
```
blegat: I would move it to another file `indicator_sos1.jl`

src/Bridges/Constraint/Constraint.jl
```diff
@@ -64,12 +64,12 @@ const RSOCtoPSD{T, OT<:MOI.ModelLike} = SingleBridgeOptimizer{RSOCtoPSDBridge{T}
...
 
 """
-    add_all_bridges(bridged_model, T::Type)
+    add_all_bridges(bridged_model, ::Type{T})
 
 Add all bridges defined in the `Bridges.Constraint` submodule to
 `bridged_model`. The coefficient type used is `T`.
 """
-function add_all_bridges(bridged_model, T::Type)
+function add_all_bridges(bridged_model, ::Type{T}) where {T}
```
blegat: Why this change ? Is it better like this ?

src/Bridges/Constraint/indicator.jl
```diff
@@ -51,3 +51,89 @@ function concrete_bridge_type(::Type{<:IndicatorActiveOnFalseBridge},
...
+The `IndicatorSOS1Bridge` replaces an indicator constraint of the following form:
+``z \\in \\mathbb{B}, z == 1 \\implies f(x) \\leq b`` with a SOS1 constraint:
+``z \\in \\mathbb{B}, w \\leq 0, f(x) + w \\leq b, SOS1(w, z)``.
+`GreaterThan` constraints are handled in a symmetric way:
+``z \\in \\mathbb{B}, z == 1 \\implies f(x) \\geq b`` is reformulated as:
+``z \\in \\mathbb{B}, w \\geq 0, f(x) + w \\geq b, SOS1(w, z)``.
+`EqualTo` constraints are handled without a bound constraint:
+``z \\in \\mathbb{B}, z == 1 \\implies f(x) == b`` is reformulated as:
+``z \\in \\mathbb{B}, w \\text{ free}, f(x) + w == b, SOS1(w, z)``.
+
```
blegat: Why not always add a free `w` so that it works for any indicator constraint (also `MOI.Interval`, ...) ?

src/Bridges/Constraint/Constraint.jl
matbesancon: `T::Type`, using the run-time value, which is a type.
`::Type{T}`: dispatching on `T`, which is a compile-time info

src/Bridges/Constraint/indicator.jl
matbesancon: Can we have a `free` fallback with all constraints but `LessThan`, `GreaterThan`, otherwise we add a free, which has to be re-bounded by two variables for `LP`& `MILP` solvers

src/Bridges/Constraint/Constraint.jl
blegat: Do you have a reference for that ? I thought I read that it recognizes this special case and it's equivalent

src/Bridges/Constraint/indicator.jl
blegat: Not sure I understood "which has to be re-bounded by two variables"

matbesancon: Sorry yes, if we create the variable `w` free, it will be replaced by `w+ >= 0, w- >= 0, w = w+ + w- ` instead of creating a bounded `w` right away

src/Bridges/Constraint/Constraint.jl
matbesancon: so it's a special parsing rule for the compiler? I wasn't aware of one

src/Bridges/Constraint/indicator.jl
blegat: That's also the case for SDP solvers with the `Variable.FreeBridge`. Which LP solver does not support free variables ? This transformation is done internally ?
For a solver supporting free variables such constraints may increase the size of the problem, we should probably have *informative* constraints but that's out of the scope of this PR.

matbesancon: > Which LP solver does not support free variables ? This transformation is done internally ?

All solvers (that I am aware of) support free variables, but the transformation needs to be done internally somehow. The simplex algorithm does require positive variables

src/Bridges/Constraint/Constraint.jl
blegat: Not sure but I think the compiler might treat `Type` variables differently than other variables. Having to write `::Type{T}` everytime an argument is a type instead of just `T` or `T::Type` seems like working around something a compiler might do itself

src/Bridges/Constraint/indicator.jl
blegat: If you want to use the standard form you have equality constraints and nonnegative variables but then the dual has free variables. Some solvers might solve the dual using the simplex in which case they need all variables to be free. Although that's the textbook simplex so I don't know whether solvers might use a more sophisticated one supporting mixing free and non-free ones.

matbesancon: > If you want to use the standard form you have equality constraints and nonnegative variables but then the dual has free variables.

My understanding is that the standard form is a form  `A x <= b, x >= 0` transformed in `Ax + s = b, x, s >= 0`. You need to have basic columns in the equality, so starting from a generic `A x = b`, it will need to change the variable system in `[A; -A] x <= [b; -b]` and then add slacks, in order to have as many basic columns as constraints.

test/Bridges/Constraint/indicator.jl
```diff
@@ -53,3 +53,92 @@ include("../utilities.jl")
         @test t.coefficient ≈ 1.0
     end
 end
+
+@testset "Indicator by SOS1" begin
+    # linear problem with indicator constraints
```
blegat: Can you also split the file in two so that we have the same files in `src` and `test` folders ?

test/Bridges/Constraint/indicator.jl
```diff
@@ -53,3 +53,92 @@ include("../utilities.jl")
         @test t.coefficient ≈ 1.0
     end
 end
+
```
blegat: We should also have a test with `basic_constraint_test` (see https://github.com/JuliaOpt/MathOptInterface.jl/issues/750) and a test against `indicator1`, `indicator2` or `indicator3` (https://github.com/JuliaOpt/MathOptInterface.jl/blob/98a0e472e1757db14b849bbba8fe84597856def2/src/Test/intlinear.jl#L609-L611) with `test_model_equals` (see https://github.com/JuliaOpt/MathOptInterface.jl/issues/820)

src/Bridges/Constraint/indicator.jl
matbesancon: done in last commit

test/Bridges/Constraint/indicator.jl
matbesancon: Done in last commit

matbesancon: of course not, it's much more fun when it all stays local :upside_down_face:

matbesancon: (just did)

test/Bridges/Constraint/indicator.jl
matbesancon: I didn't manage to make that work (yet)

matbesancon:
@blegat I can't figure out what is off here. Any tips on a debugging process for that?

blegat:
> I can't figure out what is off here. Any tips on a debugging process for that?

You mean for making `basic_constraint_tests` working ? What is the issue you are facing ? Can you share an error message ?

matbesancon:
The constraints defined in the bridge seem not to be activated.

matbesancon:
> You mean for making basic_constraint_tests working ?

Or even when testing the bridge with `Cbc`

blegat:
So the indicator constraint is added but the underlying model does not receive the SOS1 constraints ?
I would be good to add a test for passing the indicator test with a `Bridges.Constraint.IndicatortoSOS1`

matbesancon:
> So the indicator constraint is added but the underlying model does not receive the SOS1 constraints ?

Yes I assume that's it

>  I would be good to add a test for passing the indicator test with a Bridges.Constraint.IndicatortoSOS1

Agreed, that's what I was trying to figure out

src/Bridges/Constraint/indicator_sos.jl
```diff
@@ -0,0 +1,89 @@
...
+function bridge_constraint(::Type{IndicatorSOS1Bridge{T,BC,MaybeBC}}, model::MOI.ModelLike, f::MOI.VectorAffineFunction{T}, s::MOI.IndicatorSet{MOI.ACTIVATE_ON_ONE, BC}) where {T <: Real, BC, MaybeBC}
+    f_scalars = MOIU.eachscalar(f)
+    (w, bound_constraint) = _add_bound_constraint!(model, BC)
+    z = f_scalars[1].terms[1].variable_index
+    sos_vector = MOI.VectorOfVariables([w, z])
+    sos_constraint = MOI.add_constraint(model, sos_vector, MOI.SOS1{T}([0.4, 0.6]))
+    affine_func = copy(f_scalars[2])
+    push!(
+        affine_func.terms,
+        MOI.ScalarAffineTerm{T}(one(T), w)
```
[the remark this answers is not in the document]
matbesancon: done. I feel like this function is `MOIU.operate!` is not the best design, it's really obscure to read

src/Bridges/Constraint/indicator_sos.jl
```diff
@@ -0,0 +1,89 @@
...
+    return IndicatorSOS1Bridge{T, S, MOI.ConstraintIndex{MOI.SingleVariable, S}}
+end
+
+function concrete_bridge_type(::Type{<:IndicatorSOS1Bridge{T}},
+                              ::Type{<:MOI.VectorAffineFunction},
+                              ::Type{MOI.IndicatorSet{MOI.ACTIVATE_ON_ONE, S}}) where {T, S <: MOI.AbstractScalarSet}
+    return IndicatorSOS1Bridge{T, S, Nothing}
+end
+
+function concrete_bridge_type(::Type{<:IndicatorSOS1Bridge},
```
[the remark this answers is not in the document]
matbesancon: done

src/Bridges/Constraint/Constraint.jl
```diff
@@ -61,15 +61,16 @@ const RootDet{T, OT<:MOI.ModelLike} = SingleBridgeOptimizer{RootDetBridge{T}, OT
 include("soc_to_psd.jl")
 const SOCtoPSD{T, OT<:MOI.ModelLike} = SingleBridgeOptimizer{SOCtoPSDBridge{T}, OT}
 const RSOCtoPSD{T, OT<:MOI.ModelLike} = SingleBridgeOptimizer{RSOCtoPSDBridge{T}, OT}
-include("indicator.jl")
+include("indicator_activate_on_zero.jl")
+include("indicator_sos.jl")
```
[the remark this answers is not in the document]
matbesancon: done

src/Bridges/Constraint/indicator_sos.jl
```diff
@@ -0,0 +1,79 @@
...
+end
+
+function bridge_constraint(::Type{IndicatorSOS1Bridge{T,BC,MaybeBC}}, model::MOI.ModelLike, f::MOI.VectorAffineFunction{T}, s::MOI.IndicatorSet{MOI.ACTIVATE_ON_ONE, BC}) where {T <: Real, BC, MaybeBC}
+    f_scalars = MOIU.eachscalar(f)
+    (w, bound_constraint) = _add_bound_constraint!(model, BC)
+    z = convert(MOI.SingleVariable, f_scalars[1]).variable
+    sos_vector = MOI.VectorOfVariables([w, z])
+    sos_constraint = MOI.add_constraint(model, sos_vector, MOI.SOS1{T}([0.4, 0.6]))
+    affine_func = f_scalars[2]
+    MOIU.operate!(+, T, affine_func, MOI.SingleVariable(w))
```
blegat: If `f` is `VectorOfVariables`, `affine_func` is `SingleVariable` and it cannot hold the sum so to be safe, do `some_name = MOIU.operate!(...`

src/Bridges/Constraint/indicator_sos.jl
```diff
@@ -0,0 +1,89 @@
...
+function _add_bound_constraint!(model::MOI.ModelLike, ::Type{BC}) where {T <: Real, BC <: Union{MOI.LessThan{T}, MOI.GreaterThan{T}}}
+    return MOI.add_constrained_variable(model, BC(zero(T)))
+end
+
+function _add_bound_constraint!(model::MOI.ModelLike, ::Type{<:MOI.AbstractScalarSet})
+    return (MOI.add_variable(model), nothing)
+end
+
+function MOI.supports_constraint(::Type{<:IndicatorSOS1Bridge},
+                                 ::Type{<:MOI.VectorAffineFunction},
```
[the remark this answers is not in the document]
blegat: You don't have to check that the solver supports the added constraints. This is done automatically by the bridge optimizer using the `added_...` functions.

src/Bridges/Constraint/indicator_sos.jl
matbesancon: done

matbesancon:
Still no idea about the segfault on 1.0

blegat:
> Still no idea about the segfault on 1.0

You are probably just unlucky, I would recommend finishing the PR first and then see if the segfault is still there. It might disappear randomly just like it appeared.
For instance, addressing https://github.com/JuliaOpt/MathOptInterface.jl/pull/877#discussion_r341394561 will affect methods of `supports_constraint` so it might affect the segfault situation.

src/Bridges/Constraint/indicator_sos.jl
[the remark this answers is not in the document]
matbesancon: done

matbesancon:
re-opening, there was a HTTP connection error

matbesancon:
ok, back to square one, only Julia 1.0 is failing

matbesancon:
> src/Test/UnitTests/basic_constraint_tests.jl

Finally added it, I defined the corresponding `MOI.get` methods for the function and set, this should be good now

blegat:
We are still missing:

- [ ] Implements the getter for ConstraintPrimal/ConstraintPrimalStart and the setter for ConstraintPrimalStart and test them. You can see which bridges already implements it here: https://github.com/JuliaOpt/MathOptInterface.jl/issues/684
- [ ] Implements `MOI.delete` and run `test_delete_bridge`
- [ ] Run `test_model_equals` tests, see https://github.com/JuliaOpt/MathOptInterface.jl/issues/820

matbesancon:
Oh boy `test_delete_bridge` is horrible -_-

blegat:
What's the issue ?

matbesancon:
The signature is rather opaque, errors it yields hard to read

matbesancon:
Ok model equals is working, `test_delete_bridge` is way too hard to decipher for something required to implement bridges

blegat:
Does https://github.com/JuliaOpt/MathOptInterface.jl/pull/954 help ?

matbesancon:
It does yes. Sadly, the test is still failing, the number of variables is 3 at the beginning, 2 after deleting the bridge, all bridged constraints are of number 1, passing at the beginning, failing then

blegat:
The bridge does not seem to implement `ListOfVariableIndices`, `NumberOfVariables`, `ListOfConstraintIndices` and `NumberOfConstraints` which is one thing `test_delete_bridge` is testing.

matbesancon:
Ok good catch, I guess the fallback assumes 0?

matbesancon:
Tests should all be passing now

blegat:
> I guess the fallback assumes 0?

Yes :)

test/Bridges/Constraint/indicator_sos.jl
```diff
@@ -0,0 +1,168 @@
...
+    sos1: [w, z] in MathOptInterface.SOS1([0.4, 0.6])
+    wless: w <= 0.0
+    ineq: x + w <= 8.0
+    bin_cons: z in MathOptInterface.ZeroOne()
+    maxobjective: z
+    """
+    model = MOIU.Model{Float64}()
+    MOIU.loadfromstring!(model, s)
+    MOIU.test_models_equal(mock, model, var_names, ["sos1", "wless", "ineq", "bin_cons"])
+
```
blegat: We should also use `test_model_equals` on `bridged_mock`. It also tests that ListOfConstraintIndices, ... are implemented correctly as well as ConstraintFunction, ...

matbesancon: you mean testing against:
```julia
s = """
variables: z, x
indicator_cons: [z, 1.0 * x] in MathOptInterface.IndicatorSet{MathOptInterface.ACTIVATE_ON_ONE}(MathOptInterface.LessThan{Float64}(8.0))
bin_cons: z in MathOptInterface.ZeroOne()
maxobjective: z
"""

```

src/Bridges/Constraint/indicator_sos.jl
```diff
@@ -0,0 +1,185 @@
...
+    ::Type{<:IndicatorSOS1Bridge})
+    return true
+end
+
+function MOI.get(model::MOI.ModelLike, attr::MOI.ConstraintPrimal, bridge::IndicatorSOS1Bridge)
+    if attr.N == 1
+        return MOI.get(model, MOI.VariablePrimal(), bridge.z_variable_index)
+    end
+    wv = MOI.get(model, MOI.VariablePrimal(), bridge.w_variable_index)
+    lin_primal = MOI.get(model, MOI.ConstraintPrimal(), b.linear_constraint_index)
```
blegat: Replace `MOI.ConstraintPrimal()` by `attr`

src/Bridges/Constraint/indicator_sos.jl
```diff
@@ -0,0 +1,185 @@
...
+    ::MOI.ConstraintPrimalStart,
+    ::Type{<:IndicatorSOS1Bridge})
+    return true
+end
+
+function MOI.get(model::MOI.ModelLike, attr::MOI.ConstraintPrimal, bridge::IndicatorSOS1Bridge)
+    if attr.N == 1
+        return MOI.get(model, MOI.VariablePrimal(), bridge.z_variable_index)
+    end
+    wv = MOI.get(model, MOI.VariablePrimal(), bridge.w_variable_index)
```
blegat: This should be `MOI.VariablePrimal(attr.N)`

src/Bridges/Constraint/indicator_sos.jl
```diff
@@ -0,0 +1,185 @@
...
+
+function MOI.supports(
+    ::MOI.ModelLike,
+    ::MOI.ConstraintPrimalStart,
+    ::Type{<:IndicatorSOS1Bridge})
+    return true
+end
+
+function MOI.get(model::MOI.ModelLike, attr::MOI.ConstraintPrimal, bridge::IndicatorSOS1Bridge)
+    if attr.N == 1
```
blegat: Why do a special case for `N == 1` ? You should not assume anything on what are the different results and their order

src/Bridges/Constraint/indicator_sos.jl
matbesancon: how come? w is a scalar variable

src/Bridges/Constraint/indicator_sos.jl
matbesancon: This is getting the `MOI.ConstraintPrimal` for the function [z, a^T x], if this is one, it means getting the primal of a single variable z, otherwise, it's getting the primal of a Scalar Affine

matbesancon: For me this comes from the fact that we are using a vector function, instead of a tuple function: (SingleVariable, AbstractScalarFunction)

src/Bridges/Constraint/indicator_sos.jl
matbesancon: done

test/Bridges/Constraint/indicator_sos.jl
matbesancon: done

src/Bridges/Constraint/indicator_sos.jl
blegat: For the `N`th result, you have `VariablePrimal(N)`, `ConstraintPrimal(N)` and `ConstraintDual(N)`. The value of `ConstraintPrimal(N)` is the value of the function when the variables have values `VariablePrimal(N)`

src/Bridges/Constraint/indicator_sos.jl
blegat: This should return a vector of 2 values. `N` is simply the index of the result (i.e. usually it is 1). See `MOI.ResultCount`

src/Bridges/Constraint/indicator_sos.jl
matbesancon: ok, that's not really intuitive, I thought `ConstraintPrimal(N)` would be the primal of the `N`-th function in the vector function. I'll update that

src/Bridges/Constraint/indicator_sos.jl
matbesancon: so in that case there are the cases `N = 1` and `N = 2`, we can check one and deduce the other, or throw an error but this makes the functions heavier

blegat: Not sure to follow, you just need to make sure that the attributes you create in this function have the same result index. You don't assume anything on the nature of the result index.

matbesancon: oh ok, this is really not obvious eitheir, getting `attr.N` on a function value implies that you are getting the N-th member of a vector-valued function

blegat: Indeed, it is planned to be renamed, see https://github.com/JuliaOpt/MathOptInterface.jl/issues/960

matbesancon:
@blegat after thoughts, not sure `MOI.ConstraintPrimal` should be implemented. You cannot test it without a value set, and you cannot support setting a `ConstraintPrimal` without other things (see error below)

```julia
Getting primal attributes: Error During Test at /home/mbesancon/.julia/dev/MathOptInterface/test/Bridges/Constraint/indicator_sos.jl:179
  ArgumentError: `supports` is not defined for MathOptInterface.ConstraintPrimal(1), it is only defined for attributes such that `is_copyable` returns `true`.
  Stacktrace:
   [1] supports(::MathOptInterface.Utilities.Model{Float64}, ::MathOptInterface.ConstraintPrimal, ::Type{MathOptInterface.ConstraintIndex{MathOptInterface.ScalarAffineFunction{Float64},MathOptInterface.LessThan{Float64}}}) at /home/mbesancon/.julia/dev/MathOptInterface/src/attributes.jl:197
   [2] set(::MathOptInterface.Utilities.UniversalFallback{MathOptInterface.Utilities.Model{Float64}}, ::MathOptInterface.ConstraintPrimal, ::MathOptInterface.ConstraintIndex{MathOptInterface.ScalarAffineFunction{Float64},MathOptInterface.LessThan{Float64}}, ::Float64) at /home/mbesancon/.julia/dev/MathOptInterface/src/Utilities/universalfallback.jl:356

```

blegat:
If you set the variable values, the mock optimizer uses them to compute ConstraintPrimal using `get_fallback`. I dont't why this bridge should not support it.

matbesancon:
Alright `ConstraintPrimal` re-implemented and passing

src/Bridges/Constraint/indicator_sos.jl
```diff
@@ -0,0 +1,183 @@
...
+end
+
+function MOI.get(model::MOI.ModelLike, ::MOI.ConstraintPrimal, bridge::IndicatorSOS1Bridge)
+    zvalue = MOI.get(model, MOI.VariablePrimal(), bridge.z_variable_index)
+    wvalue = MOI.get(model, MOI.VariablePrimal(), bridge.w_variable_index)
+    lin_primal_start = MOI.get(model, MOI.ConstraintPrimal(), bridge.linear_constraint_index)
+    return [zvalue, lin_primal_start - wvalue]
+end
+
+function MOI.get(model::MOI.ModelLike, ::MOI.ConstraintPrimalStart, bridge::IndicatorSOS1Bridge)
```
blegat: I would merge this method and the one above into one as it is done in other bridges supporting stating values

matbesancon: In the case of `MOI.ConstraintPrimalStart`, it needs to query `MOI.VariablePrimalStart`, for `MOI.ConstraintPrimal`, it queries MOI.VariablePrimal`. Changing dispatch to reflection with `isa` is not gonna reduce complexity

blegat: Indeed, then you can keep separate methods but add `attr.N` as argument to `VariablePrimal` constructor and use `attr` instead of `MOI.ConstraintPrimal()` otherwise it won't work with multiple results (that are now exposed in JuMP with https://github.com/JuliaOpt/JuMP.jl/pull/2100)

src/Bridges/Constraint/indicator_sos.jl
```diff
@@ -158,26 +158,26 @@ function MOI.supports(
...
+    lin_primal_start = MOI.get(model, attr, bridge.linear_constraint_index)
     return [zvalue, lin_primal_start - wvalue]
 end
 
-function MOI.get(model::MOI.ModelLike, ::MOI.ConstraintPrimalStart, bridge::IndicatorSOS1Bridge)
-    zstart = MOI.get(model, MOI.VariablePrimalStart(), bridge.z_variable_index)
-    wstart = MOI.get(model, MOI.VariablePrimalStart(), bridge.w_variable_index)
-    lin_primal_start = MOI.get(model, MOI.ConstraintPrimalStart(), bridge.linear_constraint_index)
+function MOI.get(model::MOI.ModelLike, attr::MOI.ConstraintPrimalStart, bridge::IndicatorSOS1Bridge)
+    zstart = MOI.get(model, MOI.VariablePrimalStart(attr.N), bridge.z_variable_index)
```
blegat: The start attributes don't have any `result_index` field.

matbesancon: fixed in last commit

src/Bridges/Constraint/indicator_sos.jl
matbesancon: Fixed

PR #884, opened 2019-09-17
Add extension dictionary to MOIU.Model

odow:
I've found myself wanting this a few times. Particular in MathOptFormat to store options. 

We currently use a `UniversalFallback` with a new `AbstractModelAttribute`, but this requires some hacking:
https://github.com/odow/MathOptFormat.jl/blob/03a8d510aed483f1d83b8fc301da468d0ef82879/src/MPS/MPS.jl#L19-L29

There were three options:
1. What we currently do.
2. Defining a new model type in MathOptFormat and forwarding all `add_constraint` etc calls to it
3. Adding this extension dict.

Thoughts? (Mainly @blegat) Am I missing an easier way?


blegat:
1. Using the `UniversalFallback` is not appropriate here since it will also catch unsupported constraints and hide them from the inner model instead of throwing an error.
2. That would be the cleaner way to do it but it may be a lot of work compared to 3.

PR #887, opened 2019-09-19
Fix == for sets with mutable fields

blegat:
This was needed to have CDD pass `contlineartest` for `Rational{BigInt}`.
See also https://github.com/JuliaOpt/MathOptInterface.jl/issues/841

### DIFF
--- a/test/sets.jl
+++ b/test/sets.jl
@@ -18,6 +18,37 @@ end
 Base.copy(mlt::MutLessThan) = MutLessThan(Base.copy(mlt.upper))
 
 @testset "Sets" begin
+    @testset "==" begin
+        # By default, `==` redirects to `===`, it works for bits type
+        # but not for `BigInt`s. We define functions creating different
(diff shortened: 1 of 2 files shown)


test/sets.jl
```diff
@@ -18,6 +18,37 @@ end
...
+        a() = big(1)
+        b() = big(2)
+        @test a() !== a()
+        @test b() !== b()
+        for S in [MOI.LessThan, MOI.GreaterThan, MOI.EqualTo, MOI.PowerCone, MOI.DualPowerCone]
+            @test S(a()) == S(a())
+            @test S(a()) != S(b())
+            @test S(b()) == S(b())
+            @test S(b()) == S(b())
+            @test S(b()) == S(b())
```
schillic: :thinking:

schillic: Ah, this was on purpose, sorry.

blegat: Not really, I forgot to modify, thanks

PR #890, opened 2019-09-20
Add get_bounds to Utilities

joehuchette:
Computes variable lower/upper bounds set on a MOI variable

### DIFF
--- a/src/Utilities/variables.jl
+++ b/src/Utilities/variables.jl
@@ -4,21 +4,44 @@
 Return a tuple `(lb, ub)` of type `Tuple{T, T}`, where `lb` and `ub` are lower
  and upper bounds, respectively, imposed on `x` in `model`.
 """
-function get_bounds(model::MOI.ModelLike, ::Type{T}, x::MOI.VariableIndex) where {T}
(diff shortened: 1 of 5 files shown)


blegat:
What about `MOI.EqualTo`, if it's set we could return both bounds equal to the value of the set.

joehuchette:
But it's invalid to have an Interval AND a GreaterThan, correct?

odow:
> But it's invalid to have an Interval AND a GreaterThan, correct?

Yes. Here are the options:

 - Interval
 - LessThan
 - GreaterThan
 - EqualTo
 - LessThan AND GreaterThan
 - Semicontinuous
 - Semiinteger

src/Utilities/variables.jl
```diff
@@ -0,0 +1,48 @@
+"""
+    get_bounds(model::MOI.ModelLike, ::Type{T}, x::MOI.VariableIndex)
+
+Return a tuple `(lb, ub)` of type `Tuple{T, T}`, where `lb` and `ub` are lower
+ and upper bounds, respectively, imposed on `x` in `model`.
+"""
+function get_bounds(
```
blegat: It should be added somewhere to `docs/src/apireference.md`

PR #892, opened 2019-09-20
Add NumberOfThreads

remi-garcia:
I realized afterwards that I didn't need this feature. Since it's done I open this PR anyway in case you think that can be useful.

### DIFF
--- a/src/Test/UnitTests/attributes.jl
+++ b/src/Test/UnitTests/attributes.jl
@@ -61,8 +61,8 @@ function number_threads(model::MOI.ModelLike, config::TestConfig)
         @test MOI.supports(model, MOI.NumberOfThreads())
         # Get the current value to restore it at the end of the test
         value = MOI.get(model, MOI.NumberOfThreads())
-        MOI.set(model, MOI.NumberOfThreads(), 2)
-        @test MOI.get(model, MOI.NumberOfThreads()) == 2
+        MOI.set(model, MOI.NumberOfThreads(), 1)
+        @test MOI.get(model, MOI.NumberOfThreads()) == 1
         MOI.set(model, MOI.NumberOfThreads(), 3)
         @test MOI.get(model, MOI.NumberOfThreads()) == 3
         MOI.set(model, MOI.NumberOfThreads(), value)
--- a/src/Test/UnitTests/attributes.jl
+++ b/src/Test/UnitTests/attributes.jl
@@ -69,7 +69,7 @@ function number_threads(model::MOI.ModelLike, config::TestConfig)
         @test value == MOI.get(model, MOI.NumberThreads())
     end
 end
-unittests["time_limit_sec"] = time_limit_sec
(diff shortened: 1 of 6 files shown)


src/Test/UnitTests/attributes.jl
```diff
@@ -51,6 +51,26 @@ function time_limit_sec(model::MOI.ModelLike, config::TestConfig)
...
+        value = MOI.get(model, MOI.NumberThreads())
+        MOI.set(model, MOI.NumberThreads(), 2)
+        @test MOI.get(model, MOI.NumberThreads()) == 2
+        MOI.set(model, MOI.NumberThreads(), 3)
+        @test MOI.get(model, MOI.NumberThreads()) == 3
+        MOI.set(model, MOI.NumberThreads(), value)
+        @test value == MOI.get(model, MOI.NumberThreads())
+    end
+end
+unittests["time_limit_sec"] = time_limit_sec
```
blegat: `time_limit_sec` should be replaced by `number_threads`

src/attributes.jl
```diff
@@ -461,6 +461,14 @@ struct RawParameter <: AbstractOptimizerAttribute
     name::Any
 end
 
+"""
+    NumberThreads()
+
+An optimizer attribute for setting the number of threads used for an
+optimization. Values are integers and the default value is 1.
```
mlubin: 1 isn't the default for all solvers. I think many commercial solvers are multi-threaded by default. Is this intended to change the solver default?

remi-garcia: It was intended to change the solver default. I think we should have a default value so a model can be optimized with different solver with the same number of threads without any work. With that in mind 1 seemed the only choice.

blegat: We say: If `NumberOfThreads` is set to `nothing`, then it is unspecified (solver default or solver-specific attribute) and if it is set to an `Integer` then it overwrites any solver-specific attribute.

mlubin: > It was intended to change the solver default.

I don't support this. It's likely to lead to support questions like "why is JuMP slower than X?". 

> so model can be optimized with different solver with the same number of threads without any work

It's not much work at all to set `NumberOfThreads`.

+1 for `nothing` as the default value, which has no effect on the solver defaults. If `NumberOfThreads` is set to an integer then it overrides the default.

remi-garcia: > I don't support this. It's likely to lead to support questions like "why is JuMP slower than X?".

Ok, I change that

src/attributes.jl
```diff
@@ -461,6 +461,15 @@ struct RawParameter <: AbstractOptimizerAttribute
     name::Any
 end
 
+"""
+    NumberOfThreads()
+
+An optimizer attribute for setting the number of threads used for an
+optimization. When set to `nothing` use solver default. Values are positive
```
blegat: ` use` -> `, uses` ?

Issue #11, opened 2017-07-03
basis status should be specific to each result, hence numbered


chriscoey:
Currently there is only one basis status per variable and constraint, but if there are multiple results, then they could have different bases

chriscoey:
@mlubin is this resolved?

mlubin:
No, it's not resolved

PR #898, opened 2019-09-22
Add result index in ConstraintBasisStatus

blegat:
Closes https://github.com/JuliaOpt/MathOptInterface.jl/issues/11

### DIFF
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -776,13 +776,17 @@ ConstraintDual() = ConstraintDual(1)
 """
     ConstraintBasisStatus()
 
-A constraint attribute for the `BasisStatusCode` of some constraint, with
-respect to an available optimal solution basis.
+A constraint attribute for the `BasisStatusCode` of some constraint in result
+`N`, with respect to an available optimal solution basis. If `N` is omitted, it
+is 1 by default.
 
 **For the basis status of a variable, query the corresponding `SingleVariable`
 constraint that enforces the variable's bounds.**
 """
-struct ConstraintBasisStatus <: AbstractConstraintAttribute end
+struct ConstraintBasisStatus <: AbstractConstraintAttribute
+    N::Int
+end
+ConstraintBasisStatus() = ConstraintBasisStatus(1)
 
 """
     ConstraintFunction()
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -784,7 +784,7 @@ is 1 by default.
 constraint that enforces the variable's bounds.**
 """
 struct ConstraintBasisStatus <: AbstractConstraintAttribute
-    N::Int
+    result_index::Int
 end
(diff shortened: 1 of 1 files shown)


src/attributes.jl
```diff
@@ -776,13 +776,17 @@ ConstraintDual() = ConstraintDual(1)
...
+is 1 by default.
 
 **For the basis status of a variable, query the corresponding `SingleVariable`
 constraint that enforces the variable's bounds.**
 """
-struct ConstraintBasisStatus <: AbstractConstraintAttribute end
+struct ConstraintBasisStatus <: AbstractConstraintAttribute
+    N::Int
+end
+ConstraintDual() = ConstraintDual(1)
```
mlubin: `ConstraintBasisStatus`?

blegat:
> Are there any solvers that return multiple solutions with a basis for each? I'm not aware of any.

If they don't have the basis status for each solution, they should throw an error if the user get the basis status for a result that does not have the basis.

src/attributes.jl
```diff
@@ -776,13 +776,17 @@ ConstraintDual() = ConstraintDual(1)
...
+A constraint attribute for the `BasisStatusCode` of some constraint in result
+`N`, with respect to an available optimal solution basis. If `N` is omitted, it
+is 1 by default.
 
 **For the basis status of a variable, query the corresponding `SingleVariable`
 constraint that enforces the variable's bounds.**
 """
-struct ConstraintBasisStatus <: AbstractConstraintAttribute end
+struct ConstraintBasisStatus <: AbstractConstraintAttribute
+    N::Int
```
mlubin: This doesn't match the naming conventions for fields. What about `result_index`?

mlubin:
My concern is that we're creating a new code path that will almost never be used and is likely to be incorrectly implemented by all solvers. Do we have a way to test that solvers don't ignore the result_index field?

src/attributes.jl
blegat: It's `N` for `VariablePrimal`, `ConstraintPrimal` and `ConstraintDual` but `result_index` for `ObjectiveValue` and `DualObjectiveValue`. `result_index` is indeed better.

blegat:
This should be improved by https://github.com/JuliaOpt/MathOptInterface.jl/issues/900

src/attributes.jl
```diff
@@ -776,13 +776,17 @@ ConstraintDual() = ConstraintDual(1)
 """
     ConstraintBasisStatus()
 
-A constraint attribute for the `BasisStatusCode` of some constraint, with
-respect to an available optimal solution basis.
+A constraint attribute for the `BasisStatusCode` of some constraint in result
+`N`, with respect to an available optimal solution basis. If `N` is omitted, it
```
mlubin: s/`N`/`result_index`/

PR #901, opened 2019-09-22
Add tests for result count of solution attributes

### LINKED ISSUE #900: Error for `result_index` out of bound
We should have an error for `result_index` out of bound and unit tests that try to access results at index higher than `ResultCount` and check that the error is thrown.

odow:
Closes #900

### DIFF
--- a/src/Test/UnitTests/solve.jl
+++ b/src/Test/UnitTests/solve.jl
@@ -137,3 +137,23 @@ function solve_single_variable_dual_max(model::MOI.ModelLike, config::TestConfig
     end
 end
 unittests["solve_single_variable_dual_max"] = solve_single_variable_dual_max
+
+function solve_result_index(model::MOI.ModelLike, config::TestConfig)
+    MOI.empty!(model)
+    x = MOI.add_variable(model)
+    c = MOI.add_constraint(model, MOI.SingleVariable(x), MOI.GreaterThan(1.0))
+    MOI.set(model, MOI.ObjectiveSense(), MOI.MIN_SENSE)
+    MOI.set(model, MOI.ObjectiveFunction{MOI.SingleVariable}(), MOI.SingleVariable(x))
+    if config.solve
+        MOI.optimize!(model)
+        @test MOI.get(model, MOI.VariablePrimal(1), x) == 1.0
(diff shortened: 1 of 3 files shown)


src/Test/UnitTests/solve.jl
```diff
@@ -137,3 +137,27 @@ function solve_single_variable_dual_max(model::MOI.ModelLike, config::TestConfig
...
+    MOI.empty!(model)
+    x = MOI.add_variable(model)
+    c = MOI.add_constraint(model, MOI.SingleVariable(x), MOI.GreaterThan(1.0))
+    MOI.set(model, MOI.ObjectiveSense(), MOI.MIN_SENSE)
+    MOI.set(model, MOI.ObjectiveFunction{MOI.SingleVariable}(), MOI.SingleVariable(x))
+    if config.solve
+        MOI.optimize!(model)
+        @test MOI.get(model, MOI.ObjectiveValue(1)) == 1.0
+        @test_throws ErrorException MOI.get(model, MOI.ObjectiveValue(2))
+        @test MOI.get(model, MOI.DualObjectiveValue(1)) == 1.0
```
blegat: Shouldn't this be in the `if config.duals` ?

blegat:
We should define a custom error to enforce consistency between solvers wrappers, e.g.,
```julia
struct ResultIndexBoundsError{AttrType}
    attr::AttrType
    result_count::Int
end
function check_result_index_bounds(model::ModelLike, attr)
    result_count = get(model, ResultCount())
    if !(1 <= attr.result_index <= result_count)
        throw(ResultIndexBoundsError(attr, result_count)
    end
end
Base.showerror(...) = ...
```

odow:
Thoughts on making a breaking change by renaming the fields from `.N` to `.result_index`?

blegat:
> Thoughts on making a breaking change by renaming the fields from `.N` to `.result_index`?

That would break MosekTools. We could instead create a `result_index` function which could be cleaner than assuming a same field name.

rschwarz:
Is the idea here that solvers would also call `MOI.check_result_index_bounds` inside of the relevant `get` methods?

blegat:
Yes, exactly, all solvers should do that. They could also call `MOI.throw_if_not_valid` when it's a getter for an index

PR #902, opened 2019-09-23
Add Formats as a submodule

### LINKED ISSUE #28: A File-format for MathOpt
Here's a potentially naive question (considering I know nothing about conic): 

Is it possible to save any problem (based on the current standard form with linear-quadratic functions, and the current sets) in the CBF format? 

Because I was reading through the specification (incl. the new one http://cblib.zib.de/doc/format2.pdf) and it wasn't obvious to me how to have quadratic objectives/quadratic terms in the constraints.

It seems like if this is to be **the standard form** that unifies linear/conic etc, it would be nice to have one simple, easy to parse file-format that everybody agrees upon (unlike LP and MPS where nobody agrees on a canonical format and each reader/writer has weird limitations).

odow:
Move https://github.com/odow/MathOptFormat.jl to `src/Formats`.

I've kept the commit history so `git blame` still works.

- [ ] **Decision:** should we store a copy of the MOF JSONSchema, or download on `build MathOptInterface`?

Closes #28 

### Needs cherrypick

 - [ ] https://github.com/odow/MathOptFormat.jl/pull/85


mlubin:
"Serialization" or "FileFormats" would be a more standard name than "Formats".

> should we store a copy of the MOF JSONSchema, or download on build MathOptInterface?

What is the source of truth for the MOF JSONSchema?

odow:
> What is the source of truth for the MOF JSONSchema?

https://github.com/odow/MathOptFormat/releases

docs/src/file_formats.md
```diff
@@ -0,0 +1,121 @@
...
+```julia
+using JuMP
+
+jump_model = Model()
+@variable(jump_model, x, Int)
+@constraint(jump_model, my_con, 2x + 1 <= 2)
+@objective(jump_model, Max, x)
+
+mps_model = MOI.FileFormats.MPS.Model()
+MOI.copy_to(mps_model, backend(jump_model))
```
blegat: Isn't `MOI.copy_to(full_bridge_optimizer(mps_model, Float64))` needed in some cases ?

odow:
> But then why not simply usse the file joinpath(dirname(pathof(MOI)), "mof.schema.json")?

https://github.com/odow/MathOptFormat is not https://github.com/odow/MathOptFormat.jl :)

I'm moving toward adding the schema directly. It's more friction to upgrade the schema, which is probably a good thing.

odow:
Closing this since MOF has changed a little bit. I might make another attempt in the near future.

PR #906, opened 2019-09-29
Implements submit for CachingOptimizer and AbstractBridgeOptimizer

blegat:
This should allow callbacks to be used with JuMP in `AUTOMATIC` and `MANUAL` mode instead of `DIRECT` mode.
The only thing unsupported is submitting `HeuristicSolution` involving a bridged variable. We could try to determine the value of the solver variables given the value of the bridged variables by solving a linear system but let's keep it simple for now as currently the solvers with callbacks don't need variable bridges.

### DIFF
--- a/src/Bridges/bridge_optimizer.jl
+++ b/src/Bridges/bridge_optimizer.jl
@@ -1118,6 +1118,12 @@ function MOI.add_constrained_variable(b::AbstractBridgeOptimizer,
     end
 end
 
+function MOI.supports(b::AbstractBridgeOptimizer, sub::MOI.AbstractSubmittable)
+    return MOI.supports(b.model, sub)
+end
+function MOI.submit(b::AbstractBridgeOptimizer, sub::MOI.AbstractSubmittable, args...)
+    return MOI.submit(b.model, sub, bridged_function.(b, args)...)
+end
 
 """
     bridged_variable_function(b::AbstractBridgeOptimizer,
--- a/src/Utilities/cachingoptimizer.jl
+++ b/src/Utilities/cachingoptimizer.jl
@@ -575,17 +575,17 @@ end
 
 function MOI.get(m::CachingOptimizer, attr::AttributeFromOptimizer{T}) where {T <: MOI.AbstractModelAttribute}
     @assert m.state == ATTACHED_OPTIMIZER
-    return map_indices(m.optimizer_to_model_map,MOI.get(m.optimizer, attr.attr))
+    return map_indices(m.optimizer_to_model_map, MOI.get(m.optimizer, attr.attr))
 end
 
 function MOI.get(m::CachingOptimizer, attr::AttributeFromOptimizer{T}, idx::MOI.Index) where {T <: Union{MOI.AbstractVariableAttribute,MOI.AbstractConstraintAttribute}}
     @assert m.state == ATTACHED_OPTIMIZER
-    return map_indices(m.optimizer_to_model_map,MOI.get(m.optimizer, attr.attr, m.model_to_optimizer_map[idx]))
+    return map_indices(m.optimizer_to_model_map, MOI.get(m.optimizer, attr.attr, m.model_to_optimizer_map[idx]))
 end
 
 function MOI.get(m::CachingOptimizer, attr::AttributeFromOptimizer{T}, idx::Vector{<:MOI.Index}) where {T <: Union{MOI.AbstractVariableAttribute,MOI.AbstractConstraintAttribute}}
     @assert m.state == ATTACHED_OPTIMIZER
-    return map_indices(m.optimizer_to_model_map,MOI.get(m.optimizer, attr.attr, getindex.(m.model_to_optimizer_map,idx)))
+    return map_indices(m.optimizer_to_model_map, MOI.get(m.optimizer, attr.attr, getindex.(m.model_to_optimizer_map,idx)))
 end
 
 function MOI.set(m::CachingOptimizer, attr::AttributeFromModelCache{T}, v) where {T <: MOI.AbstractModelAttribute}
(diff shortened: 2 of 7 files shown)


odow:
```julia
(JuMP) pkg> st
    Status `/private/tmp/JuMP/Project.toml`
  [2e9cd046] Gurobi v0.7.2 #od/callbacks (https://github.com/JuliaOpt/Gurobi.jl.git)
  [4076af6c] JuMP v0.20.0 #master (https://github.com/JuliaOpt/JuMP.jl.git)
  [b8f27783] MathOptInterface v0.9.3 #bl/submit_cache_bridge (https://github.com/JuliaOpt/MathOptInterface.jl.git)

julia> using JuMP, Gurobi
[ Info: Recompiling stale cache file /Users/oscar/.julia/compiled/v1.1/JuMP/DmXqY.ji for JuMP [4076af6c-e467-56ae-b986-b466b2749572]
[ Info: Recompiling stale cache file /Users/oscar/.julia/compiled/v1.1/Gurobi/do9v6.ji for Gurobi [2e9cd046-0924-5485-92f1-d5272153d98b]

julia> model = Model(
           with_optimizer(
               Gurobi.Optimizer, 
               OutputFlag = 0,
               Cuts = 0,
               Presolve = 0,
               Heuristics = 0,
               LazyConstraints = 1
           )
       )
Academic license - for non-commercial use only
A JuMP Model
Feasibility problem with:
Variables: 0
Model mode: AUTOMATIC
Solver name: Gurobi

julia> @variable(model, 0 <= x <= 2.5, Int)
x

julia> @variable(model, 0 <= y <= 2.5, Int)
y

julia> @objective(model, Max, y)
y

julia> MOI.set(model, MOI.LazyConstraintCallback(), cb_data -> begin
           x_val = MOI.get(model, MOI.CallbackVariablePrimal(cb_data), x)
           y_val = MOI.get(model, MOI.CallbackVariablePrimal(cb_data), y)
           if y_val - x_val > 1 + 1e-6
               con = @build_constraint(y - x <= 1)
               MOI.submit(
                   model,
                   MOI.LazyConstraint(cb_data),
                   JuMP.moi_function(con.func),
                   con.set
               )
           elseif y_val + x_val > 3 + 1e-6
               con = @build_constraint(x + y <= 3)
               MOI.submit(
                   model,
                   MOI.LazyConstraint(cb_data),
                   JuMP.moi_function(con.func),
                   con.set
               )
           end
       end)
ERROR: MethodError: no method matching map_indices(::getfield(MathOptInterface.Utilities, Symbol("##7#8")){MathOptInterface.Utilities.IndexMap}, ::getfield(Main, Symbol("##3#4")))
Closest candidates are:
  map_indices(::AbstractDict{T<:Union{VariableIndex, ConstraintIndex},T<:Union{VariableIndex, ConstraintIndex}}, ::Any) where T<:Union{VariableIndex, ConstraintIndex} at /Users/oscar/.julia/packages/MathOptInterface/KWTd4/src/Utilities/functions.jl:73
  map_indices(::Function, ::Union{Nothing, AbstractString, DataType, Number, Tuple{Vararg{Union{Nothing, AbstractString, DataType, Number, AbstractConstraintAttribute, AbstractModelAttribute, AbstractOptimizerAttribute, AbstractSet, AbstractVariableAttribute, Enum},N} where N}, AbstractConstraintAttribute, AbstractModelAttribute, AbstractOptimizerAttribute, AbstractSet, AbstractVariableAttribute, AbstractArray{#s100,N} where N where #s100<:Union{Nothing, AbstractString, DataType, Number, Tuple{Vararg{Union{Nothing, AbstractString, DataType, Number, AbstractConstraintAttribute, AbstractModelAttribute, AbstractOptimizerAttribute, AbstractSet, AbstractVariableAttribute, Enum},N} where N}, AbstractConstraintAttribute, AbstractModelAttribute, AbstractOptimizerAttribute, AbstractSet, AbstractVariableAttribute, Enum}, Enum}) at /Users/oscar/.julia/packages/MathOptInterface/KWTd4/src/Utilities/functions.jl:80
  map_indices(::Function, ::MathOptInterface.VariableIndex) at /Users/oscar/.julia/packages/MathOptInterface/KWTd4/src/Utilities/functions.jl:82
  ...
Stacktrace:
 [1] map_indices(::MathOptInterface.Utilities.IndexMap, ::Function) at /Users/oscar/.julia/packages/MathOptInterface/KWTd4/src/Utilities/functions.jl:73
 [2] set(::MathOptInterface.Utilities.CachingOptimizer{MathOptInterface.AbstractOptimizer,MathOptInterface.Utilities.UniversalFallback{MathOptInterface.Utilities.Model{Float64}}}, ::MathOptInterface.LazyConstraintCallback, ::Function) at /Users/oscar/.julia/packages/MathOptInterface/KWTd4/src/Utilities/cachingoptimizer.jl:533
 [3] set(::Model, ::MathOptInterface.LazyConstraintCallback, ::Function) at /Users/oscar/.julia/packages/JuMP/SPKM5/src/JuMP.jl:678
 [4] top-level scope at none:0
```

odow:
So I added:
```julia
MOI.Utilities.map_indices(::MOI.Utilities.IndexMap, f::Function) = f
```
But the problem is how to map `x` inside the callback.

mlubin:
This seems pretty dangerous (i.e., prone to silently doing the wrong thing) given how the bridges renumber the variables. If an index is not correctly mapped, the user won't necessarily know about it except from incorrect results.

odow:
> If an index is not correctly mapped

Isn't that just a bug though?

blegat:
> This seems pretty dangerous (i.e., prone to silently doing the wrong thing) given how the bridges renumber the variables. If an index is not correctly mapped, the user won't necessarily know about it except from incorrect results.

If something does not implement `substitute_variables`, there will be a MethodError, it won't silently do something wrong.
An attribute of submittable value should implement `map_variables` to be usable by `CachingOptimizer` or `substitute_variables` to be usable by `AbstractBridgeOptimizer`, otherwise you get a method error telling you exactly that you should implement this function.

odow:
For the JuMP integration, can you please add:
```julia
MOI.Utilities.map_indices(::MOI.Utilities.IndexMap, f::Function) = f
MOI.is_set_by_optimize(::MOI.CallbackVariablePrimal) = true
```

blegat:
> For the JuMP integration, can you please add:
> ```julia
> MOI.Utilities.map_indices(::MOI.Utilities.IndexMap, f::Function) = f
> MOI.is_set_by_optimize(::MOI.CallbackVariablePrimal) = true
> ```

Should be done in the last commit. Strictly speaking, we should do something like
```julia
function MOI.Utilities.map_indices(d::MOI.Utilities.IndexMap, f::Function)
    rd = reverse_dict(d)
    (args...) -> f(map_indices.(Ref(rd), args)...)
end
```
in case one of the arguments of `f` is an index but this is not needed since `callback_data` does not contain any index.
As it would be costly to compute `reverse_dict`, we should maybe pass it to `map_indices` since the caching optimizer already has it but then it would be breaking.
Therefore, I suggest to just say that when functions are given as values of submittable or attributes then they will speak in the index-space of the solver and are not `index(x)` where `x` is the JuMP variables.

PR #912, opened 2019-10-07
PrimalStatus/DualStatus should be NO_SOLUTION if `result_index` is out of bounds

blegat:
It's difficult to pass both `MOIT.solve_result_index` and `MOIT.default_status_test` without this change.

### DIFF
--- a/src/Test/UnitTests/solve.jl
+++ b/src/Test/UnitTests/solve.jl
@@ -152,14 +152,14 @@ function solve_result_index(model::MOI.ModelLike, config::TestConfig)
         @test MOI.get(model, MOI.ObjectiveValue(1)) ≈ 1.0 atol=atol rtol=rtol
         @test_throws result_err(MOI.ObjectiveValue(2)) MOI.get(model, MOI.ObjectiveValue(2))
         @test MOI.get(model, MOI.PrimalStatus(1)) == MOI.FEASIBLE_POINT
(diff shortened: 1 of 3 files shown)


odow:
I thought about this, but if `TERMINATION_STATUS` is `TIME_LIMIT`, isn't it reasonable for solvers to return `NO_SOLUTION` for `PrimalStatus(1)` and `DualStatus(1)`.

Otherwise when people call `primal_status(model)` in JuMP, it'll throw an error.

blegat:
> Otherwise when people call `primal_status(model)` in JuMP, it'll throw an error.

That is the expected behavior. It may sound weird if we are not interpreting `primal_status` as "The status of the first result" but that's the semantic the users should get accustomed to.

odow:
I'm actually going to argue for `PrimalStatus` and `DualStatus` to return `NO_SOLUTION` for _all_ `result_index` if a a solution isn't present, regardless of the `ResultCount`. 

So `PrimalStatus(100)` would return `NO_STATUS`. I think that reads well.

blegat:
That sounds good, then this PR should be closed.

odow:
Or amended with an updated docstring and tests for the behavior I discussed

mlubin:
I agree with odow@. This is more user friendly, I and can't think of any negative side effects.

PR #913, opened 2019-10-08
Add complementarity constraints

### LINKED ISSUE #771: Complementarity constraint
We discussed with @frapac how to create a complementarity constraint type.
One idea would be to add a type
```julia
struct Complementarity <: MOI.AbstractVectorFunction end
MOI.dimension(::Complementarity) = 2
```
And the constraint `[x, y]`-in-`Complementarity` would constraint `x, y` to be nonnegative and `x * y` to be zero.

odow:
This PR adds the set `Complements(dimension)` which defines a mixed complementarity constraint.

The main targets are 
 - [KNITRO.jl](https://github.com/JuliaOpt/KNITRO.jl), which will implement `VectorOfVariables-in-Complements` (cc @frapac)
 - [PATH.jl](https://github.com/odow/PATH.jl), which will implement `VectorQuadraticFunction-in-Complements` and `VectorAffineFunction-in-Complements`, implementing both allows us to short-circuit the Jacobian evaluation the affine case. `VectorOfVariables-in-Complements` will be handled by bridges.

Closes #771 

I've left more obscure complementarity constraints, like conic constrained `x` compared to `lb <= x <= ub`, un-mentioned. @frapac does KNITRO just look at the bounds? Or does it do something different?

cc @xhub

### DIFF
--- a/src/Test/nlp.jl
+++ b/src/Test/nlp.jl
@@ -473,7 +473,8 @@ function test_qp_complementarity_constraint(
         @test isapprox(
             x_val,
             [1.0, 0.0, 3.5, 0.0, 0.0, 0.0, 3.0, 6.0],
-            atol = config.atol, rtol = config.rtol
+            atol = config.atol,
+            rtol = config.rtol
         )
         @test isapprox(
             MOI.get(model, MOI.ObjectiveValue()),
--- a/src/Test/nlp.jl
+++ b/src/Test/nlp.jl
@@ -337,3 +337,44 @@ const nlptests = Dict("hs071" => hs071_test,
                       )
 
 @moitestset nlp
+
+function test_linear_mcp(model::MOI.ModelLike, config::TestConfig)
+    MOI.empty!(model)
+    x = MOI.add_variables(model, 4)
+    MOI.add_constraint.(model, MOI.SingleVariable.(x), MOI.Interval(0.0, 10.0))
+    MOI.set.(model, MOI.VariablePrimalStart(), x, 0.0)
+    M = Float64[0  0 -1 -1; 0  0  1 -2; 1 -1  2 -2; 1  2 -2  4]
+    q = [2; 2; -2; -6]
+    terms = MOI.VectorAffineTerm{Float64}[]
+    for i = 1:4
+        push!(
+            terms,
+            MOI.VectorAffineTerm(4 + i, MOI.ScalarAffineTerm(1.0, x[i]))
+        )
+        for j = 1:4
+            iszero(M[i, j]) && continue
+            push!(
+                terms,
+                MOI.VectorAffineTerm(1, MOI.ScalarAffineTerm(M[i, j], x[j]))
+            )
+        end
+    end
+    MOI.add_constraint(
+        model,
+        MOI.VectorAffineFunction(terms, [q; 0.0; 0.0; 0.0; 0.0]),
+        MOI.Complements(4)
+    )
+    @test MOI.get(model, MOI.TerminationStatus()) == MOI.OPTIMIZE_NOT_CALLED
+    MOI.optimize!(model)
+    @test MOI.get(model, MOI.TerminationStatus()) == MOI.LOCALLY_SOLVED
+    x_val = MOI.get.(model, MOI.VariablePrimal(), x)
+    @test isapprox(
+        x_val, [2.8, 0.0, 0.8, 1.2], atol = config.atol, rtol = config.rtol
+    )
+end
+
+const complementaritytests = Dict(
+    "linear_mcp" => test_linear_mcp,
+)
+
+@moitestset complementarity
--- a/src/Test/nlp.jl
+++ b/src/Test/nlp.jl
@@ -361,7 +361,7 @@ function test_linear_mixed_complementarity(model::MOI.ModelLike, config::TestCon
             iszero(M[i, j]) && continue
             push!(
                 terms,
-                MOI.VectorAffineTerm(1, MOI.ScalarAffineTerm(M[i, j], x[j]))
+                MOI.VectorAffineTerm(i, MOI.ScalarAffineTerm(M[i, j], x[j]))
             )
         end
     end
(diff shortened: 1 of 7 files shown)


src/Test/nlp.jl
```diff
@@ -337,3 +337,44 @@ const nlptests = Dict("hs071" => hs071_test,
                       )
 
 @moitestset nlp
+
+function test_linear_mcp(model::MOI.ModelLike, config::TestConfig)
```
frapac: Maybe add the expression of the optimization problem there?

blegat: And the meaning of the acronym MCP

src/Test/nlp.jl
```diff
@@ -337,3 +337,44 @@ const nlptests = Dict("hs071" => hs071_test,
...
+        end
+    end
+    MOI.add_constraint(
+        model,
+        MOI.VectorAffineFunction(terms, [q; 0.0; 0.0; 0.0; 0.0]),
+        MOI.Complements(4)
+    )
+    @test MOI.get(model, MOI.TerminationStatus()) == MOI.OPTIMIZE_NOT_CALLED
+    MOI.optimize!(model)
+    @test MOI.get(model, MOI.TerminationStatus()) == MOI.LOCALLY_SOLVED
```
blegat: Should be `config.optimal_status`

src/sets.jl
```diff
@@ -669,18 +669,53 @@ end
...
+
+If the variable `x_i` is constrained in `Interval(lb, ub)`, then mathematically,
+the mixed complementarity problem is to find a solution such that at least one
+of the following holds:
+
+  1.  F_i(x) = 0, lb <= x_i <= ub_i
+  2.  F_i(x) > 0, lb == x_i
+  3.  F_i(x) < 0,       x_i == ub_i
+
+Classically, the bounding set for `x_i` is `Interval(0, Inf)`, which recovers:
```
blegat: What should the user do if the lower bound is not zero of the upper bound is not `Inf` ?

blegat:
If at some point we define `VariationalInequality(C)` that means  as @xhub suggested then `Complements(n)` will simply be `VariationalInequality(Nonnegatives(n))`.
What should we do then ?

1. Define `const Complements = VariationalInequality{Nonnegatives}`
2. Define a bridge mapping `VariationalInequality{Nonnegatives}` to `Complements` and vice versa

In MOI, we don't currently have a set for which the type is a special case of the type of another one so I think 1. would make more sense.
If we go with 1., we should make sure we don't get any backward compatibility issue by transforming `Complements` from a `struct` to a `const`.

src/sets.jl
odow:
```Julia
@variable(model, -1 <= x <= 1)
@constraint(model, [2 * x + 1, x] in Complements(1))
```
defines
```
2 * x + 1 = 0 or
2 * x + 1 > 0 if x == -1
2 * x + 1 < 0 if x == 1
```
which means the only solution is `2 * x + 1 = 0` so `x = -0.5`.

odow:
Note that our definition of complements isn't `VariationalInequality{Nonnegatives}`, it's really `VariationalInequality(Hypercube(lb, ub))` so I think a bridge is the way to go. 

> In MOI, we don't currently have a set for which the type is a special case of the type of another one 

Sure we do: `Integer` and `ZeroOne`, `GreaterThan` and `Interval`.

odow:
Really, there is nothing stopping us from interpreting the current `Complements` as `VariationalInequality`. 

The current definition is just a relation between `F(x)` and `x` with the bounds added separately. So the cone `C` can just be inferred from the constraints on `x`. Maybe we should change the docstring to reflect this.

blegat:
> The current definition is just a relation between `F(x)` and `x` with the bounds added separately.

Bridging currently assume that we can treat each constraint independently so it will be easy to bridge `VariationalInequality` to `Complements` but not `Complements` to `VariationalInequality` as you would need a mechanism for the bridge to say to the bridge optimizer to notify him when the bounds on the variable changes (we would also need something similar for the QuadtoSOC bridge to support `x^2 + y^2 <= t^2` as we need to be sure that there is a bound on `t` so maybe we could add that at some point.

I prefer `VariationalInequality(Hyperrectangle(lb, ub))` over `Complements` as all the information is inside the constraint and it does not depend on other constraint.
However, if `Complements` is a close match to Knitro and PATH interface then it makes sense to have both. It's kind of similar to Quad vs SOC. SOC (resp. VI) are preferred because they are easy to bridge into Quad (resp. Complements) and the other side depends on variable bounds.

src/sets.jl
blegat: Reading the docstring again, it seems clear and the reason I didn't understand it is due to a wrong preconception. Maybe adding the `SingleVariable`-in-`Interval` constraint in the Examples section would help.

odow:
> Bridging currently assume that we can treat each constraint independently so it will be easy to bridge VariationalInequality to Complements but not Complements to VariationalInequality

Sure. I envisaged a one-way `VariationalInequality` to `Complements` bridge.

Since we have two solvers that currently support `Complements` as written, we could keep `Complements` for now. If, in the future, we have multiple solvers that support more general variational inequalities (e.g., PATH-VI), then we can revisit this discussion, probably _after_ those solvers have experimented with the syntax as external sets.

blegat:
We also need to add it to `docs/src/apireference.md`

odow:
Before merging, we should wait for @xhub to sign-off.

frapac:
> @frapac does KNITRO just look at the bounds? Or does it do something different?

Once complementarity constraints defined inside Knitro, the solver forces the lower-bounds of the variables to be equal to 0.

blegat:
> Once complementarity constraints defined inside Knitro, the solver forces the lower-bounds of the variables to be equal to 0.

So Knitro.jl will throw an error of the variable lower are nonzero and if the upper bound is not +infinity ?

src/Test/nlp.jl
```diff
@@ -337,3 +337,142 @@ const nlptests = Dict("hs071" => hs071_test,
...
+    for i = 1:4
+        push!(
+            terms,
+            MOI.VectorAffineTerm(4 + i, MOI.ScalarAffineTerm(1.0, x[i]))
+        )
+        for j = 1:4
+            iszero(M[i, j]) && continue
+            push!(
+                terms,
+                MOI.VectorAffineTerm(1, MOI.ScalarAffineTerm(M[i, j], x[j]))
```
frapac: I have some trouble figuring out why we set `1` there...

odow: Oops. That should be `i`.

frapac: Ok, that's make sense!

src/sets.jl
```diff
@@ -669,18 +669,69 @@ end
...
+    x -in- Interval(-1, 1)
+    [2 * x - 3, x] -in- Complements(1)
+
+defines the mixed complementarity problem where at least one of the following
+holds:
+
+  1. `2 * x - 3 == 0` if `-1 < x < 1`
+  2. `2 * x - 3 >= 0` if `x == -1`
+  3. `2 * x - 3 <= 0` if `x == 1`
+
```
xhub: Here also, all the statements must hold. Note that this is more apparent in the non-monotone case.

Suppose that the problem is `[-4 * x - 3, x] -in- Complements(1)`. Then there are 3 solutions:
1. x == -3/4; F(x) = 0
2. x == -1; F(x) = 1
3. x == 1; F(x) = -7

So all statements must hold to be equivalent to the MCP. Maybe it is even better to have that example with all the solution in the documentation?

src/sets.jl
```diff
@@ -669,18 +669,69 @@ end
...
+The dimension of the vector-valued function `F` must be `2 * dimension`. This
+defines a complementarity constraint between the scalar function `F[i]` and the
+variable in `F[i + dimension]`. Thus, `F[i + dimension]` must be interpretable
+as a single variable `x_i` (e.g., `1.0 * x + 0.0`).
+
+If the variable `x_i` is constrained in `Interval(lb_i, ub_i)`, then
+mathematically, the mixed complementarity problem is to find a solution such
+that at least one of the following holds:
+
```
xhub: I would rephase the second part as follows:
"the mixed complementarity problem consists in finding `x_i` in the interval `[lb_i, ub_i]` such that the following holds:"

All statements must hold. I've discussed this a bit further later.

xhub:
I am not sure I followed the discussion of `Complements` vs `VariationalInequality`.
As @odow said, nothing prevents us from encoding a VI using `Complements`.
I understand that @blegat propose that we have a `VariationalInequality` set where the data for the set `C` is stored inside. That could be a great way to store the data.
It could be used whenever `C` can be described as `Ax + b in K` and K is a cone.

One just has to see how to educate the user about the need to specify the set `C` in this way and not via `@constraints` macros.

blegat:
> One just has to see how to educate the user about the need to specify the set `C` in this way and not via `@constraints` macros.

What do you mean ? The user can do `@constraint(model, [A*x + b; x] in MOI.VariationalInequality(V))` to model  where `F(x) = Ax + b`.

odow:
As I've said before, I think we can leave the `VariationalInequality` set for a future solver. But I think @blegat's last suggestion is the way to go.

@xhub Are the latest changes okay?

xhub:
@odow Thanks for making the changes. I am really happy with `Complements` as just some tagging information. The only part that I would maybe think about changing is the documentation for `Complements`. With the current implementation, and proper support from a solver, nothing prevents me from specifying a general polyhedral VI. Currently the doc mentions MCP, which is good. If we want to change the documentation, we could do that later.

@blegat I am not sure what `V` is in your example. Maybe it is best to move on on the `Complements`.
For `VariationalInequalities` I was thinking we could have the following
```
struct VariationalInequalities
   A::SparseMatrix
   b::AbstractVector
   K::Union{MOI.Zeros, MOI.Nonnegatives, MOI.Nonpositives}
end
```
I was thinking of a case where the user does the following in a variational inequality problem definition
```julia
@constraint(model, [A*x + b; x] in MOI.VariationalInequalities(H, k, MOI.Nonnegatives))
@constraint(model, x[2] >= 5)
```
Then, the full definition of `C` is the intersection of `Hx <= k` and `x[2] >= 5`. Then, the burden of 
I must say that I may be too pessimistic about the user's knowledge. I may have in mind people who are not optimizers by training, rather use optimization algorithm as a tool.

blegat:
> I am not sure what `V` is in your example.

It should be `C`, I have corrected it. About the `A` and `b` field in `VariationalInequality`, I would prefer keeping `VariationalInequality` simple and have a new set `AffineTransformation` and so that the user does `VariationalInequality(AffineTranformation(A, b, MOI.Nonnegatives(2)))` but that's out the scope of this PR.

odow:
> If we want to change the documentation, we could do that later.

Yeah, it's easier to change to be more general later on.

PR #918, opened 2019-10-09
Add QP complementarity test

frapac:
I stay at your disposal if you want to add other MPEC tests inside MOI

### DIFF
--- a/src/Test/nlp.jl
+++ b/src/Test/nlp.jl
@@ -415,43 +415,43 @@ function test_qp_mixed_complementarity(model::MOI.ModelLike, config::TestConfig)
     x = MOI.add_variables(model, 8)
     MOI.set.(model, MOI.VariablePrimalStart(), x, 0.0)
     for i in 1:8
-        MOI.add_constraint(model, MOI.SingleVariable(x[i]), MOI.GreaterThan(0.))
+        MOI.add_constraint(model, MOI.SingleVariable(x[i]), MOI.GreaterThan(0.0))
     end
 
     obj = MOI.ScalarQuadraticFunction(
-        [MOI.ScalarAffineTerm(-10., x[1]), MOI.ScalarAffineTerm(4., x[2])],
-        [MOI.ScalarQuadraticTerm(2., x[1], x[1]), MOI.ScalarQuadraticTerm(8., x[2], x[2])],
-        26.)
+        [MOI.ScalarAffineTerm(-10.0, x[1]), MOI.ScalarAffineTerm(4.0, x[2])],
+        [MOI.ScalarQuadraticTerm(2.0, x[1], x[1]), MOI.ScalarQuadraticTerm(8.0, x[2], x[2])],
+        26.0)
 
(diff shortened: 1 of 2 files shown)


frapac:
I understand. I still have to bridge Knitro then.

odow:
You should test the test with a `mock` optimizer. See my PR for an example.

frapac:
Just a question, however. You mentioned earlier:
> KNITRO.jl, which will implement VectorOfVariables-in-Complements

Shouldn't we test this feature as well?

test/Test/nlp.jl
```diff
@@ -16,4 +16,5 @@ const MOI = MathOptInterface
         )
     )
     MOI.Test.test_linear_mixed_complementarity(mock, config)
+    MOI.Test.test_qp_mixed_complementarity(mock, config)
```
odow: You need to set the solution in the mock optimizer like I do here:
https://github.com/JuliaOpt/MathOptInterface.jl/pull/913/files#diff-00b88401713e1c20041db74f14c5ccd1R12-R17

blegat: You probably need to add a call to `MOI.Utilities.set_mock_optimize!` as above to set the optimal solution.

src/Test/nlp.jl
```diff
@@ -371,16 +371,108 @@ function test_linear_mixed_complementarity(model::MOI.ModelLike, config::TestCon
...
+        @test isapprox(
+            x_val, [1., 0., 3.5, 0., 0., 0., 3., 6], atol = config.atol, rtol = config.rtol
+        )
+        @test MOI.get(model, MOI.ObjectiveValue()) ≈ 17.0 atol=config.atol rtol=config.rtol
+    end
 end
 
 const complementaritytests = Dict(
     "linear_mixed_complementarity" => test_linear_mixed_complementarity,
+    "qp_mixed_complementarity" => test_linear_mixed_complementarity,
```
odow: Should be `=> test_qp_mixed_complementarity`

src/Test/nlp.jl
```diff
@@ -371,16 +371,108 @@ function test_linear_mixed_complementarity(model::MOI.ModelLike, config::TestCon
...
+        x0, x1, x2, x3, x4, x5, x6, x7 >= 0
+
+```
+"""
+function test_qp_mixed_complementarity(model::MOI.ModelLike, config::TestConfig)
+    MOI.empty!(model)
+    x = MOI.add_variables(model, 8)
+    MOI.set.(model, MOI.VariablePrimalStart(), x, 0.0)
+    for i in 1:8
+        MOI.add_constraint(model, MOI.SingleVariable(x[i]), MOI.GreaterThan(0.))
```
odow: Please add a trailing `0` to `0.`. There are also multiple cases of this further down.

src/Test/nlp.jl
blegat: That's the reason the tests are passing without `MOIU.set_mock_optimize!`

PR #925, opened 2019-10-17
relax bounds in unit test for variable bound ConstraintDual

rschwarz:
- changing solve_single_variable_dual_min and solve_single_variable_dual_max
 - before, the variable was bounded 1 <= x <= 1
 - some solvers (e.g. Ipopt) consider that fixed and don't compute multipliers
   (see https://github.com/JuliaOpt/Ipopt.jl/pull/182)
 - so, I relax the bounds to 1 <= x <= 2 and 0 <= x <= 1, only changing the
   inactive bound.

### DIFF
--- a/src/Test/UnitTests/solve.jl
+++ b/src/Test/UnitTests/solve.jl
@@ -100,7 +100,7 @@ function solve_single_variable_dual_min(model::MOI.ModelLike, config::TestConfig
     MOI.empty!(model)
     x = MOI.add_variable(model)
     xl = MOI.add_constraint(model, MOI.SingleVariable(x), MOI.GreaterThan(1.0))
-    xu = MOI.add_constraint(model, MOI.SingleVariable(x), MOI.LessThan(1.0))
+    xu = MOI.add_constraint(model, MOI.SingleVariable(x), MOI.LessThan(2.0))
     MOI.set(model, MOI.ObjectiveSense(), MOI.MIN_SENSE)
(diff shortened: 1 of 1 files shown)


mlubin:
Ipopt has an option to compute duals for fixed variables (https://www.coin-or.org/Ipopt/documentation/node44.html#SECTION000114060000000000000). Does that work? If so, it's preferable to churn in the MOI tests just for one solver.

rschwarz:
I saw that parameter, but changing to another option does more than compute duals for fixed variables. It will have an effect on the whole solution process, because the variables are treated as variables, but then with relaxed bounds etc. I guess that will lead to worse performance and potentially to slightly violated bounds? If anything, I would then suggest to solve twice: first with the default option, then again with the relaxed bounds, but using the warm start.

Not sure what you mean by churn here, but I think the intention of the test is to extract duals for variable bounds. For that use case, it's not necessary to use the same value for the lower and upper bound?

rschwarz:
I might have misunderstood your suggestion. It's not necessary to change that parameter in general, for the Ipopt wrapper. I will try to set it only in the context of the MOI tests...

blegat:
> I might have misunderstood your suggestion. It's not necessary to change that parameter in general, for the Ipopt wrapper. I will try to set it only in the context of the MOI tests...

Yes, it should just affect `test/MOI_wrapper.jl`, it should not be a parameter set by default in `src/MOI_wrapper.jl`.

rschwarz:
I found that this parameter was already set in the `optimizer` used in the test suite, just not for the unit tests :see_no_evil:. Now it passes as excepted, so this change is not needed.

PR #929, opened 2019-10-25
Testing constant term in indicator constraint

matbesancon:
There was no test such that:
```
y in {0,1}
y -> a^T x + b in Set
```
with `b != 0`, this additional test is equivalent to the first indicator test, with a part of the RHS shifted to the constant term


matbesancon:
All good, see corresponding PR in SCIP:
https://github.com/SCIP-Interfaces/SCIP.jl/pull/137

PR #934, opened 2019-10-28
implement _result_index_field for NLPBlockDual

rschwarz:
- should fix failure for MINLPTests with Ipopt:
   https://travis-ci.org/JuliaOpt/MINLPTests.jl/builds/603635262

### DIFF
--- a/src/nlp.jl
+++ b/src/nlp.jl
@@ -32,6 +32,7 @@ struct NLPBlockDual <: AbstractModelAttribute
     N::Int
 end
 NLPBlockDual() = NLPBlockDual(1)
+_result_index_field(attr::NLPBlockDual) = attr.N
 
 is_set_by_optimize(::NLPBlockDual) = true
 


rschwarz:
Currently testing locally, with MINLPTests.jl

rschwarz:
See also JuliaOpt/Ipopt.jl#182 for where this is already used.

mlubin:
Which attributes need to implement `_result_index_field`? Would an attribute defined outside MOI ever need to implement it?

blegat:
> Which attributes need to implement `_result_index_field`?

Attributes for which the field is not `result_index`, see
https://github.com/JuliaOpt/MathOptInterface.jl/blob/6369fcc478fd0641ad295c2145c30fc38a6b6c0d/src/attributes.jl#L126-L129

> Would an attribute defined outside MOI ever need to implement it?

No, the field should be `result_index` by convention for `check_result_index_bounds` to work. With MOI v0.10, we might remove `_result_index_field` and rename `N` to `result_index` for the attributes for which it is `N`.

Issue #947, opened 2019-11-07
Inconsistency in dimension of complementarity set


frapac:
It seems that there is a little inconsistency with how we define the dimension of the complementarity set in MOI: 
https://github.com/JuliaOpt/MathOptInterface.jl/blob/master/src/sets.jl#L680-L683

That leads to mispecified problems when using JuMP in AUTOMATIC mode. E.g., when defining the constraint 
```
[x, y] \in MOI.Complements(1)
```
we are calling this function: 
https://github.com/JuliaOpt/MathOptInterface.jl/blob/master/src/variables.jl#L99-L101
and MOI defines a `VectorOfVariables` with length 1 instead of 2, because the dimension of the set `Complements(1)` is equal to one. 

A solution would be to define instead: 
```julia
dimension(set::Complements) = 2 * set.dimension
```
but that may lead to other issues apart. 

As a MWE, the following code breaks (but works fine once we use Knitro in direct mode):
```julia
m = JuMP.Model(with_optimizer(KNITRO.Optimizer))
@variable(m, t1 >= 0)
@variable(m, t2 >= 0)
@constraint(m, [t1; t2] in MOI.Complements(1))
@objective(m, Min, 7)
JuMP.optimize!(m)

```

blegat:
We should rename the field from `dimension` to something else as it does not follow the convention

odow:
Ah. I guess `dimension(set::Complements) = 2 * set.dimension` is the correct fix.

PR #948, opened 2019-11-07
Fix dimension of Complements

blegat:
Closes https://github.com/JuliaOpt/MathOptInterface.jl/issues/947


frapac:
Efficient. 
Thank for the fix!

PR #954, opened 2019-11-14
Add docstring for test_delete_bridge

blegat:


matbesancon:
in the docstring, maybe disinguish the `F, S` from the type signature and the `F, S` in the constraint list

PR #945, opened 2019-11-07
Fix segfault of #877

blegat:
This fixes the segfault on Julia v1.0 of https://github.com/JuliaOpt/MathOptInterface.jl/pull/877

This also give an impressive performance improvements for the benchmark added in https://github.com/JuliaOpt/MathOptInterface.jl/pull/946. This performance improvement might be explained by https://github.com/JuliaLang/julia/issues/32761
On master:
```
BenchmarkTools.Trial: 
  memory estimate:  269.80 KiB
  allocs estimate:  4658
  --------------
  minimum time:     11.074 ms (0.00% GC)
  median time:      11.251 ms (0.00% GC)
  mean time:        11.761 ms (1.45% GC)
  maximum time:     61.752 ms (79.61% GC)
  --------------
  samples:          425
  evals/sample:     1
BenchmarkTools.Trial: 
  memory estimate:  715.16 KiB
  allocs estimate:  12004
  --------------
  minimum time:     41.166 ms (0.00% GC)
  median time:      45.770 ms (0.00% GC)
  mean time:        50.720 ms (2.82% GC)
  maximum time:     203.532 ms (60.85% GC)
  --------------
  samples:          99
  evals/sample:     1
```
After this PR:
```
BenchmarkTools.Trial: 
  memory estimate:  68.59 KiB
  allocs estimate:  1861
  --------------
  minimum time:     3.819 ms (0.00% GC)
  median time:      5.165 ms (0.00% GC)
  mean time:        5.701 ms (2.53% GC)
  maximum time:     110.109 ms (91.74% GC)
  --------------
  samples:          875
  evals/sample:     1
BenchmarkTools.Trial: 
  memory estimate:  163.96 KiB
  allocs estimate:  4440
  --------------
  minimum time:     23.947 ms (0.00% GC)
  median time:      32.318 ms (0.00% GC)
  mean time:        32.113 ms (0.80% GC)
  maximum time:     83.834 ms (47.83% GC)
  --------------
  samples:          156
  evals/sample:     1
```

PR #956, opened 2019-11-15
Build bridge graph to improve performance of bridge shortest path

blegat:
Most of the time spent in the bridge shortest path algorithm is due to the type instability of bridge types accessed from the vector of bridges added to the `LazyBridgeOptimizer`.
Indeed, we then need to call `supports_constraint`, `concrete_bridge_type` and `added_...` to the bridge type and the method lookup needs to be done at runtime because the bridge type is not known at compile time.
In this PR, the subpart of the bridge graph is constructed so that the Bellman-Ford algorithm is run on a fully type-stable graph structure.
The graph is built lazily when `node` is called for new types.
Of course, the type instability still affects the `node` functions but once the graph is build, the time to run Bellman-Ford is negligible.
Running the `perf/bellman_ford.jl` benchmark gives which shows that solving the shortest path problem takes a negligible time compared to building the graph:
```
BenchmarkTools.Trial: 
  memory estimate:  10.39 KiB
  allocs estimate:  194
  --------------
  minimum time:     237.782 μs (0.00% GC)
  median time:      313.253 μs (0.00% GC)
  mean time:        386.994 μs (2.17% GC)
  maximum time:     54.999 ms (99.18% GC)
  --------------
  samples:          10000
  evals/sample:     1
BenchmarkTools.Trial: 
  memory estimate:  11.80 KiB
  allocs estimate:  284
  --------------
  minimum time:     239.733 μs (0.00% GC)
  median time:      311.878 μs (0.00% GC)
  mean time:        385.405 μs (2.63% GC)
  maximum time:     72.794 ms (99.43% GC)
  --------------
  samples:          10000
  evals/sample:     1
BenchmarkTools.Trial: 
  memory estimate:  25.72 KiB
  allocs estimate:  479
  --------------
  minimum time:     1.614 ms (0.00% GC)
  median time:      1.789 ms (0.00% GC)
  mean time:        2.268 ms (1.41% GC)
  maximum time:     52.869 ms (95.84% GC)
  --------------
  samples:          2200
  evals/sample:     1
BenchmarkTools.Trial: 
  memory estimate:  29.16 KiB
  allocs estimate:  699
  --------------
  minimum time:     1.617 ms (0.00% GC)
  median time:      1.784 ms (0.00% GC)
  mean time:        2.035 ms (1.34% GC)
  maximum time:     50.862 ms (95.61% GC)
  --------------
  samples:          2453
  evals/sample:     1
```
In comparison, the benchmark result was previously (see https://github.com/JuliaOpt/MathOptInterface.jl/pull/945):
```
BenchmarkTools.Trial: 
  memory estimate:  68.59 KiB
  allocs estimate:  1861
  --------------
  minimum time:     3.819 ms (0.00% GC)
  median time:      5.165 ms (0.00% GC)
  mean time:        5.701 ms (2.53% GC)
  maximum time:     110.109 ms (91.74% GC)
  --------------
  samples:          875
  evals/sample:     1
BenchmarkTools.Trial: 
  memory estimate:  163.96 KiB
  allocs estimate:  4440
  --------------
  minimum time:     23.947 ms (0.00% GC)
  median time:      32.318 ms (0.00% GC)
  mean time:        32.113 ms (0.80% GC)
  maximum time:     83.834 ms (47.83% GC)
  --------------
  samples:          156
  evals/sample:     1
```
This gives almost a 20x speedup !

### DIFF
--- a/src/Bridges/graph.jl
+++ b/src/Bridges/graph.jl
@@ -1,12 +1,11 @@
 const INFINITY = -1
 
 abstract type AbstractNode end
-struct ConstraintNode <: AbstractNode
-    index::Int
-end
 struct VariableNode <: AbstractNode
     index::Int
-    constraint_node::ConstraintNode
+end
+struct ConstraintNode <: AbstractNode
+    index::Int
 end
 struct ObjectiveNode <: AbstractNode
     index::Int
@@ -27,6 +26,7 @@ end
 
 mutable struct Graph
     variable_edges::Vector{Vector{Edge}}
+    variable_constraint_node::Vector{ConstraintNode}
     # variable node index -> Number of bridges that need to be used
     variable_dist::Vector{Int}
     # variable node index -> Index of bridge to be used
@@ -47,10 +47,30 @@ mutable struct Graph
 end
 function Graph()
     return Graph(
-        Vector{Edge}[], Int[], Int[], 0,
+        Vector{Edge}[], ConstraintNode[], Int[], Int[], 0,
         Vector{Edge}[], Int[], Int[], 0,
         Vector{ObjectiveEdge}[], Int[], Int[], 0,
     )
+end
+
+function Base.show(io::IO, graph::Graph)
+    print(io, "Bridge graph with ")
+    print(io, length(graph.variable_best), " variable nodes, ")
+    print(io, length(graph.constraint_best), " variable nodes and ")
+    print(io, length(graph.objective_best), " objective nodes.")
+end
+function variable_nodes(graph::Graph)
+    return MOIU.LazyMap{VariableNode}(
+        i -> VariableNode(i), eachindex(graph.variable_best))
+end
+function constraint_nodes(graph::Graph)
+    return MOIU.LazyMap{ConstraintNode}(
+        i -> ConstraintNode(i), eachindex(graph.constraint_best)
+    )
+end
+function objective_nodes(graph::Graph)
+    return MOIU.LazyMap{ObjectiveNode}(
+        i -> ObjectiveNode(i), eachindex(graph.objective_best))
 end
 
 # After `add_bridge(b, BT)`, some constrained variables `(S,)` in
@@ -65,6 +85,7 @@ end
 # if several bridges are added consecutively.
 function Base.empty!(graph::Graph)
     empty!(graph.variable_edges)
+    empty!(graph.variable_constraint_node)
     empty!(graph.variable_dist)
     empty!(graph.variable_best)
     graph.variable_last_correct = 0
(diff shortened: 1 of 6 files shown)


blegat:
There was a bug in the earlier version of the PR, it is now fixed, I have updated the benchmark, the speedup is now even better.

src/Bridges/graph.jl
```diff
@@ -0,0 +1,269 @@
...
+end
+function bridge_index(graph::Graph, node::ObjectiveNode)
+    bellman_ford!(graph)
+    return graph.objective_best[node.index]
+end
+
+# Update `b.variable_dist`, `b.constraint_dist` `b.variable_best` and
+# `b.constraint_best` for constrained variable types in `variables` and
+# constraint types in `constraints`.
+function updated_dist(graph::Graph, current, edges::Vector{<:AbstractEdge})
```
odow: Should `current` be typed here?

src/Bridges/graph.jl
```diff
@@ -0,0 +1,269 @@
...
+    push!(graph.constraint_edges[node.index], edge)
+end
+function add_edge(graph::Graph, node::ObjectiveNode, edge::ObjectiveEdge)
+    push!(graph.objective_edges[node.index], edge)
+end
+function add_variable_node(graph::Graph)
+    push!(graph.variable_edges, Edge[])
+    # Use an invalid index so that the code errors instead return something
+    # incorrect in case `set_variable_constraint_node` is not called.
+    push!(graph.variable_constraint_node, ConstraintNode(-1))
```
odow: It might be more readable to define a constant instead of using `-1` since this check is used elsewhere.

src/Bridges/lazy_bridge_optimizer.jl
```diff
@@ -402,79 +272,92 @@ end
...
     end
-    return result
+    return Objective.concrete_bridge_type(b.objective_bridge_types[index], F)
 end
 
 _func_name(::Type{Constraint.ScalarFunctionizeBridge}) = "SingleVariable", "constraint"
 _func_name(::Type{Constraint.VectorFunctionizeBridge}) = "VectorOfVariables", "constraint"
 _func_name(::Type{Objective.FunctionizeBridge}) = "SingleVariable", "objective"
-function _functionize_bridge(bridge_types, target_type)
+function __functionize_bridge(bridge_types, target_type)
```
odow: Having the single-underscore and double-underscore methods is confusing. 

Are the instances of `__functionize_bridge` above meant to be the single-underscore methods?

src/Bridges/graph.jl
```diff
@@ -0,0 +1,269 @@
...
+# Update `b.variable_dist`, `b.constraint_dist` `b.variable_best` and
+# `b.constraint_best` for constrained variable types in `variables` and
+# constraint types in `constraints`.
+function updated_dist(graph::Graph, current, edges::Vector{<:AbstractEdge})
+    bridge_index = 0
+    for edge in edges
+        if supports_added_no_update(graph, edge)
+            # `added_dist != INFINITY` because `supports_added_no_update`
+            # returned `true`.
+            dist = 1 + added_dist(graph, edge)
```
odow: So, future-proofing: if we want to add weights to the bridges, we should replace the `xxx_dist::Vector{Int}` with `xxx_dist::Vector{Float64}`, then define `weight(edge) = 1.0` and replace `dest = weight(edge) + added_dist(graph, edge)`?

Then it's a matter of solvers being able to define specific edge weights -- not sure how that would compose with different layers?

blegat: Yes exactly. If the weight is a constraint attribute, that might compose well

odow: > If the weight is a constraint attribute, that might compose well

💯 

There's something nice about `MOI.Bridges` defining a module-specific constraint attribute. We probably also need a `AbstractModelAttribute` for the cost of adding a variable.

Issue #958, opened 2019-11-20
Copy optimizer attributes


blegat:
Optimizer attributes are not copied:
https://github.com/JuliaOpt/MathOptInterface.jl/blob/1755bf88b05499aced42b4ca1c15267045b038ed/src/Utilities/copy.jl#L331-L340
See https://github.com/JuliaOpt/JuMP.jl/pull/2101#discussion_r348044668

PR #959, opened 2019-11-22
Callbacks are model attributes

blegat:
As mentioned in the docstrings:
https://github.com/JuliaOpt/MathOptInterface.jl/blob/fa4e026106a8909bd37d6a2e2e3e9fecf4534b5a/src/MathOptInterface.jl#L61-L74
https://github.com/JuliaOpt/MathOptInterface.jl/blob/fa4e026106a8909bd37d6a2e2e3e9fecf4534b5a/src/attributes.jl#L10-L12
The difference between model and optimizer attributes is that model attributes are discarded at `MOI.empty!` and are not copied at `MOI.copy_to`.

This should resolve the issue of callbacks not working in non-direct mode with JuMP. The issue was that GLPK discarded the callback in `MOI.empty!` but then did not receive it again in `MOI.copy_to`.

This does not closes https://github.com/JuliaOpt/MathOptInterface.jl/issues/958 as we should clarify in the docstring of `MOI.copy_to` that optimizer attributes are not copied.

### DIFF
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -622,7 +622,7 @@ end
 """
     abstract type AbstractCallback <: AbstractModelAttribute end
 
-Abstract type for optimizer attribute representing a callback function. The
+Abstract type for a model attribute representing a callback function. The
 value set to subtypes of `AbstractCallback` is a function that may be called
 during [`optimize!`](@ref). As [`optimize!`](@ref) is in progress, the result
 attributes (i.e, the attributes `attr` such that `is_set_by_optimize(attr)`)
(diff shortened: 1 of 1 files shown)


src/attributes.jl
```diff
@@ -620,7 +620,7 @@ function Base.showerror(io::IO, err::OptimizeInProgress)
 end
 
 """
-    abstract type AbstractCallback <: AbstractOptimizerAttribute end
+    abstract type AbstractCallback <: AbstractModelAttribute end
 
 Abstract type for optimizer attribute representing a callback function. The
```
odow: s/for optimizer attribute/for a model attribute

PR #962, opened 2019-11-23
fix incorrect increment in geomean bridge

lkapelevich:
On MOI master I get:
```julia
using JuMP
using MosekTools

n = 9
v = collect(1:n)

model = Model(with_optimizer(Mosek.Optimizer))
@variable(model, t)
@objective(model, Max, t)
@constraint(model, vcat(t, v) in MOI.GeometricMeanCone(n + 1))
optimize!(model)
objective_value(model)
prod(v)^(1 / n)

julia> objective_value(model)
2.6771584679712452

julia> prod(v)^(1 / n)
4.147166274396913
```

A test problem needs more than eight variables in the geomean to trigger this.

### DIFF
--- a/src/Bridges/Constraint/geomean.jl
+++ b/src/Bridges/Constraint/geomean.jl
@@ -159,18 +159,17 @@ function MOI.get(model::MOI.ModelLike, attr::MOI.ConstraintFunction,
     d = bridge.d
     n = d - 1
     l = ilog2(n)
-    offset = offset_next = 0
-    for i in 1:l
-        offset_next = offset + i
-        for j in 1:(1 << (i-1))
-            if i == l && 2j <= bridge.d
-                func = MOI.get(model, attr, bridge.socrc[offset + j])
-                func_scalars = MOIU.eachscalar(func)
-                f_scalars[2j] = func_scalars[1]
+    num_lvars = 1 << (l - 1)
+    offset = num_lvars - 1
+    for j in 1:num_lvars
+        if 2j <= bridge.d
+            func = MOI.get(model, attr, bridge.socrc[offset + j])
+            func_scalars = MOIU.eachscalar(func)
+            f_scalars[2j] = func_scalars[1]
+            if 2j + 1 <= bridge.d
                 f_scalars[2j + 1] = func_scalars[2]
             end
         end
-        offset = offset_next
     end
     return MOIU.vectorize(f_scalars)
 end
--- a/src/Bridges/Constraint/geomean.jl
+++ b/src/Bridges/Constraint/geomean.jl
@@ -54,7 +54,7 @@ function bridge_constraint(::Type{GeoMeanBridge{T, F, G, H}}, model,
     n = d - 1
     l = ilog2(n)
     N = 1 << l
-    xij = MOI.add_variables(model, N-1)
+    xij = MOI.add_variables(model, N - 1)
     f_scalars = MOIU.eachscalar(f)
 
     xl1 = MOI.SingleVariable(xij[1])
@@ -77,8 +77,9 @@ function bridge_constraint(::Type{GeoMeanBridge{T, F, G, H}}, model,
     socrc = Vector{CI{G, MOI.RotatedSecondOrderCone}}(undef, N - 1)
     offset = offset_next = 0
     for i in 1:l
-        offset_next = offset + i
-        for j in 1:(1 << (i - 1))
+        num_lvars = 1 << (i - 1)
+        offset_next = offset + num_lvars
+        for j in 1:num_lvars
             if i == l
                 a = _getx(2j - 1)
                 b = _getx(2j)
--- a/src/Bridges/Constraint/geomean.jl
+++ b/src/Bridges/Constraint/geomean.jl
@@ -94,6 +94,7 @@ function bridge_constraint(::Type{GeoMeanBridge{T, F, G, H}}, model,
         end
         offset = offset_next
     end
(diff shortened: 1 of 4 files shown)


src/Bridges/Constraint/geomean.jl
```diff
@@ -77,8 +77,9 @@ function bridge_constraint(::Type{GeoMeanBridge{T, F, G, H}}, model,
     socrc = Vector{CI{G, MOI.RotatedSecondOrderCone}}(undef, N - 1)
     offset = offset_next = 0
     for i in 1:l
-        offset_next = offset + i
-        for j in 1:(1 << (i - 1))
+        num_lvars = 1 << (i - 1)
+        offset_next = offset + num_lvars
+        for j in 1:num_lvars
```
blegat: We should add `@assert offset == length(socrc)` at the end of the for loop.

lkapelevich:
I added a test problem but I am confused about how the tests are supposed to work- how do values of variables in auxiliary RSOC cones get tested?

src/Test/contconic.jl
```diff
@@ -1276,8 +1276,75 @@ end
...
+    vov = MOI.VectorOfVariables([t; x])
+    if vecofvars
+        gmc = MOI.add_constraint(model, vov, MOI.GeometricMeanCone(n + 1))
+    else
+        gmc = MOI.add_constraint(model, MOI.VectorAffineFunction{Float64}(vov), MOI.GeometricMeanCone(n + 1))
+    end
+
+    cx = Vector{MOI.ConstraintIndex{MOI.ScalarAffineFunction{Float64},MOI.EqualTo{Float64}}}(undef, n)
+    for i in 1:n
+        cx[i] = MOI.add_constraint(model, MOI.ScalarAffineFunction([MOI.ScalarAffineTerm(1.0, x[i])], 0.0), MOI.EqualTo(Float64(i)))
```
blegat: x_i is constrained to be i or 1?

blegat:
When you get the ConstraintPrimal of the constrained variables, it calls the ConstraintPrimal of the added constraints which use the value of the added variables. You can also test the value of the variables of `mock` in `test/Bridges/Constraint/geomean.jl`

src/Test/contconic.jl
lkapelevich: Oops it should have been 1

lkapelevich:
Alright thanks. Regarding `test/Bridges/Constraint/geomean.jl`- if I change `[ones(4); 2; √2; √2]` in https://github.com/JuliaOpt/MathOptInterface.jl/blob/master/test/Bridges/Constraint/geomean.jl#L24 to e.g. `[ones(4); 0; 0; 0]`, should the test fail? It is passing either way.

blegat:
I wouldn't be surprised if even if the constraint primal of the added constraints depend on the value of the added variables, these entries of the vectors are ignored and it does not affect the ConstraintPrimal returned.
We don't really need to tests the value of the added variables since it's just set in the bridge tests. What would make more sense it to tests the model created with `test_model_equals`.

lkapelevich:
@blegat what constraints should be added if someone does `@constraint(model, [t, x] in MOI.GeometricMeanCone(2))`, or, `@constraint(model, [t] in MOI.GeometricMeanCone(1))`? These cases are also not handled properly at the moment.

blegat:
For GeometricMeanCone(2), it should be t <= x and x >= 0.
For GeometricMeanCone(1), it should be t <= 1? Maybe we should throw an error. How does Hypatia interpret this cone?

chriscoey:
We throw an error. MOI should throw an error too. t <= 1 isn't a cone.

lkapelevich:
What would be the right error? MOI doesn't throw an error like this for other cones.

lkapelevich:
> For GeometricMeanCone(2), it should be t <= x and x >= 0

What's the best way to add x >= 0 to the `GeoMeanBridge` struct?

blegat:
> What would be the right error? MOI doesn't throw an error like this for other cones.

We should create a new error type that the bridge and Hypatia can throw so that they have the same behavior.

> What's the best way to add x >= 0 to the GeoMeanBridge struct?

You need to add a field of type `Union{Nothing, MOI.ConstraintIndex{MOI.VectorOfVariables, MOI.Nonnegatives}}` in the bridge and you constraint `[x]` in Nonnegatives.
Note that you should also add `(VOV, Nonnegatives)` in the vector returned by `added_constraint_types`.
A similar example can be found here:
https://github.com/JuliaOpt/MathOptInterface.jl/blob/2be98a7beda61f9ab332c465bb122839f8fc9fd7/src/Bridges/Variable/rsoc_to_psd.jl#L12-L13

lkapelevich:
Couple more questions- why `[x]` in Nonnegatives rather than adding another LessThan? Also `x` could be an affine function, right?

blegat:
> Couple more questions- why `[x]` in Nonnegatives rather than adding another LessThan? Also `x` could be an affine function, right?

The rule is "use the most specific type that fits". Here `[x] in Nonnegatives` is the most specific while `x in MOI.GreaterThan(0.0)` you loose the information in the type that `x` is nonnegative as the type is just `MOI.GreaterThan{Float64}`. As the bridges reason only with types, it's best to be able to represent the most we can inside the types.

blegat:
I'll merge this PR as it contains a self-contained fix but the fix for geomean of dimension is of course welcome as a separate PR :)

lkapelevich:
Thanks. There were two TODO comments in there that actually should have been deleted.

PR #968, opened 2019-12-05
DNMY: Add support for vector-valued objective functions

odow:
This is a DNMY branch for some work I'm doing with multi-criteria solvers.

See https://github.com/JuliaOpt/JuMP.jl/issues/2099 for the full story.


odow:
Closing as stale. We can revisit this when it's time to make a push for proper multi-objective support.

PR #969, opened 2019-12-09
Add MathOptFormat.jl as FileFormats submodule

### LINKED ISSUE #28: A File-format for MathOpt

### LINKED ISSUE #633: Read/write from/to file should be IO not filenames

odow:
Closes #28
Closes #633

I can't seem to figure out how to rebase and squash the initial commits without screwing up the unrelated-histories part of Git. If we're okay adding this many commits, then I won't change anything. Otherwise I might need to try take III of this and squash the history a bit before I merge the repos.


odow:
This PR is blocked until I sort the Windows failures on JSONSchema.jl. That package could use some love, but it's not too long, so I'll wait to see if the developer is responsive, otherwise I'll fork and tidy myself.

blegat:
We should add something in the doc

odow:
This is ready on my end. I now have commit access to JSONSchema.jl and I'm planning on making some improvements, so it seems okay to add as a dependency.

docs/src/apimanual.md
```diff
@@ -792,6 +792,35 @@ becomes ``A = [3.0, 4.0]^\top`` as follows:
 modify(m, c, MultirowChange(x, [3.0, 4.0]))
 ```
 
+## File formats
+
+The `FileFormats` module provides functionality for reading and writing MOI models using
+[`write_to_file`](@ref) and [`read_from_file`](@ref).
```
mlubin: Are the extensions to `Base.write` that let you use IO objects documented?

test/runtests.jl
```diff
@@ -22,24 +22,16 @@ include("dummy.jl")
...
+    "FileFormats",
+    "Test",
+    "Utilities",
+]
+    include("$(submodule)/$(submodule).jl")
 end
 
 # Test hygiene of @model macro
 include("hygiene.jl")
+
```
blegat: Remove line

docs/src/apimanual.md
```diff
@@ -792,6 +792,35 @@ becomes ``A = [3.0, 4.0]^\top`` as follows:
...
+filename = "my_model.cbf.gz"
+dest = FileFormats.Model(filename = filename)
+MOI.copy_to(dest, src)
+MOI.write_to_file(dest, filename)
+
+src_2 = FileFormats.Model(filename = filename)
+MOI.read_from_file(src_2, filename)
+```
+Note how the compression format (GZip) is also automatically detected from the filename.
+
```
blegat: Maybe mention that one might want to do `MOI.copy_to(bridged, src)` with `bridged = MOI.Bridges.full_bridge_optimizer(dest, Float64)` with a link to the bridge section.

docs/src/apimanual.md
```diff
@@ -821,6 +821,31 @@ MOI.read_from_file(src_2, filename)
 ```
 Note how the compression format (GZip) is also automatically detected from the filename.
 
+In some cases `src` may contain constraints that are not supported by the file format (e.g.,
+binary variables in the CBF format). If so, you should copy `src` to a bridged model using
```
blegat: I would add that CBF supports integer hence even if it doesn't support binary, it can be bridged. Here it seems adding bridges solves everything, e.g. It could even make LP support PSD

odow:
Are we okay if I merge this with all 127 commits? It's a little untidy, but I don't know how to make it neater.

blegat:
> Are we okay if I merge this with all 127 commits? It's a little untidy, but I don't know how to make it neater.

Sounds good

PR #973, opened 2019-12-18
Only use variable bridge if it is optimal

blegat:
Caught by SolverTests with ECOS and SCS: https://travis-ci.com/blegat/SolverTests

### DIFF
--- a/src/Bridges/bridge_optimizer.jl
+++ b/src/Bridges/bridge_optimizer.jl
@@ -91,7 +91,7 @@ function supports_bridging_constrained_variable(
 end
 
 """
-    is_bridged_with_variable_bridge(
+    is_variable_bridged(
         b::AbstractBridgeOptimizer, S::Type{<:MOI.AbstractSet})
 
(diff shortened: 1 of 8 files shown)


src/Bridges/bridge_optimizer.jl
```diff
@@ -90,6 +90,19 @@ function supports_bridging_constrained_variable(
...
+        ::AbstractBridgeOptimizer, ::Type{<:MOI.AbstractSet})
+
+Return a `Bool` indicating whether `b` bridges constrained variable in
+`S` using a variable bridge. If it returns `false`, it means that free
+variables and a constraint on these variables should be added instead.
+The constraint is then bridged by a constraint bridge.
+"""
+function is_bridged_with_variable_bridge(
+    ::AbstractBridgeOptimizer, ::Type{<:MOI.AbstractSet})
+end
```
odow: Should this `return false` as a fallback instead of the other two methods?

Issue #705, opened 2019-04-07
complex/hermitian PSD to real PSD bridge?


chriscoey:
wanted to discuss whether this is feasible with the current design. I haven't seen a complex MOI model yet.

see https://docs.mosek.com/modeling-cookbook/sdo.html#hermitian-matrices for the transformation of a constraint from complex hermitian order-n cone to real PSD order-2n cone.

blegat:
That is feasible, `ComplexPSDBridge{T}` would transform PSD in `Complex{T}` to PSD in `T`. Another similar bridge would be equality between complex expressions into two equality constraints.

chriscoey:
yep good point, updated title to broaden the scope.

chriscoey:
Re: Hermitian PSD cone, should be a separately-named cone (like HermitianPositiveSemidefiniteConeTriangle) or the same cone type (PositiveSemidefiniteConeTriangle) but parametrized by Complex{T} insteal of T for T<:Real? In Hypatia we parametrize certain cones by T or Complex{T}. All the data is real but the complex generalizations of cones have higher dimension than their real counterparts, eg for our Hermitian PSD cone, we have one entry per diagonal element and two consecutive entries for each off-diagonal in lower triangle (complex). This has not been limiting and it works for various cones that we have generalized for complex: infinity norm, spectral norm, rootdet, logdet, and dual wsos (and their dual cones by extension). 

I'm planning to start adding support incl bridges for these complex generalizations (except wsos, which is a bit more exotic) to MOI now but need to know what conventions to follow.

blegat:
I would be tempted to say that we have the same cone type with different parametric type.
Naively I would keep the same dimension but with complex entries, would that be feasible in your opinion?

chriscoey:
[...]

PR #976, opened 2019-12-24
add nuclear norm and spectral norm cone sets (for general/nonsymmetric matrices)

chriscoey:
These cones are dual to each other. I've added the definitions in sets.jl and when we are agreed on those, I'll finish off the rest including bridges for SDP extended formulations. An important decision to make is how to handle the (general/nonsymmetric) matrix part in the definition, like whether to vectorize it row-major or column-major.


chriscoey:
Also forgot to mention: with these complex cones, some elements are complex and some are real. For example for hermitian PSD, diagonal entries are actually real, and Hypatia's realified dimension of hermitian PSD triangle cone is side^2, which is 1 real entry per diagonal and 2 real entries per under-diagonal. For another example, the complex spectral cone has a real epigraph variable and complex matrix variables, and the complex hermitian rootdet and logdet cones have real epi/hypograph (and perspective for logdet) and real diagonal and complex off-diag elements. So it's not really the case that we want a `VAF{Complex{T}} in set{Complex{T}}`, since some elements of the VAF are real and some are complex. So given all this complexity I'm starting to prefer Hypatia's way.

chriscoey:
ref: #705 for discussion of how to define complex generalizations of cones such as the spectral/nuclear cones. If that issue is resolved soon then I'll generalize for complex in this PR. The complex cones just have complex matrix entries but the epigraph variable stays real.

chriscoey:
Another question is whether these should fall under the "matrix sets", which are currently all symmetric matrix sets, or whether there should be abstract nonsymmetric matrix sets distinguished from abstract symmetric matrix sets

src/sets.jl
```diff
@@ -287,6 +287,34 @@ function Base.:(==)(set1::S, set2::S) where S <: Union{PowerCone, DualPowerCone}
     return set1.exponent == set2.exponent
 end
 
+"""
+    NormSpectralCone{T <: Real}(row_dim, column_dim)
+
+The epigraph of the matrix spectral norm (maximum singular value function) ``\\{ (t, X) \\in \\mathbb{R}^{1 + row_dim \\times column_dim} : t \\ge \\sigma_1(X) \\}`` where ``\\sigma_i`` is the ``i``th singular value of the general matrix ``X`` of row dimension `row_dim` and column dimension `column_dim`.
```
blegat: What is a "general matrix" ?
How are the singular values ordered ?

src/sets.jl
```diff
@@ -287,6 +287,34 @@ function Base.:(==)(set1::S, set2::S) where S <: Union{PowerCone, DualPowerCone}
...
+TODO how to vectorize matrix entries: column or row major?
```
blegat: I would stack the columns as done in `Base.Array` so that the user can just call `vec`
```julia
julia> A = [1 2; 3 4]
2×2 Array{Int64,2}:
 1  2
 3  4

julia> vec(A)
4-element Array{Int64,1}:
 1
 3
 2
 4
```

src/sets.jl
chriscoey: I'm trying to distinguish from the various symmetric matrix sets already there. Happy to just say "matrix" instead of "general matrix".

src/sets.jl
chriscoey: agreed

chriscoey:
What should I name the bridge file? something like norm_spec_nuc_to_psd? I'll need to rename the norm_to_lp file to something like norm_one_inf_to_lp, because "norm" isn't specific enough anymore. That's breaking I suppose?

chriscoey:
Bridges seem challenging. I can't infer the row_dim and column_dim of the matrix from the length of a vector alone, so I don't know how to implement some of the functions that other bridges implement. 

Could we merge this without any bridges (after I remove the bridge file)? Or just not implement many of the bridge functions?

blegat:
No, changing file name is not breaking.

> Bridges seem challenging. I can't infer the row_dim and column_dim of the matrix from the length of a vector alone, so I don't know how to implement some of the functions that other bridges implement.

Could you elaborate on this ?

> Could we merge this without any bridges (after I remove the bridge file)? Or just not implement many of the bridge functions?

Yes, but we should make sure now if the set definition will need to change for the bridging to work.

chriscoey:
> > Bridges seem challenging. I can't infer the row_dim and column_dim of the matrix from the length of a vector alone, so I don't know how to implement some of the functions that other bridges implement.

> Could you elaborate on this ?

I'll try. So take the `map_function` function for example. Its arguments are a bridge type and a function, but not the actual bridge or the actual set. The set definition contains the `row_dim` and `column_dim` integers that are needed to actually determine what the set looks like. It is not possible to determine `row_dim` and `column_dim` simply from the dimension of a vector (vector dimension is `1 + row_dim * column_dim`, and we do not know the number of rows and columns of the set inside the `map_function` function, which means we can't construct the extended formulation. 

I suppose an option is to parametrize the bridge type by the `row_dim` and `column_dim` integers. But might it be cleaner to just change the design of the bridge interface to give access to the fields inside the set (`row_dim` and `column_dim`) when in a function such as `map_function`? 

Several of the other sets I want to add are in the same boat: a vector dimension alone is not enough information to construct the extended formulation.

blegat:
Your bridge does not have to be a `SetMapBridges`, this interface is not the AbstractBridge interface.
SetMapBridge is a subtype of bridge for which the bridge is a linear map between two sets and the map only depends on the type of the sets. For instance, VectorizeBridge does not use this interface.

chriscoey:
Thanks @blegat. Which existing bridge would be most similar to this one, so that I can base it on a prototype?

blegat:
`NormOneBridge` would be a good start :)

chriscoey:
@blegat this is ready for review I think

chriscoey:
I'm seeing some fragility when running the new contconic spec/nuc tests with solvers, so I think I'll change those instances

src/Bridges/Constraint/norm_spec_nuc_to_psd.jl
```diff
@@ -0,0 +1,169 @@
...
+
+The `NormSpectralCone` is representable with a PSD constraint, since
+``t \\ge \\sigma_1(X)`` if and only if ``[tI X; X^\\top tI] \\succ 0``.
+"""
+struct NormSpectralBridge{T, F, G} <: AbstractBridge
+    row_dim::Int # row dimension of X
+    column_dim::Int # column dimension of X
+    psd_index::CI{F, MOI.PositiveSemidefiniteConeTriangle}
+end
+function bridge_constraint(::Type{NormSpectralBridge{T, F, G}}, model::MOI.ModelLike, f::MOI.AbstractVectorFunction, s::MOI.NormSpectralCone) where {T, F, G}
```
blegat: You can write `f::G`

src/Bridges/Constraint/norm_spec_nuc_to_psd.jl
```diff
@@ -0,0 +1,169 @@
...
+MOI.get(b::NormSpectralBridge{T, F, G}, ::MOI.NumberOfConstraints{F, MOI.PositiveSemidefiniteConeTriangle}) where {T, F, G} = 1
+MOI.get(b::NormSpectralBridge{T, F, G}, ::MOI.ListOfConstraintIndices{F, MOI.PositiveSemidefiniteConeTriangle}) where {T, F, G} = [b.psd_index]
+
+# References
+MOI.delete(model::MOI.ModelLike, c::NormSpectralBridge) = MOI.delete(model, c.psd_index)
+
+# Attributes, Bridge acting as a constraint
+function MOI.get(model::MOI.ModelLike, ::MOI.ConstraintFunction, c::NormSpectralBridge{T, F, G}) where {T, F, G}
+    psd_func = MOIU.eachscalar(MOI.get(model, MOI.ConstraintFunction(), c.psd_index))
+    # TODO is it OK to just take the t value from first diagonal index?
```
blegat: LGTM

src/Bridges/Constraint/norm_spec_nuc_to_psd.jl
```diff
@@ -0,0 +1,169 @@
...
+    row_dim::Int # row dimension of X
+    column_dim::Int # column dimension of X
+    psd_index::CI{F, MOI.PositiveSemidefiniteConeTriangle}
+end
+function bridge_constraint(::Type{NormSpectralBridge{T, F, G}}, model::MOI.ModelLike, f::MOI.AbstractVectorFunction, s::MOI.NormSpectralCone) where {T, F, G}
+    f_scalars = MOIU.eachscalar(f)
+    t = f_scalars[1]
+    row_dim = s.row_dim
+    column_dim = s.column_dim
+    @assert row_dim <= column_dim # TODO informative error if not
```
blegat: What's the issue if it's not the case

chriscoey: assuming rows <= cols allows us to keep the bridge code simpler. it's what Hypatia does, so it also avoids having to mess around with swapping indices in the MOI wrapper there. It's easy enough at the JuMP level to use ``X'`` instead of ``X`` if matrix ``X`` has more rows than columns. I'd propose to keep this restriction and if someone wants to remove it in future then they can without causing breakage.

chriscoey:
Hypatia is passing the new conic tests both with and without the bridges.

src/Bridges/Constraint/norm_spec_nuc_to_psd.jl
```diff
@@ -0,0 +1,167 @@
...
+    U::Vector{MOI.VariableIndex}
+    V::Vector{MOI.VariableIndex}
+    ge_index::CI{F, MOI.GreaterThan{T}}
+    psd_index::CI{G, MOI.PositiveSemidefiniteConeTriangle}
+end
+function bridge_constraint(::Type{NormNuclearBridge{T, F, G, H}}, model::MOI.ModelLike, f::H, s::MOI.NormNuclearCone) where {T, F, G, H}
+    f_scalars = MOIU.eachscalar(f)
+    row_dim = s.row_dim
+    column_dim = s.column_dim
+    @assert row_dim <= column_dim
```
blegat: Replace by `row_dim <= column_dim || error("Reformulation for when the number of column is larger than the number of rows is not implemented yet.")`

chriscoey: The current definition of the cone in sets.jl also includes this restriction. If we are to keep that restriction then we should error when the set is created rather than in the Bridge. I was proposing to keep the restriction in the cone definition.

blegat: What is going to fail in the bridges if `row_dim <= column_dim` is not satisfied ?

chriscoey: It looks like they still work actually. OK so I removed the restriction.

test/Bridges/Constraint/norm_spec_nuc_to_psd.jl
```diff
@@ -0,0 +1,133 @@
...
+        ]])
+
+    inv6 = inv(6)
+    rt3 = sqrt(3)
+    invrt12 = inv(2 * rt3)
+    psd_dual = [inv6, -invrt12, inv6, invrt12, -invrt12, inv6, 0, 0, 0, 0, -invrt12, invrt12, -invrt12, 0, 0.5]
+    mock.optimize! = (mock::MOIU.MockOptimizer) -> MOIU.mock_optimize!(mock, [rt3],
+        (MOI.VectorAffineFunction{Float64}, MOI.PositiveSemidefiniteConeTriangle) => [psd_dual])
+
+    MOIT.normspec1test(bridged_mock, config)
```
chriscoey: @blegat are both of the norm spec tests supposed to be run here, or only the one that is used for the mock/bridged models below?

blegat: On test is enough if you feel that it provides enough coverage

chriscoey: thanks

blegat:
Can you add ConstraintPrimalStart and ConstraintDualStart? See https://github.com/JuliaOpt/MathOptInterface.jl/issues/684

chriscoey:
If that was implemented already for NormOne then I could use that as a template. But it's not and the other bridges are set map bridges, which I don't understand. I would prefer to get this PR merged before discussing that further.

blegat:
VectorizeBridge and ScalarizeBridge are not `SetMapBridge` and implement starting values, see https://github.com/JuliaOpt/MathOptInterface.jl/pull/936/ and https://github.com/JuliaOpt/MathOptInterface.jl/pull/937/

blegat:
See also https://github.com/JuliaOpt/MathOptInterface.jl/pull/994

chriscoey:
@blegat I added constraint primal start for spectral cone and constraint dual start for nuclear cone. Not sure how to do con dual start for spectral or con primal start for nuclear.

src/Bridges/Constraint/norm_spec_nuc_to_psd.jl
```diff
@@ -0,0 +1,199 @@
...
+# constraint is (tr(U) + tr(V), 2X) in NormNuclearCone.
+function MOI.get(model::MOI.ModelLike, ::MOI.ConstraintDual, bridge::NormSpectralBridge)
+    dual = MOI.get(model, MOI.ConstraintDual(), bridge.psd_index)
+    column_dim = bridge.column_dim
+    side_dim = bridge.row_dim + column_dim
+    t = sum(dual[trimap(i, i)] for i in 1:side_dim)
+    X = 2 * dual[[trimap(i, j) for j in 1:column_dim for i in (column_dim + 1):side_dim]]
+    return vcat(t, X)
+end
+
```
blegat: For ConstraintDualStart, I would set all `trimap(i, i)` to the dual of `t` divided by `2side_dim`. It might not be the only possibility but it works.

chriscoey: OK should I make another PR for that?

blegat: I'm wrong, that won't make the matrix PSD.
In fact, it's easier to discuss the ConstraintPrimalStart of NormNuclearBridge.
So suppose you have X and t set, we can assume that t >= sum sigma_i(X)
[A X'; X B] >= 0 is equivalent to A >= X' inv(B) X. If X = U S V', then, it's equivalent to
V' * A * V <= S * U' * inv(B) * U * S
then, by choosing A = V * S * V' and B = U * S * U', you get
S <= S * inv(S) * S which is satisfied.
As the trace must sum up to t, you set A = V * S * V' * (t / tr(S)) and B = U * S * U' * (t / tr(S).

chriscoey: if that works for constraint primal start of NormNuclear then should it also work for constraint dual start of NormSpectral?

blegat: Yes, it was just easier to reason in the primal than in the dual ^^

chriscoey: OK rather than having you explain the details to me, it's probably easier if you implement those new start functions

PR #979, opened 2019-12-26
Changed attribute from SolveTime to SolveTimeSec

### LINKED ISSUE #914: Rename SolveTime to SolveTimeSec
For consistency with `TimeLimitSec`, and because it's a good idea to include the unit in the name (e.g., to help prevent issues like https://github.com/JuliaOpt/SCS.jl/pull/158).

guilhermebodin:
This closes #914

### DIFF
--- a/docs/src/apireference.md
+++ b/docs/src/apireference.md
@@ -149,7 +149,7 @@ ObjectiveValue
 DualObjectiveValue
 ObjectiveBound
 RelativeGap
-SolveTime
+SolveTimeSec
 SimplexIterations
 BarrierIterations
 NodeCount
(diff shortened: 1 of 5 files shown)


mlubin:
Thanks for this. Could you add a deprecation warning using `@deprecate`? We should provide helpful error/deprecation messages when possible.

blegat:
We should probably wait for MOI v0.10

matbesancon:
> We should probably wait for MOI v0.10

if there is a deprecation this is not breaking

blegat:
> if there is a deprecation this is not breaking

Why is that ? DataStructures just did breaking changes with deprecation on a patch release and it affected the users of their dependencies: https://github.com/JuliaAlgebra/MultivariatePolynomials.jl/issues/129
This isn't ideal to cause warning in packages depending on your package.

matbesancon:
using `@deprecate` with the previous syntax allows code which ran on the last version to run on this one, especially when the breaking change is just a name

mlubin:
This change is breaking for solver wrappers. The deprecation works only for
users of MOI.

odow:
Closing in favor of #1303

Issue #980, opened 2019-12-27
Crash on Julia v1.1


blegat:
When MOI v0.9.8 is loaded with StaticArrays on Julia v1.1.1, the `crash` function belows crashes.
We can reproduce the crash without the `StaticArrays` package as follows:
```julia
import MathOptInterface
const MOI = MathOptInterface

abstract type StaticArray end
@inline (::Type{SA})(x...) where {SA <: StaticArray} = SA(x)

function crash()
    model = MOI.Utilities.Model{Float64}()
    bridged = MOI.Bridges.LazyBridgeOptimizer(MOI.Utilities.Model{Float64}())
    MOI.copy_to(bridged, model)
end

crash()
```
It gives
```
Internal error: encountered unexpected error in runtime:
TypeError(func=:<:, context="", expected=Type, got=T<:(T where T<:MathOptInterface.ScalarQuadraticFunction{T} where T))
rec_backtrace at /buildworker/worker/package_linux64/build/src/stackwalk.c:94
record_backtrace at /buildworker/worker/package_linux64/build/src/task.c:217 [inlined]
jl_throw at /buildworker/worker/package_linux64/build/src/task.c:417
jl_type_error_rt at /buildworker/worker/package_linux64/build/src/rtutils.c:118
jl_type_error at /buildworker/worker/package_linux64/build/src/rtutils.c:125
jl_f_issubtype at /buildworker/worker/package_linux64/build/src/builtins.c:408
typeof_tfunc at ./compiler/tfuncs.jl:444
typeof_tfunc at ./compiler/tfuncs.jl:447
typeof_tfunc at ./compiler/tfuncs.jl:442
typeof_tfunc at ./compiler/tfuncs.jl:447
jl_apply_generic at /buildworker/worker/package_linux64/build/src/gf.c:2197
jl_apply at /buildworker/worker/package_linux64/build/src/julia.h:1571 [inlined]
jl_f__apply at /buildworker/worker/package_linux64/build/src/builtins.c:556
builtin_tfunction at ./compiler/tfuncs.jl:1286
builtin_tfunction at ./compiler/tfuncs.jl:1200 [inlined]
abstract_call at ./compiler/abstractinterpretation.jl:593
abstract_eval_call at ./compiler/abstractinterpretation.jl:805
abstract_eval at ./compiler/abstractinterpretation.jl:890
typeinf_local at ./compiler/abstractinterpretation.jl:1135
[...]

PR #982, opened 2019-12-28
Workaround for issue #980

blegat:
Closes https://github.com/JuliaOpt/MathOptInterface.jl/issues/980


odow:
Add a test with the previous crash?

PR #989, opened 2020-01-03
Tests for vector specialized version of MOI.delete

henriquebecker91:
Contribution based in a discussion with @odow and @mlubin. I think the tests may cover more cases but will wait suggestions. Looking at the other tests seem like the style of choice is keeping things simple.

The gist of this contribution is that MOI.delete taking a vector may be specialized to be 'smarter' (i.e., not O(n^2) in some situations as it is now) but this 'smarter' code needs extra tests.

### DIFF
--- a/src/Test/UnitTests/modifications.jl
+++ b/src/Test/UnitTests/modifications.jl
@@ -372,17 +372,17 @@ function delete_variables_in_a_batch(model::MOI.ModelLike,
         c3: z >= 1.0
     """)
     x, y, z = MOI.get(model, MOI.ListOfVariableIndices())
-    @test MOI.is_valid(model, x) == true
-    @test MOI.is_valid(model, y) == true
-    @test MOI.is_valid(model, z) == true
+    @test MOI.is_valid(model, x)
+    @test MOI.is_valid(model, y)
+    @test MOI.is_valid(model, z)
     if config.solve
         MOI.optimize!(model)
         @test MOI.get(model, MOI.ObjectiveValue()) == 6.0
     end
     MOI.delete(model, [x, z])
-    @test MOI.is_valid(model, x) == false
-    @test MOI.is_valid(model, y) == true
-    @test MOI.is_valid(model, z) == false
+    @test !MOI.is_valid(model, x)
+    @test MOI.is_valid(model, y)
+    @test !MOI.is_valid(model, z)
     if config.solve
         MOI.optimize!(model)
         @test MOI.get(model, MOI.ObjectiveValue()) == 2.0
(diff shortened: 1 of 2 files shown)


src/Test/UnitTests/modifications.jl
```diff
@@ -353,4 +353,41 @@ function delete_variable_with_single_variable_obj(model::MOI.ModelLike,
...
+    @test MOI.is_empty(model)
+    MOIU.loadfromstring!(model,"""
+        variables: x, y, z
+        minobjective: 1.0 * x + 2.0 * y + 3.0 * z
+        c1: x >= 1.0
+        c2: y >= 1.0
+        c3: z >= 1.0
+    """)
+    x, y, z = MOI.get(model, MOI.ListOfVariableIndices())
+    @test MOI.is_valid(model, x) == true
```
mlubin: `@test MOI.is_valid(model, x) == true` is redundant, you can say `@test MOI.is_valid(model, x)`.

src/Test/UnitTests/modifications.jl
```diff
@@ -353,4 +353,41 @@ function delete_variable_with_single_variable_obj(model::MOI.ModelLike,
...
+    @test MOI.is_valid(model, y) == true
+    @test MOI.is_valid(model, z) == true
+    if config.solve
+        MOI.optimize!(model)
+        @test MOI.get(model, MOI.ObjectiveValue()) == 6.0
+    end
+    MOI.delete(model, [x, z])
+    @test MOI.is_valid(model, x) == false
```
mlubin: Likewise `@test !MOI.is_valid(model, x)`

henriquebecker91:
I did test with both Gurobi.jl and CPLEX.jl, both worked when I used their branchs merged with the most recent commit in their masters against MOI 0.9.8 (the master from one, or two, weeks ago). When I updated MOI for the most recent commit, Gurobi.jl tests kept working and CPLEX.jl tests stopped working for unrelated reasons (i.e., running runtests.jl with most recent MOI and CPLEX.jl master commits gives to me a "UndefVarError: AddSubMul not defined" error).

henriquebecker91:
I added what you asked @odow.

I updated the dependencies of MOI to the versions used in MOI@0.9.9 (when I started to implement this PR the version was 0.9.8) and fixed the "UndefVarError: AddSubMul not defined" error (the problem was I was not using the last version of MutableArithmetics). I executed the 'runtests.jl' of both CPLEX.jl and Gurobi.jl and all tests related to MathOptInterface passed. In my local copy I added some 'error' calls to the MOI tests added and to the MOI.delete specializations (inside CPLEX.jl and Gurobi.jl) so to confirm they are being called. They are.

Is some benchmark test needed to check how much difference is the performance of the specializations against the fallback?

henriquebecker91:
Do you want me to come with more cases before you merge? I really agree that more tests is better, but unfortunately I need to focus on other things now (PhD is ticking), if it is really necessary I can try to write some extra tests before going back to my things.

odow:
This test is fine. It catches the most common error of non-contiguous variables being deleted.

PR #993, opened 2020-01-05
add relative entropy cone

chriscoey:
This 1+2n-dimensional cone is useful for entropy optimization and has applications in e.g. signomial positivity certificates (see [here](https://arxiv.org/pdf/1907.00814.pdf) and [here](http://users.cms.caltech.edu/~venkatc/cs_rep_mathprog17.pdf)). I've implemented a bridge to n exponential cone (3-dim) constraints and 1 LP constraint.

Hypatia supports this cone due to what is claimed to be a self-concordant barrier on p5 of "Primal-Dual Interior-Point Methods for Domain-Driven Formulations" by Karimi & Tuncel, 2019 (they didn't provide a SC proof there but I'd be interested to see one).


chriscoey:
Let me guess... I need to redo the bridge as a set map bridge?

blegat:
> Let me guess... I need to redo the bridge as a set map bridge?

No, you create extra variables and more than one constraint so it's not a set map bridge.

chriscoey:
Understood, thanks!

chriscoey:
I'll attempt the starting values here next

chriscoey:
OK so I've done my best to implement starting values. AFAICT it requires the Lambert W function. This is likely to be needed in similar contexts for future cones to be added to MOI that use Bridges with exponential cone extended formulations. The Lambert W function comes from the registered/up-to-date https://github.com/jlapeyre/LambertW.jl package, which is a small dependency I've added.

chriscoey:
@blegat this is ready for review

chriscoey:
Note btw that relative entropy is the perspective of negative log, so this cone can also be used for modeling a sum of logs, and hence is useful for a natural reformulation of geometric programming problems.

odow:
So LambertW.jl is only one file, and under MIT, so potentially we could just include that file in MOI.Utilities? It hasn't been updated in > 1 year, and its `Project.toml` file is out-of-date.

src/Bridges/Constraint/relentr_to_exp.jl
```diff
@@ -0,0 +1,126 @@
...
+import LambertW
+function MOI.set(model::MOI.ModelLike, ::MOI.ConstraintDualStart, bridge::RelativeEntropyBridge, value)
+    MOI.set(model, MOI.ConstraintDualStart(), bridge.ge_index, value[1])
+    w_start = 1 + length(bridge.y)
+    for i in eachindex(bridge.y)
+        s_value = value[w_start + i]
+        t_value = value[1 + i]
+        r_value = exp(LambertW.lambertw(s_value / (t_value * ℯ))) * -t_value * ℯ
+        @assert -r_value * exp(s_value / r_value - 1) - t_value ≈ 0
+        MOI.set(model, MOI.ConstraintDualStart(), bridge.exp_indices[i], [r_value, s_value, t_value])
```
blegat: As `y` is free, its dual should be zero.
So we have `-r_i + value[1] == 0` since `y[i]` appears on both. Therefore `r_i == value[1]` so we don't need to use LambertW, we can just use `value[1]`.
Does this make sense ?

chriscoey:
@blegat's solution is better than mine so I changed that. thanks!

src/Bridges/Constraint/relentr_to_exp.jl
```diff
@@ -0,0 +1,116 @@
...
+    for i in eachindex(bridge.y)
+        exp_dual_i = MOI.get(model, attr, bridge.exp_indices[i])
+        dual[1 + i] = exp_dual_i[3]
+        dual[w_start + i] = exp_dual_i[2]
+    end
+    return dual
+end
+# Given constraint dual start of (u, v, w), constraint dual on GreaterThan is u
+# and on exponential cone constraint i is (r_i, w_i, v_i), but since y_i is free,
+# its dual is 0, so we have -r_i + value[1] == 0 hence r_i = value[1].
```
blegat: Add that r_i could also be computed using the Lambert W function but it we choose to use value[1] instead

docs/src/apimanual.md
```diff
@@ -92,6 +92,10 @@ sets recognized by the solver.
   ``\{ x \in \mathbb{R}^\mbox{dimension} : x \ge 0 \}``
 * **[`Nonpositives(dimension)`](@ref MathOptInterface.Nonpositives)**:
   ``\{ x \in \mathbb{R}^\mbox{dimension} : x \le 0 \}``
+* **[`NormInfinityCone(dimension)`](@ref MathOptInterface.NormInfinityCone)**:
```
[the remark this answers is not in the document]
blegat: Forget it, I misread

src/Bridges/Constraint/relentr_to_exp.jl
blegat: Otherwise, when we read this later, we might wonder why we are not computing it the other way

chriscoey: Done, good idea

PR #994, opened 2020-01-05
Implement starting values for NormOneBridge

blegat:
See https://github.com/JuliaOpt/MathOptInterface.jl/issues/684

### DIFF
--- a/test/Bridges/Constraint/norm_to_lp.jl
+++ b/test/Bridges/Constraint/norm_to_lp.jl
@@ -100,13 +100,17 @@ end
     MOIT.normone1vtest(bridged_mock, config)
     MOIT.normone1ftest(bridged_mock, config)
 
+    var_names = ["x", "y", "z"]
+    MOI.set(bridged_mock, MOI.VariableName(), MOI.get(bridged_mock, MOI.ListOfVariableIndices()), var_names)
+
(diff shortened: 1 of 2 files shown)


test/Bridges/Constraint/norm_to_lp.jl
```diff
@@ -100,13 +100,17 @@ end
     MOIT.normone1vtest(bridged_mock, config)
```
chriscoey: just a question: what's the point of running this test set here? it's not relevant to all the below tests right? I ask because in #976 I have multiple test sets but currently only run one of them in the bridge test file

blegat: It does not affect tests below but it still adds coverage. `normone1f` uses VAF functions and `normone1v` uses VOV. So this tests that the bridging and the getters of ConstraintPrimal and ConstrainDual work when the input function is VOV

chriscoey: OK I didn't realize it wasn't called elsewhere. I'll add both tests for spectral and for nuclear cones then.

chriscoey: wait no I misunderstood. it only works for the same test modulo the vov/vaf difference. so scrap what I said.

blegat: Yes, so in your case you'll have to redo the mock dual and primal values so probably not worth it

PR #996, opened 2020-01-06
RFC/WIP: add complex norm-1 and norm-infinity cones

chriscoey:
Hypatia supports real and complex norm-1 and norm-infinity cones (and real/complex versions of various other cones, such as spectral and PSD/rootdet/logdet). This is the least-complicated way to support these two cones, I think. If we agree on this soon then I can generalize the spectral/nuclear cones in a similar way after #976 is merged, and then we can discuss how to do it for Hermitian PSD/rootdet/logdet.

### DIFF
--- a/src/sets.jl
+++ b/src/sets.jl
@@ -211,6 +211,28 @@ end
 dual_set(s::NormOneCone) = NormInfinityCone(dimension(s))
 
 """
+    ComplexNormInfinityCone(dimension)
+
+The ``\\ell_\\infty``-norm of complex vectors cone ``\\{ (t,x) \\in \\mathbb{R}\\times\\mathbb{C}^{n} : t \\ge \\lVert x \\rVert_\\infty = \\max_i \\lvert x_i \\rvert \\}`` of real `dimension```{}=1+2n``.
+"""
+struct ComplexNormInfinityCone <: AbstractVectorSet
+    dimension::Int
+end
+
+dual_set(s::ComplexNormInfinityCone) = ComplexNormOneCone(dimension(s))
+
+"""
+    ComplexNormOneCone(dimension)
+
+The ``\\ell_1``-norm of complex vectors cone ``\\{ (t,x) \\in \\mathbb{R}\\times\\mathbb{C}^{n} : t \\ge \\lVert x \\rVert_\\infty_1 = \\sum_i \\lvert x_i \\rvert \\}`` of real `dimension```{}=1+2n``.
+"""
+struct ComplexNormOneCone <: AbstractVectorSet
+    dimension::Int
+end
+
+dual_set(s::ComplexNormOneCone) = ComplexNormInfinityCone(dimension(s))
+
+"""
     SecondOrderCone(dimension)
 
 The second-order cone (or Lorenz cone or ``\\ell_2``-norm cone) ``\\{ (t,x) \\in \\mathbb{R}^{dimension} : t \\ge \\lVert x \\rVert_2 \\}`` of dimension `dimension`.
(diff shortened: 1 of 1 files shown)


chriscoey:
I've expanded the descriptions of the sets as stacking real and imaginary parts of complex elements. This is the same idea Hypatia uses for these cones and for complex spectral/nuclear and hermitian PSD/logdet/rootdet cones (only complex elements ie not epigraphs or hermitian diagonals get two stacked elements).

blegat:
I am still undecided between this and using complex coefficients in the functions.
The argument against complex coefficients is that you need to convert the real diagonal elements of hermitian matrices to a complex value but it's similar to the SingleVariable functions that are converted to affine functions in Complements and Indicator to fit in a VAF.
The argument for complex coefficients is that you need to duplicate each set definition and bridges, even if the bridge work in both the complex and real case.

chriscoey:
It's not clear to me that the bridges won't need significant changes to work for real and complex cases anyway. 
Duplicating cone definitions isn't such a big deal if we structure the descriptions of the complex cases to reference the real cases and just describe the differences.

odow:
So my hesitations with adding these cones, and the other ones in this series of PRs are:

1) we don't know we're going to support complex numbers in JuMP, so it seems JuMPing (if you will) the gun to add complex cones.
2) We're adding cones that aren't obvious how widely they will be supported or used
3) Adding lots of cones increases the long-term maintenance burden of MOI
4) We probably have to add these cones to MathOptFormat

I get that these cones are useful for modeling, and Hypatia supports them. But I'm strongly in favor of adding them in Hypatia, or in a ConicModelingExtensions.jl package first, letting the definitions and issues settle down in that package, and then adding them to MOI in the long-term once it's obvious we want to support them.

cc @mlubin thoughts?

mlubin:
@blegat will have a better estimate than I do of the extra maintenance burden from additional sets.

In general, if a set is supported by only one solver and we have no short-term plans to provide special modeling support for it in JuMP or another upstream modeling package, then I don't see a strong motivation for including it in MOI.

@chriscoey, is there a technical reason why it helps you to have these cones defined in MOI instead of in Hypatia or a hypothetical ConicModelingExtensions package?

chriscoey:
@mlubin No there's no reason I need them in MOI, I was doing it because I believe they are useful modeling constructs and bridges for the community to have. Some are more useful than others of course, and there's a chicken and egg problem in that no one will use them if they don't exist.

One goal of the complex cones discussion here is deciding how to handle the Hermitian PSD cone, for which there exists a less efficient real PSD reformulation.

odow:
> I believe they are useful modeling constructs and bridges for the community to have.

I think we all agree with this.

> no one will use them if they don't exist.

We're debating between
```Julia
using JuMP, Hypatia
model = Model()
@variable(model, x in MOI.MyCoolCone())
```
and 
```Julia
using JuMP, Hypatia, ConicModelingExtensions
model = Model()
@variable(model, x in ConicModelingExtensions.MyCoolCone())
# or even
@variable(model, x in MyCoolCone())
```

One goal of MOI was to make it extensible so that people could add new cones without having to modify JuMP/MOI.

chriscoey:
So would you propose moving sets like logdet and rootdet and geometric mean to that package? What would be the criteria for which cones belong where?

odow:
If it's in CVX(py), it's a solid candidate: e.g., https://www.cvxpy.org/api_reference/cvxpy.atoms.other_atoms.html#geo-mean

I don't think we should necessarily move cones out of MOI. I'm just advocating adding them more carefully in future.

ericphanson:
> @mlubin No there's no reason I need them in MOI, I was doing it because I believe they are useful modeling constructs and bridges for the community to have. Some are more useful than others of course, and there's a chicken and egg problem in that no one will use them if they don't exist.

FWIW, I think Convex.jl could definitely depend on a hypothetical ConicModelingExtensions package and using those cones for formulating its problems, which might negate some of the visibility problem. Convex.jl also has complex-number support (although we just reformulate everything as real before passing to MOI, currently).

chriscoey:
this can be revisited later if there is a desire. depends on what the plan is for complex modeling in MOI.

Issue #977, opened 2019-12-24
Set request: WeightedPowerMeanCone


chriscoey:
~~The power cone in MOI is only 3-dimensional, but in some places it is defined in a much more general way, such as in the Conic Benchmark Format (see [CBF docs](http://cblib.zib.de/doc/format3.pdf) page 13). I'd propose to use that definition (Hypatia [uses it too](https://github.com/chriscoey/Hypatia.jl/blob/master/src/Cones/power.jl)), even though arguably the ability to use the norm on the RHS might be of questionable utility.~~ edit: out of scope

The geometric mean cone in MOI is (n+1)-dimensional but only allows powers 1/n. ~~, when really it could use any vector of powers on the unit simplex (positive values summing to 1 - see "Constructing Self-Concordant Barriers for Convex Cones" by Nesterov; Hypatia [supports it](https://github.com/chriscoey/Hypatia.jl/blob/master/src/Cones/hypogeomean.jl)). We could either generalize the geomean cone for more general powers, or add a new cone (called what? maybe GeneralizedGeometricMeanCone? I don't love it).~~ edit: out of scope

edit: Proposal here is for the (1+n)-dim WeightedPowerMeanCone ``(y, x) in R x R_{>= 0}^n : y^sum(alpha) <= prod x_i^alpha_i``. This is the hypograph of a concave product of powers on unit simplex ``alpha/sum(alpha)``.

chriscoey:
Anyone have thoughts on this?

chriscoey:
Given how long it takes to add a cone and associated bridges, I lack the appetite for adding the CBFv3 n-dimensional power cone of questionable utility, but I will try adding the generalized geomean cone in #997

blegat:
We could have
* PowerCone1: prod t_i^(alpha_i)  >= ||x||_2, alpha in interior of simplex
* PowerCone2: prod t_i^(alpha_i) >= x, alpha in interior of simplex

Then we have the bridges:
1) PowerCone1 is bridged to PowerCone2 and SOC
2) PowerCone2 is bridged to PowerCone
3) GeometricMeanCone is bridged to PowerCone2 (just sets all alpha_i to 1/n)

Then we may not need https://github.com/JuliaOpt/MathOptInterface.jl/issues/406 as it's just a combination of bridges 2) and 3).

We need to find good names as PowerCone1 and PowerCone2 cannot both be named GeneralizedPowerCone

chriscoey:
[...]

PR #997, opened 2020-01-06
RFC/WIP: add weighted power mean cone

chriscoey:
This is a generalization of the geometric mean cone for any vector of positive exponents (rather than just [1/n, 1/n, ..., 1/n]). It gives pretty close to all the utility (that I can see) of n-dimensional power cones. Hypatia supports this cone thanks to the existence of a SC barrier. When the definition is decided, I can add bridges, either to 3D exponential cone constraints or to 3D MOI power cone constraints.

### DIFF
--- a/src/sets.jl
+++ b/src/sets.jl
@@ -235,11 +235,22 @@ dual_set(s::RotatedSecondOrderCone) = copy(s)
 """
     GeometricMeanCone(dimension)
 
-The geometric mean cone ``\\{ (t,x) \\in \\mathbb{R}^{n+1} : x \\ge 0, t \\le \\sqrt[n]{x_1 x_2 \\cdots x_n} \\}`` of dimension `dimension```{}=n+1``.
+The geometric mean cone ``\\{ (t,x) \\in \\mathbb{R}^{1+n} : x \\ge 0, t \\le \\sqrt[n]{x_1 x_2 \\cdots x_n} \\}`` of dimension `dimension```{}=1+n``.
 """
 struct GeometricMeanCone <: AbstractVectorSet
     dimension::Int
 end
+
+"""
+    GeneralizedGeometricMeanCone{T <: Real}(exponents::Vector{T})
+
+The generalized geometric mean cone ``\\{ (t,x) \\in \\mathbb{R}^{1+n} : x \\ge 0, t^{sum(exponents)} \\le \\prod_i x_i^{exponents_i} \\}`` of dimension ``1+n`` with parameter vector `exponents```{}\\in \\mathbb{R}_{++}^{n}``.
+"""
+struct GeneralizedGeometricMeanCone{T <: Real} <: AbstractVectorSet
+    exponents::Vector{T}
+end
+
+dimension(s::GeneralizedGeometricMeanCone) = 1 + length(s.exponents)
 
 """
     ExponentialCone()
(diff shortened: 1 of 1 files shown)


src/sets.jl
```diff
@@ -235,12 +235,23 @@ dual_set(s::RotatedSecondOrderCone) = copy(s)
...
+The geometric mean cone ``\\{ (t,x) \\in \\mathbb{R}^{1+n} : x \\ge 0, t \\le \\sqrt[n]{x_1 x_2 \\cdots x_n} \\}`` of dimension `dimension```{}=1+n``.
 """
 struct GeometricMeanCone <: AbstractVectorSet
     dimension::Int
 end
 
+"""
+    GeneralizedGeometricMeanCone{T <: Real}(exponents::Vector{T})
+
+The generalized geometric mean cone ``\\{ (t,x) \\in \\mathbb{R}^{1+n} : x \\ge 0, t^{sum(exponents)} \\le \\prod_i x_i^{exponents_i} \\}`` of dimension ``1+n`` with parameter vector `exponents```{}\\in \\mathbb{R}_{++}^{n}``.
```
blegat: What is `R_{++}` ?

blegat: Maybe `R_{>0}` is easier to understand

src/sets.jl
```diff
@@ -235,12 +235,23 @@ dual_set(s::RotatedSecondOrderCone) = copy(s)
...
 
-The geometric mean cone ``\\{ (t,x) \\in \\mathbb{R}^{n+1} : x \\ge 0, t \\le \\sqrt[n]{x_1 x_2 \\cdots x_n} \\}`` of dimension `dimension```{}=n+1``.
+The geometric mean cone ``\\{ (t,x) \\in \\mathbb{R}^{1+n} : x \\ge 0, t \\le \\sqrt[n]{x_1 x_2 \\cdots x_n} \\}`` of dimension `dimension```{}=1+n``.
 """
 struct GeometricMeanCone <: AbstractVectorSet
     dimension::Int
 end
 
+"""
+    GeneralizedGeometricMeanCone{T <: Real}(exponents::Vector{T})
```
blegat: How would we name the power cone where we have ||x|| in the rhs ?

blegat: According to https://en.wikipedia.org/wiki/Generalized_mean#Geometric_mean, we should call it WeightedPowerMeanCone

chriscoey:
xref discussion at #977

src/sets.jl
```diff
@@ -235,12 +235,23 @@ dual_set(s::RotatedSecondOrderCone) = copy(s)
...
     dimension::Int
 end
 
+"""
+    WeightedPowerMeanCone(exponents::Vector{<:Real})
+
+The generalized geometric mean cone ``\\{ (t,x) \\in \\mathbb{R}^{1+n} : x \\ge 0, t^{sum(exponents)} \\le \\prod_i x_i^{exponents_i} \\}`` of dimension ``1+n`` with positive parameter vector `exponents```{}\\in \\mathbb{R}_{>0}^n``.
+"""
+struct WeightedPowerMeanCone{T <: Real} <: AbstractVectorSet
+    exponents::Vector{<:Real}
```
blegat: Shouldn't this be `Vector{T}` ?

blegat:
So in `MOIU.Model` macro definition, we should add something like `WeightedPowerMeanCone{Int}` in the untyped vector cones, and `WeightedPowerMeanCone` in the typed vector cones.

chriscoey:
this can be revisited later if there is a desire. xref #977

Issue #863, opened 2019-08-30
Abstract / alternative linear operator storage


matbesancon:
The current way of representing `A x` is with a `VectorAffineFunction`, which has the following structure:

```julia
MOI.VectorAffineFunction(
    [MOI.VectorAffineTerm(
        row_idx,
        ScalarAffineTerm(
            coeff,
            col_idx,
        )
    )],
    [constants]
)
```

This has a lot of nested structure and imposes a fixed concrete layout:  

- one cannot take advantage of specific matrix structures
- data gets moved around twice, from the matrix to a `VectorAffineTerm`, then from `VAT` to the solver.

Could we have an abstract linear operator function, `VectorLinearOperator`, with possibly a bridge to `VectorAffineTerm` to avoid excessive burden on solvers. 
It would let solvers query the information as they want, `rows(vector_linear.A)` or `cols(vector_linear.A)` or through scalar `vector_linear.A[i, j]`

blegat:
Do you know a solver which could take advantage of knowing the matrix structure ?

chriscoey:
When Hypatia is released, I will chime in here. Not quite ready to yet.

matbesancon:
From the Gurobi doc:
> A brief note for users of the Gurobi MATLAB® and R interfaces: our interfaces to these languages are built around the assumption that you will use the rich matrix-oriented capabilities of the underlying languages to build your optimization models. Thus, our examples for these languages don’t attempt to show you how to build models. We have instead chosen to provide a few simple examples that demonstrate how to pass matrices into our interface.

With this example in R:
https://www.gurobi.com/documentation/8.1/examples/mip_r.html

Where the matrix is passed to the model, instead of constructing row by row

blegat:
Is Gurobi actually using this matrix or is it just transforming it and for the algorithm it won't change anything whether you give this matrix or a VectorAffineFunction ? Is it just to gain model generation time or also solve time ?

matbesancon:
Not sure there is a way to know, I would say model generation time only.

mtanneau:
> Do you know a solver which could take advantage of knowing the matrix structure ?

Tulip can. Actually, knowing the matrix structure is the main reason for having coded Tulip in the first place :)

> Is it just to gain model generation time or also solve time ?

I'm only 99.99% sure it would only affect model generation time.

matbesancon:
> Is it just to gain model generation time or also solve time ?

[...]

PR #998, opened 2020-01-06
WIP/DNMY: add linear matrix inequality cone

chriscoey:
Hypatia supports this cone thanks to an SC barrier. It is parametrized by a vector of symmetric matrices of equal size. The dimension of the cone is the number of such matrices, which is potentially and in many practical cases much smaller than the dimension of the PSD cone that would be needed in the SDP reformulation (which I will add as a Bridge after the definition is settled).

### DIFF
--- a/src/sets.jl
+++ b/src/sets.jl
@@ -288,6 +288,17 @@ function Base.:(==)(set1::S, set2::S) where S <: Union{PowerCone, DualPowerCone}
 end
 
 """
+    LinearMatrixInequalityCone{T <: Real}(component::Vector{Symmetric(T, Matrix{T})})
+
+The linear matrix inequality (LMI) cone ``\\{ x \\in \\mathbb{R}^n : sum_i x_i component_i \\succeq 0 \\}`` with `n`-dimensional parameter vector `component` of symmetric matrices of equal dimensions.
+"""
+struct LinearMatrixInequalityCone{T <: Real} <: AbstractVectorSet
+    component::Vector{Symmetric(T, Matrix{T})}
+end
+
+dimension(s::LinearMatrixInequalityCone) = length(component)
+
+"""
     abstract type AbstractSymmetricMatrixSetTriangle <: AbstractVectorSet end
 
 Abstract supertype for subsets of the (vectorized) cone of symmetric matrices,


blegat:
The cone `[x, y] in LMICone(A, B)` can be written as `Ax + By in PSD` so the LMI cone seems to be a way to rewrite a VAF-in-PSD into a VOV constraint.
Why is it needed and why would we do this for the PSD cone and not for other cones ?
We could also have a `LinearVectorSOC` for which `[x, y] in LVSOC(A, B)` is equivalent to `xA + yB in SOC`

chriscoey:
I agree that there are many many potential cone definitions we could add to MOI, and that we don't want that sort of proliferation. So the goal is to add common ones that are convenient and provide efficiency advantages over their equivalent extended formulation representations. 

It's obvious that this LMI set has a SDP extended formulation. There are several advantages to being able to use this set, in order from least convincing to most convincing in my mind:
1. This LMI set is a very commonly-used set, and it is convenient to have it because it is slightly easier to construct than the equivalent SDP extended formulation.
2. This representation in MOI is more efficient than the equivalent SDP representation. Think about the case where the side dimension of the symmetric matrix parameters is large, but the number of such parameters n (the dimension of the variable vector x) is small. It will be faster and more memory-efficient to construct the LMI function than the PSD function.
3. The SC barrier evaluations for this set are generally much more efficient than barrier evaluations for the PSD set, especially in the case of large side dimension of the parameter matrices, because we have knowledge of the structure of the set that is lost when the SDP formulation is used. The algorithmic efficiency gain is analogous to that of e.g. WSOS cone formulations vs PSD formulations for WSOS problems.

We'll provide computational results in future that should support the 3rd argument. Until then let's leave this PR unfinished and unmerged.

blegat:
> It will be faster and more memory-efficient to construct the LMI function than the PSD function.

Why is that ? The VAF-in-PSD and VOV-in-LMI are represented the same way in memory if the matrices of LMI are sparse matrices, aren't they ?
Since the LMI set of this PR uses dense matrices instead of sparse matrices, it will in fact use more memory than VAF-in-PSD.

chriscoey:
We are planning to relax the current type restriction on the components, to allow for a mixture of identity, diagonal, block/sparse/dense symmetric, etc etc. For each component matrix, we just need it to have symmetry, even if it is not a Symmetric type. Then we can dispatch to whatever is most efficient for each component in the barrier evaluations. I'll let you know when we make that generalization in Hypatia.

chriscoey:
@blegat see some of the combinations of sparse/dense/structured matrix types that work as parameters to this cone in Hypatia: https://github.com/chriscoey/Hypatia.jl/blob/f1070b6d49b236de07f85feef65c19e7a5cb12f4/test/nativeinstances.jl#L1027

We are working on applied examples to demonstrate computational efficiency gains and I'll update you when we have results

blegat:
I understand that there may be benefits with structured matrices since it allows to communicate the structure while VAF-in-PSD forces the structure to be SparseMatrixCSC.
However, we could say the same for other cones, a solver might be able to exploit `A * x in SOC` where `A` has some specific structure.
The faith of this PR seems to be linked to this discussion: https://github.com/JuliaOpt/MathOptInterface.jl/issues/863
Until we find an agreement on how we go with https://github.com/JuliaOpt/MathOptInterface.jl/issues/863, this set could start as a solver specific one.

chriscoey:
That's reasonable. I'll close the PR and keep it in mind for the future.

Issue #987, opened 2020-01-01
[RFC] supports_constrained_variable


blegat:
We had a talk with @odow and @mlubin about constrained variables, here is what we came up with:

Currently, there is no way to differentiate whether a constrained variable can be added or whether constraint can be added on existing variables.
However,
* solvers such as SDPNAL, CSDP, SDPA can create constrained variables in Nonnegatives or PSD but cannot create VectorOfVariables-in-Nonnegatives or PSD constraints
* SDPT3 can create constrained variables in Nonnegatives, SecondOrderCone or PSD but cannot create VectorOfVariables-in-Nonnegatives, SecondOrderCone or PSD constraints
* Mosek can create constrained variables in PSD but cannot create VectorOfVariables-in-PSD constraints
* SDPNAL can create SingleVariable-in-Interval constraints but cannot create constrained variables in Interval (as free variables are not supported). The SingleVariable-in-Interval need to be added on Nonnegatives or PSD variables.

Because of these issues, if you call `MOI.add_constrained_variable(optimizer::LazyBridgeOptimizer{SDPNAL.Optimizer}, MOI.Interval(1.0, 2.0))`, the bridges will let the call go through as `MOI.supports(::SDPNAL.Optimizer, ::Type{SingleVariable}, ::Type{Interval{Float64}})` is true and then there will be an error.
Another issue is that the user could create constrained variables in PSD and then constrain some of them in SecondOrderCone in a CachingOptimizer. Then when the cache is copied from the cache to Mosek, it might add constrained variables in SecondOrderCone then then the PSD constraint will need to be added as a constraint as some of the variables have already been added. Then it will give an error.

To solve this issue, we can add
```julia
supports_constrained_variable(model::ModelLike, S::Type{<:AbstractScalarSet}) = supports_constraint(model, SingleVariable, S)
supports_constrained_variables(model::ModelLike, S::Type{<:AbstractVectorSet}) = supports_constraint(model, VectorOfVariables, S)
```
so that a solver can override it if it's not the same.

Then in `LazyBridgeOptimizer`, `supports_constrained_variable` will be used instead of `supports_constraint` when appropriate.
And in `copy_to`, we need to first add as constrained variables the set that are supported as constrained variables but not as constraint.
For the rest of the set, we could compute, for each set type, the difference between the bridging cost of adding constrained variables vs the bridging cost of adding constraints.
[...]

PR #1004, opened 2020-01-18
Add supports_constrained_variable(s)

blegat:
See https://github.com/JuliaOpt/MathOptInterface.jl/issues/987

### DIFF
--- a/docs/src/apimanual.md
+++ b/docs/src/apimanual.md
@@ -1146,7 +1146,7 @@ then it should only implement [`add_constrained_variable`](@ref) and not
 [`add_variable`](@ref) nor [`add_constraint`](@ref) for
 [`SingleVariable`](@ref)-in-`typeof(set)`. In addition, it should implement
 `supports_add_constrained_variables(::Optimizer, ::Type{Reals})` and return
-`false` so that free variables are bridged, see
+`false` so that these variables are bridged, see
 [`supports_add_constrained_variables`](@ref).
 
 ### Handling duplicate coefficients
--- a/docs/src/apimanual.md
+++ b/docs/src/apimanual.md
@@ -864,7 +864,7 @@ read!(io, src_2)
 ### Duals
 
 Conic duality is the starting point for MOI's duality conventions. When all functions are affine (or coordinate projections), and all constraint sets are closed convex cones, the model may be called a conic optimization problem.
-For conic-form minimization problems, the primal is:
+For a minimization problem in geometric conic form, the primal is:
 
 ```math
 \begin{align}
@@ -874,7 +874,7 @@ For conic-form minimization problems, the primal is:
 \end{align}
 ```
 
-and the dual is:
+and the dual is a maximization problem in standard conic form:
 
 ```math
 \begin{align}
@@ -888,7 +888,7 @@ and the dual is:
 
 where each ``\mathcal{C}_i`` is a closed convex cone and ``\mathcal{C}_i^*`` is its dual cone.
 
-For conic-form maximization problems, the primal is:
+For a maximization problem in geometric conic form, the primal is:
 ```math
 \begin{align}
 & \max_{x \in \mathbb{R}^n} & a_0^T x + b_0
@@ -897,7 +897,7 @@ For conic-form maximization problems, the primal is:
 \end{align}
 ```
 
-and the dual is:
+and the dual is a minimization problem in standard conic form:
 
 ```math
 \begin{align}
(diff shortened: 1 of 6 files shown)


src/variables.jl
```diff
@@ -34,6 +34,39 @@ done in the current state of the model `model`.
...
+
+By default, this function falls back to
+`supports_add_constrained_variables(model, Reals) &&
+supports_constraint(model, MOI.SingleVariable, S)` which is the correct
+definition for most models.
+
+## Example
+
+Suppose that a solver supports only two kind of variables: binary variables
+and continuous variables with a lower bound. If the solver decide not to
```
mlubin: s/decide/decides/

src/variables.jl
```diff
@@ -34,6 +34,39 @@ done in the current state of the model `model`.
...
+support `SingleVariable`-in-`Binary` and `SingleVariable`-in-`GreaterThan`
+constraints, it only has to implement `add_constrained_variable` for these
+two sets which prevents the user to add both a binary constraint and a
+lower bound on the same variable. Moreover, if the user adds a
+`SingleVariable`-in-`GreaterThan` constraint, it will transparently be bridged
```
mlubin: "it will be transparently bridged" depends on how bridges are set up. It's more precise to say that implementing this interface (i.e., `supports_add_constrained_variables`) enables the constraint to be transparently bridged.

src/variables.jl
```diff
@@ -55,6 +88,59 @@ function add_constrained_variable(model::ModelLike, set::AbstractScalarSet)
...
+created with [`add_constraint`](@ref).
+
+By default, if `S` is `Reals` then this function returns `true` and otherwise,
+it falls back to `supports_add_constrained_variables(model, Reals) &&
+supports_constraint(model, MOI.VectorOfVariables, S)` which is the correct
+definition for most models.
+
+## Example
+
+In the standard conic form, the variables are grouped into several cones
```
mlubin: Which standard conic form is this? A reference or link to an example (e.g., SCS) might help.

src/variables.jl
```diff
@@ -55,6 +88,59 @@ function add_constrained_variable(model::ModelLike, set::AbstractScalarSet)
...
+and the constraints are affine equality constraints.
+If `Reals` is not one of the cones supported by the solvers then it needs
+to implement `supports_add_constrained_variables(::Optimizer, ::Type{Reals}) = false`
+as free variables are not supported.
+The solvers should then implement
+`supports_add_constrained_variables(::Optimizer, ::Type{<:SupportedCones}) = true`
+where `SupportedCones` is the union of all cone types that are supported
+but it should not implement
```
mlubin: This is a run-on sentence and a bit confusing. You need a comma before "but". A semi-colon would work also.
"should not implement" doesn't follow from "should return `false`". The solver could implement it to return false (but that's not necessary because it's the default). Please tidy up the logic.

src/variables.jl
```diff
@@ -55,6 +88,59 @@ function add_constrained_variable(model::ModelLike, set::AbstractScalarSet)
...
+`supports_constraint(::Type{VectorOfVariables}, Type{<:SupportedCones})`
+as it should return `false`.
+This prevents the user to constrain the same variable in two different cones.
+When a `VectorOfVariables`-in-`S` is added, the variables of the vector
+have already been created so they already belong to given cones.
+The constraint will therefore be bridged by adding slack variables in `S`
```
mlubin: "If bridges are enabled"

src/variables.jl
```diff
@@ -55,6 +88,59 @@ function add_constrained_variable(model::ModelLike, set::AbstractScalarSet)
...
+and equality constraints ensuring that the slack variables are equal to the
+corresponding variables of the given constraint function.
+
+Note that there may also be sets for which
+`!supports_add_constrained_variables(model, S)` and
+`supports_constraint(model, MOI.VectorOfVariables, S)`.
+For instance, suppose a solver supports positive semidefinite variable
+constraints and two types of variables: binary variables and nonnegative
+variables. Then the solver should support adding
+`VectorOfVariables`-in-`PositiveSemidefiniteConeTriangle` constraints but it
```
mlubin: Nit: comma before "but"

src/variables.jl
```diff
@@ -55,6 +88,59 @@ function add_constrained_variable(model::ModelLike, set::AbstractScalarSet)
...
+should not support creating variables constrained to belong to the
+`PositiveSemidefiniteConeTriangle` as free variables are not supported.
```
mlubin: Reword mention of "free variables". 
Maybe: "because variables in `PositiveSemidefiniteConeTriangle` are not necessarily binary or non-negative".

docs/src/apimanual.md
```diff
@@ -1127,25 +1127,27 @@ MOI.set(model, MyPackage.PrintLevel(), 0)
...
-In addition, it should implement `supports_add_constrained_variables(::Optimizer,
-::Type{Reals})` and return `false` so that free variables are bridged,
-see [`supports_add_constrained_variables`](@ref).
+if `model` does not require that variables be constrained when they are created.
+However, if `model` requires that variables be constrained when they're created,
+then it should only implement [`add_constrained_variable`](@ref) and not
+[`add_variable`](@ref) nor [`add_constraint`](@ref) for
+[`SingleVariable`](@ref)-in-`typeof(set)`. In addition, it should implement
+`supports_add_constrained_variables(::Optimizer, ::Type{Reals})` and return
+`false` so that free variables are bridged, see
```
mlubin: "free variables" slipped through. I think "these variables" works fine here.

PR #1008, opened 2020-01-22
Add OptimizerWithAttributes

blegat:
See https://github.com/JuliaOpt/JuMP.jl/issues/2138#issuecomment-575680880
We might consider the name `ParametrizedOptimizer` instead of `UninstantiatedOptimizer`.

### DIFF
--- a/src/instantiate.jl
+++ b/src/instantiate.jl
@@ -90,7 +90,8 @@ Create an instance of optimizer represented by [`UninstantiatedOptimizer`](@ref)
 with all the bridges defined in the MathOptInterface.Bridges submodule enabled
 with coefficient type `T`.
 If `!MOI.Utilities.supports_default_copy_to(optimizer, with_names)` then a
-[`CachingOptimizer`](@ref) is added to store a cache of the bridged model.
+[`Utilities.CachingOptimizer`](@ref) is added to store a cache of the bridged
+model.
 """
 function instantiate_with_bridges(uninstantiated, with_names::Bool=false,
                                   T::Type=Float64)
--- a/src/instantiate.jl
+++ b/src/instantiate.jl
@@ -13,7 +13,7 @@ struct OptimizerWithAttributes
     # * `Function`: a function, or
     # * `DataType`: a type, or
     # * `UnionAll`: a type with missing parameters.
-    optimizer
+    optimizer_constructor
     params::Vector{Pair{AbstractOptimizerAttribute,Any}}
 end
 
@@ -72,7 +72,7 @@ Create an instance of optimizer represented by [`OptimizerWithAttributes`](@ref)
 Then check that the type returned is an empty [`AbstractOptimizer`](@ref).
 """
 function _instantiate_and_check(optimizer_constructor::OptimizerWithAttributes)
-    optimizer = _instantiate_and_check(optimizer_constructor.optimizer)
+    optimizer = _instantiate_and_check(optimizer_constructor.optimizer_constructor)
     for param in optimizer_constructor.params
         set(optimizer, param.first, param.second)
     end
@@ -85,15 +85,18 @@ end
                 with_names::Bool=false)
 
 Creates an instance of optimizer either by calling
-`optimizer_constructor.optimizer()` and setting the parameters in
+`optimizer_constructor.optimizer_constructor()` and setting the parameters in
 `optimizer_constructor.params` if `optimizer_constructor` is a
 [`OptimizerWithAttributes`](@ref) or by calling `optimizer_constructor()`
 if `optimizer_constructor` is function.
 
 If `with_bridge_type` is not `nothing`, it enables all the bridges defined in
 the MathOptInterface.Bridges submodule with coefficient type
-`with_bridge_type`. If
-`!MOI.Utilities.supports_default_copy_to(optimizer, with_names)` then a
+`with_bridge_type`.
+
+If the optimizer created by `optimizer_constructor` does not support loading the
+problem incrementally or does not support names (see
+[`Utilities.supports_default_copy_to`](@ref)) then a
 [`Utilities.CachingOptimizer`](@ref) is added to store a cache of the bridged
 model.
 """
--- a/src/instantiate.jl
+++ b/src/instantiate.jl
@@ -22,12 +22,6 @@ _to_param(param::Pair{String}) = RawParameter(param.first) => param.second
 function _to_param(param::Pair)
     error("Expected an optimizer attribute or a string, got `$(param.first)` which is a `$(typeof(param.first))`.")
 end
-function _add_params(params::Vector{Pair{AbstractOptimizerAttribute,Any}}) end
-function _add_params(params::Vector{Pair{AbstractOptimizerAttribute,Any}},
-                     param::Pair, other_params::Vararg{Pair, N}) where N
-    push!(params, _to_param(param))
-    _add_params(params, other_params...)
-end
 
 """
     OptimizerWithAttributes(optimizer_constructor, params::Pair...)
@@ -36,24 +30,13 @@ Create an [`OptimizerWithAttributes`](@ref) with the parameters `params`.
 """
 function OptimizerWithAttributes(optimizer_constructor, args::Vararg{Pair, N}) where N
     if !applicable(optimizer_constructor)
-        error(_instantiate_not_callable_message)
+        error(_INSTANTIATE_NOT_CALLABLE_MESSAGE)
     end
-    params = Pair{AbstractOptimizerAttribute,Any}[]
-    _add_params(params, args...)
+    params = Pair{AbstractOptimizerAttribute,Any}[_to_param(arg) for arg in args]
     return OptimizerWithAttributes(optimizer_constructor, params)
 end
 
-function (optimizer_constructor::OptimizerWithAttributes)(;
-    with_bridges::Bool=false, with_names::Bool=false, T::Type=Float64)
-    if with_bridges
-        return instantiate_with_bridges(optimizer_constructor;
-                                        with_names=with_names, T=T)
-    else
-        return instantiate(optimizer_constructor)
-    end
-end
-
-const _instantiate_not_callable_message =
+const _INSTANTIATE_NOT_CALLABLE_MESSAGE =
     "The provided `optimizer_constructor` is invalid. It must be callable with zero" *
     "arguments. For example, \"Ipopt.Optimizer\" or" *
     "\"() -> ECOS.Optimizer()\". It should not be an instantiated optimizer " *
@@ -61,13 +44,14 @@ const _instantiate_not_callable_message =
     "(Note the difference in parentheses!)"
 
 """
-    instantiate(optimizer_constructor)
+    instantiate_and_check(optimizer_constructor)
 
 Create an instance of optimizer by calling `optimizer_constructor`.
+Then check that the type returned is an empty [`AbstractOptimizer`](@ref).
 """
-function instantiate(optimizer_constructor)
+function instantiate_and_check(optimizer_constructor)
     if !applicable(optimizer_constructor)
-        error(_instantiate_not_callable_message)
+        error(_INSTANTIATE_NOT_CALLABLE_MESSAGE)
     end
     optimizer = optimizer_constructor()
     if !isa(optimizer, AbstractOptimizer)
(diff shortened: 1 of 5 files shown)


src/instantiate.jl
```diff
@@ -0,0 +1,104 @@
+"""
+    struct UninstantiatedOptimizer
+        optimizer
```
mlubin: I know I suggested this name, but it's misleading because it's a function that returns an optimizer, not an optimizer itself. What about `optimizer_constructor`?

src/instantiate.jl
```diff
@@ -0,0 +1,104 @@
...
+        params::Vector{Pair{AbstractOptimizerAttribute,<:Any}}
+    end
+
+Object representing an optimizer, created by `optimizer()` parametrized by
```
mlubin: It's ambiguous to say that it "represents" an optimizer. I would say that it groups an optimizer constructor with a set of optimizer attributes.

src/instantiate.jl
```diff
@@ -0,0 +1,104 @@
...
+_to_param(param::Pair{Symbol}) = _to_param(string(param.first) => param.second)
+function _add_params(params::Vector{Pair{AbstractOptimizerAttribute,Any}}) end
+function _add_params(params::Vector{Pair{AbstractOptimizerAttribute,Any}},
+                     param::Pair, other_params::Vararg{Pair, N}) where N
+    push!(params, _to_param(param))
+    _add_params(params, other_params...)
+end
+
+"""
+    parametrize(uninstantiated, params::Pair...)
```
mlubin: What is `uninstantiated`? `optimizer_constructor` is clearer.

mlubin: To parametrize means to describe in terms of parameters, i.e., listing all possible parameters. This is different in my mind from specifying some subset of parameters.
I would suggest `optimizer_with_parameters`.

odow:
So, clarifying our discussion earlier, I am suggesting something like:
```Julia
struct OptimizerWithAttributes
    optimizer
    attributes::Vector{Pair{AbstractOptimizerAttribute,<:Any}}
    function OptimizerWithAttributes(optimizer::ModelLike, args::Pair...)
        attributes = Pair{AbstractOptimizerAttribute,<:Any}[]
        for arg in args
            if typeof(arg.first) == String
                push!(attributes, MOI.RawParameter(arg.first) => arg.second)
            elseif typeof(arg.first) <: AbstractOptimizerAttribute
                push!(attributes, arg)
            else
                error("A nice error message.")
            end
        end
        return new(optimizer, attributes)
    end
end

function (o::OptimizerWithAttributes)(; bridge_constraints::Bool = false)
    opt = o.optimizer()
    optimizer = bridge_constraints ? Bridges.full_bridge_optimizer(opt, Float64) : opt    
    for (attr, value) in o.attributes
        set(optimizer, attr, value)
    end
    return optimizer
end

Model(Gurobi.Optimizer)
Model(() -> Gurobi.Optimizer(env))
Model(OptimizerWithAttributes(Gurobi.Optimizer, "OutputFlag" => 0))
Model(OptimizerWithAttributes(() -> Gurobi.Optimizer(env), "OutputFlag" => 0))
```

blegat:
We might want to keep the `instantiate` function in addition to the function call syntax because if `OT <: MOI.AbstractOptimizer` then
* `OT()` creates an optimizer but does not do any check while `instantiate(OT)` gives a nice error message if `OT` is not callable, or it does not create `AbstractOptimizer` or if the model created is not empty
* `OT(; with_bridges)` will fail but `instantiate_with_bridges(OT)` works.

odow:
I'd be fine with
```Julia
MOI.instantiate(::Type{<:MOI.ModelLike}; with_bridges::Bool = true)
MOI.instantiate(::MOI.OptimizerWithAttributes; with_bridges::Bool = true)
```
preferred over the call syntax.

src/instantiate.jl
```diff
@@ -0,0 +1,117 @@
...
+    OptimizerWithAttributes(optimizer_constructor, params::Pair...)
+
+Create an [`OptimizerWithAttributes`](@ref) with the parameters `params`.
+"""
+function OptimizerWithAttributes(optimizer_constructor, args::Vararg{Pair, N}) where N
+    if !applicable(optimizer_constructor)
+        error(_instantiate_not_callable_message)
+    end
+    params = Pair{AbstractOptimizerAttribute,Any}[]
+    _add_params(params, args...)
```
odow: Why not just:
```Julia
params = Pair{AbstractOptimizerAttribute,Any}[_to_param(arg) for arg in args]
```
Instead of all this `_add_params` indirection?

src/instantiate.jl
```diff
@@ -0,0 +1,117 @@
...
+    with_bridges::Bool=false, with_names::Bool=false, T::Type=Float64)
+    if with_bridges
+        return instantiate_with_bridges(optimizer_constructor;
+                                        with_names=with_names, T=T)
+    else
+        return instantiate(optimizer_constructor)
+    end
+end
+
+const _instantiate_not_callable_message =
```
odow: Constants should be in `_UPPER_CASE`.

src/instantiate.jl
blegat: The compiler might prefer `_add_params` recursion tuples of mixed type rather than for loops. In your approach, I would guess the dispatch to the `_to_param` method is done at runtime.

odow: This is very unlikely to be the bottleneck, and the current approach takes a little while to understand what's going on.

blegat:
> `instantiate(Gurobi.Optimizer; with_bridge_type = Float64)`

It's nice to merge two arguments in one but I am bit concerned that it reads "the type of the bridge is Float64" while the type of the bridge is more like `MOI.Bridges.Constraint.SplitInterval{Float64}`.
Maybe `with_bridge_coefficient_type` but it's quite long, or `with_bridge_of_type` (but it's not so much better) ?
The keyword argument `T` is not so explicit either so `with_bridge_type` is still an improvement IMO

src/instantiate.jl
```diff
@@ -0,0 +1,108 @@
...
+end
+
+"""
+    instantiate_and_check(optimizer_constructor::OptimizerWithAttributes)
+
+Create an instance of optimizer represented by [`OptimizerWithAttributes`](@ref).
+Then check that the type returned is an empty [`AbstractOptimizer`](@ref).
+"""
+function instantiate_and_check(optimizer_constructor::OptimizerWithAttributes)
+    optimizer = instantiate(optimizer_constructor.optimizer)
```
mlubin: Should this be calling `instantiate_and_check`?

src/instantiate.jl
```diff
@@ -0,0 +1,108 @@
...
+    return optimizer
+end
+
+"""
+    instantiate(optimizer_constructor,
+                with_bridge_type::Union{Nothing, Type}=nothing,
+                with_names::Bool=false)
+
+If `with_bridge_type` is `nothing`, it is equivalent to
+`instance(optimizer_constructor)`. Otherwise, it creates an instance of
```
mlubin: It's equivalent to `instantiate_and_check`

blegat:
> I'm confused by `instantiate_and_check` vs. `instantiate`

`instantiate_and_check` should be an internal function to avoid confusion

src/instantiate.jl
mlubin: This comment still applies.

src/instantiate.jl
```diff
@@ -0,0 +1,112 @@
...
+Creates an instance of optimizer either by calling
+`optimizer_constructor.optimizer()` and setting the parameters in
+`optimizer_constructor.params` if `optimizer_constructor` is a
+[`OptimizerWithAttributes`](@ref) or by calling `optimizer_constructor()`
+if `optimizer_constructor` is function.
+
+If `with_bridge_type` is not `nothing`, it enables all the bridges defined in
+the MathOptInterface.Bridges submodule with coefficient type
+`with_bridge_type`. If
+`!MOI.Utilities.supports_default_copy_to(optimizer, with_names)` then a
```
mlubin: What does this mean for users? How should they decide whether to set `with_names` to true or false?

src/instantiate.jl
```diff
@@ -0,0 +1,115 @@
...
+`optimizer_constructor.params` if `optimizer_constructor` is a
+[`OptimizerWithAttributes`](@ref) or by calling `optimizer_constructor()`
+if `optimizer_constructor` is function.
+
+If `with_bridge_type` is not `nothing`, it enables all the bridges defined in
+the MathOptInterface.Bridges submodule with coefficient type
+`with_bridge_type`.
+
+If the optimizer created by `optimizer_constructor` does not support loading the
+problem incrementally or does not support names (see
```
mlubin: This doesn't provide clear guidance on what value to set for `with_names`.

src/instantiate.jl
```diff
@@ -0,0 +1,116 @@
...
+    params = Pair{AbstractOptimizerAttribute,Any}[_to_param(arg) for arg in args]
+    return OptimizerWithAttributes(optimizer_constructor, params)
+end
+
+const _INSTANTIATE_NOT_CALLABLE_MESSAGE =
+    "The provided `optimizer_constructor` is invalid. It must be callable with zero" *
+    "arguments. For example, \"Ipopt.Optimizer\" or" *
+    "\"() -> ECOS.Optimizer()\". It should not be an instantiated optimizer " *
+    "like \"Ipopt.Optimizer()\" or \"ECOS.Optimizer()\"." *
+    "(Note the difference in parentheses!)"
```
odow: Needs spaces at end of each line otherwise the text runs together:
```Julia
julia> _INSTANTIATE_NOT_CALLABLE_MESSAGE
"The provided `optimizer_constructor` is invalid. It must be callable with zeroarguments. For example, \"Ipopt.Optimizer\" or\"() -> ECOS.Optimizer()\". It should not be an instantiated optimizer like \"Ipopt.Optimizer()\" or \"ECOS.Optimizer()\".(Note the difference in parentheses!)"
```

src/instantiate.jl
```diff
@@ -0,0 +1,116 @@
...
+"""
+function _instantiate_and_check(optimizer_constructor)
+    if !applicable(optimizer_constructor)
+        error(_INSTANTIATE_NOT_CALLABLE_MESSAGE)
+    end
+    optimizer = optimizer_constructor()
+    if !isa(optimizer, AbstractOptimizer)
+        error("The provided `optimizer_constructor` returned an object of type " *
+              "$(typeof(optimizer)). Expected a " *
+              "MathOptInterface.AbstractOptimizer.")
```
odow: Does this mean that the user can't provide a `MOI.ModelLike`? What's the reasoning for this? They might want to write out the model to file without solving.

src/instantiate.jl
```diff
@@ -0,0 +1,116 @@
...
+"""
+    instantiate(optimizer_constructor,
+                with_bridge_type::Union{Nothing, Type}=nothing,
+                with_names::Bool=false)
+
+Creates an instance of optimizer either by calling
+`optimizer_constructor.optimizer_constructor()` and setting the parameters in
+`optimizer_constructor.params` if `optimizer_constructor` is a
+[`OptimizerWithAttributes`](@ref) or by calling `optimizer_constructor()`
+if `optimizer_constructor` is function.
```
odow: "function" might be the wrong word here. Is `GLPK.Optimizer` a function? I guess so, but it's also a type. `if optimizer_constructor is callable`?

src/instantiate.jl
blegat: I copy pasted this from JuMP, see https://github.com/JuliaOpt/JuMP.jl/pull/2141. I agree that we can accept any `ModelLike`

blegat: @mlubin what's your take on this as you added it in https://github.com/JuliaOpt/JuMP.jl/pull/2090

mlubin: If it's not an `AbstractOptimizer`, then it can't take `AbstractOptimizerAttribute`s, so it doesn't make sense to put it inside an `OptimizerWithAttributes`. It also doesn't make sense to pass something that doesn't return an optimizer to `JuMP.set_optimizer`, unless we want to rename the function.

For the use case of writing a model to a file without solving, you could either use `JuMP.direct_model` or build a model and copy it using `MOI.copy_to`. Do we need to support setting it via `set_optimizer`?

odow: @mlubin makes good points. I remove my objection.

Issue #1000, opened 2020-01-10
Tests fail on 1.4 due to changes in printing


KristofferC:
It seems the test fails on 1.4 due to how something is printed:

```
Testing /home/pkgeval/.julia/packages/MathOptInterface/C1XBe/test/Bridges/Variable/vectorize.jl
MultirowChange: Test Failed at /home/pkgeval/.julia/packages/MathOptInterface/C1XBe/test/Bridges/Variable/vectorize.jl:121
  Expression: MOI.modify(bridged_mock, cis[1], change)
    Expected: MathOptInterface.ModifyConstraintNotAllowed{MathOptInterface.VectorAffineFunction{Float64},MathOptInterface.ExponentialCone,MathOptInterface.MultirowChange{Float64}}(MathOptInterface.ConstraintIndex{MathOptInterface.VectorAffineFunction{Float64},MathOptInterface.ExponentialCone}(12345677), MathOptInterface.MultirowChange{Float64}(MathOptInterface.VariableIndex(-1), [(3, 0.0)]), "The change MathOptInterface.MultirowChange{Float64}(MathOptInterface.VariableIndex(-1), Tuple{Int64,Float64}[(3, 0.0)]) contains variables bridged into a function with nonzero constant.")
      Thrown: MathOptInterface.ModifyConstraintNotAllowed{MathOptInterface.VectorAffineFunction{Float64},MathOptInterface.ExponentialCone,MathOptInterface.MultirowChange{Float64}}(MathOptInterface.ConstraintIndex{MathOptInterface.VectorAffineFunction{Float64},MathOptInterface.ExponentialCone}(12345677), MathOptInterface.MultirowChange{Float64}(MathOptInterface.VariableIndex(-1), [(3, 0.0)]), "The change MathOptInterface.MultirowChange{Float64}(MathOptInterface.VariableIndex(-1), [(3, 0.0)]) contains variables bridged into a function with nonzero constant.")
Stacktrace:
 [1] top-level scope at /home/pkgeval/.julia/packages/MathOptInterface/C1XBe/test/Bridges/Variable/vectorize.jl:121
 [2] top-level scope at /workspace/srcdir/usr/share/julia/stdlib/v1.4/Test/src/Test.jl:1116
 [3] top-level scope at /home/pkgeval/.julia/packages/MathOptInterface/C1XBe/test/Bridges/Variable/vectorize.jl:117
[...]

PR #1010, opened 2020-01-27
Fix tests for Julia v1.4-rc1

blegat:
Closes https://github.com/JuliaOpt/MathOptInterface.jl/issues/1000


blegat:
The tests are failing now due to the change in https://github.com/JuliaLang/julia/pull/33205

PR #1017, opened 2020-02-08
DNMY: add bridge for geomean to relative entropy

chriscoey:
as discussed with @blegat.
relative entropy can bridge to exponential cones, so we have an exponential cone extended formulation for geomean.
can later be extended easily for more general weighted power means from #977.
@blegat the dual transformations here used a bit of guesswork - perhaps you can check them?
I added dual solution tests to geomean1 test.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,7 @@ MathOptInterface (MOI) release notes
 v0.9.10 (January 31, 2020)
 ---------------------
 
+- Added `GeoMeantoRelEntrBridge` to bridge a geomean constraint to a relative entropy constraint.
 - Added `OptimizerWithAttributes` grouping an optimizer constructor and a list
   of optimizer attributes (#1008).
 - Added `RelativeEntropyCone` with corresponding bridge into exponential cone
@@ -14,7 +15,7 @@ v0.9.10 (January 31, 2020)
 - Added `dual_set_type` (#1002).
 - Added tests for vector specialized version of `delete` (#989, #1011).
 - Added PSD3 test (#1007).
-- Clarifed dual solution of `Tests.pow1v` and `Tests.pow1f` (#1013).
+- Clarified dual solution of `Tests.pow1v` and `Tests.pow1f` (#1013).
 - Added support for `EqualTo` and `Zero` in
   `Bridges.Constraint.SplitIntervalBridge` (#1005).
 - Fixed `Utilities.vectorize` for empty vector (#1003).
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,10 +1,14 @@
 MathOptInterface (MOI) release notes
 ====================================
 
-v0.9.10 (January 31, 2020)
+v0.9.11 (February 9, 2020)
 ---------------------
 
 - Added `GeoMeantoRelEntrBridge` to bridge a geomean constraint to a relative entropy constraint.
+
+v0.9.10 (January 31, 2020)
+---------------------
+
 - Added `OptimizerWithAttributes` grouping an optimizer constructor and a list
   of optimizer attributes (#1008).
 - Added `RelativeEntropyCone` with corresponding bridge into exponential cone
--- a/src/Bridges/Constraint/geomean_to_relentr.jl
+++ b/src/Bridges/Constraint/geomean_to_relentr.jl
@@ -73,7 +73,7 @@ function MOI.set(model::MOI.ModelLike, attr::MOI.ConstraintPrimalStart, bridge::
     return
 end
 # Given a is dual on y >= 0 and (b, c, d) is dual on RelativeEntropyCone constraint,
-# dual on (u, w) in GeometricMeanCone is (-a, c)
+# dual on (u, w) in GeometricMeanCone is (-a, c).
 function MOI.get(model::MOI.ModelLike, attr::Union{MOI.ConstraintDual, MOI.ConstraintDualStart}, bridge::GeoMeantoRelEntrBridge)
     u_dual = -MOI.get(model, attr, bridge.nn_index)[1]
     relentr_dual = MOI.get(model, attr, bridge.relentr_index)
--- a/src/Bridges/Constraint/geomean_to_relentr.jl
+++ b/src/Bridges/Constraint/geomean_to_relentr.jl
@@ -1,0 +1,92 @@
+"""
+    GeoMeantoRelEntrBridge{T}
+
+The `geometric mean cone` is representable with a relative entropy constraint and a nonnegative auxiliary variable,
+since ``u \\le \\prod_{i=1}^n w_i^{1/n}`` if and only if there exists a ``y \\ge 0`` such that
+``0 \\ge \\sum_{i=1}^n (u + y) \\log (\\frac{u + y}{w_i})``, or equivalently
+``(0, w, (u + y) e) \\in RelativeEntropyCone(1 + 2n)``, where ``e`` is a vector of ones.
+"""
+struct GeoMeantoRelEntrBridge{T, F, G, H} <: AbstractBridge
+    y::MOI.VariableIndex
+    ge_index::CI{F, MOI.GreaterThan{T}} # for y >= 0
+    relentr_index::CI{G, MOI.RelativeEntropyCone}
+end
+function bridge_constraint(::Type{GeoMeantoRelEntrBridge{T, F, G, H}}, model::MOI.ModelLike, f::H, s::MOI.GeometricMeanCone) where {T, F, G, H}
+    f_scalars = MOIU.eachscalar(f)
+    y = MOI.add_variable(model)
+    ge_index = MOI.add_constraint(model, y, MOI.GreaterThan(zero(T)))
+    relentr_func = MOIU.operate(vcat, T, zero(MOI.ScalarAffineFunction{Float64}), f_scalars[2:end], fill(MOIU.operate(+, T, f_scalars[1], MOI.SingleVariable(y)), MOI.dimension(s) - 1)...)
+    relentr_index = MOI.add_constraint(model, relentr_func, MOI.RelativeEntropyCone(MOI.output_dimension(relentr_func)))
+    return GeoMeantoRelEntrBridge{T, F, G, H}(y, ge_index, relentr_index)
+end
+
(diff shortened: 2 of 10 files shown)


NEWS.md
```diff
@@ -4,6 +4,7 @@ MathOptInterface (MOI) release notes
 v0.9.10 (January 31, 2020)
 ---------------------
 
+- Added `GeoMeantoRelEntrBridge` to bridge a geomean constraint to a relative entropy constraint.
```
blegat: This won't be in v0.9.10, create a new entry for v0.9.11

test/Bridges/Constraint/geomean.jl
```diff
@@ -9,7 +9,7 @@ const MOIB = MathOptInterface.Bridges
 include("../utilities.jl")
 
 mock = MOIU.MockOptimizer(MOIU.Model{Float64}())
-config = MOIT.TestConfig()
+config = MOIT.TestConfig(duals = false)
```
blegat: This is not enough, if we add the duals in the test, we need to implement it for `GeoMeanBridge` since otherwise, all solvers that are currently passing GeoMean tests thanks to this bridge will now fail the test

src/Bridges/Constraint/geomean_to_relentr.jl
```diff
@@ -0,0 +1,92 @@
...
+``(0, w, (u + y) e) \\in RelativeEntropyCone(1 + 2n)``, where ``e`` is a vector of ones.
+"""
+struct GeoMeantoRelEntrBridge{T, F, G, H} <: AbstractBridge
+    y::MOI.VariableIndex
+    ge_index::CI{F, MOI.GreaterThan{T}} # for y >= 0
+    relentr_index::CI{G, MOI.RelativeEntropyCone}
+end
+function bridge_constraint(::Type{GeoMeantoRelEntrBridge{T, F, G, H}}, model::MOI.ModelLike, f::H, s::MOI.GeometricMeanCone) where {T, F, G, H}
+    f_scalars = MOIU.eachscalar(f)
+    y = MOI.add_variable(model)
```
blegat: I would do `MOI.add_constrained_variables(model, MOI.Nonnegatives(1))`. Otherwise, you loose problem structure. Here for instance, a solver not supporting free variables will need to bridge `y` with the FreeBridge.

src/Bridges/Constraint/geomean_to_relentr.jl
```diff
@@ -0,0 +1,92 @@
...
+    ge_index = MOI.add_constraint(model, y, MOI.GreaterThan(zero(T)))
+    relentr_func = MOIU.operate(vcat, T, zero(MOI.ScalarAffineFunction{Float64}), f_scalars[2:end], fill(MOIU.operate(+, T, f_scalars[1], MOI.SingleVariable(y)), MOI.dimension(s) - 1)...)
```
blegat: Splatting with a vector that can have arbitrary size seems to be a been crazy. You need to compile a new vcat method for each length of (u + y) e`.
`MOIU.vectorize` seems more appropriate here. Or call `MOIU.vectorize` on `(u + y) e` and then merge with `(0, w` with `vcat`.

src/Bridges/Constraint/geomean_to_relentr.jl
```diff
@@ -0,0 +1,92 @@
...
+
+# Attributes, Bridge acting as a constraint
+function MOI.get(model::MOI.ModelLike, ::MOI.ConstraintFunction, bridge::GeoMeantoRelEntrBridge{T, F, G, H}) where {T, F, G, H}
+    relentr_func = MOIU.eachscalar(MOI.get(model, MOI.ConstraintFunction(), bridge.relentr_index))
+    d = div(length(relentr_func) - 1, 2)
+    u_func = MOIU.remove_variable(MOIU.operate(-, T, relentr_func[end], MOI.SingleVariable(bridge.y)), bridge.y)
+    w_func = relentr_func[2:(1 + d)]
+    return MOIU.convert_approx(H, MOIU.operate(vcat, T, u_func, w_func))
+end
+MOI.get(model::MOI.ModelLike, ::MOI.ConstraintSet, bridge::GeoMeantoRelEntrBridge) = MOI.GeometricMeanCone(1 + div(MOI.dimension(MOI.get(model, MOI.ConstraintSet(), bridge.relentr_index)) - 1, 2))
```
blegat: The line seems to be a bit long

src/Bridges/Constraint/geomean_to_relentr.jl
```diff
@@ -0,0 +1,92 @@
...
+    u_pos = -min(zero(u_val), u_val)
+    d = length(value) - 1
+    MOI.set(model, MOI.VariablePrimalStart(), bridge.y, u_pos)
+    MOI.set(model, attr, bridge.ge_index, u_pos)
+    MOI.set(model, attr, bridge.relentr_index, vcat(0, value[2:end], fill(u_pos + u_val, d)))
+    return
+end
+# Given a is dual on y >= 0 and (b, c, d) is dual on RelativeEntropyCone constraint,
+# dual on (u, w) in GeometricMeanCone is (-a, c)
+function MOI.get(model::MOI.ModelLike, attr::Union{MOI.ConstraintDual, MOI.ConstraintDualStart}, bridge::GeoMeantoRelEntrBridge)
```
blegat: Here is how I get the dual:
A bridge can always be represented as a transformation from an F-in-S constraint into creating auxiliary vars y and constraints in F1-in-S1, F2-in-S2, ...
So you have:
x in S <=> there exists y such that Ax + By in S1 x S2 x ... x Sn
Theorem:
[I; 0] * S* = [A'; B'] (S1 x S2 x ... x Sn)*
So you have
S* = A' * (S1 x S2 x ... x Sn)*
and
0 = B' * (S1 x S2 x ... x Sn)* -> This is in fact the dual constraint due to the fact that y is free
In this case, you have S1 = RelEntr and S2 = R+
A = [0 0; 0 I; 0 e; 0 0]
B = [0; 0; e; 1]
So given a dual z = [b; c; d; a], you get
A' * z = [e' * d; c]
B' * z = e' * d + a
The first equation says that what you are looking for is [e' * d; c]
The second says that e' * d = -a so the (-a, c) you are using is indeed correct

src/Bridges/Constraint/geomean_to_relentr.jl
```diff
@@ -0,0 +1,92 @@
...
+    u_dual = -MOI.get(model, attr, bridge.ge_index)[1]
+    relentr_dual = MOI.get(model, attr, bridge.relentr_index)
+    d = div(length(relentr_dual) - 1, 2)
+    w_dual = relentr_dual[2:(d + 1)]
+    return vcat(u_dual, w_dual)
+end
+# Given constraint dual start of (u, w), constraint dual on RelativeEntropyCone
+# constraint is (1/d, -u/d * e, -w) and dual on y >= 0 constraint is -u.
```
blegat: From above, we have u = -a = e'd and w = c
So c = w, a = -u and  e'd = u.
To deduce that b = 1/d and d = u/n * e, we would need to know what is the dual of the RelativeEntropyCone.
Here `(1/d, -u/d * e, -w)` seems incorrect, `(1/d, w, u/d * e)` might work though. The test should have failed with the current implementation

blegat: So you need to find b, c, d. You know that c and you know sum(d) and you know that (b, c_i, d_i) in DualExp for each i.
You can decouple the problem. First, you find b such that
-b sum exp(c_i/b) <= exp(1) * sum d_i
which can be obtained by minimizing -b sum exp(c_i/b)
Then you split sum d_i according to the weight -b exp(c_i/b) so you give `d_i = -b exp(c_i/b) / (-b sum exp(c_i/b)) * sum d_i`

blegat: The univariate minimization should be easy since `sum -b exp(c_i/b)` is convex as it is the sum of convex functions.

test/Bridges/Constraint/geomean.jl
chriscoey: the geomean -> RSOC bridge seems more painful to get duals for, given it's slightly awkward structure. do you have any leads on that?

src/Bridges/Constraint/geomean_to_relentr.jl
chriscoey: just amending: A = [0 0; 0 I; e 0; 0 0].
and for extra details: x = (u, w).

chriscoey: thanks for this very helpful explanation @blegat

src/Bridges/Constraint/geomean_to_relentr.jl
chriscoey: are you proposing to use say Optim.jl to solve this univariate convex minimization problem?

chriscoey: For now I've updated it with your first suggestion of `(1/d, w, u/d * e)`, and added a TODO

test/Bridges/Constraint/geomean.jl
chriscoey: @blegat I'm going to leave that to you since you wrote the original bridge. What I can do is add dual solution tests for the other geomean tests in contconic if that's helpful - let me know.

src/Bridges/Constraint/geomean_to_relentr.jl
blegat: > are you proposing to use say Optim.jl to solve this univariate convex minimization problem?

Since it's stricktly unimodal (since it's convex), we can implement a [golden-section search](https://en.wikipedia.org/wiki/Golden-section_search) which is implemented in a few lines. A TODO is ok

chriscoey:
closing in favor of  #1021

src/Bridges/Constraint/geomean_to_relentr.jl
blegat: The dual for the RelativeEntropyCone I gave was incorrect. When changing it, we get a closed form expression for `b` so no need to solve an univariate optimization problem anymore, see https://github.com/JuliaOpt/MathOptInterface.jl/pull/1028

PR #1018, opened 2020-02-08
Add dual in RelativeEntropyCone docstring

blegat:
It can be obtained from the bridge to exponential cone.
[I 0]^(-1) * RelEntr = [A B]^(-1) [Exp1 x ... x Expn x R+]
so
[I; 0] * RelEntr* = [A'; B'] * [Exp1* x ... x Expn* x R+]
B' * R+ = 0 says that the u_i are all equal.
Then A' * [Exp1* x ... x Expn*] gives the dual exponential cone constraints with the same u_i


chriscoey:
``-u \\exp (v_i/u) \\le \\exp(1) w_i`` is equivalent to ``v_i \\le u (\\log(w_i/-u) + 1)``, not sure if worth noting both.

PR #1021, opened 2020-02-10
add duals to geomean tests, and add geomean to exponential cone bridge

chriscoey:
As mentioned by @blegat in #1017, we need to add dual transformations for the existing GeoMean bridge (which is to RSOC constraints).


chriscoey:
The dual transformations now work and pass tests when running with and without bridges in Hypatia.

Remaining TODO is to get the dual transformations for the GeoMean (to RSOC) bridge going. @blegat?

chriscoey:
Waiting for #1025 to be merged

chriscoey:
also I'll modify the geomean to rsoc bridge tests here to test what was added in #1025

src/Bridges/Constraint/geomean_to_exp.jl
```diff
@@ -0,0 +1,125 @@
...
+# Given a is dual on u - y <= 0, b is dual on sum(z) >= 0, and (c_i, d_i, e_i) is dual on ExponentialCone constraint i,
+# dual on (u, w) in GeometricMeanCone is (a, e).
+function MOI.get(model::MOI.ModelLike, attr::Union{MOI.ConstraintDual, MOI.ConstraintDualStart}, bridge::GeoMeantoExpBridge)
+    u_dual = MOI.get(model, attr, bridge.le_y_index)
+    w_dual = [MOI.get(model, attr, exp_index_i)[3] for exp_index_i in bridge.exp_indices]
+    return vcat(u_dual, w_dual)
+end
+# Given constraint dual start of (u, w), constraint dual on LessThan constraint is a = u,
+# on GreaterThan constraint is b, and on exponential cone constraint i is (c_i = -b, d_i, e_i = w_i).
+# We must have sum(d) = a = u. We let b = -c_i = -u/n and d_i = -b (log(w_i/b) + 1).
```
blegat: Why is c_i equal to u / n?

chriscoey: It was a choice (hence the "We let ...")

chriscoey: I chose it because it allowed me to make that d_i value work. My math was sketchy though - it's worth double checking.

blegat: You still have the same issue as in https://github.com/JuliaOpt/MathOptInterface.jl/pull/1017 that for the duals to be correct, you need to have `sum(d_i) = a`. With your choice of `d_i`, it's not necessarily the case.

chriscoey: `sum(d) = sum_i -b (log(w_i/b) + 1) = -b (n + sum_i log(w_i/b)) = -b (n + log(prod(w_i/b))) = -(-u/n) (n + log(prod(w_i * -n/u))) = u + u/n * log(prod(w_i) * prod(-n/u))` which equals `u` if `prod(w_i) = prod(-u/n)` i.e. `-u = n * geomean(w)` i.e. if the dual geomean cone inequality holds at equality. but it's true that I can't assume that from `(u, w)`. so instead what if i let `b = -c_i = geomean(w)`. then I think it works?

chriscoey: and `(-b, -b (log(w_i/b) + 1), w_i)` is still feasible for the dual exp cone `-u exp(v/u) <= exp(1) w` because:
```
b exp(-b (log(w_i/b) + 1) / -b) <= exp(1) w_i
b exp(log(w_i/b) + 1) <= exp(1) w_i
b * w_i/b * exp(1) <= exp(1) w_i
w_i <= w_i
```

chriscoey: OK try again:

Given constraint dual start of `(u, w)`, constraint dual on LessThan constraint is `a = u`, on GreaterThan constraint is `b`, and on exponential cone constraint i is `(c_i = -b, d_i, e_i = w_i)`. We must have `sum(d) = a = u`. We let `b = -c_i = -u/n` and `d_i = -b * (log(w_i/geomean(w)) + 1)`. Note by dual geomean cone inequality, `-u <= n * geomean(w)` i.e. `geomean(w) >= -u/n`. Hence the dual on exponential cone constraint i is `(-b = u/n, d_i = u/n * (log(w_i/geomean(w)) + 1), w_i)`, which is feasible for the DualExponentialCone `(x,y,z): -x exp(y/x) <= exp(1) z`, since we have
```
-u/n * exp(u/n * (log(w_i/geomean(w)) + 1) / (u/n)) 
= -u/n * w_i / geomean(w) * exp(1) 
<= geomean(w) * w_i / geomean(w) * exp(1) 
= exp(1) * w_i.
```

chriscoey: I think that's it

lkapelevich:
is there anything this is waiting on?

blegat:
I would argue that https://github.com/JuliaOpt/MathOptInterface.jl/pull/1017 should be merged instead.
This bridge is the combination of the bridge in #1017 and the RelEntrtoExpBridge and the RelEntr formulation of GeoMean preserve more structure than the Exp formulation. As a result, the RelEntr cone has dimension 2n + 1 while the Exp formulation has dimension 3n + 2.

odow:
But doesn't a relative entropy cone need to be bridged further?  @chriscoey has done a lot of work here, and it'd be good to get it merged.

blegat:
I'm planning to take a look and merge either this one or #1017

chriscoey:
@blegat if you could make a decision on this PR vs #1017 that would be helpful. 

BTW there may be some related fixes on one of these two branches that aren't on the other, so we can compare and move over anything that's useful.

chriscoey:
updated. Hypatia passes the relevant MOI geomean conic tests with and without the bridge (including dual values).

blegat:
I'm ready to merge #1017 if it's reopened and rebased.
As mentioned in https://github.com/jump-dev/MathOptInterface.jl/pull/1021#issuecomment-589976775, this bridge is a combination of two other bridges.
So we have a bridge graph where the nodes are the different constraint types.
We have edges between these nodes and if we start adding an edge from a to c whenever there is an edge from a to b and an edge from b to c, this would be unmaintainable. This is the reason we have a sophisticated bridge selection mechanism that automatically finds how to combine bridges.

chriscoey:
closing in favor of #1178

PR #1023, opened 2020-02-11
[WIP] Evaluation distance to set

matbesancon:
Fairly straightforward, just like `eval_variables` evaluates the value of a function with a given value mapping of the variables, this PR adds the evaluation of values belonging to a set.

I reused the `Base.in` function, which seemed appropriate here since it's not used elsewhere with this signature.

One thing to work out as pointed by @mtanneau is the propagation of absolute and relative tolerance


mlubin:
IMO for this to be generally useful it needs to support tolerances in some form.

matbesancon:
Agreed, I wasn't sure yet how to propagate them, `rtol` and `atol` keywords similar to `isapprox`?

odow:
I would prefer a `distance` function which returned the distance from a point to feasibility inside the set.
```Julia
function distance(x::T, set::LessThan{T}) where {T}
    return max(zero(T), x - set.upper)
end

function distance(x::T, set::Integer) where {T}
    return x - round(Int, x)
end

function distance(x::Vector{T}, set::SOS1{T}) where {T}
    _, index = findmax(abs.(x))
    return sqrt(sum(x[i]^2 for i = 1:length(x) if i != index))
end
```

Feasibility checkers could then use these distances and some notion of tolerances.

matbesancon:
a notion of distance makes sense for convex sets, integer sets get more ambiguous

odow:
> integer sets get more ambiguous

Sure. But if we have a notion of belonging to a set with a tolerance, then we need some corresponding distance to check against.

For some sets, e.g., SOSII, it isn't immediately obvious what a good distance is. L2 norm of the elements excluding the two largest adjacent absolute values?

We might even consider returning a vector to feasibility, rather than the norm. So a PSD constraint is the vector of eigenvalues `min.(0.0, eigvals(X))`.

That would let the feasibility checker choose an appropriate norm as well. In the SOSII case, it isn't apparent if you want to use the L2 norm.

mtanneau:
> integer sets get more ambiguous

That can be the distance from `x` to the lattice, rather than from `x` to a continuous set.

Then, as odow seems to suggest, which norm to use could be up to the user.
In a flavor similar to `norm(x, 2)` or `norm(x, Inf)`.

mtanneau:
One (probably obvious) point, maybe useful for the docs: for a constraint `f(x) ∈ S`, here we are talking of "feasibility" as the distance from `f(x)` to `S`.
The distance from `x` to the feasible set `X = {x | f(x) ∈ S}` can be something else entirely.

Thus, here we are really addressing the question of "how close is this constraint to being satisfied". (which is 100% fine by me)

matbesancon:
just to be sure: you mean we deal with distance in the image space, and not in the input space right?

matbesancon:
One more corner case, in the case of dimension mismatch, I simply returned false for set belonging. When using a distance, this would be +Inf I guess

matbesancon:
> We might even consider returning a vector to feasibility, rather than the norm. So a PSD constraint is the vector of eigenvalues min.(0.0, eigvals(X)).

+1

mtanneau:
> in the case of dimension mismatch

Why not throw a `DimensionMismatch`?

matbesancon:
Not a fan of proliferating thrown errors all over the place, but yeah seems appropriate with this change

matbesancon:
I've started a re-write up to `NormSpectralCone`.
One open question is whether  to use `min(a, b, f(x) - S)` when requiring, say, a >= 0, b >= 0, f(x) in S

matbesancon:
note here that `_distance` returns an unsigned distance, which is then converted in `set_distance`

src/Utilities/set_eval.jl
```diff
@@ -1,114 +1,121 @@
 
-Base.in(v, s::MOI.LessThan) = v <= s.upper
-Base.in(v, s::MOI.GreaterThan) = v >= s.lower
-Base.in(v, s::MOI.EqualTo) = v ≈ s.value
+function set_distance(v::Real, s)
```
mtanneau: `set_distance` sounds like a setter.

Why not just `distance`, or (more explicit) `distance_to_set`?

matbesancon: fixed in last commit

src/Utilities/set_eval.jl
```diff
@@ -0,0 +1,175 @@
...
+end
+
+function _distance(v::AbstractVector{<:Real}, s::MOI.NormNuclearCone)
+    t = v[1]
+    m = reshape(v[2:end], (s.row_dim, s.column_dim))
+    s1 = sum(LinearAlgebra.svd(m).S)
+    return s1 - t
+end
+
+# return the second largest absolute value (=0 if SOS1 respected)
```
odow: Or, return a vector of values with the largest value set to `0`. Every element should be 0.

src/Utilities/set_eval.jl
```diff
@@ -0,0 +1,175 @@
...
+function _distance(v::AbstractVector{<:Real}, ::MOI.NormOneCone)
+    t = v[1]
+    xs = v[2:end]
+    return t - sum(abs, xs)
+end
+
+function _distance(v::AbstractVector{<:Real}, ::MOI.SecondOrderCone)
+    t = v[1]
+    xs = v[2:end]
+    return max(-t, dot(xs, xs) - t^2) # avoids sqrt
```
odow: This isn't correct. `max(0.0, norm(v[2:end], 2) - v[1])`

src/Utilities/set_eval.jl
```diff
@@ -0,0 +1,175 @@
...
+
+function distance_to_set(v::AbstractVector{T}, s) where {T <: Real}
+    length(v) != MOI.dimension(s) && throw(DimensionMismatch("Mismatch between v and s"))
+    d = _distance(v, s)
+    return [di < 0 : zero(T) : di for di in d]
+end
+
+_distance(v, s::MOI.LessThan) = v - s.upper
+_distance(v, s::MOI.GreaterThan) = s.lower - v
+_distance(v, s::MOI.EqualTo) = abs(v - s.value)
```
odow: These arguments should be typed.

src/Utilities/set_eval.jl
```diff
@@ -0,0 +1,175 @@
...
+    return d < 0 : zero(d) : d
+end
+
+function distance_to_set(v::AbstractVector{T}, s) where {T <: Real}
+    length(v) != MOI.dimension(s) && throw(DimensionMismatch("Mismatch between v and s"))
+    d = _distance(v, s)
+    return [di < 0 : zero(T) : di for di in d]
+end
+
+_distance(v, s::MOI.LessThan) = v - s.upper
```
odow: I think this should be `max(zero(T), v - s.upper)`

matbesancon:
@odow the _distance function remains internal and computed the unsigned distance to avoid checking if >= 0 in every single method. 
The external function is the top one, computing the distance, and then returning max(0, dist)

odow:
I understand the current split. I'm suggesting a public method for each set, with no re-direction to the zero-checker. Maybe a few more lines of code, but easier to understand for each set in isolation.

matbesancon:
Ok yes I see your point, will do that. I still think we should keep the non-checked version, which would be comparable to a `noargcheck or `@inbounds``

odow:
The zero check is essentially free. No one should be using this in any performance-critical code, and even if they are, `max(0, x)` isn't going to be the bottleneck. I'm in favor of simple as possible, one-way-to-do-things. You should wait for someone else to chime in though, in case I'm by myself on this.

matbesancon:
> Maybe a few more lines of code, but easier to understand for each set in isolation.

the bother with this is mostly that it means maintaining boilerplate over all sets instead of just at one place. That's why one function could do the dimension check, possibly throwing an error, call the set-specific method and return the cleaned-up version, otherwise these three steps have to be copy-pasted all over the place

src/Utilities/set_eval.jl
```diff
@@ -0,0 +1,178 @@
...
+function _correct_distance_to_set(d::Real)
+    return ifelse(d < 0, zero(d), d)
+end
+
+function _correct_distance_to_set(d::AbstractVector{T}) where {T <: Real}
+    return [ifelse(di < 0, zero(T), di) for di in d]
+end
+
+function distance_to_set(v, s)
+    length(v) != MOI.dimension(s) && throw(DimensionMismatch("Mismatch between v and s"))
```
blegat: I would write `Mismatch between vector and set`

src/Utilities/set_eval.jl
```diff
@@ -0,0 +1,178 @@
+
+function _correct_distance_to_set(d::Real)
+    return ifelse(d < 0, zero(d), d)
+end
+
+function _correct_distance_to_set(d::AbstractVector{T}) where {T <: Real}
+    return [ifelse(di < 0, zero(T), di) for di in d]
```
blegat: You can do `[_correct_distance_to_set(di) for di in d]`

src/Utilities/set_eval.jl
```diff
@@ -0,0 +1,178 @@
...
+    return _correct_distance_to_set(d)
+end
+
+_distance(v::Real, s::MOI.LessThan) = v - s.upper
+_distance(v::Real, s::MOI.GreaterThan) = s.lower - v
+_distance(v::Real, s::MOI.EqualTo) = abs(v - s.value)
+
+_distance(v::Real, s::MOI.Interval) = max(s.lower - v, v - s.upper)
+
+_distance(v::AbstractVector{<:Real}, ::MOI.Reals) = false * v
```
blegat: Why not simply `zero.(v)` ?

src/Utilities/set_eval.jl
blegat: `norm(v[2:end], 2) - v[1]` also works as negative values are set to zero afterwards

blegat:
In the current advantage, it's indeed nice that the dimension checks and `max(0, .)` are factored out and done only once.
Once concern is that MOI extension extending new sets should implement `_distance` which is an internal function.
We could rename `_distance` to `signed_distance_to_set` and require sets to implement this function while the user  should use `distance_to_set`. Then the users might still be willing to use `signed_distance_to_set` and it's annoying that it does not do the `DimensionMismatch`. So we could have `no_check_signed_distance_to_set` or create a `check_dimension` that should be called from each `signed_distance_to_set` so that we have a consistent error message (similar to ` throw_if_not_valid` or `check_result_index_bounds`)

odow:
> Once concern is that MOI extension extending new sets should implement _distance which is an internal function.

:100:

> We could rename _distance to signed_distance_to_set and require sets to implement this function while the user should use distance_to_set. Then the users might still be willing to use signed_distance_to_set and it's annoying that it does not do the DimensionMismatch. So we could have no_check_signed_distance_to_set 

The issue I have with this is that there is one concept, the distance from a point to the set, and yet we have a plethora of functions and ways to call it. I vote simplicity.

> or create a check_dimension that should be called from each signed_distance_to_set so that we have a consistent error message (similar to throw_if_not_valid or check_result_index_bounds)

:100:

I propose
```Julia
function _check_dimension(x::AbstractVector, set::AbstractVectorSet)
    if length(x) != dimension(set)
        throw(DimensionMismatch("dimension of `x` does not match dimension of `set`."))
    end
end

function distance(x::T, set::LessThan{T}) where {T}
    return min(zero(T), x - set.upper)
end

function distance(x::T, set::EqualTo{T}) where {T}
    return abs(x - set.value)
end

function distance(x::AbstractVector{T}, set::NonNegatives) where {T}
    _check_dimension(x, set)
    return min.(zero(T), x)
end
```

mlubin:
> Once concern is that MOI extension extending new sets should implement _distance which is an internal function.

Yes, this is an important concern.

Also, I think this belongs in MOI proper, not in `Utilities`. These are *definitions* of distances that don't immediately follow from the definitions of the sets.

src/Utilities/set_eval.jl
matbesancon: `v` can also be a scalar, `Mismatch between value and set`?

src/Utilities/set_eval.jl
matbesancon: I'll replace with `zeros(length(v))`, I'm not sure a broadcast would be eliminated

src/Utilities/set_eval.jl
blegat: That would work

odow:
Can we include these distance definitions in `sets.jl`? This is what we did with `dual_set`.

I'm going to repeat my ask that we only have one `distance_to_set(point, set)` function. The extra boilerplate doesn't matter. These definitions are not going to change.

mlubin:
I agree that the two levels of indirection (`distance_to_set` redirects to `unsigned_distance` and `_correct_distance_to_set`) and extra complexity in the API that MOI extensions need to implement are not worth saving the 1-2 lines of code per function. These are fundamental definitions in MOI, and the code should be as clear and simple as reasonably possible.

src/set_distances.jl
```diff
@@ -0,0 +1,183 @@
...
+
+Compute the distance of a value to a set.
+For some vector-valued sets, can return a vector of distances.
+When `v ∈ s`, the distance is zero (or all individual distances are zero).
+
+Each set `S` implements `distance_to_set(v::T, s::S)` with `T` of appropriate type.
+"""
+function distance_to_set end
+
+distance_to_set(v::Real, s::LessThan) = _correct_distance_to_set(v - s.upper)
```
odow: We meant:

a) Move this to https://github.com/JuliaOpt/MathOptInterface.jl/blob/e022626b962be01d19fb9b9fe89c77990a0b96c8/src/sets.jl#L168
b) Write it as 
```Julia
distance_to_set(x::T, s::LessThan{T}) where {T} = max(zero(T), x - s.upper)
```

src/set_distances.jl
```diff
@@ -0,0 +1,183 @@
...
+end
+
+function distance_to_set(v::AbstractVector{<:Real}, s::Zeros)
+    length(v) != dimension(s) && throw(DimensionMismatch("Mismatch between value and set"))
+    return abs.(v)
+end
+
+function distance_to_set(v::AbstractVector{<:Real}, s::Nonnegatives)
+    length(v) != dimension(s) && throw(DimensionMismatch("Mismatch between value and set"))
+    return _correct_distance_to_set(-v)
```
odow:
```Julia
function distance_to_set(x::AbstractVector{T}, s::Nonnegatives) where {T<:Real}
    _check_dimension(x, s)
    return max.(zero(T), -x)
end

Comment (matbesancon) [reply src/set_distances.jl]: we need consistency somewhere in the decision. Either calling helper functions is allowed or we re-write everything in the function body. Why having all the `max(zero(T), d)` in the function body but add an indirection for the `_check_dimension`?

File: src/set_distances.jl (hunk 9 above)
Comment (matbesancon) [reply src/set_distances.jl]: `distance_to_set(x::T, s::LessThan{T})` this signature adds another constraint on types, not just that the types are comparable (which is enough to compute the distance), but equal

File: src/set_distances.jl (hunk 10 above)
Comment (odow) [reply src/set_distances.jl]: The `max(0, -x)` is the definition of the distance, so it should be in the function body.

`_check_dimension` is not part of the definition of distance, and is just to help users from making mistakes. There isn't a need for it to be in the function body explicitly.

File: src/set_distances.jl (hunk 9 above)
Comment (odow) [reply src/set_distances.jl]: I guess the type restriction isn't critical. I meant more the `max(zero(T), x - s.upper)` part.

File: src/MathOptInterface.jl [hunk 11]
```diff
@@ -113,7 +113,6 @@ include("error.jl")
 include("indextypes.jl")
 include("functions.jl")
 include("sets.jl")
-include("attributes.jl")
```
mlubin: Was this deletion intentional?

src/sets.jl
```diff
@@ -80,6 +80,22 @@ function dual_set_type end
 
 dual_set_type(S::Type{<:AbstractSet}) = error("Dual type of $S is not implemented.")
 
+"""
+    distance_to_set(v, s)
+
+Compute the distance of a value to a set.
+For some vector-valued sets, can return a vector of distances.
```
mlubin: A vector of distances isn't a distance, mathematically speaking. What is meant here precisely?

src/sets.jl
```diff
@@ -251,6 +301,14 @@ end
 dual_set(s::SecondOrderCone) = copy(s)
 dual_set_type(::Type{SecondOrderCone}) = SecondOrderCone
 
+function distance_to_set(v::AbstractVector{<:Real}, s::SecondOrderCone)
+    _check_dimension(v, s)
+    t = v[1]
+    xs = v[2:end]
+    result = LinearAlgebra.norm(xs) - t
+    return max(result, zero(result)) # avoids sqrt
```
mlubin: Why return a vector of distances for `Nonnegatives` and a scalar for `SecondOrderCone`? Both are convex cones.

src/sets.jl
```diff
@@ -263,6 +321,18 @@ end
 dual_set(s::RotatedSecondOrderCone) = copy(s)
 dual_set_type(::Type{RotatedSecondOrderCone}) = RotatedSecondOrderCone
 
+function distance_to_set(v::AbstractVector{<:Real}, s::RotatedSecondOrderCone)
+    _check_dimension(v, s)
+    t = v[1]
+    u = v[2]
+    xs = v[3:end]
+    return [
```
mlubin: Why return a scalar for `SecondOrderCone` and a vector for `RotatedSecondOrderCone`?

src/sets.jl
```diff
@@ -345,6 +480,15 @@ end
 dual_set(s::NormSpectralCone) = NormNuclearCone(s.row_dim, s.column_dim)
 dual_set_type(::Type{NormSpectralCone}) = NormNuclearCone
 
+function distance_to_set(v::AbstractVector{<:Real}, s::NormSpectralCone)
+    _check_dimension(v, s)
+    t = v[1]
+    m = reshape(v[2:end], (s.row_dim, s.column_dim))
+    s1 = LinearAlgebra.svd(m).S[1]
```
mlubin: I'm not an expert on this, but I know that computing the full SVD is an extremely expensive way to compute the smallest singular value. This is probably not suitable for non-toy applications. There are a few packages that do this, e.g., https://jutho.github.io/KrylovKit.jl/latest/man/svd/. We should evaluate which, if any, are suitable to add as dependencies of MOI.

src/sets.jl
odow: I guess we want `distance_to_set(point, set::AbstractSet; norm::Int = 2)`. Or even just say that we're using the L2 norm.

src/MathOptInterface.jl
matbesancon: absolutely not, reverted

src/sets.jl
matbesancon: the intuition I had for it is giving an element-wise distance when it makes sense, giving which element to "improve" to reduce the distance to the set.
We can also always normalize as @odow mentions

src/sets.jl
matbesancon: SOC yields one result (there is one inequality to respect to belong to the set), versus Nonnegative where `dimension` inequalities must be respected

src/sets.jl
matbesancon: agreed, I just kept the dependency-free version for now. MOI can either add a dependency or vendor (copy and acknowledge) a single function from another package

matbesancon:
ping for the discussed comments from the JuMP call

odow:
I believe the outcome of the call was that we pick an opinionated distance for each set. It's up to solvers to convert between their internal representation and the MOI distance if so required. For each set, we should survey the landscape for what existing solvers measure, but this shouldn't be a blocker for us coming up with something reasonable and justifiable for the long-term.

For example:
```Julia
function distance_to_set(x::Vector{<:Real}, set::Nonnegatives)
    _check_dimension(x, set)
    return norm(min.(0, x), 2)
end
```

In particular, `distance_to_set(x, set)` should return a non-negative Float64.

joaquimg:
Have you seen this package: https://github.com/JuliaStats/Distances.jl
There is one basic entry point for computing distances which is `evaluate` and the distance type is passed as an argument.
This would make the interface very flexible, and then user could even create their own distances.
We could have a default distance dict that maps Sets to favorite distances and we allow the user to pass their dict of preferred distances.

odow:
I'm not in favor of adding extra dependencies. But we could define
```Julia
abstract type AbstractInfeasibilityDistance end
struct DefaultInfeasibilityDistance end

function distance_to_set(
    x::Vector, 
    set::Nonnegatives, 
    ::DefaultInfeasibilityDistance
)
    return norm(min.(0, x), 2)
end
```

This could let solvers define their own distances.

Then a future infeasibility checker could go
```Julia
function return_infeasible_constraints(
    model::MOI.ModelLike, 
    x::Dict{MOI.VariableIndex, Float64},
    distance::AbstractInfeasibilityDistance = DefaultInfeasibilityDistance();
    atol::Float64 = eps(Float64),
    rtol::Float64 = 0.0
)
    for (F, S) in ...
        ...
    end
end
```

joaquimg:
sure, that was my idea!
My suggestion was not to add https://github.com/JuliaStats/Distances.jl as a dependency.
It was jus an example of people dealing with multiple distances :)

matbesancon:
I thought the conclusion from previous discussions is that we wanted an opinionated unique distance to sets and not multiple values as presented there

odow:
The opinionated distance is the first element of the returned tuple. I just retuned the nearest element inside the set for the sake of exposition. We would really only have `return value(d)`. 

The bigger issue is that we can't rely on solvers (since they have tolerance issues themselves), and some solvers might not support `SecondOrderCone` + `S`.

We really want explicit formulae for each set that corresponds to the optimal solution of that optimization problem.

matbesancon:
one of the only bothers I have with norm2 is for cases where you end up comparing potentially unrelated quantities with radically different scales. Example with the exponential cone: 
`𝑦 exp(𝑥/𝑦) ≤ 𝑧, 𝑦 > 0`
comparing the quantities `z - 𝑦 exp(𝑥/𝑦)` and `y` have potentially different scales, etc

odow:
> one of the only bothers I have with norm2 is for cases where you end up comparing potentially unrelated quantities with radically different scales

Exactly, which is why I think we want the norm2 distance to the nearest projection to the set.

odow:
For example, with a bit of algebra you can show
```Julia
function distance_to_set(x::Vector{T}, ::MOI.SecondOrderCone) where {T <: Real}
    c = LinearAlgebra.norm2(x[i] for i = 2:length(x))
    if x[1] >= c
        return zero(T)
    end
    y = similar(x)
    y[1] = (x[1] + c) / 2
    d = y[1] / (2y[1] - x[1])
    for i = 2:length(x)
        y[i] = d * x[i]
    end
    return LinearAlgebra.norm2(x[i] - y[i] for i = 1:length(x))
end
```

Or, if you don't want to allocate `y`:
```julia
function distance_to_set(x::Vector{T}, ::MOI.SecondOrderCone) where {T <: Real}
    c = LinearAlgebra.norm2(x[i] for i = 2:length(x))
    if x[1] >= c
        return zero(T)
    end
    y1 = 0.5 * (x[1] + c)
    d = (x[1] - y1)^2 / (2y1 - x[1])^2
    dist = (y1 - x[1])^2
    for i = 2:length(x)
        dist += d * x[i]^2
    end
    return sqrt(dist)
end
```

src/sets.jl
joaquimg: If we add the distance type parameter we can easily allow the user to switch from the norm2 thing to a norm infinity which is pretty reasonable

matbesancon:
so conclusion would be to agree whenever possible on the following distance:
```
dist(x, S) ==
min norm2(x - y)
s.t. y in S
```

matbesancon:
weird sets can define their distance in whatever "best" way for them (indicator constraints are an example)

matbesancon:
Open questions for distances to set:
To extend the distance system, do we need a fallback for a new distance to the default one?
To extend sets, do you extend only the default distance?

odow:
Yes, presumably we have a fallback like:
```Julia
function distance_to_set(::AbstractDistanceFunction, x, s)
    return distance_to_set(DefaultDistance(), x, s)
end
```

Note that `DistanceFunction` should probably be renamed `AbstractDistanceFunction`.

matbesancon:
Another issue comes from complementarity constraints:

`MOI.Complements()` requires access to variable bounds to determine if a given vector of values 
respects the complementarity constraint, which is not possible with the current signature.
A possible fix would be accessing the variable bounds in a generic way.
Another would be to specify complementarity constraints with the associated bounds Complementarity(Vector{<:AbstractSet}).

odow:
The main reason why complementary bounds are defined separately is to avoid issues like the following, where declare bounds twice.
```Julia
model = Model()
@variable(model, 1 <= x <= 2)
@constraint(model, 2x - 1 ⟂ { x >= 0})
```

It seems reasonable that we wouldn't implement the `DefaultDistance` for complements. They're a bit if a special case. You could imagine defining a new distance function that required the `l` and `u` bound vectors as input.

matbesancon:
Yes agreed, even though it makes Complements a bit of a corner case for many things outside of distance, because of this non-self-contained constraint

src/sets.jl
```diff
@@ -690,6 +895,23 @@ struct SOS1{T <: Real} <: AbstractVectorSet
...
+        return d
+    end
+    # removing greatest distance
+    for i in eachindex(d)
+        @inbounds if d[i] == m
+            d[i] = zero(T)
+            return d
+        end
+    end
+end
```
odow, line 913: I don't think this function is correct. Something like:
```Julia
function distance_to_set(::DefaultDistance, v::AbstractVector{T}, ::SOS1{T}) where {T <: Real}
    _check_dimension(v, s)
    _, i = findmax(abs.(v))
    return LinearAlgebra.norm2([v[j] for j = 1:length(v) if j != i])
end
```

src/sets.jl
```diff
@@ -251,6 +329,14 @@ end
 dual_set(s::SecondOrderCone) = copy(s)
 dual_set_type(::Type{SecondOrderCone}) = SecondOrderCone
 
+function distance_to_set(::DefaultDistance, v::AbstractVector{<:Real}, s::SecondOrderCone)
+    _check_dimension(v, s)
+    t = v[1]
+    xs = v[2:end]
+    result = LinearAlgebra.norm2(xs) - t
+    return max(result, zero(result)) # avoids sqrt
```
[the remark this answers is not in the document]
mtanneau: I personally use the epigraph violation `d = x1 - norm(x[2:end], 2)`, mainly because
* it's easy to compute
* it tells by how much the conic constraint should be "relaxed" for the point to be feasible (where "relaxed" means that you shift the cone's apex)
* if the SOC constraint is written as `x2^2 + ... xn^2 <= x1^2`, the epigraph violation is the square root of this quadratic constraint's absolute violation.

Looks like [COSMO](https://github.com/oxfordcontrol/COSMO.jl/blob/d7913018a8375989fab1d6108be8043288150ea9/src/convexset.jl#L98) uses euclidean distance for computing projections.

src/sets.jl
```diff
@@ -80,6 +82,40 @@ function dual_set_type end
 
 dual_set_type(S::Type{<:AbstractSet}) = error("Dual type of $S is not implemented.")
 
+
+"""
+    AbstractDistance
+
+Distance function used to evaluate the distance from a point to a set.
+New subtypes of `AbstractDistance` must implement fallbacks for sets they don't cover and implement
+`distance_to_set(::Distance, v, s::S)` for sets they override the distance for.
```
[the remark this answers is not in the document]
matbesancon: > This is still an issue.
 
mistake on my part this should be removed

src/sets.jl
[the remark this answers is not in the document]
odow: What about sets like the exponential cone needing domain constraints like `y > 0`?

src/sets.jl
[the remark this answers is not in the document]
matbesancon: > I thought we were providing the AbstractDistance fallback?

we can provide one d(::AbstractDistance, v, s) = d(::DefaultDistance, v, s) but not be more specific because of method ambiguity

src/sets.jl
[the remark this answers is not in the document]
matbesancon: > What is the comment # avoids sqrt for?

legacy comment, just removed. For domain constraints it can be an open question, because some parts of the distance may be non-computable / infinite or complex if the domain is not respected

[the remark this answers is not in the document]
odow: Okay, so for conic sets "Epigraph violation or `Inf` if point fails feasibility bounds" seems like an easily computable and sensible definition.

src/sets.jl
```diff
@@ -263,6 +349,16 @@ end
 dual_set(s::RotatedSecondOrderCone) = copy(s)
 dual_set_type(::Type{RotatedSecondOrderCone}) = RotatedSecondOrderCone
 
+function distance_to_set(::DefaultDistance, v::AbstractVector{<:Real}, s::RotatedSecondOrderCone)
+    _check_dimension(v, s)
+    t = v[1]
+    u = v[2]
+    xs = v[3:end]
+    return LinearAlgebra.norm2(
+        (max(-t, zero(t)), max(-u, zero(u)), max(dot(xs,xs) - 2 * t * u))
```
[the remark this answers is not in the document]
matbesancon: The core issue here is that the cone is characterized by three constraints.
With vector-distance, this would have been:
```
max(-u, 0)
max(-v, 0)
max(-2tu + norm_squared(x) , 0)
```

since we don't have that, the l2-norm of these three epigraph violations is the thing computed here.

chriscoey:
My comment here isn't actually very helpful but: Hypatia doesn’t use a notion of distance to set anywhere. It’s tough because there are many notions of distance and the right one depends on context. Also not all conic sets are typically defined by epigraph/hypographs, e.g. the PSD cone or the set of nonnegative polynomials or the copositive cone.

chriscoey:
Also for many sets the Euclidean distance can be hard to compute (eg for exponential cone in general, though some "cases" are just violation on variable nonnegativity)

src/sets.jl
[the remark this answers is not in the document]
joaquimg: And you call it EpigraphDistance or something like...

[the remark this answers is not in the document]
joaquimg: For norm two distances you can use what proximal solvers do:
https://web.stanford.edu/~boyd/papers/pdf/prox_algs.pdf page 183 (section 6.3)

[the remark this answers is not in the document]
joaquimg: So, SOC is easy, PSD you just need eigen decomposition, Power and Exponential are well posed even with the extra constraints, but need a Newton...

joaquimg:
I will put it here, so it does not gets lost in the inner conversation:

For norm two distances you can use what proximal solvers do:
https://web.stanford.edu/~boyd/papers/pdf/prox_algs.pdf page 183 (section 6.3)

So, SOC is easy, PSD you just need eigen decomposition, Power and Exponential are well posed even with the extra constraints, but need a Newton...

joaquimg:
Many of these projections are implemented here: https://github.com/kul-forbes/ProximalOperators.jl/tree/master/src/functions

odow:
Considering this PR is approaching the longest PR in the history of MOI (which @matbesancon also holds first and second place in #709 and #877), maybe this is a sign it should be in a package first. Then @matbesancon  can experiment with the different set definitions in peace, not worry about pulling in heavy dependencies for the proximal operators or the SVD, and see what works.

matbesancon:
> maybe this is a sign it should be in a package first

do you mean the distances themselves or the advanced features?

Dear 2-month-ago self, what what you done?

odow:
I mean the distances, the feasibility checker, everything. Call it `matbesancon/InfeasibilityToolkit.jl`.

This approach worked for `Dualization.jl`, and we brought the `dual_set` definitions in once they stabilized.

matbesancon:
I should have a skeleton more or less there for Friday's meeting

matbesancon:
So, with the heart broken, we can close this PR which is replaced by https://github.com/matbesancon/MathOptSetDistances.jl

PR #1024, opened 2020-02-11
Typo in sets

dourouc05:
Needed part for https://github.com/JuliaOpt/JuMP.jl/pull/2051, to gather some feedback.


src/sets.jl
```diff
@@ -807,6 +807,123 @@ end
...
 dimension(set::Complements) = 2 * set.dimension
 
+"""
+    AllDifferent(dimension::Int)
+
+The set corresponding to an all-different constraint.
+
+All expressions of a vector-valued function are enforced to take distinct
+values in the solution: for all pairs of expressions, their values must
+differ. This constraint is only defined for integer-valued expressions.
```
odow: I would remove `This constraint is only defined ...`. 

Presumably solvers will reject malformed models. But there's nothing stopping me from writing 
```Julia
model = Model()
@variable(model, x[1:2], Int)
@constriant(model, [x[1] + 1.0, 2x[2] - 2.0] in AllDifferent())
```

src/sets.jl
```diff
@@ -807,6 +807,123 @@ end
...
+    Domain{T <: Int}(values::Set{T})
+
+The set corresponding to an enumeration of constant values.
+
+The value of a scalar function is enforced to take a value from this set of
+values.
+
+This constraint is sometimes called `in`.
+"""
+struct Domain{T <: Int} <: AbstractScalarSet
```
odow: Again, why restrict to `Int`? Also, `Int` here is a concrete type, so `T <: Int` is just `Int`.

src/sets.jl
```diff
@@ -807,6 +807,123 @@ end
...
+
+The value of a scalar function is enforced to take a value from this set of
+values. These values are given by expressions potentially depending on other
+variables for the problem.
+
+This constraint is sometimes called `in_set`.
+"""
+struct Membership <: AbstractScalarSet
+    values::Set{AbstractScalarFunction}
+end
```
odow: Hmm... We need to think about this one. Set's shouldn't have functions as parameters. Consider the `Complements` and `IndicatorSet`s as examples.

src/sets.jl
```diff
@@ -807,6 +807,123 @@ end
...
+`S` has to be a sub-type of `AbstractScalarSet`.
+
+This set serves to find out whether a given constraint is satisfied. The only possible values are
+0 and 1.
+"""
+struct ReificationSet{S <: AbstractScalarSet} <: AbstractScalarSet
+    set::S
+end
+
+function Base.copy(set::ReificationSet{S <: AbstractScalarSet}{S}) where S
```
odow: One too many `{S}` here.

dourouc05:
What do you think of the latest version?

src/sets.jl
odow: How about:
```Julia
"""
    Membership(dimension)

The first element of a function of dimension `dimension` must equal at least 
one of the following `dimension - 1` elements of the function.

## Example

    [x, y, z] in Membership(dimension)
    # enforces `x == y` OR `x == z`.
"""
struct Membership <: AbstractVectorSet
    dimension::Int
end
```

src/sets.jl
```diff
@@ -807,6 +807,127 @@ end
...
+    Domain{T <: Real}(values::Set{T})
+
+The set corresponding to an enumeration of constant values.
+
+The value of a scalar function is enforced to take a value from this set of
+values.
+
+This constraint is sometimes called `in`.
+"""
+struct Domain{T <: Real} <: AbstractScalarSet
```
odow: Is the `Real` restriction necessary?

src/sets.jl
```diff
@@ -807,6 +807,127 @@ end
...
+
+"""
+    DifferentFrom(value::AbstractScalarFunction)
+
+The set excluding the single point ``x \\in \\mathbb{R}`` where ``x`` is given by `value`.
+All other values in the domain of ``x`` are allowed. `value` is a generic scalar function.
+"""
+struct DifferentFrom <: AbstractScalarSet
+    value::AbstractScalarFunction
+end
```
odow:
```Julia
"""
    DifferentFrom(dimension)

The first element of a function of dimension `dimension` cannot equal any
of the following `dimension - 1` elements of the function.

## Example

    [x, y, z] in DifferentFrom(dimension)
    # enforces `x != y` AND `x != z`.
"""
struct DifferentFrom <: AbstractVectorSet
    dimension::Int
end
```

src/sets.jl
```diff
@@ -807,6 +807,127 @@ end
...
+
+"""
+    StrictlyGreaterThan{T <: Real}(lower::T)
+
+The set ``(lower,\\infty) \\subseteq \\mathbb{R}``. It is mostly useful if `lower` is an integer.
+"""
+struct StrictlyGreaterThan{T <: Real} <: AbstractScalarSet
+    lower::T
+end
```
odow: Another option is 

```Julia
struct Strictly{S <: Union{LessThan, GreaterThan}} <: AbstractScalarSet
    set::S
end
```
So `Strictly(LessThan(1))`

Make it easier if we want to add `Strictly{Nonnegatives}`, `Strictly{Interval{Int}}`, etc.

dourouc05:
Better this way? I also added Sort, SortPermutation, BinPacking, and CapacitatedBinPacking. I think this set of constraints is enough as a first step.

dourouc05:
I completely agree with your objection. 

Regarding the JuMP PR (https://github.com/JuliaOpt/JuMP.jl/pull/2051), I suppose it makes no sense to have support in JuMP for sets that are not in MOI. This PR could then be reduced to the minimum to allow implementing things, with the actual constraints being implemented in that other package (say ConstraintProgrammingExtensions.jl, or another one so that solvers do not have a dependency on JuMP, like JuCP.jl or anything). 

(Also tagging @blegat on this, due to his feedback on the JuMP PR.)

Also, would that ConstraintProgrammingExtensions.jl package be a good place to implement bridges, like in MOI?

odow:
You could even have JuMP as a dependency to `ConstraintProgrammingExtensions`, and export a macro like
```Julia
@logicalconstraint(model, all_different(x))
```
which re-writes to
```Julia
@constraint(model, x in CP.AllDifferent(length(x))
```

I guess the points I want to make are:
1. If you want to work on constraint programming for JuMP, we want to help make that happen
2. In the long-term, we're supportive of integrating constraint programming into JuMP
3. In the near-term, I don't know if we've explored the space enough to understand what works and what doesn't. It'd be nice to understand what syntax decisions mini zinc made, why they made them, and what types of constraints they support.
4. Developing a constraint programming package now with the view to bringing some of it into MOI and some of it into JuMP in the future seems like the way to go.

So I would close #2051, but more because the timing is off, rather than us not wanting to add it.

> Also, would that ConstraintProgrammingExtensions.jl package be a good place to implement bridges, like in MOI?

Yes, perfect place for it. CP -> MIP bridges are probably a good way of getting solvers hooked up to test various things.

blegat:
> Developing a constraint programming package now with the view to bringing some of it into MOI and some of it into JuMP in the future seems like the way to go.

I agree, MOI already integrated 3 packages: MathOptFormat, MathOptInterfaceTests and MathOptInterfaceBridges.
It was easier to start with separate packages and experiment and once they matured merge them to MOI.
For instance, in MOI, we don't like making breaking releases hence we block any breaking changes. That might not be appropriate when you want to experiment and iterate with user feedback in the early stage.
It's also probably simpler if you make one package depending on both JuMP and MOI that defines sets, bridges and extension of JuMP macros.
Examples are [PolyJuMP](https://github.com/JuliaOpt/PolyJuMP.jl) and [SumOfSquares](https://github.com/JuliaOpt/SumOfSquares.jl) which defines new sets, new bridges and extend JuMP macros. In this case, we don't plan to merge them to MOI. First because they have a lot of dependencies in JuliaAlgebra, and second because MOI is designed such that there is no disadvantage of having extensions outside of MOI/JuMP. The goal is to have a distributed ecosystem to support the distributed and explorative nature of the world of research and academics :)

dourouc05:
Thanks for your thoughts! I just created a new package for these sets (not yet registered): https://github.com/dourouc05/ConstraintProgrammingExtensions.jl

I repurposed this PR just to correct two typos in IndicatorSet.

src/sets.jl
```diff
@@ -711,9 +711,9 @@ end
 """
     IndicatorSet{A, S <: AbstractScalarSet}(set::S)
 
-``\\{((y, x) \\in \\{0, 1\\} \\times \\mathbb{R}^n : y = 0 \\implies x \\in set\\}``
```
matbesancon: you added opening parentheses, which is the closing one?

dourouc05: Actually, I removed parentheses that were not closed anywhere.

odow:
Thanks @dourouc05! Let's keep discussing things in your extension package.

PR #1028, opened 2020-02-17
Add dual of GeometricMeanCone

blegat:
According to https://github.com/JuliaOpt/MathOptInterface.jl/pull/1017#discussion_r376703101, the dual of the GeometricMeanCone is the set of `(u, v)` such that there exists `b < 0` and `d_i` such that `sum(d_i) = u` and
`-b exp(d_i/b) <= exp(1) v_i` for all `i`.
Multiplying them all gives
`(-b)^n exp(u/b) <= exp(1)^n prod(w_i)` (1)
Let `k` such that `exp(k)= exp(1) sqrt[n](prod(w_i))` and `a = -b >= 0`
We have
`-u <= a * (n*k - n * log(a))`.
The maximum of the rhs is for `a = exp(k - 1)`.
Injecting this value of `a` in (1), we have
`exp(u / (n * b)) <= exp(1)`
hence
`-u <= -n * b`
and simplifying gives `-u <= n * sqrt[n](prod(w_i))`.

### DIFF
--- a/src/sets.jl
+++ b/src/sets.jl
@@ -271,7 +271,7 @@ The geometric mean cone ``\\{ (t,x) \\in \\mathbb{R}^{n+1} : x \\ge 0, t \\le \\
 ### Duality note
 
 The dual of the geometric mean cone is
-``\\{ (u, v) \\in \\mathbb{R}^{n+1} : v \\ge 0, u \\le n \\sqrt[n]{\\prod_i v_i} \\}`` of dimension `dimension```{}=n+1``.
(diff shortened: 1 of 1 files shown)


chriscoey:
we can think of a geomean cone as a power cone plus the nonpositive cone just for u. so wouldn't that mean the dual geomean cone is the dual power cone intersect the nonpositives for u? like:
```(u, w) : u <= 0, -u <= n * sqrt[n](prod(w_i))```
otherwise it seems both cones are too "large" i.e. both include more than R_- x R_+^(n-1)

blegat:
Yes indeed, I corrected the docstring and the reasoning of my post so that I arrive at the correct conclusion.

Issue #1030, opened 2020-02-18
Add option to make `MOIU.@model` and AbstractOptimizer


odow:
Currently, `@model` makes an `AbstractModel` subtype
https://github.com/JuliaOpt/MathOptInterface.jl/blob/fb303dceaa85dfff8f97e7fe13542c70fdef7754/src/Utilities/model.jl#L927
where `AbstractModel` is
https://github.com/JuliaOpt/MathOptInterface.jl/blob/fb303dceaa85dfff8f97e7fe13542c70fdef7754/src/Utilities/model.jl#L85

However, solvers like [PATH.jl](https://github.com/odow/PATH.jl/issues/4) and [AmplNLWriter](https://github.com/JuliaOpt/AmplNLWriter.jl/issues/97) use `@model` as an `AbstractOptimizer`. We should add a flag to let the super type be chosen by the constructor.

Alternatively, is there any reason not to make `AbstractMode{T} <: MOI.AbstractOptimizer`?

Re-reading, https://github.com/JuliaOpt/MathOptInterface.jl/issues/203, it's not obvious why we have the `ModelLike`/`AbstractOptimizer` distinction.

This feels like it's a legacy of the MPB solver/instance distinction.

cc @chkwon

blegat:
I have no objection for adding an option `@model`, it can even be part of a non-breaking release if we add a default value.
Removing `ModelLike` would be a separate discussion since it is a breaking change. Not sure what to think about this one.

odow:
Closed by #1031

PR #1031, opened 2020-02-19
Make AbstractModel <: AbstractOptimizer

odow:
### Commit 1

One potential fix for #1030 

Pros: this passes tests and is simple.

Cons: 
- we don't define `optimize!`, which might violate the contract of `AbstractOptimizer`. 
- Can this break code? 
I don't think it breaks any of mine. I'm also not aware of any code that assumes that a `@model` is `<: ModelLike`, but not `<: AbstractOptimizer`.

### Commit 2/3

Alternative fix for #1030

Adds a flag to switch between `MOIU.AbstractModelLike` and `MOIU.AbstractOptimizer`.

Pros: flexible.

Cons: will this break code that uses `MOIU.AbstractModel{T}`?

### DIFF
--- a/src/Utilities/model.jl
+++ b/src/Utilities/model.jl
@@ -851,8 +851,11 @@ addition to being different between constraints `F`-in-`S` for the same types
 use the the value of the index directly in a dictionary representing a mapping
 between constraint indices and something else.
 
-If `is_optimizer = true`, the resulting stuct is a subtype of
-`MOIU.AbstractOptimizer`, otherwise, it is a subtype of `MOIU.AbstractModelLike`.
+If `is_optimizer = true`, the resulting struct is a subtype of
+of `MOIU.AbstractOptimizer`, which is a subtype of
+[`MathOptInterface.AbstractOptimizer`](@ref), otherwise, it is a subtype of
+`MOIU.AbstractModelLike`, which is a subtype of
+[`MathOptInterface.ModelLike`](@ref).
 
 ### Examples
 
@@ -866,8 +869,8 @@ The model describing an linear program would be:
       (),                                                         # untyped scalar functions
       (MOI.ScalarAffineFunction,),                                #   typed scalar functions
       (MOI.VectorOfVariables,),                                   # untyped vector functions
-      (MOI.VectorAffineFunction,)                                 #   typed vector functions
-      true
+      (MOI.VectorAffineFunction,),                                #   typed vector functions
+      false
     )
 ```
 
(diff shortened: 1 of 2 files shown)


src/Utilities/model.jl
```diff
@@ -838,6 +851,9 @@ addition to being different between constraints `F`-in-`S` for the same types
 use the the value of the index directly in a dictionary representing a mapping
 between constraint indices and something else.
 
+If `is_optimizer = true`, the resulting stuct is a subtype of
```
blegat: stuct -> struct

src/Utilities/model.jl
```diff
@@ -838,6 +851,9 @@ addition to being different between constraints `F`-in-`S` for the same types
...
+`MOIU.AbstractOptimizer`, otherwise, it is a subtype of `MOIU.AbstractModelLike`.
```
blegat: The reader does not know what these abstract types are

src/Utilities/model.jl
```diff
@@ -850,7 +866,9 @@ The model describing an linear program would be:
       (),                                                         # untyped scalar functions
       (MOI.ScalarAffineFunction,),                                #   typed scalar functions
       (MOI.VectorOfVariables,),                                   # untyped vector functions
-      (MOI.VectorAffineFunction,))                                #   typed vector functions
+      (MOI.VectorAffineFunction,)                                 #   typed vector functions
```
blegat: Missing a comma at the end

src/Utilities/model.jl
```diff
@@ -850,7 +866,9 @@ The model describing an linear program would be:
...
+      true
```
blegat: I would write `false` as we write above "The **model** describing ..."

test/Utilities/model.jl
```diff
@@ -25,8 +25,28 @@ module TestExternalModel
         (NewFunction,),
         (),
         (),
-        ()
+        (),
```
blegat: Remove comma

test/Utilities/model.jl
```diff
@@ -25,8 +25,28 @@ module TestExternalModel
...
+    )
+end
+
+@testset "Super-types" begin
+    model = TestExternalModel.ExternalModel{Float64}()
+    optimizer = TestExternalModel.ExternalOptimizer{Float64}()
+    @test isa(model, MOIU.AbstractModelLike{Float64})
+    @test !isa(model, MOIU.AbstractOptimizer{Float64})
+    @test !isa(optimizer, MOIU.AbstractModelLike{Float64})
+    @test isa(optimizer, MOIU.AbstractOptimizer{Float64})
```
blegat: Also check with `MOI.AbstractOptimizer` to make sure the subtyping of the MOIU abstract types is correctly set up

odow:
I've tested this change on PATH.jl: https://github.com/odow/PATH.jl/pull/5

Seems to work fine.

PR #1033, opened 2020-02-21
Add tests for semi continuous and semi integer variables

rafabench:
I've added some tests for semi-continuous and semi-integer variables.
It's a simple model:
 min x 
     st   x >= y
           x ∈ {0.0} U [2.0,3.0]
           y = 0.0
And changed the fixed value of y to see the result.
I've tested on both CPLEX and Xpress solvers.

### DIFF
--- a/src/Test/intlinear.jl
+++ b/src/Test/intlinear.jl
@@ -688,12 +688,24 @@ function _semitest(model::MOI.ModelLike, config::TestConfig{T}, int::Bool) where
     @test MOIU.supports_default_copy_to(model, #=copy_names=# false)
     @test MOI.supports(model, MOI.ObjectiveFunction{MOI.ScalarAffineFunction{T}}())
     @test MOI.supports(model, MOI.ObjectiveSense())
-    @test MOI.supports_constraint(model, MOI.SingleVariable, MOI.Semicontinuous{T})
+    if !int
+        @test MOI.supports_constraint(model, MOI.SingleVariable, MOI.Semicontinuous{T})
+    else
+        @test MOI.supports_constraint(model, MOI.SingleVariable, MOI.Semiinteger{T})
+    end
 
     # 2 variables
+    #
+    # If int == false
     # min  x
     # st   x >= y
     #      x ∈ {0.0} U [2.0,3.0]
+    #      y = 0.0
+    #
+    # If int == true
+    # min  x
+    # st   x >= y
+    #      x ∈ {0.0} U {2.0} U {3.0}
     #      y = 0.0
 
     MOI.empty!(model)
@@ -702,9 +714,14 @@ function _semitest(model::MOI.ModelLike, config::TestConfig{T}, int::Bool) where
     v = MOI.add_variables(model, 2)
     @test MOI.get(model, MOI.NumberOfVariables()) == 2
 
-    vc1 = MOI.add_constraint(model, MOI.SingleVariable(v[1]), MOI.Semicontinuous(T(2), T(3)))
-    @test MOI.get(model, MOI.NumberOfConstraints{MOI.SingleVariable,MOI.Semicontinuous{T}}()) == 1
-
+    if !int
+        vc1 = MOI.add_constraint(model, MOI.SingleVariable(v[1]), MOI.Semicontinuous(T(2), T(3)))
+        @test MOI.get(model, MOI.NumberOfConstraints{MOI.SingleVariable,MOI.Semicontinuous{T}}()) == 1
+    else
+        vc1 = MOI.add_constraint(model, MOI.SingleVariable(v[1]), MOI.Semiinteger(T(2), T(3)))
+        @test MOI.get(model, MOI.NumberOfConstraints{MOI.SingleVariable,MOI.Semiinteger{T}}()) == 1
+    end
+    
     vc2 = MOI.add_constraint(model, MOI.SingleVariable(v[2]), MOI.EqualTo(zero(T)))
     @test MOI.get(model, MOI.NumberOfConstraints{MOI.SingleVariable,MOI.EqualTo{T}}()) == 1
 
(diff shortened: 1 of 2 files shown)


blegat:
Why did you close, it is a welcome addition

rafabench:
Sorry, it had a bug in the code. I was going to make another PR.

src/Test/intlinear.jl
```diff
@@ -681,6 +681,307 @@ function indicator4_test(model::MOI.ModelLike, config::TestConfig)
...
 end
 
+function semiconttest(model::MOI.ModelLike, config::TestConfig{T}) where T
+    atol = config.atol
+    rtol = config.rtol
+
+    @test MOIU.supports_default_copy_to(model, #=copy_names=# false)
+    @test MOI.supports(model, MOI.ObjectiveFunction{MOI.ScalarAffineFunction{T}}())
+    @test MOI.supports(model, MOI.ObjectiveSense())
+    @test MOI.supports_constraint(model, MOI.SingleVariable, MOI.MathOptInterface.Semicontinuous{T})
```
blegat: MOI.MathOptInterface

rafabench: Thank you! I will change that, the line 843 has the same error.

src/Test/intlinear.jl
```diff
@@ -681,6 +681,305 @@ function indicator4_test(model::MOI.ModelLike, config::TestConfig)
...
+
+    @test MOI.get(model, MOI.ObjectiveSense()) == MOI.MIN_SENSE
+
+    if config.solve
+        @test MOI.get(model, MOI.TerminationStatus()) == MOI.OPTIMIZE_NOT_CALLED
+
+        MOI.optimize!(model)
+
+        @test MOI.get(model, MOI.TerminationStatus()) == config.optimal_status
+
```
blegat: Can you remove the empty lines in between these lines to be consistent with other tests ?

test/Test/intlinear.jl
```diff
@@ -42,3 +42,51 @@ MOIU.set_mock_optimize!(mock,
...
+    (mock::MOIU.MockOptimizer) -> begin
+    MOI.set(mock, MOI.ObjectiveBound(), 3.0)
+    MOIU.mock_optimize!(mock, [3.0,3.0])
+    end,
+    (mock::MOIU.MockOptimizer) -> MOI.set(mock, MOI.TerminationStatus(), MOI.INFEASIBLE)
+)
+MOIT.semiconttest(mock,config)
+MOIU.set_mock_optimize!(mock,
+    (mock::MOIU.MockOptimizer) -> begin
+    MOI.set(mock, MOI.ObjectiveBound(), 0.0)
```
blegat: Indent between `begin` and `end`

src/Test/intlinear.jl
```diff
@@ -681,6 +681,305 @@ function indicator4_test(model::MOI.ModelLike, config::TestConfig)
...
+        @test MOI.get(model, MOI.ResultCount()) >= 1
+
+        @test MOI.get(model, MOI.PrimalStatus()) == MOI.FEASIBLE_POINT
+
+        @test MOI.get(model, MOI.ObjectiveValue()) ≈ 2.0 atol=atol rtol=rtol
+
+        @test MOI.get(model, MOI.VariablePrimal(), v) ≈ [2.0,2.0] atol=atol rtol=rtol
```
blegat: Add a space after the comma

src/Test/intlinear.jl
rafabench: You would like to remove only the lines between the `MOI.optimize!(model) `and `@test MOI.get(model, MOI.TerminationStatus()) == config.optimal_status` (lines 725,747, ...) or all empty lines in the `if config.solve` block (723, 725,727, ...) ?
I don't like these empty lines too, but I was trying to make it look the same as the other tests, like in lines 363, 303, ...

blegat: You can leave 723 and 725 bu remove 727, 729, ..., 737

src/Test/intlinear.jl
```diff
@@ -681,6 +681,244 @@ function indicator4_test(model::MOI.ModelLike, config::TestConfig)
     end
 end
 
+function semiconttest(model::MOI.ModelLike, config::TestConfig{T}) where T
```
blegat: These tests seem to differ in only a few places, could you refactor with an auxiliary function
```julia
function _semitest(model::MOI.ModelLike, config::TestConfig{T}, int::Bool) where T
    [...]
end
semiconttest(model::MOI.ModelLike, config::TestConfig) = _semitest(model, config, false)
semiinttest(model::MOI.ModelLike, config::TestConfig) = _semitest(model, config, true)
```

src/Test/intlinear.jl
```diff
@@ -681,6 +681,135 @@ function indicator4_test(model::MOI.ModelLike, config::TestConfig)
...
+
+    @test MOIU.supports_default_copy_to(model, #=copy_names=# false)
+    @test MOI.supports(model, MOI.ObjectiveFunction{MOI.ScalarAffineFunction{T}}())
+    @test MOI.supports(model, MOI.ObjectiveSense())
+    @test MOI.supports_constraint(model, MOI.SingleVariable, MOI.Semicontinuous{T})
+
+    # 2 variables
+    # min  x
+    # st   x >= y
+    #      x ∈ {0.0} U [2.0,3.0]
```
blegat: This should depend on `int`

rafabench: Something like this?
    # 2 variables
    #
    # If int == false
    # min  x
    # st   x >= y
    #      x ∈ {0.0} U [2.0,3.0]
    #      y = 0.0
    #
    # If int == true
    # min  x
    # st   x >= y
    #      x ∈ {0.0} U {2.0} U {3.0}
    #      y = 0.0

blegat: Yes but you can do the if only around the semi... constraint. Also, do `int` and `!int`

blegat:
Thanks for the contribution !

PR #1034, opened 2020-02-24
Bump BenchmarkTools version

ericphanson:
BenchmarkTools made a semver-breaking release although did not really break anything (just removed Julia 0.7 support and added features).

I could make a PR to add CompatHelper support if that would be desired.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -16,7 +16,7 @@ Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 Unicode = "4ec0a83e-493e-50e2-b9ac-8f72acf5a8f5"
 
 [compat]
-BenchmarkTools = "0.4"
+BenchmarkTools = "0.4, 0.5"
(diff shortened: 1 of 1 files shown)


odow:
CompatHelper would be good. I haven't looked at it yet.

ericphanson:
CompatHelper just checks once a day if there are new releases of dependencies and opens a PR to broaden compat with each package who got a new release (but does not check that they work). Ideally, then your CI tests the PR with the new version, and if it passes, you merge.

That doesn't work perfectly though, because you need to check that the CI actually tested with the new version of the package (because it might not if some other package is holding back the dependency whose compat you just broadened; ref https://github.com/bcbi/CompatHelper.jl/issues/183). Even though it's not perfect, I think it's still useful to at least know there is a new version of a dependency, and you can check the CI logs to see if it uses the new version as well.

odow:
Yes, at the very least, being notified of new releases seems useful.

PR #1037, opened 2020-02-26
[MOF] fix writing NLPBlock

odow:
Two bugs:
1. We were still assuming `NLPBlock` used `x[1]` instead of `x[VariableIndex(1)]`
2. NLP objectives were being overwritten by empty linear/quadratic ones because the JSON file for the test was wrong :(


odow:
Codecov is annoying again. 

@benoit objections to merging this and then tagging 0.9.12?

PR #1041, opened 2020-03-03
Optimality Checker 

yashcodes:
Hey! 
This PR is adding a function to check whether it is really possible to proof by the linear combination of Constraints that whether the Global Optimum value which we got from the Solver really sets the Min/Max limit on the Objective function in feasible set of x
eg.
Let the given the Min value of objective function  f(x) = x1 + x2 is 3
with constraints 
2x1 + x2 >= 5
x1 + x2 >= 6
x1 >= 1
Now  to check is 3 is really Minimum value or not, we can check is it possible that linear combination of constraints can prove the minimum value of objective function is 3
we can do it in following way
2x1 + x2 >= 5   x y1 
x1 + x2 >= 6.    x y2
x1 >= 1.             x y3
by adding all three ineqaulities gives (1)
(2y1 + y2 + y3) x1 + (y1 + y2) x2  >= 5y1 + 6y2 + y3. -- (1)
Now, campare ineqaulity (1) with (2) with 
x1 + x2 >= 3 --(2)
which gives the following set Linear equation
2y1 + y2 + y3 = 1
y1 + y2 = 1
5y1 + 6y2 + y3 = 3
So, if the following set of linear equations gives **non-negative** solutions of y1, y2, y3  then it is prooved that the minimum value of the objective function is 3 in a feasible set of x.
or we can call this approach is also called LP Duality - Complementery slackness condition.


odow:
Hi there! 

What is the goal of this? 

A prerequisite is https://github.com/JuliaOpt/MathOptInterface.jl/pull/1023.

Once that PR is moved, the next step is to write a function that takes a `ModelLike` object and a primal solution vector, and then returns the list of constraints that violate primal feasibility to some tolerance.

odow:
Do you can return the dual variables `y` directly from the solver using `ConstraintDual`.

You want a function along the lines of:
```Julia
function check_complementary_slackness(model::MOI.AbstractOptimizer)
    # 1) Check is linear program

    # 2) Get variable primal
    x = MOI.get(model, MOI.ListOfVariableIndices())
    x_val = [MOI.get(model, MOI.VariablePrimal(), xi) for xi in x]

    # 3) Get constraint dual/primal
    constraint_types = 
    for (F, S) in MOI.get(model, MOI.ListOfConstraints())
        for c in MOI.get(model, MOI.ListOfConstraintIndices{F, S}())
            c_val = MOI.get(model, MOI.ConstraintPrimal(), c)
            c_dual = MOI.get(model, MOI.ConstraintDual(), c)
        end
    end

    # 4) Check complementary slackness
    is_satisfied = ...
    return is_satisfied
end
```

yashcodes:
Thanks @odow , I am planning to do something similar.

odow:
Closing because the `distance_to_set` discussion was never resolved. Once we have something, we can revisit this.

PR #1042, opened 2020-03-05
Adding functions to return infeasible constraints

yashcodes:


odow:
See `MathOptInterface.Utilities.eval_variables`.
https://github.com/JuliaOpt/MathOptInterface.jl/blob/bab76823c14d421183a1e6a6da7cd09323d5d93b/src/Utilities/functions.jl#L6-L13

yashcodes:
Yes, I didn't know it is already Implemented. Anyways, I think I should start by calculating distance using these `eval_variables` function and functions in  [#1023](https://github.com/JuliaOpt/MathOptInterface.jl/pull/1023)

yashcodes:
@odow , after the merge of [PR #1023](https://github.com/JuliaOpt/MathOptInterface.jl/pull/1023), `return_infeasible_constraint` can be used to list out infeasbile constraints.
Please have a look.

joaquimg:
@yashcodes you should take a look in the code base of MOI.
This way you can follow patterns that were already applied elsewhere.
Also take a look a the JuMP style guide.

yashcodes:
Can you please be a little bit more specific what I did wrong in the last commit.

src/Utilities/checking_feasibility.jl
```diff
@@ -0,0 +1,22 @@
+function generate_varval(x)
+    return  eval(quote
+        t -> begin
+            return $x[t.value]
+            end
+        end)
+end
```
odow: Using `eval` is a sign that something is seriously wrong. You can just go:
```Julia
function generate_varval(x)
    return t -> x[t.value]
end
```

odow: There is also a problem that variable `.value` fields need not be contiguous, or ordered from `1:N`.

src/Utilities/Utilities.jl
```diff
@@ -3,6 +3,8 @@ module Utilities
 using LinearAlgebra # For dot
 
 using MathOptInterface
+
+import Base: invokelatest, push!
```
odow: You shouldn't use `invokelatest`, and you don't need to import `push!`.

src/Utilities/checking_feasibility.jl
```diff
@@ -0,0 +1,22 @@
...
+    
+    infeasible_constraints = Vector{Tuple{F, S}}()
+
+    for i in 1:length(dist_to_sets)
+        if dist_to_sets[i] > tol
+            push!(feasible_constraints, constraints[i])
+        end
+    end
+    return infeasible_constraints
+end
```
odow: We want a function like:
```Julia
function infeasible_constraints(model::MOI.ModelLike, x::Dict{VariableRef, Float64}, tol)
    violated_constraints = ConstraintRef[]
    # ...
    return violated_constraints
end
```

yashcodes:
@odow , Any more corrections Now?

yashcodes:
@joaquimg, @odow, This is the most basic version of `function return_dual_pair_gap` which is only can use to get dual_pair gap of LP optimization problems, with having constraints and Objective function type  MOI.ScalarAffineFunction.

odow:
Closing because we don't have a good notion of infeasibility. See also https://github.com/jump-dev/MathOptInterface.jl/pull/1041.

PR #1045, opened 2020-03-07
add L1/Linf cone tests and fix bridge when VAF has nonzero constants

chriscoey:
as discussed with @blegat. I have similar fixes incoming for some other bridges that use `MOIU.normalize_and_add_constraint` but I want to get this merged before I make those PRs.

### DIFF
--- a/src/Bridges/Constraint/norm_to_lp.jl
+++ b/src/Bridges/Constraint/norm_to_lp.jl
@@ -88,9 +88,9 @@ function bridge_constraint(::Type{NormOneBridge{T, F, G}}, model::MOI.ModelLike,
     return NormOneBridge{T, F, G}(y, nn_index)
 end
 
-MOI.supports_constraint(::Type{NormOneBridge{T}}, ::Type{<:MOI.AbstractVectorFunction}, ::Type{MOI.NormOneCone}) where T = true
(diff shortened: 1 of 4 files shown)


lkapelevich:
Bump for review

src/Bridges/Constraint/norm_to_lp.jl
```diff
@@ -70,69 +70,61 @@ The `NormOneCone` is representable with LP constraints, since
...
-    f_new = MOIU.operate(vcat, T, ub, lb)
-    nn_index = MOI.add_constraint(model, f_new, MOI.Nonnegatives(2d - 2))
-    return NormOneBridge{T, F, G, H}(y, ge_index, nn_index)
+    f_new = MOIU.operate(vcat, T, ge, ub, lb)
+    nn_index = MOI.add_constraint(model, f_new, MOI.Nonnegatives(2d - 1))
+    return NormOneBridge{T, F, G}(y, nn_index)
 end
 
-MOI.supports_constraint(::Type{NormOneBridge{T}}, ::Type{<:MOI.AbstractVectorFunction}, ::Type{MOI.NormOneCone}) where T = true
+MOI.supports_constraint(::Type{<:NormOneBridge{T}}, ::Type{<:MOI.AbstractVectorFunction}, ::Type{MOI.NormOneCone}) where T = true
```
blegat: We don't use `<:` apart from the indicator bridge (which should be changed). I'd rather keep the same thing for everywhere to ensure a consistent behavior of bridges

Issue #1029, opened 2020-02-17
MOI.Test: make it easy to understand what tests are run


odow:
In discussions with @kdheepak and @jd-lara:

Calling `unittest(solver, config)` (etc) is a little opaque with respect to what tests are actually run.

- Add `verbose` flag that prints out the name of the tests 
- Make it easy to obtain a list of tests that will be run so you can add them to `exclude` when developing the wrapper of a new solver
- Or document that the function calls the dictionary of the same name, e.g.,:
```Julia
julia> using MathOptInterface

julia> MathOptInterface.Test.unittests
Dict{String,Function} with 44 entries:
  "number_threads"        => number_threads
  "solve_qcp_edge_cases"  => solve_qcp_edge_cases
  "solve_zero_one_with_b… => solve_zero_one_with_bounds_1
  "solve_with_lowerbound" => solve_with_lowerbound
  "delete_nonnegative_va… => delete_nonnegative_variables
  "solve_affine_greatert… => solve_affine_greaterthan
  "add_variable"          => add_variable
  "solve_singlevariable_… => solve_singlevariable_obj
  "solve_duplicate_terms… => solve_duplicate_terms_vector_affine
  "solve_unbounded_model" => solve_unbounded_model
  "solve_constant_obj"    => solve_constant_obj
  "solve_single_variable… => solve_single_variable_dual_max
  "solve_single_variable… => solve_single_variable_dual_min
  "solver_name"           => solver_name
  "delete_variable"       => delete_variable
  "solve_time"            => solve_time
  "solve_duplicate_terms… => solve_duplicate_terms_obj
  "delete_variables"      => delete_variable
  "raw_status_string"     => raw_status_string
  "add_variables"         => add_variables
  "feasibility_sense"     => feasibility_sense
  "solve_affine_equalto"  => solve_affine_equalto
  "max_sense"             => max_sense
  "variablenames"         => variablenames
  "solve_integer_edge_ca… => solve_integer_edge_cases
  "silent"                => silent
  "update_dimension_nonn… => update_dimension_nonnegative_variables
  "solve_blank_obj"       => solve_blank_obj
  "min_sense"             => min_sense
  "solve_zero_one_with_b… => solve_zero_one_with_bounds_3
  "solve_affine_interval" => solve_affine_interval
  "solve_with_upperbound" => solve_with_upperbound
  "solve_duplicate_terms… => solve_duplicate_terms_scalar_affine
[...]

PR #1050, opened 2020-03-11
Updating docs for MOI.Test.unittest

yashcodes:
Solves [#1029](https://github.com/JuliaOpt/MathOptInterface.jl/issues/1029)

### DIFF
--- a/docs/src/apimanual.md
+++ b/docs/src/apimanual.md
@@ -1325,18 +1325,11 @@ end
 
 const BRIDGED = MOI.instantiate(OPTIMIZER_CONSTRUCTOR, with_bridge_type = Float64)
 const CONFIG = MOIT.TestConfig(atol=1e-6, rtol=1e-6)
-```
-`MOI.Test.unittests` is a dictionary having "test name":test function as a key:value pair and `MOIT.unittest(model, config)` runs all of these tests. You can use `keys(MOI.Test.unittests)` to obtain a list of keys. To exclude any particular tests you can pass vector of there keys to the `exclude` keyword argument of unittest.
-```julia
-@testset "Unit" begin
-    MOIT.unittest(BRIDGED, CONFIG)
-end
 
 @testset "Unit" begin
-    MOIT.unittest(BRIDGED, CONFIG, exclude =  ["number_threads", "solve_qcp_edge_cases"])
+    MOIT.unittest(BRIDGED, CONFIG, ["number_threads", "solve_qcp_edge_cases"]) # test all the functions included in dictionary MOI.Test.unittests except functions number_threads and solve_qcp_edge_cases"
 end
-```
-```julia
+
 @testset "Modification" begin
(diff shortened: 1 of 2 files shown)


odow:
I've added a comment to the issue explaining our preferred resolution: https://github.com/JuliaOpt/MathOptInterface.jl/issues/1029#issuecomment-597638306

yashcodes:
@odow Please check!!

src/Test/config.jl
```diff
@@ -39,7 +39,7 @@ macro moitestset(setname, subsets=false)
         runtest = :( f(model, config) )
     end
     esc(:(
-        function $testname(model::$MOI.ModelLike, config::$MOI.Test.TestConfig, exclude::Vector{String} = String[])
+        function $testname(model::$MOI.ModelLike, config::$MOI.Test.TestConfig; exclude::Vector{String} = String[])
```
odow: This breaks every solver. Please leave as an optional positional argument.

docs/src/apimanual.md
```diff
@@ -1325,11 +1325,18 @@ end
 
 const BRIDGED = MOI.instantiate(OPTIMIZER_CONSTRUCTOR, with_bridge_type = Float64)
 const CONFIG = MOIT.TestConfig(atol=1e-6, rtol=1e-6)
-
+```
+`MOI.Test.unittests` is a dictionary having "test name":test function as a key:value pair and `MOIT.unittest(model, config)` runs all of these tests. You can use `keys(MOI.Test.unittests)` to obtain a list of keys. To exclude any particular tests you can pass vector of there keys to the `exclude` keyword argument of unittest.
```
odow: This applies, not just for `unittests`, but all of the `MOI.Test.XXX` functions.

You want something like

Test functions like `MOI.Test.unittest` and `MOI.Test.modificationtest` are wrappers around corresponding dictionaries `MOI.Test.unittests` and `MOI.Test.modificationtests`. The keys of each dictionary (strings describing the test) map to functions that take two arguments: an optimizer and a `MOI.Test.TestConfig` object. Exclude tests by passing a vector of strings corresponding to the test keys you want to exclude as the third positional argument to the test function (e.g., `MOI.Test.unittest`). Print a list of all keys using `println.(keys(MOI.Test.unittests))`.

You could also put this paragraph below the code, so that someone can copy-paste the code block in one piece.

docs/src/apimanual.md
```diff
@@ -1325,11 +1325,18 @@ end
...
+```julia
 @testset "Unit" begin
     MOIT.unittest(BRIDGED, CONFIG)
 end
 
+@testset "Unit" begin
+    MOIT.unittest(BRIDGED, CONFIG, exclude =  ["number_threads", "solve_qcp_edge_cases"])
+end
```
odow: No need to double. Just add the positional argument to `unites` above.

yashcodes:
@odow Hope this is what we needed.

PR #1055, opened 2020-03-26
Print bridging scheme

yashcodes:
@blegat @matbesancon 
This PR adds a function `print_bridging_strategy` which prints out the strategy in which the unsupported constraint (unsupported by inner model ) bridged to supported constraint type by added bridges in`LazyBridgeOptimizer` (if it could bridge it) by showing the order in which bridges are used (i.e the strategy got from bellman-ford)   
I also added test cases . Please have a look 
I think It is the second step of the project Visualisation of brigde graph.

### DIFF
--- a/test/Bridges/lazy_bridge_optimizer.jl
+++ b/test/Bridges/lazy_bridge_optimizer.jl
@@ -608,7 +608,7 @@ Objective function of type `MOI.ScalarQuadraticFunction{$T}` is not supported an
 """
         MOIB.add_bridge(bridged, MOIB.Variable.RSOCtoPSDBridge{T})
         MOIB.add_bridge(bridged, MOIB.Constraint.ScalarFunctionizeBridge{T})
-        @test debug_string(MOIB.print_supporting_scheme, attr) == String("Objective function of type `MOI.ScalarQuadraticFunction{$T}` is not supported and" 
+        @test debug_string(MOIB.print_supporting_scheme, attr) == String("Objective function of type `MOI.ScalarQuadraticFunction{$T}` is not supported and"
                                                * " cannot be bridged into a supported objective function by adding only supported"
                                                * " constrained variables and constraints. To see detatils use debug_unsupported.")
         @test debug_string(MOIB.debug_supports, attr) =="""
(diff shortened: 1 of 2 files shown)


yashcodes:
I dont understand why codecov/patch is failing, Can anyone tell whts wrong?

yashcodes:
@odow I think this PR is good to go now!!

test/Bridges/lazy_bridge_optimizer.jl
```diff
@@ -610,7 +629,19 @@ Objective function of type `MOI.ScalarQuadraticFunction{$T}` is not supported an
   (5) `MOI.ScalarQuadraticFunction{$T}`-in-`MOI.LessThan{$T}` constraints are not supported
 """
         MOIB.add_bridge(bridged, MOIB.Constraint.ScalarizeBridge{T})
-        @test debug_string(MOIB.debug_supports, attr) == "Objective function of type `MOI.ScalarQuadraticFunction{$T}` is supported.\n"
+        @test debug_string(MOIB.debug_supports, attr) == "Objective function of type `MOI.ScalarQuadraticFunction{$T}` is supported. To see bridging strategy use print_supporting_scheme.\n"
+        @test debug_string(MOIB.print_supporting_scheme, attr) == """
+|1| objective function of type `MOI.ScalarQuadraticFunction{$T}` is bridged (distance 12) by MOIB.Objective.SlackBridge{$T,MOI.ScalarQuadraticFunction{$T},MOI.ScalarQuadraticFunction{$T}} to
+  (1) `MOI.ScalarQuadraticFunction{$T}`-in-`MOI.GreaterThan{$T}` constraints are bridged (distance 5) by MOIB.Constraint.QuadtoSOCBridge{$T} to
+    (2) `MOI.VectorAffineFunction{$T}`-in-`MOI.RotatedSecondOrderCone` constraints are bridged (distance 4) by MOIB.Constraint.VectorSlackBridge{$T,MOI.VectorAffineFunction{$T},MOI.RotatedSecondOrderCone} to
```
blegat: Here this block is printed twice. To avoid this, we should have something like
```
|1| ...
   |*| objective function for type `MOI.SingleVariable` is natively supported
   (1) ... constraints are bridged
(1) ... constraints are bridged (distance 5) by ...
```
here we use a star `|*|` and not a number because since this node is not natively supported, we won't detail how it's bridge.
In `(1) ... constraints are bridged` we just say it's bridged, no other details, the reader is refered to `(1)` for more details.

odow:
Closing as stale. We can revisit in a new PR.

PR #1056, opened 2020-03-26
Interface from IIS.

dourouc05:
Exact copy from the common bits in the existing solvers, except that `compute_conflict!` now has a !. 

https://github.com/JuliaOpt/CPLEX.jl/blob/2756458f5f6c02b784e5c2fb4594c930d6febf60/src/MOI/MOI_wrapper.jl#L2745

https://github.com/JuliaOpt/Xpress.jl/blob/master/src/MOI/MOI_wrapper.jl#L3291

https://github.com/JuliaOpt/Gurobi.jl/blob/master/src/MOI_wrapper.jl#L2498

This follows https://github.com/JuliaOpt/JuMP.jl/issues/1035#issuecomment-509636707.


src/attributes.jl
```diff
@@ -910,6 +910,24 @@ A model attribute for the number of results available.
...
+A model attribute for the `TerminationStatusCode` explaining why the conflict
+refiner stopped when computing the conflict `N`.
+
+If a minimal conflict is found, it will return `OPTIMAL`.
+If the problem is feasible, it will return `INFEASIBLE`.
+If `compute_conflict` has not been called yet, it will return
+`OPTIMIZE_NOT_CALLED`.
+"""
+struct ConflictStatus <: AbstractModelAttribute
+    N::Int
```
joaquimg: we are moving towards result_index instead of N: https://github.com/JuliaOpt/MathOptInterface.jl/pull/999

src/attributes.jl
```diff
@@ -1117,6 +1135,19 @@ function throw_set_error_fallback(::ModelLike, ::ConstraintSet,
...
 
+"""
+    ConstraintConflictStatus()
+    ConstraintConflictStatus(N)
+
+A Boolean constraint attribute indicating whether the constraint participates
+in conflict `N`. Some solvers call the conflict IIS.
+"""
+struct ConstraintConflictStatus <: AbstractConstraintAttribute
+    N::Int
```
joaquimg: same here

dourouc05:
Thanks for the pointer. 

Is it required to add tests for this? I'm not sure exactly what to test… And it looks like not all attributes are tested either.

dourouc05:
@blegat @joaquimg Is anything else required to make this PR go on?

blegat:
> Is anything else required to make this PR go on?

Oops, I did comments on my phone but forgot to click on "Submit", you should see them now, thanks for the reminder.

src/MathOptInterface.jl
```diff
@@ -32,6 +32,22 @@ Start the solution procedure.
 """
 function optimize! end
 
+"""
+    compute_conflict!(optimizer::AbstractOptimizer)
+
+Compute a minimal subset of the constraints and variables that keep the model
```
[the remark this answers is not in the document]
dourouc05: The new commit uses your new formulation, even though I avoided the wording "infeasibility certificate": MOI does not really care how the solver does its job (for instance, `optimize!` does not mandate the use of B&B or interior point).

src/attributes.jl
```diff
@@ -910,6 +910,24 @@ A model attribute for the number of results available.
 """
 struct ResultCount <: AbstractModelAttribute end
 
+"""
+    ConflictStatus(result_index)
+    ConflictStatus()
+
+A model attribute for the `TerminationStatusCode` explaining why the conflict
+refiner stopped when computing the conflict `result_index`.
```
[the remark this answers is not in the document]
dourouc05: There may be several minimal sets, even though as far as I know only Xpress can compute several IISes. (See https://www.msi-jp.com/xpress/learning/square/optimizer-2015.pdf, p. 38: `XPRSiisfirst` and `XPRSiisnext`.)

dourouc05:
GitHub should really warn you more when you do not post all your messages, but at least they were kept!

src/attributes.jl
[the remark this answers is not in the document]
blegat: Then it might help to have an equivalent of `ResultCount` for conflict. Also, as there is no relation between the index of the result and the index of the conflict, it might be better to use another name than `result_index`. `ConflictCount` and `conflict_index` maybe ?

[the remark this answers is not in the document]
dourouc05: Indeed, for `ResultCount`. For `result_index`, I just followed @joaquimg and https://github.com/JuliaOpt/MathOptInterface.jl/pull/999. An IIS can be still thought of as a result of some computation, so I was wondering if it would not ease some parts of the code to keep the same field name?

src/MathOptInterface.jl
```diff
@@ -32,6 +32,22 @@ Start the solution procedure.
 """
 function optimize! end
 
+"""
+    compute_conflict!(optimizer::AbstractOptimizer)
+
+Computes a minimal subset of constraints such that the model with the other
+constraint removed is still infeasible.
+
+See also `ConflictStatus` and `ConstraintConflictStatus`.
```
blegat, line 41: Put these around brackets and add `(@ref)`.

src/attributes.jl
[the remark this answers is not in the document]
blegat: I'm hesitant. @odow what's your take on this ?

src/attributes.jl
```diff
@@ -910,6 +910,31 @@ A model attribute for the number of results available.
...
+
+If a minimal conflict is found, it will return `OPTIMAL`.
+If the problem is feasible, it will return `INFEASIBLE`.
+If `compute_conflict` has not been called yet, it will return
+`OPTIMIZE_NOT_CALLED`.
```
blegat, line 923: It might be better to create a new enum with three entries. `OPTIMIZE_NOT_CALLED` and `INFEASIBLE` are confusing here.

dourouc05:
I just did this in the last commit. I also added an enum for the status, so that all the information the solver gives is accessible through MOI.

dourouc05:
@blegat I just changed the value for the enum. However, for the indenting, you had a look at an old version :).

src/MathOptInterface.jl
```diff
@@ -32,6 +32,23 @@ Start the solution procedure.
...
+constraint removed is still infeasible.
+
+See also [`ConflictStatus`](@ref), [`ConflictCount`](@ref), and
+[`ConstraintConflictStatus`](@ref).
+
+### Note
+
+If the model is modified after a call to `compute_conflict!`, the implementor
+is not obliged to purge the conflict. Any calls to the above attributes may
+return values for the original conflict without a warning.
```
odow, line 48: There is a flip-side to this as well. If the model is modified, the conflict may be discarded.

src/attributes.jl
```diff
@@ -910,6 +910,45 @@ A model attribute for the number of results available.
...
 
+"""
+    ConflictStatusCode
+
+An Enum of possible values for the `ConflictStatus` attribute. This attribute
+is meant to explain the reason why the conflict finder stopped executing in the
+most recent call to [`compute_conflict!`](@ref).
+
+Possible values are:
+* `NO_CONFLICT `: the function [`compute_conflict!`](@ref) has
```
odow, line 921: `COMPUTE_CONFLICT_NOT_CALLED`?

src/attributes.jl
```diff
@@ -910,6 +910,45 @@ A model attribute for the number of results available.
...
+  not yet been called
+* `FEASIBLE`: there is no conflict because the problem is feasible
```
odow, line 923: `NO_CONFLICT_FOUND`?

src/attributes.jl
```diff
@@ -910,6 +910,45 @@ A model attribute for the number of results available.
...
+* `FOUND`: at least one conflict could be found
```
odow, line 924: `CONFLICT_FOUND`?

src/attributes.jl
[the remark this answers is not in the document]
odow: Do we need to add multiple conflicts? It seems that this is a pretty niche part of the API. I vote just one for now, we can always add more later. If people are wanting to look at multiple IISs, they should use the solver API directly.

src/attributes.jl
```diff
@@ -1117,6 +1156,40 @@ function throw_set_error_fallback(::ModelLike, ::ConstraintSet,
...
 
+"""
+    ConflictParticipationStatusCode
+
+An Enum of possible values for the `ConstraintConflictStatus` attribute.
+This attribute is meant to indicate whether a given constraint participates
+or not in the last computed conflict.
+
+Possible values are:
+* `EXCLUDED`: the constraint does not participate in the conflict
```
odow, line 1167: `NOT_IN_CONFLICT`?

src/attributes.jl
```diff
@@ -1117,6 +1156,40 @@ function throw_set_error_fallback(::ModelLike, ::ConstraintSet,
...
+* `INCLUDED`: the constraint participates in the conflict
```
odow, line 1168: `IN_CONFLICT`

src/attributes.jl
```diff
@@ -1117,6 +1156,40 @@ function throw_set_error_fallback(::ModelLike, ::ConstraintSet,
...
+* `POSSIBLY_INCLUDED`: the constraint may participate in the conflict,
```
odow, line 1169: `MAYBE_IN_CONFLICT`?

dourouc05:
I modified the PR accordingly. Indeed, having just one conflict simplifies things quite a bit.

src/attributes.jl
```diff
@@ -910,6 +910,29 @@ A model attribute for the number of results available.
...
+    ConflictStatusCode
+
+An Enum of possible values for the `ConflictStatus` attribute. This attribute
+is meant to explain the reason why the conflict finder stopped executing in the
+most recent call to [`compute_conflict!`](@ref).
+
+Possible values are:
+* `COMPUTE_CONFLICT_NOT_CALLED`: the function [`compute_conflict!`](@ref) has
+  not yet been called
+* `NO_CONFLICT_FOUND`: there is no conflict because the problem is feasible
```
odow, line 923: It may also be that the IIS failed to find a conflict, despite the problem being infeasible.

src/attributes.jl
```diff
@@ -910,6 +910,29 @@ A model attribute for the number of results available.
...
+* `CONFLICT_FOUND`: at least one conflict could be found
+"""
+@enum(ConflictStatusCode, COMPUTE_CONFLICT_NOT_CALLED, NO_CONFLICT_FOUND, CONFLICT_FOUND)
+
+"""
+    ConflictStatus()
+
+A model attribute for the `ConflictStatusCode` explaining why the conflict
```
odow, line 931: ````[`ConflictStatusCode`](@ref)````

src/attributes.jl
```diff
@@ -1117,6 +1140,31 @@ function throw_set_error_fallback(::ModelLike, ::ConstraintSet,
...
+  the solver was not able to prove that the constraint can be excluded from
+  the conflict
+"""
+@enum(ConflictParticipationStatusCode, NOT_IN_CONFLICT, IN_CONFLICT, MAYBE_IN_CONFLICT)
+
+"""
+    ConstraintConflictStatus()
+
+A constraint attribute indicating whether the constraint participates
+in the conflict. Its type is `ConflictParticipationStatusCode`.
```
odow, line 1163: ````[`ConflictPartitipationStatusCode`](@ref)````

src/attributes.jl
```diff
@@ -1117,6 +1140,31 @@ function throw_set_error_fallback(::ModelLike, ::ConstraintSet,
...
+Some solvers call the conflict IIS.
```
odow, line 1164: This note is probably best put in `compute_conflict!` as a `Note: some solvers call a set of conflicting constraints an Irreducible Inconsistent Subsystem (IIS).`

src/attributes.jl
dourouc05: I'm adding a third status code (only available through Xpress) to catch that case. I guess most solvers will still use `NO_CONFLICT_FOUND`, though.

dourouc05:
What tests would be needed, here? I'm not sure what to test for MOI… Or you mean a few things for the solver to check their compatibility in MOIT? 

For the documentation, is there anything to add in `docs/src/apireference.md`? Maybe a section in https://www.juliaopt.org/MathOptInterface.jl/dev/apimanual/#Advanced-1 could be useful?

src/attributes.jl
```diff
@@ -1117,6 +1146,30 @@ function throw_set_error_fallback(::ModelLike, ::ConstraintSet,
...
+* `MAYBE_IN_CONFLICT`: the constraint may participate in the conflict,
+  the solver was not able to prove that the constraint can be excluded from
+  the conflict
+"""
+@enum(ConflictParticipationStatusCode, NOT_IN_CONFLICT, IN_CONFLICT, MAYBE_IN_CONFLICT)
+
+"""
+    ConstraintConflictStatus()
+
+A constraint attribute indicating whether the constraint participates
```
blegat, line 1168: Say that the value is a `Bool`

dourouc05: It's no more a Bool, but rather a `ConflictParticipationStatusCode` (just above) :).

dourouc05:
Is this enough?

docs/src/apireference.md
```diff
@@ -169,6 +170,15 @@ The value of the attribute is of type `TerminationStatusCode`.
 TerminationStatusCode
 ```
 
+### Conflict Status
+
+The `ConflictStatus` attribute indicates why the optimizer stopped executing.
+The value of the attribute is of type `ConflictStatusCode`.
+
+```@docs
+ConflictStatusCode
```
blegat, line 179: I would add `compute_conflict!` here and move all attributes to this section

dourouc05: Better this way?

dourouc05: 👋

docs/src/apireference.md
```diff
@@ -169,6 +169,18 @@ The value of the attribute is of type `TerminationStatusCode`.
...
+### Conflict Status
+
+The `ConflictStatus` attribute indicates why the conflict finder stopped executing.
+The value of the attribute is of type `ConflictStatusCode`.
+
+```@docs
+compute_conflict!
+ConflictStatus
+ConflictStatusCode
+ConstraintConflictStatus
```
blegat, line 181: Travis is failing because you are missing `ConflictParticipationStatusCode` here

dourouc05: Fixed, thanks!

dourouc05:
By the way, should I include an entry for NEWS in this PR?

blegat:
We usually collect NEWS entry separately as otherwise all PRs will conflict with each other.

PR #1057, opened 2020-03-29
Adding keyword argument `io`  to print_graph and documentation

yashcodes:
This PR adds missing documentation of `print_graph` and add keyword argument `io`


yashcodes:
Oh that's one embarrassing mistake :p

yashcodes:
The PR was intended to let users use `print_graph` and other debugging functions in a similar fashion, I think there must be similarity in a way to make a call to `print_graph` and `debug_supported`.

odow:
Closing because the new method is redundant.

Issue #990, opened 2020-01-05
Doctests fail on julia 1.3


mlubin:
It looks like there were printing changes between Julia 1.0 and 1.3. This is causing the docs to fail to build.

```
┌ Error: doctest failure in src/apireference.md:588-594
│ 
│ ```jldoctest bridged_variable_function
│ MOI.Bridges.bridged_variable_function(bridged_model, bridged_variable)
│ 
│ # output
│ 
│ MOI.ScalarAffineFunction{Float64}(MOI.ScalarAffineTerm{Float64}[ScalarAffineTerm{Float64}(1.0, VariableIndex(1))], 1.0)
│ ```
│ 
│ Subexpression:
│ 
│ MOI.Bridges.bridged_variable_function(bridged_model, bridged_variable)
│ 
│ Evaluated output:
│ 
│ MathOptInterface.ScalarAffineFunction{Float64}(MathOptInterface.ScalarAffineTerm{Float64}[MathOptInterface.ScalarAffineTerm{Float64}(1.0, MathOptInterface.VariableIndex(1))], 1.0)
│ 
│ Expected output:
│ 
│ MOI.ScalarAffineFunction{Float64}(MOI.ScalarAffineTerm{Float64}[ScalarAffineTerm{Float64}(1.0, VariableIndex(1))], 1.0)
│ 
│   diff = MOI.ScalarAffineFunction{Float64}(MOI.ScalarAffineTerm{Float64}[ScalarAffineTerm{Float64}(1.0, VariableIndex(1))], MathOptInterface.ScalarAffineFunction{Float64}(MathOptInterface.ScalarAffineTerm{Float64}[MathOptInterface.ScalarAffineTerm{Float64}(1.0, MathOptInterface.VariableIndex(1))], 1.0)
└ @ Documenter.DocTests ~/.julia/packages/Documenter/BZNsm/src/DocTests.jl:364
┌ Error: doctest failure in src/apireference.md:597-603
│ 
│ ```jldoctest bridged_variable_function
│ MOI.Bridges.unbridged_variable_function(bridged_model, inner_variables[1])
│ 
│ # output
│ 
│ MOI.ScalarAffineFunction{Float64}(MOI.ScalarAffineTerm{Float64}[ScalarAffineTerm{Float64}(1.0, VariableIndex(-1))], -1.0)
│ ```
│ 
│ Subexpression:
│ 
[...]

PR #1058, opened 2020-04-02
Fix Documenter on Julia 1.3

frapac:
The issue arises because of missing `MOI` prefixes in `# output`.
Should solve #990

### DIFF
--- a/docs/src/apireference.md
+++ b/docs/src/apireference.md
@@ -597,7 +597,7 @@ MOI.Bridges.bridged_variable_function(bridged_model, bridged_variable)
 
 # output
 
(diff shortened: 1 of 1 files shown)


blegat:
The issue is that this makes it fail for Julia v1.0. I don't know a way that works for both

odow:
Maybe we should just close this (and the issue), and deal with it when we want to update the version that the docs build on.

blegat:
Yes, the doc should be built on Julia v1.0 and is not expected to work on other versions

frapac:
That makes sense. I will close this PR then. Thanks for your feedbacks.

PR #1066, opened 2020-04-08
Adding get function for MOI.ConstraintSet for NonposToNonnegBridge

yashcodes:
Adding a get function to for `MOI.ConstraintSet` attribute for `FlipSignBridge` . Test is also added.

### DIFF
--- a/src/Bridges/Variable/flip_sign.jl
+++ b/src/Bridges/Variable/flip_sign.jl
@@ -48,8 +48,8 @@ function MOI.delete(model::MOI.ModelLike, bridge::FlipSignBridge, i::IndexInVect
 end
 
 # Attributes, Bridge acting as a constraint
-function MOI.get(model::MOI.ModelLike, attr::MOI.ConstraintSet, bridge::FlipSignBridge)
-    return MOI.Nonpositives(length(bridge.flipped_variables))
+function MOI.get(model::MOI.ModelLike, attr::MOI.ConstraintSet, bridge::FlipSignBridge{T, S1}) where {T, S1<:MOI.AbstractVectorSet}
+    return S1(length(bridge.flipped_variables))
 end
 
 function MOI.get(model::MOI.ModelLike,
--- a/src/Bridges/Variable/flip_sign.jl
+++ b/src/Bridges/Variable/flip_sign.jl
@@ -48,6 +48,9 @@ function MOI.delete(model::MOI.ModelLike, bridge::FlipSignBridge, i::IndexInVect
 end
 
 # Attributes, Bridge acting as a constraint
+function MOI.get(model::MOI.ModelLike, attr::MOI.ConstraintSet, bridge::FlipSignBridge)
+    return MOI.Nonpositives(length(bridge.flipped_variables))
+end
 
 function MOI.get(model::MOI.ModelLike,
                  attr::Union{MOI.ConstraintPrimal, MOI.ConstraintDual},
--- a/test/Bridges/Variable/flip_sign.jl
+++ b/test/Bridges/Variable/flip_sign.jl
@@ -31,59 +31,49 @@ config = MOIT.TestConfig()
         x, y_flipped, z, s = MOI.get(mock, MOI.ListOfVariableIndices())
         @test MOI.get(mock, MOI.VariablePrimalStart(), y_flipped) == -1
         @test MOI.get(bridged_mock, MOI.VariablePrimalStart(), y) == 1
-    end
 
-    @testset "Test Mock model" begin
-        var_names = ["a", "b", "c", "d"]
+        var_names = ["x", "y", "z", "w"]
         MOI.set(mock, MOI.VariableName(), MOI.get(mock, MOI.ListOfVariableIndices()), var_names)
-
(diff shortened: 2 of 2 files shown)


src/Bridges/Variable/flip_sign.jl
```diff
@@ -48,6 +48,9 @@ function MOI.delete(model::MOI.ModelLike, bridge::FlipSignBridge, i::IndexInVect
 end
 
 # Attributes, Bridge acting as a constraint
+function MOI.get(model::MOI.ModelLike, attr::MOI.ConstraintSet, bridge::FlipSignBridge)
+    return MOI.Nonpositives(length(bridge.flipped_variables))
```
blegat, line 52: This is only valid for `NonposToNonnegBridge`, you can do
```julia
..., bridge::FlipSignBridge{T, S1}) where {T, S1}
return S1(...)
```

test/Bridges/Variable/flip_sign.jl
```diff
@@ -24,6 +24,9 @@ config = MOIT.TestConfig()
         vis = MOI.get(bridged_mock, MOI.ListOfVariableIndices())
         y = vis[4]
         @test y.value == -1
+        
+        con = MOI.get(bridged_mock, MOI.ListOfConstraintIndices{MathOptInterface.VectorOfVariables, MathOptInterface.Nonpositives}())[1]
```
blegat, line 28: Use `test_models_equal` to test the getter of `ConstraintSet`. See
https://github.com/JuliaOpt/MathOptInterface.jl/blob/5b25a050d99efb1a57a796d4ca2ffc5cdcf44e17/test/Bridges/Constraint/geomean.jl#L29-L72

src/Bridges/Variable/flip_sign.jl
yashcodes: Yeah cool, that would much nicer.

yashcodes: But I think abstract type `FlipSignBridge{T, S1}` has only `NonposToNonnegBridge` as a subtype, there is no `NonnegToNonposBridge`.So, I think using `MOI.Nonpositives(length(bridge.flipped_variables))` would also be ok.

test/Bridges/Variable/flip_sign.jl
```diff
@@ -33,6 +33,59 @@ config = MOIT.TestConfig()
         @test MOI.get(bridged_mock, MOI.VariablePrimalStart(), y) == 1
     end
 
+    @testset "Test Mock model" begin
```
blegat, line 36: Put this testset inside the `lin2v` testset.

test/Bridges/Variable/flip_sign.jl
```diff
@@ -33,6 +33,59 @@ config = MOIT.TestConfig()
...
+        minobjective: 3*a + -2*b + -4*c
+        """
+        model = MOIU.Model{Float64}()
+        MOIU.loadfromstring!(model, s)
+        MOIU.test_models_equal(mock, model, var_names, ["cd", "cb", "cc", "cex"])
+    end
+
+    @testset "Test Bridged model" begin
+        var_names = ["x", "y", "z", "w"]
+        MOI.set(bridged_mock, MOI.VariableName(), MOI.get(bridged_mock, MOI.ListOfVariableIndices()), var_names)
```
blegat, line 64: Move this outside and before both testsets

test/Bridges/Variable/flip_sign.jl
```diff
@@ -33,6 +33,59 @@ config = MOIT.TestConfig()
         @test MOI.get(bridged_mock, MOI.VariablePrimalStart(), y) == 1
     end
 
+    @testset "Test Mock model" begin
+        var_names = ["a", "b", "c", "d"]
+        MOI.set(mock, MOI.VariableName(), MOI.get(mock, MOI.ListOfVariableIndices()), var_names)
```
blegat, line 38: Only set the name of the bridged variables, the other variables should have the names `x`, `z` and `w` already set.

test/Bridges/Variable/flip_sign.jl
```diff
@@ -33,6 +33,59 @@ config = MOIT.TestConfig()
...
+
+        con_d = MOI.get(mock, MOI.ListOfConstraintIndices{MathOptInterface.VectorOfVariables, MathOptInterface.Zeros}())[1]
+        con_bc = MOI.get(mock, MOI.ListOfConstraintIndices{MathOptInterface.VectorOfVariables, MathOptInterface.Nonnegatives}())
+        con_ex = MOI.get(mock, MOI.ListOfConstraintIndices{MathOptInterface.VectorAffineFunction{Float64}, MathOptInterface.Zeros}())[1]
```
blegat, line 42: Same than for variable names, you should not set the name here of the constraint that are not bridged. It should be set only once for both `mock` and `bridged_mock`.

test/Bridges/Variable/flip_sign.jl
yashcodes: Only this or both testsets inside in line2v?

PR #1079, opened 2020-04-28
Feature to read/write integer constraints

### LINKED ISSUE #1074: Integer variable section for SDPA Format
Since #1071 and #1073 are made for SDPA, I would like to add a suggestion for the feature to indicate integer variables in SDPA format. Here is what has been proposed: http://www.opt.tu-darmstadt.de/scipsdp/downloads/data_format.txt

kibaekkim:
I have added a feature to read/write integer restrictions on single variables. This resolves #1074.
@blegat @odow, it would be great if you can review this PR.

The feature is based on the extension in http://www.opt.tu-darmstadt.de/scipsdp/downloads/data_format.txt and should be backward compatible.

### DIFF
--- a/src/FileFormats/SDPA/SDPA.jl
+++ b/src/FileFormats/SDPA/SDPA.jl
@@ -8,7 +8,7 @@ import MathOptInterface
 const MOI = MathOptInterface
 
 MOI.Utilities.@model(Model,
-    (),
+    (MOI.Integer,),
     (),
     (MOI.Nonnegatives, MOI.PositiveSemidefiniteConeTriangle),
     (),
@@ -25,6 +25,10 @@ function MOI.supports_constraint(
 ) where {T}
 
     return false
+end
+function MOI.supports_constraint(
+    ::Model, ::Type{MOI.SingleVariable}, ::Type{MOI.Integer})
+    return true
 end
 
 function MOI.supports(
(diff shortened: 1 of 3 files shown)


src/FileFormats/SDPA/SDPA.jl
```diff
@@ -8,7 +8,7 @@ import MathOptInterface
 const MOI = MathOptInterface
 
 MOI.Utilities.@model(Model,
-    (),
+    (MOI.Integer,),
```
blegat, line 11: You don't have to add that, SingleVatiable constraints are supported by default

src/FileFormats/SDPA/SDPA.jl
```diff
@@ -251,14 +268,28 @@ function Base.read!(io::IO, model::Model{T}) where T
...
+        end
         if !num_variables_read
             if isempty(line)
                 continue
             end
             num_variables_read = true
             # According to http://plato.asu.edu/ftp/sdpa_format.txt,
             # additional text after the number of variables should be ignored.
-            MOI.add_variables(model, parse(Int, split(line)[1]))
+            scalar_vars = MOI.add_variables(model, parse(Int, split(line)[1]))
```
blegat, line 292: Can you modify `MOI.VariableIndex(matrix)`  below into `scalars_vars[matrix]`?

kibaekkim: Both suggestions were handled and passed. Thanks!

PR #1082, opened 2020-05-04
add Hessian-vector product callback for HS071

frapac:
This PR implements a callback to compute Hessian vector product for the test HS071. 

It currrently runs fine with Knitro. I do not add this test to the [default NLP tests](https://github.com/JuliaOpt/MathOptInterface.jl/blob/master/src/Test/nlp.jl#L349-L361) as that would break the testing for all solvers not supporting Hessian-vector product (including Ipopt).

Implementing a callback for Jacobian vector-product would be straightforward, but I am not aware of any NLP solver supporting this feature.

### DIFF
--- a/src/Test/nlp.jl
+++ b/src/Test/nlp.jl
@@ -3,8 +3,10 @@ const VI = MOI.VariableIndex
 struct HS071 <: MOI.AbstractNLPEvaluator
     enable_hessian::Bool
     enable_hessian_vector_product::Bool
-    function HS071(enable_hessian::Bool,
-                   enable_hessian_vector_product::Bool=false)
+    function HS071(
+        enable_hessian::Bool,
+        enable_hessian_vector_product::Bool = false,
+    )
         return new(enable_hessian, enable_hessian_vector_product)
     end
 end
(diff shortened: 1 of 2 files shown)


src/Test/nlp.jl
```diff
@@ -2,6 +2,7 @@ const VI = MOI.VariableIndex
 
 struct HS071 <: MOI.AbstractNLPEvaluator
     enable_hessian::Bool
+    enable_hessian_vector_product::Bool
 end
```
odow, line 6: Can we add a constructor like
```Julia
function HS071(enable_hessian::Bool, enable_hessian_vector_product::Bool = false)
    return new(enable_hessian, enable_hessian_vector_product)
end
```
so we don't break the existing tests?

src/Test/nlp.jl
```diff
@@ -172,6 +200,7 @@ end
 
 hs071_test(model, config) = hs071test_template(model, config, HS071(true))
 hs071_no_hessian_test(model, config) = hs071test_template(model, config, HS071(false))
+hs071_hessian_vector_product_test(model, config) = hs071test_template(model, config, HS071(false, true))
```
blegat, line 203: Why not add it to the `@moitestset` ?

frapac: I thought that as Ipopt does not support Hessian vector product, it would be better to let `hs071_hessian_vector_product_test` out of the `@moitestset`.

frapac:
Thanks for merging this PR!

odow:
No worries. I added extra tests and tried with ipopt.

PR #1084, opened 2020-05-05
Fix add_constrained_variable(s) for a CachingOptimizer with an attached optimizer

### LINKED ISSUE #1083: add_constrained_variable(s) can break for CachingOptimizer if an optimizer has been attached
If a `CachingOptimizer` has an attached optimizer, and `add_constrained_variable(opt, set)` is called, it defaults to the fallback
```julia
function add_constrained_variable(model::ModelLike, set::AbstractScalarSet)
    variable = add_variable(model)
    constraint = add_constraint(model, SingleVariable(variable), set)
    return variable, constraint
end
```
in variables.jl. If the attached optimizer only supports the `set` through `add_constrained_variable`, then this errors. Here is a MWE:
```julia
using MathOptInterface
const MOI = MathOptInterface
const MOIU = MOI.Utilities

struct DummySet <: MOI.AbstractScalarSet end

MOI.supports_add_constrained_variable(::MOIU.MockOptimizer, ::Type{DummySet}) = true
function MOI.add_constrained_variable(::MOIU.MockOptimizer, ::DummySet)
    # Dummy support
    return MOI.VariableIndex(1), MOI.ConstraintIndex{MOI.SingleVariable, DummySet}(1)
end

cache = MOIU.UniversalFallback(MOIU.Model{Float64}())
m = MOIU.CachingOptimizer(cache, MOIU.MANUAL)
s = MOIU.MockOptimizer(MOIU.Model{Float64}(), supports_names=false)
MOIU.reset_optimizer(m, s)
# Catched by universal fallback
MOI.add_constrained_variable(m, DummySet())
# Properly calls add_constrained_variable during copy
MOIU.attach_optimizer(m)
# Adding constrained var after optimizer has been attached
# errors even though optimizer supports the dummy set
if MOI.supports_add_constrained_variable(m.optimizer, DummySet)
    MOI.add_constrained_variable(m, DummySet())
end
```
```julia
ERROR: MathOptInterface.UnsupportedConstraint{MathOptInterface.SingleVariable,DummySet}: `MathOptInterface.SingleVariable`-in-`DummySet` constraint is not supported by the model.
Stacktrace:
 [1] add_constraint(::MathOptInterface.Utilities.Model{Float64}, ::MathOptInterface.SingleVariable, ::DummySet) at /home/mbiel/.julia/dev/MathOptInterface/src/Utilities/model.jl:539
 [2] add_constraint(::MathOptInterface.Utilities.MockOptimizer{MathOptInterface.Utilities.Model{Float64}}, ::MathOptInterface.SingleVariable, ::DummySet) at /home/mbiel/.julia/dev/MathOptInterface/src/Utilities/mockoptimizer.jl:113
 [3] add_constraint(::MathOptInterface.Utilities.CachingOptimizer{MathOptInterface.AbstractOptimizer,MathOptInterface.Utilities.UniversalFallback{MathOptInterface.Utilities.Model{Float64}}}, ::MathOptInterface.SingleVariable, ::DummySet) at /home/mbiel/.julia/dev/MathOptInterface/src/Utilities/cachingoptimizer.jl:262
 [4] add_constrained_variable(::MathOptInterface.Utilities.CachingOptimizer{MathOptInterface.AbstractOptimizer,MathOptInterface.Utilities.UniversalFallback{MathOptInterface.Utilities.Model{Float64}}}, ::DummySet) at /home/mbiel/.julia/dev/MathOptInterface/src/variables.jl:88
```
A more relevant example would be a custom set that requires a variable bridge and must be added through `add_constrained_variable`. The problem holds for vector sets as well.

A suggested fix is to implement `add_constrained_variable(s)` for `CachingOptimizer` in the same way as `add_variable` and `add_constraint`.

martinbiel:
Resolves #1083 

I have not added any tests yet as I am not sure where it would be best to do so. The "CachingOptimizer MANUAL mode" testset includes adding constraints after the optimizer has been attached, so it could be tested for there.

### DIFF
--- a/src/Utilities/cachingoptimizer.jl
+++ b/src/Utilities/cachingoptimizer.jl
@@ -239,6 +239,68 @@ function MOI.add_variables(m::CachingOptimizer, n)
     return vindices
 end
 
+function MOI.add_constrained_variable(m::CachingOptimizer, set::MOI.AbstractScalarSet)
+    if m.state == MOIU.ATTACHED_OPTIMIZER
+        if m.mode == MOIU.AUTOMATIC
+            try
+                vindex_optimizer, cindex_optimizer =
+                    MOI.add_constrained_variable(m.optimizer, set)
+            catch err
+                if err isa MOI.NotAllowedError
+                    reset_optimizer(m)
+                else
+                    rethrow(err)
+                end
+            end
+        else
+            vindex_optimizer, cindex_optimizer =
+                MOI.add_constrained_variable(m.optimizer, set)
+        end
+    end
+    vindex = MOI.add_variable(m.model_cache)
+    cindex = MOI.add_constraint(m.model_cache, MOI.SingleVariable(vindex), set)
+    if m.state == MOIU.ATTACHED_OPTIMIZER
+        m.model_to_optimizer_map[vindex] = vindex_optimizer
+        m.optimizer_to_model_map[vindex_optimizer] = vindex
+        m.model_to_optimizer_map[cindex] = cindex_optimizer
+        m.optimizer_to_model_map[cindex_optimizer] = cindex
+    end
+    return vindex, cindex
+end
+
+function MOI.add_constrained_variables(m::CachingOptimizer, set::MOI.AbstractVectorSet)
+    if m.state == ATTACHED_OPTIMIZER
+        if m.mode == AUTOMATIC
+            try
+                vindices_optimizer, cindices_optimizer =
+                    MOI.add_constrained_variables(m.optimizer, set)
+            catch err
+                if err isa MOI.NotAllowedError
+                    reset_optimizer(m)
+                else
+                    rethrow(err)
+                end
+            end
+        else
+            vindices_optimizer, cindices_optimizer =
+                MOI.add_constrained_variables(m.optimizer, set)
+        end
+    end
+    vindices = MOI.add_variables(m.model_cache, MOI.dimension(set))
+    cindices = MOI.add_constraint(m.model_cache, MOI.VectorOfVariables(vindices), set)
+    if m.state == ATTACHED_OPTIMIZER
+        for (vindex, vindex_optimizer) in zip(vindices, vindices_optimizer)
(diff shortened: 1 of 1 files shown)


src/Utilities/cachingoptimizer.jl
```diff
@@ -239,6 +239,68 @@ function MOI.add_variables(m::CachingOptimizer, n)
...
+        end
+    end
+    vindices = MOI.add_variables(m.model_cache, MOI.dimension(set))
+    cindices = MOI.add_constraint(m.model_cache, MOI.VectorOfVariables(vindices), set)
+    if m.state == ATTACHED_OPTIMIZER
+        for (vindex, vindex_optimizer) in zip(vindices, vindices_optimizer)
+            m.model_to_optimizer_map[vindex] = vindex_optimizer
+            m.optimizer_to_model_map[vindex_optimizer] = vindex
+        end
+        for (cindex, cindex_optimizer) in zip(cindices, cindices_optimizer)
```
blegat, line 296: There is only one constraint index here, just do
```julia
m.model_to_optimizer_map[cindex] = cindex_optimizer
m.optimizer_to_model_map[cindex_optimizer] = cindex
```

martinbiel: Fixed!

src/Utilities/cachingoptimizer.jl
```diff
@@ -239,6 +239,66 @@ function MOI.add_variables(m::CachingOptimizer, n)
...
+        m.optimizer_to_model_map[cindex_optimizer] = cindex
+    end
+    return vindex, cindex
+end
+
+function MOI.add_constrained_variables(m::CachingOptimizer, set::MOI.AbstractVectorSet)
+    if m.state == ATTACHED_OPTIMIZER
+        if m.mode == AUTOMATIC
+            try
+                vindices_optimizer, cindices_optimizer =
```
blegat, line 275: `cindices_optimizer` -> `cindex_optimizer`

martinbiel: Fixed.

PR #1086, opened 2020-05-06
Add support for custom objective functions in UniversalFallback.

### LINKED ISSUE #1085: Feature request: Support custom objective functions in UniversalFallback
I can create a custom `MOI.AbstractFunction` and add support to it through bridges. The custom function can then be included in constraints, before attaching a bridged optimizer, because the `UniversalFallback` supports any constraints. It would be nice if this also worked for the objective function.

Currently, if I add a custom objective function that is not supported by the `MOIU.Model` it ends up in the `modattr` dict of `UniversalFallback`. When JuMP queries the objective function it first gets the `ObjectiveFunctionType` from the `MOIU.Model` so that ultimately a `zero(ScalarAffineFunction{Float64})` will be returned, even if I have added support for my custom function in JuMP and MOI through bridges.

`UniversalFallback` could handle custom objectives that are not supported by `MOIU.Model` in the same way that it currently handles custom constraints.

martinbiel:
Quick prototype that resolves #1085 

Tests are missing.

### DIFF
--- a/src/Utilities/universalfallback.jl
+++ b/src/Utilities/universalfallback.jl
@@ -258,7 +258,7 @@ function MOI.get(uf::UniversalFallback, listattr::MOI.ListOfOptimizerAttributesS
 end
 function MOI.get(uf::UniversalFallback, listattr::MOI.ListOfModelAttributesSet)
     list = MOI.get(uf.model, listattr)
-    if !uf.objective === nothing
+    if !(uf.objective === nothing)
         push!(list, MOI.ObjectiveFunction{typeof(uf.objective)}())
     end
     for attr in keys(uf.modattr)
--- a/src/Utilities/universalfallback.jl
+++ b/src/Utilities/universalfallback.jl
@@ -2,11 +2,11 @@
     UniversalFallback
 
 The `UniversalFallback` can be applied on a [`MathOptInterface.ModelLike`](@ref)
-`model` to create the model `UniversalFallback(model)` supporting *any* objective,
-constraint, and attribute. This allows to have a specialized implementation in
-`model` for performance critical objective/constraints and attributes while still
+`model` to create the model `UniversalFallback(model)` supporting *any*
+constraint and attribute. This allows to have a specialized implementation in
+`model` for performance critical constraints and attributes while still
 supporting other attributes with a small performance penalty. Note that `model`
-is unaware of objective/constraints and attributes stored by `UniversalFallback` so this
+is unaware of constraints and attributes stored by `UniversalFallback` so this
 is not appropriate if `model` is an optimizer (for this reason,
 [`MathOptInterface.optimize!`](@ref) has not been implemented). In that case,
 optimizer bridges should be used instead.
@@ -160,7 +160,7 @@ function MOI.delete(uf::UniversalFallback, vi::VI)
     for d in values(uf.varattr)
         delete!(d, vi)
     end
-    if !(uf.objective === nothing)
+    if uf.objective !== nothing
         uf.objective = remove_variable(uf.objective, vi)
     end
     for (_, constraints) in uf.constraints
@@ -174,7 +174,7 @@ function MOI.delete(uf::UniversalFallback, vis::Vector{VI})
             delete!(d, vi)
         end
     end
-    if !(uf.objective === nothing)
+    if uf.objective !== nothing
         uf.objective = remove_variable(uf.objective, vis)
     end
     for (_, constraints) in uf.constraints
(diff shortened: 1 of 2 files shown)


src/Utilities/universalfallback.jl
```diff
@@ -54,11 +57,12 @@ function Base.show(io::IO, U::UniversalFallback)
 end
 
 function MOI.is_empty(uf::UniversalFallback)
-    return MOI.is_empty(uf.model) && isempty(uf.constraints) &&
+    return MOI.is_empty(uf.model) && isnothing(uf.objective) && isempty(uf.constraints) &&
```
blegat, line 60: Don't use `isnothing` as it is not available in Julia v1.0

src/Utilities/universalfallback.jl
```diff
@@ -2,17 +2,18 @@
     UniversalFallback
 
 The `UniversalFallback` can be applied on a [`MathOptInterface.ModelLike`](@ref)
-`model` to create the model `UniversalFallback(model)` supporting *any*
-constaint and attribute. This allows to have a specialized implementation in
-`model` for performance critical constraints and attributes while still
+`model` to create the model `UniversalFallback(model)` supporting *any* objective,
```
blegat, line 5: Technically, the objective is an attribute. The objective attribute needs a special handling by `UniversalFallback` but it's still an attribute.

src/Utilities/universalfallback.jl
```diff
@@ -156,6 +160,9 @@ function MOI.delete(uf::UniversalFallback, vi::VI)
     for d in values(uf.varattr)
         delete!(d, vi)
     end
+    if !(uf.objective === nothing)
```
blegat, line 163: You can use `!==`

src/Utilities/universalfallback.jl
```diff
@@ -268,6 +281,52 @@ function MOI.get(uf::UniversalFallback, listattr::MOI.ListOfConstraintAttributes
     return list
 end
 
+# Objective
+MOI.supports(uf::UniversalFallback, ::MOI.ObjectiveSense) = true
```
blegat, line 285: This is already covered by
https://github.com/JuliaOpt/MathOptInterface.jl/blob/74d74433c783a980fa118ffbda1fbe9f5f0f1d8c/src/Utilities/universalfallback.jl#L339

src/Utilities/universalfallback.jl
```diff
@@ -268,6 +281,52 @@ function MOI.get(uf::UniversalFallback, listattr::MOI.ListOfConstraintAttributes
...
+end
+function MOI.get(uf::UniversalFallback,
+                 attr::MOI.ObjectiveFunction{F})::F where F
+    if uf.objective === nothing
+        return MOI.get(uf.model, attr)
+    else
+        return uf.objective
+    end
+end
+MOI.supports(uf::UniversalFallback, ::MOI.ObjectiveFunction{F}) where F <: MOI.AbstractScalarFunction = true
```
blegat, line 309: This is already covered by
https://github.com/JuliaOpt/MathOptInterface.jl/blob/74d74433c783a980fa118ffbda1fbe9f5f0f1d8c/src/Utilities/universalfallback.jl#L339

src/Utilities/universalfallback.jl
```diff
@@ -268,6 +281,52 @@ function MOI.get(uf::UniversalFallback, listattr::MOI.ListOfConstraintAttributes
...
+function MOI.set(uf::UniversalFallback,
+                 attr::MOI.ObjectiveFunction,
+                 func::MOI.AbstractScalarFunction)
+    if MOI.supports(uf.model, attr)
+        MOI.set(uf.model, attr, func)
+        uf.objective = nothing
+    else
+        uf.objective = copy(func)
+        uf.model.objectiveset = false
```
blegat, line 318: `uf.model` is not necessarily an `MOIU.AbstractModel`. Here, you can do:
```julia
sense = MOI.get(uf.model, MOI.ObjectiveSense())
MOI.set(uf.model, MOI.ObjectiveSense(), MOI.FEASIBILITY_SENSE) # clear objective function if any was set
MOI.set(uf.model, MOI.ObjectiveSense(), sense)
```

blegat:
Implementation looks good. Just missing tests

martinbiel:
I added some simple tests to `universalfallback.jl`, but more importantly I believe 
```
function MOI.supports(
    ::ModelForUniversalFallback{T},
    ::Type{MOI.ObjectiveFunction{MOI.ScalarAffineFunction{T}}}) where T
    return false
end
```
should give a lot of coverage from the unit tests. I also rebased the branch on master.

PR #1088, opened 2020-05-10
deterministic ordering of constraints -> variable indices

Wikunia:
Fixes https://github.com/JuliaOpt/JuMP.jl/issues/2238
Uses idea by @odow 
Thanks again for pointing me where the randomness happens

### DIFF
--- a/src/Utilities/universalfallback.jl
+++ b/src/Utilities/universalfallback.jl
@@ -13,7 +13,7 @@ optimizer bridges should be used instead.
 """
 mutable struct UniversalFallback{MT} <: MOI.ModelLike
     model::MT
-    constraints::OrderedDict{Tuple{DataType, DataType}, Dict} # See https://github.com/JuliaOpt/JuMP.jl/issues/1152 and https://github.com/JuliaOpt/JuMP.jl/issues/2238
+    constraints::OrderedDict{Tuple{DataType, DataType}, OrderedDict} # See https://github.com/JuliaOpt/JuMP.jl/issues/1152 and https://github.com/JuliaOpt/JuMP.jl/issues/2238
     nextconstraintid::Int64
     con_to_name::Dict{CI, String}
     name_to_con::Union{Dict{String, MOI.ConstraintIndex}, Nothing}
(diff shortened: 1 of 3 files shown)


src/Utilities/universalfallback.jl
```diff
@@ -13,7 +13,7 @@ optimizer bridges should be used instead.
 """
 mutable struct UniversalFallback{MT} <: MOI.ModelLike
     model::MT
-    constraints::Dict{Tuple{DataType, DataType}, Dict} # See https://github.com/JuliaOpt/JuMP.jl/issues/1152
+    constraints::OrderedDict{Tuple{DataType, DataType}, Dict} # See https://github.com/JuliaOpt/JuMP.jl/issues/1152 and https://github.com/JuliaOpt/JuMP.jl/issues/2238
```
odow, line 16: Let's have a think about this type. Do we gain anything from `Tuple{DataType, DataType}` and `Dict`? Why not just `OrderedDict{Any, Any}`? I can't imagine the compiler can specialize.

blegat: At least it helps in readability. What would be the advantage of using `Any`?

Wikunia:
It seems like this is not good enough. I am not familiar with the structure here. 
Does the key give the constraint type and the value is another dictionary storing the actual constraints? The value type of the dict must be `OrderedDict` as well as far as I can see.

blegat:
Yes, exactly, it should be `OrderedDict` as well. Could you add a test ?

Wikunia:
Sure can you guide me a bit of how and where to test it?
Adding some constraints and check whether the order preserves the insertion ordering?

blegat:
You can add a testset in `test/Utilities/universalfallback.jl` like:
```julia
@testset "Duplicate names" begin
    F = MOI.ScalarAffineFunction{Float64}
    _affine(vi) = convert(F, MOI.SingleVariable(vi))
    set1 = MOI.EqualTo(1.0)
    set2 = MOI.GreaterThan(1.0)
    for sets in [[set1, set2], [set2, set1]]
        model = ModelForUniversalFallback{Float64}()
        uf = MOIU.UniversalFallback(model)
        x = MOI.add_variable(uf)
        y = MOI.add_variable(uf)
        cx1 = MOI.add_constraint(uf, _affine(x), set1)
        cx2 = MOI.add_constraint(uf, _affine(x), set2)
        cy1 = MOI.add_constraint(uf, _affine(y), set1)
        cy2 = MOI.add_constraint(uf, _affine(y), set2)
        @test MOI.get(uf, MOI.ListOfConstraints()) == [(F, typeof(set1)), (F, typeof(set2))]
        @test MOI.get(uf, MOI.ListOfConstraintIndices{F, typeof(set1)}()) = ...
        ... typeof(set2) ...
    end
end
```

PR #1093, opened 2020-05-29
Bugfix for set ConstraintFunction in ScalarizeBridge.

martinbiel:
I found a small bug in ScalarizeBridge. This should fix it.

### DIFF
--- a/src/Bridges/Constraint/scalarize.jl
+++ b/src/Bridges/Constraint/scalarize.jl
@@ -122,7 +122,7 @@ function MOI.modify(model::MOI.ModelLike, bridge::ScalarizeBridge,
     nothing
 end
 function MOI.set(model::MOI.ModelLike, ::MOI.ConstraintFunction,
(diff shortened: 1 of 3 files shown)


martinbiel:
The `bridge.constants[i] != old_constants[i]` case is probably not encountered in tests as this would have triggered a runtime error.

blegat:
Thanks! Could you add a test ?
It could be as simple as a custom test in https://github.com/JuliaOpt/MathOptInterface.jl/blob/3f872f9edb50b1a97cb65011430a0160fd56111e/test/Bridges/Constraint/scalarize.jl or even a new test in https://github.com/JuliaOpt/MathOptInterface.jl/blob/3f872f9edb50b1a97cb65011430a0160fd56111e/src/Test/UnitTests/modifications.jl that is called from https://github.com/JuliaOpt/MathOptInterface.jl/blob/3f872f9edb50b1a97cb65011430a0160fd56111e/test/Bridges/Constraint/scalarize.jl.

PR #1096, opened 2020-05-31
Small simplification for adding SingleVariable constraints

blegat:


src/Utilities/model.jl
```diff
@@ -624,30 +624,28 @@ function MOI.supports_constraint(
...
     return true
 end
+
 function MOI.add_constraint(
     model::AbstractModel{T},
     f::MOI.SingleVariable,
-    s::MOI.AbstractScalarSet,
+    s::SUPPORTED_VARIABLE_SCALAR_SETS{T},
 ) where {T}
-    if MOI.supports_constraint(model, MOI.SingleVariable, typeof(s))
```
odow, line 632: Don't we still need the `supports_constraint` check here? Or should it be moved here:
https://github.com/jump-dev/MathOptInterface.jl/blob/6c5fd6bac5d8bfd6c15f723e4c87bbbfdc9a21b2/src/Utilities/parser.jl#L286-L292

blegat: I don't understand what you mean. This PR is removing this check. What is the relation with the parser ?

odow: Turns out there was also a bug in the parser, which added constraints without checking `supports_constraint`, but tests passed because they relied on this check to work.

blegat: Ah ok, it's a bit surprising as it should fall back to the default `add_constraint` which throws a UnsupportedConstraint:
```julia
julia> using MathOptInterface
[ Info: Precompiling MathOptInterface [b8f27783-ece8-5eb3-8dc8-9495eed66fee]

julia> const MOI = MathOptInterface
MathOptInterface

julia> model = MOI.Utilities.Model{Float64}()
MOIU.Model{Float64}

julia> fx = MOI.SingleVariable(x)
MathOptInterface.SingleVariable(MathOptInterface.VariableIndex(1))

julia> struct A <: MOI.AbstractScalarSet end

julia> MOI.add_constraint(model, fx, A())
ERROR: MathOptInterface.UnsupportedConstraint{MathOptInterface.SingleVariable,A}: `MathOptInterface.SingleVariable`-in-`A` constraint is not supported by the model.
Stacktrace:
 [1] add_constraint(::MathOptInterface.Utilities.Model{Float64}, ::MathOptInterface.SingleVariable, ::A) at /home/blegat/.julia/dev/MathOptInterface/src/Utilities/model.jl:648
 [2] top-level scope at REPL[7]:1
```

blegat: Tried to see the failing tests but Travis says "There was an error while trying to fetch the log."
https://travis-ci.com/github/jump-dev/MathOptInterface.jl/jobs/390570606

odow: We don't use Travis anymore. Here's the log: https://github.com/jump-dev/MathOptInterface.jl/runs/1554087843#step:6:358

odow:
@blegat good to merge?

PR #1101, opened 2020-06-02
Supports add constrained variable

ilancoulon:
I am trying to tackle #987 with this PR.

As it's my first one, I am still trying to comprehend MOI's overall design and I am aware this PR still has a lot of flaws. I hope you guys will help me correct them.

Here is what issues I still have and where you guys could help me:
- Thanks to @blegat , I created a fallback for the `get(model, attr)` method in order to be able to give a default value to the newly created attribute `VariableBridgingCost`. Not creating an intermediary `get_fallback` method resulted in method ambiguities since model-specific methods were already defined, such as `MOI.get(mock::MockOptimizer, attr::MOI.AbstractModelAttribute)`. The issue here is that I still had to define a [specific method for the CachingOptimizer](https://github.com/ilancoulon/MathOptInterface.jl/blob/supports_add_constrained_variable/src/Utilities/cachingoptimizer.jl#L645), otherwise an error was thrown during the tests when no optimizer was attached (thrown [there](https://github.com/ilancoulon/MathOptInterface.jl/blob/supports_add_constrained_variable/src/Utilities/cachingoptimizer.jl#L639)). I must find a way to get rid of those definitions.
- I'm not sure I really understood how the "bridging graph" works, and I'd like your opinion on what I edited in `lazy_bridge_optimizer.jl`.
- In `copy.jl`, since the `NLPBlock` assumes no change in the order of the variables, I didn't use any bridging cost for it, am I right?
- In `copy.jl`, the copying still takes the vector variables first. I did that to change as little as possible. Should we keep that?
- I added a `ConstraintBridgingCost` too, as @blegat first stated in #987, but I'm not sure how I should use it in the `default_copy_to` function.
- I'm not confident about where I put everything, such as the definition of the new attributes in `attributes.jl`, the default values in `model.jl`. 
- Finally, I don't know where I should write the tests for this new feature, I'd be grateful if you could point me in the right direction on that.

That is a lot of questions, but I hope this PR will eventually be useful.

PS: For now, you can't really use this feature with JuMP, and we will eventually need to add this to JuMP.

### DIFF
--- a/src/Utilities/copy.jl
+++ b/src/Utilities/copy.jl
@@ -320,11 +320,13 @@ function default_copy_to(dest::MOI.ModelLike, src::MOI.ModelLike, copy_names::Bo
     else
         # Order the copying of the variables by 1) their variable bridging cost 2) by starting with the vectors, as what was done before
         # See issue #987
-        sorted_by_cost = sortperm(constraint_types; by=((F, S),) -> (MOI.get(dest, MOI.VariableBridgingCost{S}()), F == MOI.SingleVariable))
+        single_or_vector_variables_types = [(F, S) for (F, S) in constraint_types
+                                            if F == MOI.SingleVariable || F == MOI.VectorOfVariables]
+        sorted_by_cost = sortperm(single_or_vector_variables_types; by=((F, S),) -> (MOI.get(dest, MOI.VariableBridgingCost{S}()) - MOI.get(dest, MOI.ConstraintBridgingCost{F, S}()), F == MOI.SingleVariable))
         vector_of_variables_not_added = Vector{Array{MOI.ConstraintIndex{MOI.VectorOfVariables, <:MOI.AbstractVectorSet}}}()
         single_variable_not_added = Vector{Array{MOI.ConstraintIndex{MOI.SingleVariable, <:MOI.AbstractScalarSet}}}()
         for i in sorted_by_cost
-            F, S = constraint_types[i]
+            F, S = single_or_vector_variables_types[i]
             if F == MOI.VectorOfVariables
                 push!(vector_of_variables_not_added, copy_vector_of_variables(dest, src, idxmap, S))
             elseif F == MOI.SingleVariable
--- a/src/Utilities/copy.jl
+++ b/src/Utilities/copy.jl
@@ -302,13 +302,15 @@ function default_copy_to(dest::MOI.ModelLike, src::MOI.ModelLike, copy_names::Bo
 
     vis_src = MOI.get(src, MOI.ListOfVariableIndices())
     constraint_types = MOI.get(src, MOI.ListOfConstraints())
-    single_variable_types = [S for (F, S) in constraint_types
-                             if F == MOI.SingleVariable]
-    vector_of_variables_types = [S for (F, S) in constraint_types
-                                 if F == MOI.VectorOfVariables]
+    single_variable_types = Type{<:MOI.AbstractScalarSet}[]
+    vector_of_variables_types = Type{<:MOI.AbstractVectorSet}[]
 
     # The `NLPBlock` assumes that the order of variables does not change (#849)
     if MOI.NLPBlock() in MOI.get(src, MOI.ListOfModelAttributesSet())
+        single_variable_types = [S for (F, S) in constraint_types
+                                 if F == MOI.SingleVariable]
+        vector_of_variables_types = [S for (F, S) in constraint_types
+                                     if F == MOI.VectorOfVariables]
         vector_of_variables_not_added = [
             MOI.get(src, MOI.ListOfConstraintIndices{MOI.VectorOfVariables, S}())
             for S in vector_of_variables_types
@@ -329,8 +331,10 @@ function default_copy_to(dest::MOI.ModelLike, src::MOI.ModelLike, copy_names::Bo
             F, S = single_or_vector_variables_types[i]
             if F == MOI.VectorOfVariables
                 push!(vector_of_variables_not_added, copy_vector_of_variables(dest, src, idxmap, S))
+                push!(vector_of_variables_types, S)
             elseif F == MOI.SingleVariable
                 push!(single_variable_not_added, copy_single_variable(dest, src, idxmap, S))
+                push!(single_variable_types, S)
             end
         end
     end
--- a/src/Utilities/copy.jl
+++ b/src/Utilities/copy.jl
@@ -318,14 +318,17 @@ function default_copy_to(dest::MOI.ModelLike, src::MOI.ModelLike, copy_names::Bo
             for S in single_variable_types
         ]
     else
-        vector_of_variables_not_added = [
-            copy_vector_of_variables(dest, src, idxmap, S)
-            for S in vector_of_variables_types
-        ]
-        single_variable_not_added = [
-            copy_single_variable(dest, src, idxmap, S)
-            for S in single_variable_types
-        ]
+        sorted_permutation = sortperm(constraint_types; by=((F, S),) -> (MOI.get(dest, MOI.VariableBridgingCost{S}()), F == MOI.SingleVariable))
+        vector_of_variables_not_added = Vector{Array{MOI.ConstraintIndex{MOI.VectorOfVariables, <:MOI.AbstractVectorSet}}}()
+        single_variable_not_added = Vector{Array{MOI.ConstraintIndex{MOI.SingleVariable, <:MOI.AbstractScalarSet}}}()
+        for i in sorted_permutation
+            F, S = constraint_types[i]
+            if F == MOI.VectorOfVariables
+                push!(vector_of_variables_not_added, copy_vector_of_variables(dest, src, idxmap, S))
+            elseif F == MOI.SingleVariable
+                push!(single_variable_not_added, copy_single_variable(dest, src, idxmap, S))
+            end
+        end
     end
 
     copy_free_variables(dest, idxmap, vis_src, MOI.add_variables)
--- a/src/Utilities/copy.jl
+++ b/src/Utilities/copy.jl
@@ -318,10 +318,12 @@ function default_copy_to(dest::MOI.ModelLike, src::MOI.ModelLike, copy_names::Bo
             for S in single_variable_types
         ]
     else
-        sorted_permutation = sortperm(constraint_types; by=((F, S),) -> (MOI.get(dest, MOI.VariableBridgingCost{S}()), F == MOI.SingleVariable))
+        # Order the copying of the variables by 1) their variable bridging cost 2) by starting with the vectors, as what was done before
+        # See issue #987
+        sorted_by_cost = sortperm(constraint_types; by=((F, S),) -> (MOI.get(dest, MOI.VariableBridgingCost{S}()), F == MOI.SingleVariable))
         vector_of_variables_not_added = Vector{Array{MOI.ConstraintIndex{MOI.VectorOfVariables, <:MOI.AbstractVectorSet}}}()
         single_variable_not_added = Vector{Array{MOI.ConstraintIndex{MOI.SingleVariable, <:MOI.AbstractScalarSet}}}()
-        for i in sorted_permutation
+        for i in sorted_by_cost
             F, S = constraint_types[i]
             if F == MOI.VectorOfVariables
                 push!(vector_of_variables_not_added, copy_vector_of_variables(dest, src, idxmap, S))
--- a/src/Utilities/model.jl
+++ b/src/Utilities/model.jl
@@ -540,9 +540,7 @@ function MOI.add_constraint(model::AbstractModel{T}, f::MOI.SingleVariable,
     end
 end
 function MOI.add_constraint(model::AbstractModel, f::F, s::S) where {F<:MOI.AbstractFunction, S<:MOI.AbstractSet}
-    if (MOI.supports_constraint(model, F, S)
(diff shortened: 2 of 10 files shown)


src/Utilities/copy.jl
```diff
@@ -318,14 +318,19 @@ function default_copy_to(dest::MOI.ModelLike, src::MOI.ModelLike, copy_names::Bo
...
-            copy_vector_of_variables(dest, src, idxmap, S)
-            for S in vector_of_variables_types
-        ]
-        single_variable_not_added = [
-            copy_single_variable(dest, src, idxmap, S)
-            for S in single_variable_types
-        ]
+        # Order the copying of the variables by 1) their variable bridging cost 2) by starting with the vectors, as what was done before
+        # See issue #987
+        sorted_by_cost = sortperm(constraint_types; by=((F, S),) -> (MOI.get(dest, MOI.VariableBridgingCost{S}()), F == MOI.SingleVariable))
```
blegat, line 323: In `constraint_types` you also have the constraints for which `F` is not `VectorOfVariables` or `SingleVariable`, you may want to filter that first.

blegat: The cost should be `MOI.get(dest, MOI.VariableBridgingCost{S}()) - MOI.get(dest, MOI.ConstraintBridgingCost{F, S}())`.
If bridging a variable is cheap but bridging it as a constraint is even cheaper then it should not have priority.

src/attributes.jl
```diff
@@ -1348,6 +1349,14 @@ end
 DualStatus() = DualStatus(1)
 _result_index_field(attr::DualStatus) = attr.N
 
+
+# Cost of bridging constrained variable in S
+struct VariableBridgingCost{S <: AbstractSet} <: AbstractOptimizerAttribute
```
blegat, line 1354: If you use `AbstractModelAttribute` here you won't have the issue with `CachingOptimizer`.

src/variables.jl
```diff
@@ -63,7 +63,7 @@ lower bound on the same variable. Moreover, if the user adds a
 bridged into a supported constraint.
 """
 function supports_add_constrained_variable(model::ModelLike,
-                                       S::Type{<:AbstractScalarSet})
+                                       S::Type{<:AbstractSet})
```
blegat, line 66: I would not change this. There is `supports_add_constrained_variables` below for vector sets.

ilancoulon: I agree with that, I forgot I changed it. Then, in [`node(b::LazyBridgeOptimizer, S::Type{<:MOI.AbstractSet})`](https://github.com/ilancoulon/MathOptInterface.jl/blob/1dec6d2092ad5b8408a77a0efcd1185dcd3cacd5/src/Bridges/lazy_bridge_optimizer.jl#L116), should I call `supports_add_constrained_variables` if S is an `AbstractvectorSet`, or should I write something like this in `variables.jl`?:
`supports_add_constrained_variable(model::ModelLike, S::Type{<:AbstractVectorSet}) = supports_add_constrained_variables(model, S)`

The real question could be: could we use only one fonction with two methods instead of two functions?

ilancoulon: I did the latter in the last commits, just for the tests to pass.

src/attributes.jl
ilancoulon: Thank you for that advice! I changed it.

src/Utilities/copy.jl
ilancoulon: I think I get it now. I changed the cost and filtered `constraint_types`.

src/Utilities/model.jl
```diff
@@ -714,6 +714,12 @@ function load_constraint(::AbstractModel, ::CI, ::MOI.AbstractFunction,
                          ::MOI.AbstractSet)
 end
 
+# Bridging cost fallback values
+MOI.get_fallback(model::MOI.ModelLike, ::MOI.VariableBridgingCost{S}) where {S<:MOI.AbstractScalarSet} = MOI.supports_add_constrained_variable(model, S) ? 0.0 : Inf
+MOI.get_fallback(model::MOI.ModelLike, ::MOI.VariableBridgingCost{S}) where {S<:MOI.AbstractVectorSet} = MOI.supports_add_constrained_variables(model, S) ? 0.0 : Inf
+MOI.get_fallback(model::MOI.ModelLike, ::MOI.ConstraintBridgingCost{F, S}) where {F<:MOI.AbstractFunction, S<:MOI.AbstractSet} = MOI.supports_constraint(model, F, S) ? 0.0 : Inf
```
blegat, line 720: Move this to `src/attributes.jl` after the definition of the attributes

src/variables.jl
```diff
@@ -67,6 +67,7 @@ function supports_add_constrained_variable(model::ModelLike,
     return supports_add_constrained_variables(model, Reals) &&
         supports_constraint(model, SingleVariable, S)
 end
+supports_add_constrained_variable(model::ModelLike, S::Type{<:AbstractVectorSet}) = supports_add_constrained_variables(model, S)
```
blegat, line 70: You can remove this now

src/variables.jl
blegat: You should do the former :)

src/variables.jl
ilancoulon: Done!

src/variables.jl
ilancoulon: Ahah I changed it and filtered in `node()`.

blegat:
The implementation looks good, testing now :)
I suggest you add tests in `test/Utilities/copy.jl`. You can fill a `src = MOIU.Model{Float64}()` with some constraints/variables, then you create some custom model that you create a `dest` model that you define with `MOIU.@model` and customize with `supports_constrained_variable(s)` and then you call `index_map MOI.copy_to(dest, src)`. You can then check in `index_map` the indices of the constraints in `dest` and check that their value is what is expected. The indices of a `MOIU.AbstractModel` are unique even across different constraint type so that allows to see the order of creation easily.
You can try different examples of custom model types for `dest` with different combinations of supported/unsupported variables/constrained and check that the order is modifed accordingly.

ilancoulon:
I wrote a test for supports_add_constrained_variable**s** since the indices of AbstractModel are not unique with constraints on AbstractScalarSet. Should I write my own model to be able to test that or is this test enough?

test/Utilities/copy.jl
```diff
@@ -157,3 +157,32 @@ end
...
+@testset "Create variables using supports_add_constrained_variable(s) (#987)" begin
+    src = MOIU.Model{Float64}()
+    a, c1 = MOI.add_constrained_variables(src, MOI.Nonpositives(3))
+    c2 = MOI.add_constraint(src, a, MOI.Nonnegatives(3))
+    dest = OrderingConstrainedVariablesModel{Float64}()
+
+    index_map = MOI.copy_to(dest, src)
+
+
+    @test index_map[c1].value == 2
```
blegat, line 186: Could you also try with a model that supports Nonpositives variables and not Nonnegatives variables. Otherwise, i the cost is not use, there is still 50 % chance that the test pass if we are lucky.

ilancoulon: I had to edit `model.jl` in order to not have an error thrown.

blegat: Indeed, this is an issue. For `dest`, it's probably better to use a custom `MOI.ModelLike` then. It can be a simple struct with an `MOIU.Model{Float64}` inside which redirects all calls to it except the `supports_...` calls.

ilancoulon: I created a custom `ModelLike` and fixed an issue with the types as well.

test/Utilities/copy.jl
```diff
@@ -159,30 +159,77 @@ end
...
 MOI.supports_constraint(::OrderingConstrainedVariablesModel, ::Type{MOI.VectorOfVariables}, ::Type{MOI.Nonnegatives}) = false
 MOI.supports_add_constrained_variables(::OrderingConstrainedVariablesModel, ::Type{MOI.Nonnegatives}) = true
+
+MOI.supports_constraint(::OrderingConstrainedVariablesModel, ::Type{MOI.VectorOfVariables}, ::Type{MOI.Nonnegatives}) = true
 MOI.supports_add_constrained_variables(::OrderingConstrainedVariablesModel, ::Type{MOI.Nonpositives}) = false
 
+MOI.supports_constraint(::OrderingConstrainedVariablesModel, ::Type{MOI.SingleVariable}, ::Type{MOI.GreaterThan}) = false
+MOI.supports_add_constrained_variable(::OrderingConstrainedVariablesModel, ::Type{MOI.GreaterThan}) = true
+
+MOI.supports_constraint(::OrderingConstrainedVariablesModel, ::Type{MOI.SingleVariable}, ::Type{MOI.LessThan}) = true
```
blegat: `Type{MOI.LessThan}` should be `Type{<:MOI.LessThan}`

ilancoulon: Indeed, this commit didn't work, I just changed that.

test/Utilities/copy.jl
```diff
@@ -157,3 +157,70 @@ end
...
+    # With vectors
+    src = MOIU.Model{Float64}()
+    a, c1 = MOI.add_constrained_variables(src, MOI.Nonpositives(3))
+    c2 = MOI.add_constraint(src, a, MOI.Nonnegatives(3))
+    dest = OrderingConstrainedVariablesModel()
+
+    index_map = MOI.copy_to(dest, src)
+
+    @test typeof(c1) == typeof(dest.constraintIndices[2])
+    @test typeof(c2) == typeof(dest.constraintIndices[1])
```
blegat, line 214: Could you try a model where the order is reversed as suggested in https://github.com/JuliaOpt/MathOptInterface.jl/pull/1101#discussion_r434667027 ? You can save most of the code between the models with an abstract types between the two as only the supports functions change?

ilancoulon: Is the new commit what you meant?

blegat: Yes, exactly !

test/Utilities/copy.jl
```diff
@@ -157,3 +157,97 @@ end
...
+
+    dest = OrderConstrainedVariablesModel()
+    index_map = MOI.copy_to(dest, src)
+    @test typeof(c1) == typeof(dest.constraintIndices[2])
+    @test typeof(c2) == typeof(dest.constraintIndices[1])
+
+    dest = ReverseOrderConstrainedVariablesModel()
+    index_map = MOI.copy_to(dest, src)
+    @test typeof(c1) == typeof(dest.constraintIndices[1])
+    @test typeof(c2) == typeof(dest.constraintIndices[2])
```
blegat, line 252: Now, if you run the test with `dest = full_bridge_optimizer(ReverseOrderConstrainedVariablesModel(), Float64)` or `dest = full_bridge_optimizer(OrderConstrainedVariablesModel(), Float64)`, I guess one of the two will fail as with the bridges they both support both constrained variables.
To fix this, we should implement the attribute for `LazyBridgeOptimizer` and return the bridge cost.

ilancoulon: I'm not sure what you meant here. Could you maybe write the corresponding test and I'd try to make it work?

blegat: I mean something like
```julia
dest = OrderConstrainedVariablesModel()
bridged_dest = MOI.Bridges.full_bridge_optimizer(dest, Float64)
index_map = MOI.copy_to(bridged_dest, src)
@test typeof(c1) == typeof(dest.constraintIndices[2])
@test typeof(c2) == typeof(dest.constraintIndices[1])

dest = ReverseOrderConstrainedVariablesModel()
bridged_dest = MOI.Bridges.full_bridge_optimizer(dest, Float64)
index_map = MOI.copy_to(bridged_dest, src)
@test typeof(c1) == typeof(dest.constraintIndices[1])
@test typeof(c2) == typeof(dest.constraintIndices[2])
```

ilancoulon: By trying the following lines:
``` julia
src = MOIU.Model{Float64}()
a, c1 = MOI.add_constrained_variables(src, MOI.Nonpositives(3))
c2 = MOI.add_constraint(src, a, MOI.Nonnegatives(3))
dest = OrderConstrainedVariablesModel()
bridged_dest = MOI.Bridges.full_bridge_optimizer(dest, Float64)
index_map = MOI.copy_to(bridged_dest, src)
```
I get an error stating `MathOptInterface.UnsupportedConstraint{MathOptInterface.VectorOfVariables,MathOptInterface.Nonpositives}: MathOptInterface.VectorOfVariables-in-MathOptInterface.Nonpositives constraint is not supported by the model.`.
This is because no bridge of type `::Type{MathOptInterface.VectorOfVariables}, ::Type{MathOptInterface.Nonpositives}` exists in the LazyBridgeOptimizer.

I must say I did not really grasp how the bridge types work, and I'm not sure how I should go from here. Maybe that is why the issue was not tagged as a good first issue :sweat_smile:

I tried implemeting the attributes with:
```julia
MOI.get(b::LazyBridgeOptimizer, c::MOI.VariableBridgingCost{S}) where S<:MOI.AbstractScalarSet = MOI.get(b.model, c)
MOI.get(b::LazyBridgeOptimizer, c::MOI.VariableBridgingCost{S}) where S<:MOI.AbstractVectorSet = MOI.get(b.model, c)
function MOI.get(b::LazyBridgeOptimizer, c::MOI.ConstraintBridgingCost{F,S}) where {S <: MOI.AbstractSet, F <: MOI.AbstractFunction}
    MOI.get(b.model, c)
end
```

But this did not work better, I think this shows how I don't actually understand how all of this works.

PR #1118, opened 2020-07-10
[RFC] Add CanonicalConstraintFunction attribute

blegat:
For models that knows that the constraint function is already canonical they can just
```julia
function MOI.get(model::ModelType, ::Union{MOI.ConstraintFunction, MOI.CanonicalConstraintFunction}, ci)
    ...
end
```
For `MOI.Utilities.Model`, the function are not canonicalized when added (should we ?) so we canonicalize them then `CanonicalConstraintFunction` is called to spare the `copy` that is done in `MOI.Utilities.canonical` as it can slows things down dramatically (see https://github.com/jump-dev/SCS.jl/issues/186).
Alternatively, we could say that the functions added to `MOI.Utilities.Model` are automatically canonicalized in addition of being copied by replacing this `copy(f)` by `canonical(f)`:
https://github.com/jump-dev/MathOptInterface.jl/blob/26e0af44a254b545276b528298a390d34c378586/src/Utilities/model.jl#L550

### DIFF
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -1103,6 +1103,23 @@ end
 ConstraintBasisStatus() = ConstraintBasisStatus(1)
 
 """
+    CanonicalConstraintFunction()
+
+A constraint attribute for a canonical representation of the `AbstractFunction` object used to define the constraint.
+It is guaranteed to be equivalent but not necessarily identical to the function provided by the user.
+
+By default, `MOI.get(model, MOI.CanonicalConstraintFunction(), ci)` fallbacks to
+`MOI.Utilities.canonical(MOI.get(model, MOI.ConstraintFunction(), ci))`.
+However, if `model` knows that the constraint function is canonical then it can
+implement a specialized method that directly return the function without calling
+[`Utilities.canonical`](@ref). Therefore, the value returned **cannot** be
+assumed to be a copy of the function stored in `model`.
+Moreover, [`Utilities.Model`](@ref) calls [`Utilities.canonicalize!`](@ref)
+to the function stored internally and returns this function without copy.
+"""
+struct CanonicalConstraintFunction <: AbstractConstraintAttribute end
+
+"""
     ConstraintFunction()
 
 A constraint attribute for the `AbstractFunction` object used to define the constraint.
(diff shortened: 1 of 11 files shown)


src/attributes.jl
```diff
@@ -1102,6 +1102,23 @@ struct ConstraintBasisStatus <: AbstractConstraintAttribute
...
+`MOI.Utilities.canonical(MOI.get(model, MOI.ConstraintFunction(), ci))`.
+However, if `model` knows that the constraint function is canonical then it can
+implement a specialized method that directly return the function without calling
+[`Utilities.canonical`](@ref). Therefore, the value returned **cannot** be
+assumed to be a copy of the function stored in `model`.
+Moreover, [`Utilities.Model`](@ref) calls [`Utilities.canonicalize!`](@ref)
+to the function stored internally and returns this function without copy.
+"""
+struct CanonicalConstraintFunction <: AbstractConstraintAttribute end
+
```
joaquimg, line 1121: Just add default fallback

odow: We talked about instead MOIU models keeping the constraints canonicalized, and solvers checking `is_canonical` instead of just calling `canonical`.

joaquimg:
Many times we dont need cannonicalized functions.
I think that forcing that all the time is a overkill.
Better to let one use is_cannonical if needed.

blegat:
Any objection to merge ?

src/attributes.jl
```diff
@@ -1102,6 +1102,30 @@ struct ConstraintBasisStatus <: AbstractConstraintAttribute
...
+[`Utilities.canonical`](@ref). Therefore, the value returned **cannot** be
+assumed to be a copy of the function stored in `model`.
+Moreover, [`Utilities.Model`](@ref) checks with [`Utilities.is_canonical`](@ref)
+whether the function stored internally is already canonical and if it's the case,
+then it returns the function stored internally instead of a copy.
+"""
+struct CanonicalConstraintFunction <: AbstractConstraintAttribute end
+
+function get_fallback(model::ModelLike, ::CanonicalConstraintFunction, ci::ConstraintIndex)
+    return Utilities.canonical(get(model, ConstraintFunction(), ci))
```
odow, line 1126: Shouldn't the `is_canonical` check go in this fallback? It would save the checks in `Model` and `UniversalFallback`.

blegat: Done :)

PR #1124, opened 2020-07-11
Fixed unsupported objectives in FileFormats

### LINKED ISSUE #1116: Model Writing Error Messages
Given this model,
```
m = Model()
@variable(m, x)
@objective(m, Min, x^2)
```

The command, `write_to_file(m, "model.lp")` yields,
```
ERROR: MethodError: no method matching write_function(::IOStream, ::MathOptInterface.FileFormats.LP.Model{Float64}, ::MathOptInterface.ScalarQuadraticFunction{Float64}, ::Dict{MathOptInterface.VariableIndex,String})
```

The command, `write_to_file(m, "model.mps")` yields,
```
ERROR: MethodError: no method matching extract_terms(::MathOptInterface.FileFormats.MPS.Model{Float64}, ::Dict{String,Array{Tuple{String,Float64},1}}, ::String, ::MathOptInterface.ScalarQuadraticFunction{Float64}, ::Set{String})
```

Could these messages be improved?  Maybe something like, format X does not support constraint type Y?

egbuck:
Closes #1116

### DIFF
--- a/src/Utilities/model.jl
+++ b/src/Utilities/model.jl
@@ -416,7 +416,10 @@ function MOI.supports(
         MOI.ScalarQuadraticFunction{T}}}) where T
     return true
 end
-function MOI.set(model::AbstractModel, ::MOI.ObjectiveFunction, f::MOI.AbstractFunction)
+function MOI.set(model::AbstractModel, attr::MOI.ObjectiveFunction, f::MOI.AbstractFunction)
+    if !MOI.supports(model, attr)
+        throw(MOI.UnsupportedAttribute(attr))
+    end
     model.objectiveset = true
     # f needs to be copied, see #2
     model.objective = copy(f)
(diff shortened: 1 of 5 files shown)


src/Utilities/model.jl
```diff
@@ -416,7 +416,10 @@ function MOI.supports(
         MOI.ScalarQuadraticFunction{T}}}) where T
     return true
 end
-function MOI.set(model::AbstractModel, ::MOI.ObjectiveFunction, f::MOI.AbstractFunction)
+function MOI.set(model::AbstractModel, attr::MOI.ObjectiveFunction, f::MOI.AbstractFunction)
+    if !MOI.supports(model, attr)
+        throw(MOI.UnsupportedAttribute(attr))
+    end
```
odow, line 422: @blegat how breaking is this? It turns out, even if you said you didn't support it, it would still accept it.

PR #1126, opened 2020-07-12
Implement DoubleDicts

joaquimg:
Improve performance of IndexMap for constraints.

Open for name suggestions.


src/Utilities/DoubleDict.jl
```diff
@@ -0,0 +1,314 @@
...
+const MOI = MathOptInterface
+const CI{F,S} = MOI.ConstraintIndex{F,S}
+
+"""
+    DoubleDict
+
+Special dictionary to map ConstraintIndex to ConstraintIndex with minimal
+performance loss from the type instability due different constraint types.
+"""
+mutable struct DoubleDict{D, D2}
```
blegat, line 14: It seems weird to do `D` here and `D{Int, Int}` in the field, I would expect something like
`D<:AbstractDict{Int, Int}, D2<:AbstractDict{Tuple{DataType, DataType}, D}` and a field `dict::D2`

src/Utilities/DoubleDict.jl
```diff
@@ -0,0 +1,314 @@
...
+        " instead.")
+end
+# function Base.sizehint!(d::DoubleDict, ::Type{F}, ::Type{S}, n) where {F,S}
+#     inner = lazy_get(d, F, S)::Dict{Int,Int}
+#     sizehint!(inner, n)
+# end
+
+function Base.length(d::DoubleDict)
+    len = 0
+    for inner in d.dict
```
blegat, line 31: I think it should be `values(d.dict)`

src/Utilities/DoubleDict.jl
```diff
@@ -0,0 +1,314 @@
...
+const CI{F,S} = MOI.ConstraintIndex{F,S}
+
+"""
+    DoubleDict
+
+Special dictionary to map ConstraintIndex to ConstraintIndex with minimal
+performance loss from the type instability due different constraint types.
+"""
+mutable struct DoubleDict{D, D2}
+    dict::D2{Tuple{DataType, DataType}, D{Int, Int}}
```
blegat, line 15: We should maybe have `D <: AbstractDict{CI{F, S}, CI{F, S}}`. `getindex` would not be type-stable for `D2` but that would allow `D` to be used in a standalone way. When you are treating `F`-in-`S` constraints, you get the corresponding inner dict and then use this one in a standalone way to have a type-stable operations. `D` would even be a `DenseDict`.

src/Utilities/DoubleDict.jl
joaquimg: yup, that was messy, thanks for the cleaup!

src/Utilities/DoubleDict.jl
joaquimg: good catch

src/Utilities/DoubleDict.jl
joaquimg: Thats similiar to what I tried to do with the `WithType`

src/Utilities/DoubleDicts.jl
```diff
@@ -0,0 +1,381 @@
...
+which is a concrete type.
+
+When optimal performance or type stability is required its possible to obtain a
+fully type stable dictionary with values of type
+`CI{MOI.SingleVariable, MOI.Integers}` and keys of type
+`CI{MOI.SingleVariable, MOI.Integers}` from the dictionary `dict`, for instance:
+
+    ``inner = dict[MOI.SingleVariable, MOI.Integers]``
+"""
+struct IndexDoubleDict{V<:Int64, D<:AbstractDict{Int64, Int64}, D2<:AbstractDict{Tuple{DataType, DataType}, D}} <: AbstractDoubleDict{V, D, D2}
```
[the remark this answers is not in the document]
joaquimg: good idea! will double check, but makes total sense

src/Utilities/DoubleDicts.jl
```diff
@@ -0,0 +1,381 @@
+module DoubleDicts
+
+import MathOptInterface
+
+const MOI = MathOptInterface
+const CI{F,S} = MOI.ConstraintIndex{F,S}
+
+abstract type AbstractDoubleDict{V, D<:AbstractDict{Int64, V}, D2<:AbstractDict{Tuple{DataType, DataType}, D}} end
```
blegat, line 8: Isn't it an `AbstractDict` ? I would maybe do `AbstractDoubleDict{K, V, ...}` <: `AbstractDict{K, V}`.
For `D`, I would just do `D <: AbstractDict{Int64}`. Indeed, for instance, you might want `DoubleDict` to contain `F`. In this case you have `D = AbstractDict{Int64, F}` so the type of the dict is different for each `(F, S)`. If you impose it to be `Dict{Int64, MO.AbstractFunction}` then this will convert the inner type-stable dicts to type unstable dicts `Dict{Int64, MO.AbstractFunction}`. This is the case for UniversalFallback for instance which stores functions, sets pairs

joaquimg: I don't think it can be an `AbstractDict`, if we add the key type `K` to the parameters it will always be not concrete because the are `CI{F,S}` for multiple `F,S` pairs. By not typing the key type at all we have a special structure that can dispatch properly and well typed for those "unstable" keys.

joaquimg:
@blegat now there is a FunctionSetDoubleDict that can be used in universallfallback.
What do you think of the current state?

PR #1129, opened 2020-07-26
Improve indexing performance

joaquimg:
Improves:
- dense dict type stability 
- indexmap copying and reversing in attach optimizer

cc @blegat

### DIFF
--- a/src/Utilities/cachingoptimizer.jl
+++ b/src/Utilities/cachingoptimizer.jl
@@ -145,25 +145,32 @@ errors can be thrown.
 """
 function attach_optimizer(model::CachingOptimizer)
     @assert model.state == EMPTY_OPTIMIZER
-    # We do not need to copy names because name-related operations are handled by `m.model_cache`
+    # We do not need to copy names because name-related operations are handled
+    # by `m.model_cache`
     indexmap = MOI.copy_to(model.optimizer, model.model_cache, copy_names=false)
     model.state = ATTACHED_OPTIMIZER
-    # MOI does not define the type of index_map, so we might have to copy it
-    # into a concrete container. Also load the reverse map.
-    model.model_to_optimizer_map = convert(IndexMap, indexmap)#IndexMap()
-    model.optimizer_to_model_map = reverse_index_map(indexmap)#IndexMap()
-    nothing
+    # MOI does not define the type of index_map, so we have to convert it
+    # into an actual IndexMap. Also load the reverse IndexMap.
+    model.model_to_optimizer_map = convert(IndexMap, indexmap)
+    model.optimizer_to_model_map = reverse_index_map(indexmap)
+    return nothing
 end
 
 function reverse_index_map(src::IndexMap)
     dest = IndexMap()
     sizehint!(dest.varmap, length(src.varmap))
-    # sizehint!(dest.conmap, length(src.conmap))
-    reverse_dict(dest.varmap, src.varmap)
-    reverse_dict(dest.conmap, src.conmap)
+    _reverse_dict(dest.varmap, src.varmap)
+    _reverse_dict(dest.conmap, src.conmap)
     return dest
 end
-function reverse_dict(dest::AbstractDict, src::AbstractDict)
+
+"""
+    _reverse_dict(dest::AbstractDict, src::AbstractDict)
+
+Reverse dictionary so that values of `src` are key of `dest` and vice-versa.
+`dest` must be empty. Also the values of `src` are assumed to be unique.
+"""
+function _reverse_dict(dest::AbstractDict, src::AbstractDict)
     for (k,v) in src
         dest[v] = k
     end
@@ -182,15 +189,18 @@ function Base.convert(::Type{IndexMap}, d::IndexMap)
     # if there was a dense dict inside
     # the solution would be to allow automatically swtiching
     # a ClevelDenseDict...
-    return IndexMap(standard_dict(d.varmap), d.conmap)
+    return IndexMap(_standard_dict(d.varmap), d.conmap)
 end
-function standard_dict(d::D)::D where
-    {D<:AbstractDict{MOI.VariableIndex, MOI.VariableIndex}}
+function _standard_dict(
+    d::D
+)::D where {D<:Dict{MOI.VariableIndex, MOI.VariableIndex}}
     return d
 end
-function standard_dict(d::D
-    )::Dict{MOI.VariableIndex, MOI.VariableIndex} where
-    {D<:AbstractDict{MOI.VariableIndex, MOI.VariableIndex}}
+function _standard_dict(
+    d::D
+)::Dict{MOI.VariableIndex, MOI.VariableIndex} where {
+    D<:AbstractDict{MOI.VariableIndex, MOI.VariableIndex}
+}
     ret = Dict{MOI.VariableIndex, MOI.VariableIndex}()
     sizehint!(ret, length(d))
     for (k,v) in d
--- a/src/Utilities/cachingoptimizer.jl
+++ b/src/Utilities/cachingoptimizer.jl
@@ -184,28 +184,6 @@ function _standardize(d::AbstractDict{MOI.Index, MOI.Index})
     return map
 end
 function _standardize(d::IndexMap)
-    # return d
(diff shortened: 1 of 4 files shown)


src/Utilities/cachingoptimizer.jl
```diff
@@ -148,16 +148,57 @@ function attach_optimizer(model::CachingOptimizer)
...
-    model.model_to_optimizer_map = IndexMap()
-    model.optimizer_to_model_map = IndexMap()
-    for k in keys(indexmap)
-        model.model_to_optimizer_map[k] = indexmap[k]
-        model.optimizer_to_model_map[indexmap[k]] = k
+    # MOI does not define the type of index_map, so we might have to copy it
+    # into a concrete container. Also load the reverse map.
+    model.model_to_optimizer_map = convert(IndexMap, indexmap)#IndexMap()
+    model.optimizer_to_model_map = reverse_index_map(indexmap)#IndexMap()
+    nothing
```
odow, line 155: `return` instead of `nothing`

src/Utilities/cachingoptimizer.jl
```diff
@@ -148,16 +148,57 @@ function attach_optimizer(model::CachingOptimizer)
...
-    # concrete container. Also load the reverse map.
-    model.model_to_optimizer_map = IndexMap()
-    model.optimizer_to_model_map = IndexMap()
-    for k in keys(indexmap)
-        model.model_to_optimizer_map[k] = indexmap[k]
-        model.optimizer_to_model_map[indexmap[k]] = k
+    # MOI does not define the type of index_map, so we might have to copy it
+    # into a concrete container. Also load the reverse map.
+    model.model_to_optimizer_map = convert(IndexMap, indexmap)#IndexMap()
+    model.optimizer_to_model_map = reverse_index_map(indexmap)#IndexMap()
```
odow, line 154: Delete the trailing comments and edit the above comment.

src/Utilities/cachingoptimizer.jl
```diff
@@ -148,16 +148,57 @@ function attach_optimizer(model::CachingOptimizer)
...
+    nothing
+end
+
+function reverse_index_map(src::IndexMap)
+    dest = IndexMap()
+    sizehint!(dest.varmap, length(src.varmap))
+    # sizehint!(dest.conmap, length(src.conmap))
```
odow, line 161: Delete.

src/Utilities/cachingoptimizer.jl
```diff
@@ -148,16 +148,57 @@ function attach_optimizer(model::CachingOptimizer)
...
+    reverse_dict(dest.varmap, src.varmap)
+    reverse_dict(dest.conmap, src.conmap)
+    return dest
+end
+function reverse_dict(dest::AbstractDict, src::AbstractDict)
```
odow, line 166: These methods should begin with underscores. And this one could do with a docstring that it assumes the values are unique.

src/Utilities/cachingoptimizer.jl
```diff
@@ -148,16 +148,57 @@ function attach_optimizer(model::CachingOptimizer)
...
+        map[k] = v
+    end
+    return map
+end
+function Base.convert(::Type{IndexMap}, d::IndexMap)
+    # return d
+    # if we return d as is, its not possible to add variables
+    # if there was a dense dict inside
+    # the solution would be to allow automatically swtiching
+    # a ClevelDenseDict...
```
odow, line 184: I don't understand this comment or this function. What does it mean to convert an `IndexMap` to an `IndexMap`?

src/Utilities/cachingoptimizer.jl
```diff
@@ -148,16 +148,57 @@ function attach_optimizer(model::CachingOptimizer)
...
+    return IndexMap(standard_dict(d.varmap), d.conmap)
+end
+function standard_dict(d::D)::D where
+    {D<:AbstractDict{MOI.VariableIndex, MOI.VariableIndex}}
```
odow, line 188:
```Julia
function standard_dict(
    d::D
)::D where {D<:AbstractDict{MOI.VariableIndex, MOI.VariableIndex}}
```

src/Utilities/cachingoptimizer.jl
```diff
@@ -148,16 +148,57 @@ function attach_optimizer(model::CachingOptimizer)
...
+    return d
+end
+function standard_dict(d::D
+    )::Dict{MOI.VariableIndex, MOI.VariableIndex} where
+    {D<:AbstractDict{MOI.VariableIndex, MOI.VariableIndex}}
```
odow, line 193: Ditto

src/Utilities/copy.jl
```diff
@@ -63,9 +63,10 @@ error in case `copy_to` is called with `copy_names` equal to `true`.
 """
 supports_default_copy_to(model::MOI.ModelLike, copy_names::Bool) = false
 
-const DenseVariableDict{V} = DenseDict{MOI.VariableIndex, V, typeof(MOI.index_value), typeof(MOI.VariableIndex)}
+_index_to_variable(i) = MOI.VariableIndex(i)
```
odow, line 66: This could do with a docstring explaining why it is necessary

src/Utilities/cachingoptimizer.jl
joaquimg: the varmap field might be a standard dict or a dense dict. dense dict cannot grow, they are fixed.
One possible smooth solution for that is doing something similar from CleverDict and automating the switching between the inner dict types.

joaquimg:
> Benchmarks?

https://github.com/jump-dev/MOIPaperBenchmark/pull/3

src/Utilities/dense_dict.jl
```diff
@@ -22,15 +22,21 @@ end
...
     if itr === nothing
         return nothing
     else
         el, i = itr
-        return d.inverse_hash(el) => d.map[el], i
+        return inverse_hash(d, el)::K => d.map[el]::V, i
     end
 end
+
+function inverse_hash(d::DenseDict{K, V, F, I}, el)::K where {K,V,F,I}
```
blegat, line 35: Is this still needed with `_index_to_variable` ?

src/Utilities/cachingoptimizer.jl
odow: `_standardize(::IndexMap)`

src/Utilities/dense_dict.jl
joaquimg: Indeed not needed anymore

src/Utilities/dense_dict.jl
```diff
@@ -22,15 +22,16 @@ end
...
 # Base.empty(::DenseDict, ::Type{K}, ::Type{V}) not implemented
-function Base.iterate(d::DenseDict, args...)
+function Base.iterate(d::DenseDict{K,V}, args...) where {K,V}
     itr = iterate(d.set, args...)
     if itr === nothing
         return nothing
     else
         el, i = itr
-        return d.inverse_hash(el) => d.map[el], i
+        return d.inverse_hash(el)::K => d.map[el]::V, i
```
blegat, line 31: Are the `::K` and `::V` needed ?

blegat:
This is taking the majority of time for a benchmark with https://github.com/jump-dev/SCS.jl/pull/192
indexmap_copy

joaquimg:
> This line:

Did you run with this PR?
This line is not there anymore.

blegat:
No, I run with master, my point is that this PR is needed

src/Utilities/dense_dict.jl
joaquimg: Just tested, Julia does not seem to infer those correctly. Therefore the is a very minor difference a little less than 1%.
I vote for leaving it there, since its not restricting anything.
However, if we really want to avoid that, I can remove.

joaquimg:
@blegat ready to merge this?
I had to revert a commit, so it would be better to squash when merging.

src/Utilities/dense_dict.jl
blegat: No if Julia inference is not doing it correctly then leaving it there is what we should do

PR #1133, opened 2020-07-28
Add name option for basic_constraint_test

blegat:
It would be helpful to use these tests for Ipopt but it does not support names


odow:
Did you miss a commit? You don't actually use the `name` option.

blegat:
Indeed, thanks for catching this, should be fixed now

Issue #853, opened 2019-08-26
Add VariablePrimalStart test


odow:
Because Gurobi does not support `ConstraintPrimalStart`, it skips the `start_values_test`. 

But this led to a bug: https://github.com/JuliaOpt/Gurobi.jl/pull/238

It would be good to have separate tests for `VariablePrimalStart`, `ConstraintPrimalStart`, and `ConstraintDualStart`.

Currently, we just have
https://github.com/JuliaOpt/MathOptInterface.jl/blob/6ac4e6d6ee47486062ab9a14816aca6d51328760/src/Test/modellike.jl#L321-L381

odow:
Closed by https://github.com/jump-dev/MathOptInterface.jl/pull/1136

Issue #794, opened 2019-07-16
Test getting `VariablePrimalStart`


odow:
See: https://discourse.julialang.org/t/warning-invalid-warm-start-basis-discarded/26403/2
See: https://github.com/JuliaOpt/Gurobi.jl/issues/225

We should add a test like the following to MOIT.
```Julia
model = Gurobi.Optimizer()
x = MOI.add_variables(model, 3)
MOI.set(
    model, 
    MOI.ObjectiveFunction{MOI.ScalarAffineFunction{Float64}}(), 
    MOI.ScalarAffineFunction(
        MOI.ScalarAffineTerm.(1.0, x),
        0.0
    )
)
MOI.add_constraint(model, MOI.SingleVariable(x[2]), MOI.GreaterThan(1.0))
MOI.set(model, MOI.VariablePrimalStart(), x[1], 1.0)
MOI.set(model, MOI.VariablePrimalStart(), x[2], 0.0)
MOI.set(model, MOI.VariablePrimalStart(), x[3], 2.0)
@test MOI.get.(model, MOI.VariablePrimalStart(), x) == [1.0, 0.0, 2.0]

MOI.set(model, MOI.VariablePrimalStart(), x[2], nothing)
@test MOI.get.(model, MOI.VariablePrimalStart(), x) == [1.0, nothing, 2.0]
```

QunLin0:
I got the following message with the above code (noting that I use @show, instead of @test):

Warning, invalid warm-start basis discarded

ArgumentError: ModelLike of type Gurobi.Optimizer does not support accessing the attribute MathOptInterface.VariablePrimalStart()

odow:
@QunLin0, correct. The reason to add these tests is so that the solvers will all support them. At present, we aren't testing setting warm-starts properly, which is why some solvers haven't implemented them correctly.

odow:
Closed by https://github.com/jump-dev/MathOptInterface.jl/pull/1136

Issue #649, opened 2019-01-29
Document that nothing can be a value provided to set!


frapac:
Hi, I was testing the package [MINLPTests.jl](https://github.com/lanl-ansi/MINLPTests.jl ) with KNITRO, and when running this code: 
```julia
optimizer = with_optimizer(KNITRO.Optimizer)
m = Model(optimizer)

@variable(m, x, start=0.1)
@variable(m, y)

@objective(m, Min, -x-y)
@NLconstraint(m, exp(x-2.0) - 0.5 <= y)
@NLconstraint(m, log(x) + 0.5 >= y)

optimize!(m)
````
I get the following error: 
`ERROR: MathOptInterface.SetAttributeNotAllowed{MathOptInterface.VariablePrimalStart}`. 

I get the same error with Ipopt. After some investigation, it appears that the error does not arise 
in `DirectMode`. In fact, it happens when we set the `VariablePrimalStart` for all variables all in once here:
https://github.com/JuliaOpt/MathOptInterface.jl/blob/master/src/attributes.jl#L286

In this example, as the primal start of `y` is not specified, the type of the array `vector_of_values` in MOI is 
`Union{Nothing, Float64}` 

However, both Ipopt and KNITRO support only `Real` in the setter of `VariablePrimalStart` : 
```julia
function MOI.set(model::Optimizer, ::MOI.VariablePrimalStart, vi::MOI.VariableIndex, value::Real)
...
end
```

The question is: 

- shall we modify the setter in KNITRO and Ipopt to discard by hand primal start of type `Nothing`?
- or shall we filter directly `Nothing` inside MOI?

frapac:
Thanks for the response and the PR! I will mirror the behavior of Ipopt to update KNITRO. If you want help to update MOI documentation, tell me.

mlubin:
Leaving this open so that we don't forget to fix the documentation.

joaquimg:
We want that explicit in the docstring for `set!`?
Because #1136 already added that do `VariablePrimalStart`'s docstring.

odow:
I think it is sufficient to clarify for each attribute where it applies. This won't always be the case, so it isn't appropriate for `set`.

PR #1136, opened 2020-08-03
Fixing `set_start_value` only accept numbers (#2237)

henriquebecker91:
Intends to close #2237 and some related open issues (#649, #794, #853).

Some guidance is needed. For now, I just edited what I found to be the relevant piece of documentation, and added a new tests as well flags to an old test (allowing more granular selection of which attributes are tested). I would like to know:

1. There are other places where the documentation needs to be changed? `JuMP` probably will need a change, but there are other points inside MOI?
2. Seems to me that allowing `nothing` to be passed is something done at the wrapper level, and the generic methods in MOI already allow it. The test will probably force the wrappers to allow this behavior.
3. How is the best way to check if the test is working? Should I download a bunch of wrappers and make they use my version of MOI, and then call their tests?

No hurry, @odow, I do not know if I will work during the week, but I will try finish it next weekend if I am sure of what needs to be done yet.

### DIFF
--- a/src/Test/modellike.jl
+++ b/src/Test/modellike.jl
@@ -451,7 +451,7 @@ function start_values_unset_test(model::MOI.ModelLike, config)
         # determined behaviour, there are two possibilities about the behaviour
         # of ListOfConstraintAttributesSet and ListOfVariableAttributesSet:
         #
-        # 1) They list all attributes ever set, even after unsetted.
+        # 1) They list all attributes ever set, even after attributes are unset.
         # 2) They list only the attributes that, at the moment, are set.
         #
         # The current behavior is (1), but we are not sure if this will not
--- a/src/Test/modellike.jl
+++ b/src/Test/modellike.jl
@@ -386,7 +386,13 @@ function failcopytestca(dest::MOI.ModelLike)
     @test_throws MOI.UnsupportedAttribute MOI.copy_to(dest, BadConstraintAttributeModel())
 end
 
-function start_values_test(dest::MOI.ModelLike, src::MOI.ModelLike)
+function start_values_test(
+    dest::MOI.ModelLike, src::MOI.ModelLike;
+    skip_variable_primal = false,
+    skip_vp_unsetting = false,
+    skip_constraint_primal = false,
+    skip_constraint_dual = false
+)
     MOI.empty!(dest)
     @test MOIU.supports_default_copy_to(src, #=copy_names=# false)
     x, y, z = MOI.add_variables(src, 3)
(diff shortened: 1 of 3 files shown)


src/attributes.jl
```diff
@@ -965,6 +965,8 @@ struct VariableName <: AbstractVariableAttribute end
     VariablePrimalStart()
 
 A variable attribute for the initial assignment to some primal variable's value that the optimizer may use to warm-start the solve.
+
+The value `nothing` means the solver is free to decide which value to use. This attribute can be set back to `nothing` after being set to a numerical value.
```
mlubin, line 969: The `VariablePrimalStart` is a hint that the optimizer "may" use to warm-start the solve when it's set. It's redundant to say the solver "is free to decide which value to use" when it's set to `nothing`, because this is also true when it's set to a number. Just say "May be a number or `nothing` (unset)."

odow:
1. We should update the documentation for `ConstraintDualStart` and `ConstraintPrimaStart` as well:
https://github.com/jump-dev/MathOptInterface.jl/blob/a6274e63f8cc020e16cb99a0b2d52bd8062e1613/src/attributes.jl#L1046-L1053

2. Rather than modifying the current test, you should add new tests
```Julia
function test_variable_primal_start(model, config)
    MOI.empty!(model)
    x = MOI.add_variable(model)
    @test MOI.get(model, MOI.VariablePrimalStart(), x) === nothing
    MOI.set(model, MOI.VariablePrimalStart(), x, 1.0)
    @test MOI.get(model, MOI.VariablePrimalStart(), x) == 1.0
    MOI.set(model, MOI.VariablePrimalStart(), x, nothing)
    @test MOI.get(model, MOI.VariablePrimalStart(), x) === nothing
end
```

3. You can add a test for the test here: https://github.com/jump-dev/MathOptInterface.jl/blob/master/test/Test/modellike.jl
Something like
```julia
model = MOIU.MockOptimizer(MOIU.UniversalFallback(MOIU.Model{Float64}()))
MOI.Test.test_variable_primal_start(model, MOI.Test.TestConfig())
```

henriquebecker91:
@odow, you asked to do the same modification of documentation to `ConstraintPrimalStart` and `ConstraintDualStart`, ok, I did (not pushed yet), but then you suggest creating just a `function test_variable_primal_start(model, config)`, tests should be added for `ConstraintPrimalStart` and `ConstraintDualStart`, no?

henriquebecker91:
I made a new commit. It updates the documentation of `VariablePrimalStart`, `ConstraintDualStart`, and `ConstraintPrimalStart` to what was suggested by @mlubin.

I also reverted the test I modified and instead created a new test, but not only for `VariablePrimalStart` (also for the constraint start values), and it is a little more involved than the test suggested by @odow.

I call the new test in the fashion suggested by @odow. The `config` parameter is not used, should it be used? 

My new test method has 5 commented `@test` calls. They fail if uncommented. I am not sure if this is desired behavior or not. I expect a decision from the maintainers about removing or uncommenting them. In the current situation (with these 5 commented) all tests pass.

henriquebecker91:
Should I be doing something here?

henriquebecker91:
Right now, I have a window of opportunity to make any changes necessary for this to be approved, while I wait for some reviews of my last paper. Any directions?

src/attributes.jl
```diff
@@ -964,7 +964,7 @@ struct VariableName <: AbstractVariableAttribute end
 """
     VariablePrimalStart()
 
-A variable attribute for the initial assignment to some primal variable's value that the optimizer may use to warm-start the solve.
+A variable attribute for the initial assignment to some primal variable's value that the optimizer may use to warm-start the solve. May be a number or nothing (unset).
```
blegat, line 967: nothing -> `nothing`

src/Test/modellike.jl
```diff
@@ -386,6 +386,87 @@ function failcopytestca(dest::MOI.ModelLike)
...
+        @test all(MOI.get.((model,), (vpattr,), (x, y, z)) .== (0.0, 1.0, 2.0))
+        @test all(MOI.get.((model,), (cpattr,), (a, b)) .== (0.0, 1.0))
+        @test all(MOI.get.((model,), (cdattr,), (a, b)) .== (0.0, 1.0))
+        @test MOI.get(model, cpattr, c) == [2.0]
+        @test MOI.get(model, cdattr, c) == [2.0]
+
+        MOI.set(model, vpattr, y, nothing)
+        MOI.set(model, cpattr, a, nothing)
+        MOI.set(model, cdattr, a, nothing)
+        MOI.set(model, cpattr, c, [nothing])
```
blegat, line 447: I would expect it to be `nothing`, not `[nothing]`

src/Test/modellike.jl
```diff
@@ -386,6 +386,87 @@ function failcopytestca(dest::MOI.ModelLike)
...
+        MOI.set(model, cdattr, c, [nothing])
+
+        #@test cpattr ∉ MOI.get(model, MOI.ListOfConstraintAttributesSet{F2, S2}())
```
blegat, line 450: Why is this commented ?

henriquebecker91: As I said above:

> My new test method has 5 commented @test calls. They fail if uncommented. I am not sure if this is desired behavior or not. I expect a decision from the maintainers about removing or uncommenting them. In the current situation (with these 5 commented) all tests pass.

src/Test/modellike.jl
henriquebecker91: I will check this. I noticed now that I do not query the `cpattr` of `c` before setting it the first time. If the initial value of `cpattr` for `c` is `[nothing]`, then it would make sense keep it that way, to avoid breaking the reasoning that initially motivated this behavior, that is, ``if variables with an attribute unset return `nothing` when queried, then setting them with the returned value (i.e., nothing) should restore the original/unset state''. 

I have to admit I did not deal directly with this kinda of constraint, so I do not know what to expect here.
```julia
c = MOI.add_constraint(model, MOIU.operate(vcat, Float64, 2.0fz + 1.0), MOI.Nonnegatives(1))
```

src/attributes.jl
henriquebecker91: Already fixed in my local copy. Will be together with the fix of the other problems.

src/Test/modellike.jl
henriquebecker91: Oh, okay, I was already checking it before and it has value `nothing`. So the `MOI.set` should take `nothing` and `MOI.get` should return `nothing`. I changed the tests and it works. Should we have some tests for `nothing` values inside the arrays that may be used to set this attribute, or setting the whole constraint to `nothing` and getting `nothing` back is enough?

src/Test/modellike.jl
henriquebecker91: Expanding a little on the above:

The documentation for [`ListOfVariableAttributesSet`](https://jump.dev/MathOptInterface.jl/stable/apireference/#MathOptInterface.ListOfVariableAttributesSet) and for [`ListOfConstraintAttributesSet`](https://jump.dev/MathOptInterface.jl/stable/apireference/#MathOptInterface.ListOfConstraintAttributesSet) are both not very clear (at least to me) on what happens if the attributes are unset.

* If I search for the attributes in these lists from a newly created model, they do not exist in the lists, as it is expected.
* If I search for them in these lists after setting them  in the model, they do exist in the lists, as it is expected.
* The problem is that after *unsetting* the attributes, they keep existing in these lists, I do not know if this is desired behaviour.

More specifically, I do not know if this is a list for every attribute that was ever set (i.e., `MOI.set` was called with it), or just for the attributes with a value different from `nothing` at the moment the list is queried.

blegat: Yes, this is a bit ambiguous, we don't want the definition to be to constrained here as it would make it difficult to implement and it does not change anything for the main use case which is using it for copying.
You can remove these tests or add a comment explaining that it currently fails but we may want to enable these in the future.

src/Test/modellike.jl
henriquebecker91: Just changed the tests to expect `nothing` and set `nothing`. No test for `nothing` elements inside a Vector.

src/Test/modellike.jl
henriquebecker91: I kept them commented and added a comment explaining why they are commented referencing this thread.

henriquebecker91:
I followed all @blegat suggestions and did a minor cosmetic change (changed `!(... \in ...)` to `... \notin ...`).

henriquebecker91:
Screenshot of tests passing:
[](https://postimg.cc/cKW5pLBj)

src/Test/modellike.jl
blegat: It's ok if `nothing` element inside vector for a vector constraint is undefined behavior for now, this is out of the scope of this PR.

src/Test/modellike.jl
```diff
@@ -386,6 +386,100 @@ function failcopytestca(dest::MOI.ModelLike)
...
+        MOI.set(model, cpattr, a, nothing)
+        MOI.set(model, cdattr, a, nothing)
+        MOI.set(model, cpattr, c, nothing)
+        MOI.set(model, cdattr, c, nothing)
+
+        # The tests below are commented because it was decided to not pin a
+        # determined behaviour, there are two possibilities about the behaviour
+        # of ListOfConstraintAttributesSet and ListOfVariableAttributesSet:
+        #
+        # 1) They list all attributes ever set, even after unsetted.
```
blegat, line 454: unsetted -> unset

blegat (approved):
Can you remove the "Draft" state of the PR ?

henriquebecker91:
Sorry, forgot about that. Now it is ready for review. I will make that small comment change.

src/attributes.jl
henriquebecker91: Done.

PR #1139, opened 2020-08-06
Double dict reversion

joaquimg:
Following #1129

This PR applies dictionary reversion directly do double dicts.
Now it is type stable and more performant.

### DIFF
--- a/src/Utilities/DoubleDicts.jl
+++ b/src/Utilities/DoubleDicts.jl
@@ -73,24 +73,17 @@ end
     v
 end
 
-function MOIU._reverse_dict(dest::IndexDoubleDict, src::IndexDoubleDict)
+# reversing IndexDoubleDict is ok because they map CI to CI
+function MOIU._reverse_dict(
+    dest::IndexDoubleDict{DI}, src::IndexDoubleDict{DI}
+) where DI
     for (k, v) in src.dict
-        dest.dict[k] = MOIU._reverse_dict(v)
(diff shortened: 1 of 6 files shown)


src/Utilities/DoubleDicts.jl
```diff
@@ -72,6 +73,18 @@ end
     v
 end
 
+# reversing IndexDoubleDict is ok because they map CI to CI
+function MOIU._reverse_dict(
+    dest::IndexDoubleDict{DI}, src::IndexDoubleDict{DI}
+) where DI
+    for (k, v) in src.dict
+        dest.dict[k] = DI(values(v) .=> keys(v))
```
blegat, line 81: Why not call `MOIU._reverse_dict(dest.dict[k], v)` ? That way, it would automatically specialize for the best way to reverse the inner dicts as well.

joaquimg: I will add a single input version, because `dest.dict[k]` is initially empty and its more efficient to pass data on construction instead of building an empty one and then populating.

src/Utilities/Utilities.jl
```diff
@@ -26,6 +26,9 @@ function print_with_acronym(io::IO, s::AbstractString)
     print(io, s)
 end
 
+
+function _reverse_dict end
```
odow, line 30: This could probably use a docstring explaining what it does.

joaquimg: done

odow:
LGTM.

PR #1140, opened 2020-08-06
Indexing performance + lazy dense dicts

joaquimg:
Follows: #1129 

This avoids the conversion from dense dict to standard dict in copy functions.

### DIFF
--- a/src/Utilities/cachingoptimizer.jl
+++ b/src/Utilities/cachingoptimizer.jl
@@ -145,25 +145,32 @@ errors can be thrown.
 """
 function attach_optimizer(model::CachingOptimizer)
     @assert model.state == EMPTY_OPTIMIZER
-    # We do not need to copy names because name-related operations are handled by `m.model_cache`
+    # We do not need to copy names because name-related operations are handled
+    # by `m.model_cache`
     indexmap = MOI.copy_to(model.optimizer, model.model_cache, copy_names=false)
     model.state = ATTACHED_OPTIMIZER
-    # MOI does not define the type of index_map, so we might have to copy it
-    # into a concrete container. Also load the reverse map.
-    model.model_to_optimizer_map = convert(IndexMap, indexmap)#IndexMap()
-    model.optimizer_to_model_map = reverse_index_map(indexmap)#IndexMap()
-    nothing
+    # MOI does not define the type of index_map, so we have to convert it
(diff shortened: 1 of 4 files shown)


blegat:
I remember we hesitated between adding a dict in DenseDict or using DenseDict inside a CleverDict to switch from DenseDict to Dict in case of deletion. I thought we decided the second one but I don't remember the details

blegat:
I think we should
1) replace the vector by a DenseDict in CleverDict
2) add the possibility to add a new contiguous index to DenseDict
3) implement setindex! for CleverDict similarly to this PR

joaquimg:
> I remember we hesitated between adding a dict in DenseDict or using DenseDict inside a CleverDict to switch from DenseDict to Dict in case of deletion. I thought we decided the second one but I don't remember the details

I think we only hesitated because it could be too similar to CleverDict.

joaquimg:
We need to be careful with CleverDict because of the downstream usage.
We need to check if theses changes will be breaking.

joaquimg:
one problem is that:
CleverDicts requires the methods `index_to_key` and `key_to_index`
https://github.com/jump-dev/MathOptInterface.jl/blob/746b2e101728724f47402838630e921f2917b677/src/Utilities/CleverDicts.jl#L9

As methods to be overloaded, while DenseDicts get `hash` and `inverse_hash` in the type constructor.
This would be breaking for solvers like GLPK that overload the CleverDict methods:

https://github.com/jump-dev/GLPK.jl/blob/41e649cd81864e90065ac78f42317edcebcd9951/src/MOI_wrapper/MOI_wrapper.jl#L34

blegat:
We could use these functions as default for the map and inverse map of DenseDict so that it is non-breaking

odow:
It is a bit confusing that we have so many types of dictionaries. It's probably okay to get some performance improvement for now, but at some point I'd like to go through and tidy up a lot of this.

joaquimg:
So instead of adding a DenseDict inside CleverDict I will merge the two of them.
As @blegat that can be done without breaking. And DenseDict was never released anyway.

blegat:
Merging them sounds good

blegat:
@joaquimg Is this closed by https://github.com/jump-dev/MathOptInterface.jl/pull/1142 ?

PR #1142, opened 2020-08-09
Merge Clever and Dense dicts

joaquimg:


src/Utilities/CleverDicts.jl
```diff
@@ -33,7 +37,11 @@ Note that querying a `LinearIndex` immediately after deleting a key via
 Store an item `val` using `add_item(c::CleverDict, val)`. `add_item` returns a
 key corresponding to the stored item.
 
-Overload the functions `index_to_key` and `key_to_index` to enable mappings
+It is possible to initialize the CleverDict as `CleverDict{K, V}(n)` so that
+`n` elements are pre-allocated and can be efficiently added even out of order
+as long as theirkey hashes are between ´1´ and `n`.
```
blegat, line 42: Missing space after their

src/Utilities/CleverDicts.jl
```diff
@@ -33,7 +37,11 @@ Note that querying a `LinearIndex` immediately after deleting a key via
...
+
+Overload the functions `index_to_key` and `key_to_index` to enable vectorpings
```
blegat, line 44: "vectorpings"?

src/Utilities/CleverDicts.jl
```diff
@@ -46,11 +54,28 @@ index_to_key(::Type{MyKey}, i::Int) = MyKey(i)
 key_to_index(key::MyKey) = key.x
 ```
 """
-mutable struct CleverDict{K, V} <: AbstractDict{K, V}
+mutable struct CleverDict{K, V, F, I} <: AbstractDict{K, V}
```
blegat, line 57: `F<:Function`

src/Utilities/CleverDicts.jl
```diff
@@ -63,68 +88,119 @@ function index_to_key end
...
+        if !haskey(c, key)
+            throw(KeyError(key))
+        end
+        return c.vector[c.hash(key)::Int]
+    else
+        return c.dict[key]
+    end
+end
+function Base.setindex!(c::CleverDict{K, V}, value, key)::V where {K, V}
+    h = c.hash(key)::Int
```
blegat, line 151: `Int64`

src/Utilities/CleverDicts.jl
odow: Also ´1´ -> `1`

src/Utilities/CleverDicts.jl
odow: Is this going to break the downstream solvers?

src/Utilities/CleverDicts.jl
```diff
@@ -63,68 +88,119 @@ function index_to_key end
 """
     key_to_index(key::K)
 
-Map `key` to an integer valued index, assuming that there have been no
+vector `key` to an integer valued index, assuming that there have been no
```
odow, line 91: ?

src/Utilities/CleverDicts.jl
joaquimg: Haven't tested, but it is possible.
After writing the entire thing I think that it ok to use the function overloading for now.

joaquimg: Yet another non-breaking possibility is to have AbstractCleverDict with 4 params and the Clever Dict with 2, no super good though...

odow: What's the hit of just making them `::Function`? Or making `index_to_key` dependent on the `K-V` pair: `key_to_index(::CleverDict{K, V}, x::K)::Int` and `index_to_key(::CleverDict{K, V}, v::Int)::K`,

joaquimg: just `::Function` is not good for inference, and the type wont be concrete.

I like:
> making index_to_key dependent on the K-V pair: key_to_index(::CleverDict{K, V}, x::K)::Int and index_to_key(::CleverDict{K, V}, v::Int)::K

odow: If you go the second route, we will need a deprecated fallback.

joaquimg:
@odow it turns out that the way it currently is does not break GLPK, CPLEX, Gurobi nor Xpress.
Any other solver I should test?

They might have a small inference penalty because their structs will be referencing an abstract type.
A simple straight forward PR fixes this non-breaking problem.

If this one gets merged I can quickly PR an optimized `_reverse_dict` for `CleverDict`.

blegat:
> They might have a small inference penalty because their structs will be referencing an abstract type.

This is because we added type arguments to CleverDict right? So adding it make it concrete?

joaquimg:
> This is because we added type arguments to CleverDict right? So adding it make it concrete?

Yes, the PRs only have to add 2 parameters to the type definition of the CleverDict in the optimizer structs.

blegat:
Any objection to merge?

joaquimg:
It is good for me!

PR #1146, opened 2020-08-23
Update to MathOptFormat v0.5

odow:
x-ref: https://github.com/jump-dev/MathOptFormat/pull/14

### DIFF
--- a/src/FileFormats/MOF/MOF.jl
+++ b/src/FileFormats/MOF/MOF.jl
@@ -7,7 +7,7 @@ import JSONSchema
 import MathOptInterface
 
 const MOI = MathOptInterface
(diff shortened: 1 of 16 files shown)


odow:
Any other changes we want?

blegat:
What is going to happen when you try to read an old file with the new MOI ? It will print an error message saying that the version does not match ?

odow:
Ah. Good catch. This piece of code doesn't really make sense:
https://github.com/jump-dev/MathOptInterface.jl/blob/284ab043fd6de62c2d73b822b36732d891f33800/src/FileFormats/MOF/read.jl#L16-L22

Hold off merging, and I will push a fix.

blegat (approved):

PR #1161, opened 2020-09-16
v0.9.16

odow:


odow:
@blegat do you have any other items for 0.9.16?

blegat:
I added two PRs in the milestone that are needed for Mosek to pass the tests on MOI master. Once these are merged, we can to a SolverTests run and tag if all is green.

blegat (approved):

PR #1172, opened 2020-10-04
Tweak wording of LazyConstraintCallback docs

odow:
x-ref: https://github.com/jump-dev/GLPK.jl/issues/146

GLPK sometimes asks for a lazy constraint at an integer-infeasible point. We can either tweak these docs to drop the mention of primal feasibility, or tweak GLPK to check for integer feasibility before asking for a lazy constraint.

### DIFF
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -649,7 +649,7 @@ solution by submitting a [`LazyConstraint`](@ref). For instance, it may be
 called at an incumbent of a mixed-integer problem. Note that there is no
 guarantee that the callback is called at *every* feasible primal solution.
 
-The feasible primal solution is accessed through
+The current primal solution is accessed through
 [`CallbackVariablePrimal`](@ref). Trying to access other result
 attributes will throw [`OptimizeInProgress`](@ref) as discussed in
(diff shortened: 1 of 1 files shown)


blegat:
Adding a warning in the doc or code or a fix to GLPK would help as this is a difference from JuMP v0.18 so it might cause issues for code transitioning.
In JuMP v0.18, the callback had to set `state` to `MIPSol` if it was integer feasible, e.g.
https://github.com/JuliaOpt/GLPKMathProgInterface.jl/blob/26f755770f8b70a81ade83cd1e6d85fa8c8254c9/src/GLPKInterfaceMIP.jl#L99-L110
Then, if it was fractional and `fractional` was not set explicitly to `true` when creating the callback with the JuMP macro, it was not called:
https://github.com/jump-dev/JuMP.jl/blob/a9de4d0de1b34b85d35c3624bea236afec7e5bec/src/callbacks.jl#L77-L78

odow:
There's an issue for that: https://github.com/jump-dev/MathOptInterface.jl/issues/1112

blegat (approved):

PR #1176, opened 2020-10-08
fix real types in TestConfig

chriscoey:
Is the `::Float64` necessary? I like to use `tol = sqrt(sqrt(eps(T)))` but the type constraint forced me to write `tol = sqrt(sqrt(Float64(eps(T))))`.

Side note, `atol = rtol = sqrt(eps(T))` (equal to `1.4901161193847656e-8` for Float64) might be a reasonable default to put here instead of `1e-8`, but that is likely to be more controversial.


chriscoey:
pinging anyone to review/merge

src/Test/config.jl
```diff
@@ -1,6 +1,6 @@
 struct TestConfig{T <: Number}
-    atol::Float64 # absolute tolerance for ...
-    rtol::Float64 # relative tolerance for ...
+    atol::Real # absolute tolerance for ...
```
blegat, line 2: Why not `atol::T` ?

blegat: Here, the type of the field is not concrete

chriscoey: Can I leave it as ::Real in the constructor function and use ::T just for the field?

chriscoey: IIRC it will automatically convert when setting the field in construction

blegat: Yes, that sounds good, what's weird is that we allow `T` to be `Number` but it does not makes sense to have a complex tolerance

chriscoey: do we ever want any complex TestConfigs? could we change Number to Real?

chriscoey:
Travis fails after the second commit https://travis-ci.com/github/jump-dev/MathOptInterface.jl/jobs/400457944
```
EqualTo{Int64}: Error During Test at /home/travis/build/jump-dev/MathOptInterface.jl/test/Bridges/Constraint/interval.jl:106
  InexactError: Int64(1.0e-8)
  Stacktrace:
   [1] Int64 at ./float.jl:710 [inlined]
   [2] convert at ./number.jl:7 [inlined]
   [3] MathOptInterface.Test.TestConfig{Int64}(; atol::Float64, rtol::Float64, solve::Bool, query::Bool, modify_lhs::Bool, duals::Bool, dual_objective_value::Bool, infeas_certificates::Bool, optimal_status::MathOptInterface.TerminationStatusCode, basis::Bool) at /home/travis/build/jump-dev/MathOptInterface.jl/src/Test/config.jl:19
   [4] MathOptInterface.Test.TestConfig{Int64}() at /home/travis/build/jump-dev/MathOptInterface.jl/src/Test/config.jl:19
   [5] top-level scope at /home/travis/build/jump-dev/MathOptInterface.jl/test/Bridges/Constraint/interval.jl:112
   [6] top-level scope at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Test/src/Test.jl:1190
``` 
because at https://github.com/jump-dev/MathOptInterface.jl/blob/3d801ebf6d257dec7437e57360e689046ae34bdd/test/Bridges/Constraint/interval.jl#L112 T is Int in one test case.

Does it really make sense to parametrize the tolerances the same as the TestConfig? I can see cases where the tolerances should be float or rational, but maybe not Int and probably not Complex. And again, are there really cases where we would want a complex TestConfig? I am not sure what the plan is for complex support in MOI. Fundamentally it seems to be that models should be "real", for example the objective function needs to be real because the complex numbers aren't ordered

mtanneau:
I agree. I would even go as far as to say we could drop the type parametrization for `TestConfig`.

* Currently, the type parameterization `TestConfig{T}` is redundant: `atol` and `rtol` are hard-coded `Float64`, and  `T` is not used anywhere in the constructor.
* Setting default values for `atol` and `rtol` to `sqrt(eps(T))` seems very reasonable to me. Users are responsible for setting their own tolerances, anyway.
* I don't see a compelling reason to force the `T` of `TestConfig{T}` to be the same as the `T` of `Optimizer{T}`

chriscoey:
I agree with @mtanneau. @blegat is there a reason for parametrizing TestConfig?

mtanneau:
There are quite a few test functions (in `MOI.Tests`) that have a signature like
```julia
function linear1test(model::MOI.ModelLike, config::TestConfig{T}) where T
    ...
    @test MOI.supports_constraint(model, MOI.SingleVariable, MOI.EqualTo{T})
```

Dropping the parametrization of `TestConfig` would break those tests, but one could argue that it would make more sense to have something like
```julia
function linear1test(model::MOI.ModelLike{T}, config::TestConfig) where T
    ...
    @test MOI.supports_constraint(model, MOI.SingleVariable, MOI.EqualTo{T})
```
so that the test uses the model's arithmetic, not the `TestConfig`'s.

So it may be a smoother transition to do `atol::T` and `rtol::T` in `TestConfig{T}`.

src/Test/config.jl
```diff
@@ -12,7 +12,7 @@ struct TestConfig{T <: Number}
     optimal_status::MOI.TerminationStatusCode
     basis::Bool # can get variable and constraint basis status
     function TestConfig{T}(;
-        atol::Float64 = 1e-8, rtol::Float64 = 1e-8, solve::Bool = true,
+        atol::Real = 1e-8, rtol::Real = 1e-8, solve::Bool = true,
```
blegat, line 15: I think `atol = Base.rtoldefault(one(T), one(T), 0)` and `rtol = atol` would make `atol` and `rtol` have type `T`.

blegat:
I would use `T <: Real`, `TestConfig{T}`  means that the real type to use in the test is `T` and the complex type is `Complex{T}`.
`atol` and `rtol` cannot be `Float64`, they should be allow to be more accurate if, e.g. `T` is `BigFloat`.

src/Test/config.jl
chriscoey: `sqrt(eps(T))` gives the same result and type, and is shorter and more transparent, so I used that.

src/Test/config.jl
```diff
@@ -12,10 +12,10 @@ struct TestConfig{T <: Number}
     optimal_status::MOI.TerminationStatusCode
     basis::Bool # can get variable and constraint basis status
     function TestConfig{T}(;
-        atol::Float64 = 1e-8, rtol::Float64 = 1e-8, solve::Bool = true,
+        atol::Real = sqrt(eps(T)), rtol::Real = sqrt(eps(T)), solve::Bool = true,
```
blegat, line 15: Isn't eps only defined for AbstractFloat?

chriscoey: yep, OK `Base.rtoldefault` it is

chriscoey: I think `Base.rtoldefault(T)` is OK instead of `Base.rtoldefault(one(T), one(T), 0)`, just looking at https://github.com/JuliaLang/julia/blob/e58bc723ed7b1c10da28f55385c243f6df5499cf/base/floatfuncs.jl#L299

blegat (approved):

chriscoey:
if we can tag a new version of MOI soon, Hypatia can start to run more MOI tests with non-Float64 reals

PR #1178, opened 2020-10-19
add geomean duals tests and geomean to entropy bridge

chriscoey:
replaces #1021 and #1017.


blegat:
See CI failure
https://travis-ci.com/github/jump-dev/MathOptInterface.jl/jobs/401824880#L460

chriscoey:
thanks, this will take a bit more time. there are things i need to bring over from #1021

chriscoey:
This passes MOI contconic tests (including duals) using with the combination of bridges at https://github.com/chriscoey/Hypatia.jl/blob/318a20dd7f9c19ac62eb6914e1bdf6624439c25e/test/moi.jl#L74

src/Bridges/Constraint/geomean_to_relentr.jl
```diff
@@ -0,0 +1,95 @@
+"""
+    GeoMeantoRelEntrBridge{T}
+
+The `geometric mean cone` is representable with a relative entropy constraint and a nonnegative auxiliary variable,
+since ``u \\le \\prod_{i=1}^n w_i^{1/n}`` if and only if there exists a ``y \\ge 0`` such that
+``0 \\ge \\sum_{i=1}^n (u + y) \\log (\\frac{u + y}{w_i})``, or equivalently
```
blegat, line 6: Can you detail a bit more where this reduction come from or give a ref ?

src/Bridges/Constraint/geomean_to_relentr.jl
```diff
@@ -0,0 +1,95 @@
...
+    u_val = value[1]
+    u_pos = -min(zero(u_val), u_val)
+    d = length(value) - 1
+    MOI.set(model, MOI.VariablePrimalStart(), bridge.y, u_pos)
+    MOI.set(model, attr, bridge.nn_index, [u_pos])
+    MOI.set(model, attr, bridge.relentr_index, vcat(0, value[2:end], fill(u_pos + u_val, d)))
+    return
+end
+# Given a is dual on y >= 0 and (b, c, d) is dual on RelativeEntropyCone constraint,
+# dual on (u, w) in GeometricMeanCone is (-a, c).
```
blegat, line 76: I have checked and I have the same thing with (2.8) of https://drive.google.com/file/d/1Y8kvUsHmvWSvgnEBS8T7GI6ofvuWng6U/view?usp=sharing
I would add in the comment that we have `sum(d) = -a` so we could also use `sum(d)` instead of `-a`.

src/Bridges/Constraint/geomean_to_relentr.jl
chriscoey: I didn't see this anywhere. But I can try to show why the math works

chriscoey: See updated explanation

src/Bridges/Constraint/geomean_to_relentr.jl
chriscoey: See updated comment

blegat (approved):
Thanks for all the hard work on this bridge! It might be a good fit for advanced dual mapping for bridges in the doc or in a blog post.

PR #1182, opened 2020-10-23
Implement conversion between ScalarAffineFunction of different coeffi…

blegat:
…cent type


blegat:
I checked manually that all lines were covered by adding `@show @__LINE__` statements. They are all covered, the coverage drop should be due to inlining.

PR #1184, opened 2020-10-30
Doctests work on both julia 1.5.2 and 1.0.5

shadiakiki1986:
Related to issue #990

### DIFF
--- a/docs/src/apireference.md
+++ b/docs/src/apireference.md
@@ -3,6 +3,14 @@ CurrentModule = MathOptInterface
 DocTestSetup = quote
     using MathOptInterface
     const MOI = MathOptInterface
+
+    # For compatibility with both Julia 1.0.5 and 1.5.2
+    # Upon the Julia LTS version becoming 1.6, these imports could be dropped,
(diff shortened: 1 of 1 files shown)


docs/src/apireference.md
```diff
@@ -3,6 +3,10 @@ CurrentModule = MathOptInterface
 DocTestSetup = quote
     using MathOptInterface
     const MOI = MathOptInterface
+
+    # For compatibility with both Julia 1.0.5 and 1.5.2
```
blegat, line 7: This is needed for Julia v1.0 but not for Julia v1 right ?

blegat: Sorry, yes I mean v1.5 by v1. Having it work or both v1.0 and v1.5 is ideal of course but it would be nice to leave a comment here detailing whether this is needed for v1.0 or v1.5 so that we know what we do once v1.6 is the new LTS and we drop v1.0

shadiakiki1986: Added more comments as discussed

blegat (approved):

PR #1190, opened 2020-11-05
Add comprehensive tests for Farkas certificates

odow:
Implemented in solvers:
- https://github.com/jump-dev/Clp.jl/pull/105
- https://github.com/jump-dev/CPLEX.jl/pull/337
- https://github.com/jump-dev/GLPK.jl/pull/155
- https://github.com/jump-dev/Gurobi.jl/pull/363

Solvers like Xpress probably still need to be updated.


blegat (approved):
Does this closes https://github.com/jump-dev/MathOptInterface.jl/issues/433 ?

odow:
I think it resolves the general issue and adds tests. I'm pretty happy with the convention of removing objective constants. 

I don't understand the question about constants in `ConstraintPrimal` or `ObjectiveValue`.

blegat:
This is for unbounded rays, in that case, the equivalent of "removing objective coefficients" is "remove constraints constants".

odow:
Does it make sense to query `ConstraintPrimal` and `ObjectiveValue` for a ray though? Isn't it really just `VariablePrimal`?

blegat:
Can't you query `DualObjectiveValue` for infeasibility ray ? I need it to create infeasibility cut in SDDP. There is a `get_fallback` for ObjectiveValue and ConstraintPrimal if the solver does not provide that.

odow:
Ah. `DualObjectiveValue` is a reasonable thing to query. I guess the decision is whether to query the "beta" term here, or whether to query the complete RHS.

image

blegat:
If you have the problem
c1: x + y = 1
c2: x >= 1
c3: y >= 1
The dual is
max y1 + y2 + y3
y1 + y2 == 0
y1 + y3 == 0
y2, y3 >= 0
You are wondering whether we should include y2 + y3 in the objective ?
An infeasibility ray is y = (-1, 1, 1). If we only include y1, we have -1 and with the full objective, we have 1. The correct answer is 1 IMO

odow:
Yes, the `DualObjectiveValue` should be 1.

Let's merge this for now. It adds tests that the solvers already support. I'll need to go through and fix `DualObjectiveValue` in the solvers before we can add tests in MOI, so that can be done in a separate PR.

PR #1191, opened 2020-11-06
tests: add unit test for add_constraint 

shadiakiki1986:
Tests for input nan, for lower-bound already set, and for upper-bound already set. This was prompted by PR 230 in Ipopt ([link](https://github.com/jump-dev/Ipopt.jl/pull/230/files)):  Fix error in add_constraint NaN handling for EqualTo

I figured I'd add a unit test for the checks here, and they'd be useful for other solvers as well. I checked ECOS and SCS, but they don't implement `add_constraint`. OTOH, GLPK.jl does implement it ([ref](https://github.com/jump-dev/GLPK.jl/blob/master/src/MOI_wrapper/MOI_wrapper.jl#L736)), but it needs this new test to be added to the unit test array ([ref](https://github.com/jump-dev/GLPK.jl/blob/master/test/MOI_wrapper.jl#L13)).

### DIFF
--- a/src/Test/UnitTests/constraints.jl
+++ b/src/Test/UnitTests/constraints.jl
@@ -361,3 +361,33 @@ function solve_zero_one_with_bounds_3(model::MOI.ModelLike, config::TestConfig)
     end
 end
 unittests["solve_zero_one_with_bounds_3"] = solve_zero_one_with_bounds_3
+
+# Taken from https://github.com/jump-dev/Ipopt.jl/pull/230/files#
+"""
+    add_constraint_nan(model::MOI.ModelLike, config::TestConfig)
+
+Add a NaN constraint. This should throw a proper error message.
+"""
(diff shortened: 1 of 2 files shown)


odow:
So we already check for finiteness at the JuMP level: https://github.com/jump-dev/JuMP.jl/pull/2272

I don't know if we want to make every solver check every coefficient for a `NaN`. That seems like a lot of overhead.

odow:
Closing since there are already tests for `LowerBoundAlreadySet`: https://github.com/jump-dev/MathOptInterface.jl/blob/162049e15415b47a40c29d9a0aa9d6acf58a2d72/src/Test/modellike.jl#L711-L734
and I don't think we need all solvers checking for `isnan` on every element. JuMP already does this, and either the lower-level solver should do it, or the person choosing to bypass JuMP and use MOI should do it.

PR #1197, opened 2020-11-11
fix Grisu usage on upcoming 1.6 and fix a few printing tests

KristofferC:
This fixes the usage of Base.Grisu on the upcoming 1.6 where this is removed.

Also, it fixes a few tests that test different printing.
Tests written like this are quite sensitive to small printing tweaks made to e.g. types in Julia.
In the tests I fixed, I tried to show that it is better to interpolate the whole type
in the string so that Julia will use its own printing for it instead of manually copy pasting
the type printing for a single Julia version.

After fixing one file, there was a lot of similar test errors in other files but at least
this gets it a bit on the way.


src/FileFormats/LP/LP.jl
```diff
@@ -4,6 +4,12 @@ import ..FileFormats
 import MathOptInterface
 const MOI = MathOptInterface
 
+if isdefined(Base, :Grisu)
+    import Base.Grisu
+else
+    import Grisu
+end
```
odow, line 11: What's the advantage of using the `Base` version over just this package?

odow:
> there was a lot of similar test errors in other files but at least
this gets it a bit on the way.

Thanks! I'll take it from here.

odow:
Tests will fail due to some issues in MutableArithmetics.

src/FileFormats/LP/LP.jl
KristofferC: It avoids having to load the package, which is a bit faster. I think it loads pretty quickly though so perhaps doesn't matter.

KristofferC:
Another option instead of using Grisu is to transition to the new Rye floating point printer that is in Julia Base:

https://github.com/JuliaLang/julia/blob/93fb785831dcfcc442f82fab8746f0244c5274ae/base/ryu/Ryu.jl#L19-L51

Of course, even better would be if it is possible to do this whole thing without using Julia internals. I haven't looked into it enough to know if such high level of control is needed that calling e.g. the internal Grisu is a requirement.

odow:
It seems like we used Grisu in due to some issues with Julia printing floats via `@printf` in 2014, and it stayed around through various refactoring: https://github.com/jump-dev/JuMP.jl/issues/102, https://github.com/jump-dev/JuMP.jl/pull/191.  `Base.print` seems to work fine for floats now.

odow:
Tests passing on 1.6.

KristofferC:
Would be nice with a new release with this whenever possible.

blegat:
Let's do it: https://github.com/jump-dev/MathOptInterface.jl/pull/1204

PR #1202, opened 2020-11-29
Add support for conflicts in MockOptimizer

dourouc05:


odow:
Can we test this? I don't remember if there are already tests for conflict status in MOI.Test, if not, we should add some.

dourouc05:
You mean, something like https://github.com/jump-dev/MathOptInterface.jl/blob/master/src/Test/UnitTests/solve.jl#L63, with an infeasible model, `test_model_solution` being replaced by a call to `compute_conflict!` and tests of some attributes? In that case, that might go into another PR, as it's not really related to MockOptimizer.

dourouc05:
I just added a direct test for the new functionality.

blegat:
I understand the need to define `compute_conflict!` but for other attributes, you can use a `UniversalFallback` as inner model.

dourouc05:
Actually, I tried to set attributes like `MOI.ConflictStatus`, but I had messages saying the operation is unsupported. Adding the `get`/`set` methods solved the issue.

odow:
Merging this because I need it to fix https://github.com/jump-dev/MathOptInterface.jl/issues/1203

PR #1205, opened 2020-12-04
Use JuliaFormatter to format /src/FileFormats

odow:
A quick scan suggests no real issues. The `/src/Utilities/` diff is much scarier (longer).

### DIFF
--- a/src/FileFormats/FileFormats.jl
+++ b/src/FileFormats/FileFormats.jl
@@ -51,11 +51,10 @@ The `filename` argument is only needed if `format == FORMAT_AUTOMATIC`.
 
 `kwargs` are passed to the underlying model constructor.
 """
-function Model(
-    ;
+function Model(;
     format::FileFormat = FORMAT_AUTOMATIC,
-    filename::Union{Nothing, String} = nothing,
-    kwargs...
+    filename::Union{Nothing,String} = nothing,
+    kwargs...,
 )
     if format == FORMAT_CBF
         return CBF.Model(; kwargs...)
@@ -78,7 +77,7 @@ function Model(
             (".mof.json", MOF.Model),
             (".mps", MPS.Model),
             (".dat-s", SDPA.Model),
-            (".sdpa", SDPA.Model)
+            (".sdpa", SDPA.Model),
         ]
             if endswith(filename, ext) || occursin("$(ext).", filename)
                 return model(; kwargs...)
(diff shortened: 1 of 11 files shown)


matbesancon (approved):
Once this is an established practice, it could be interesting to add a reference to the use of JuliaFormatter in the documentation, and possibly a bot for PR checks

odow:
Yes. This is going to be standard practice, but we need to fix a looot of code first. Some I'm just doing it piece by piece.

src/FileFormats/FileFormats.jl
```diff
@@ -88,23 +87,18 @@ function Model(
...
-    LP.Model,
-    MOF.Model,
-    MPS.Model,
-    SDPA.Model,
-}
+const MATH_OPT_FORMATS =
+    Union{CBF.Model,LP.Model,MOF.Model,MPS.Model,SDPA.Model}
 
 function MOI.write_to_file(model::MATH_OPT_FORMATS, filename::String)
     compressed_open(filename, "w", AutomaticCompression()) do io
```
matbesancon, line 94: `return` in `do` blocks is a bit weird, is there a way to avoid them?

src/FileFormats/FileFormats.jl
```diff
@@ -51,11 +51,10 @@ The `filename` argument is only needed if `format == FORMAT_AUTOMATIC`.
...
 `kwargs` are passed to the underlying model constructor.
 """
-function Model(
-    ;
+function Model(;
     format::FileFormat = FORMAT_AUTOMATIC,
-    filename::Union{Nothing, String} = nothing,
-    kwargs...
+    filename::Union{Nothing,String} = nothing,
+    kwargs...,
```
blegat, line 57: It's a bit weird to add a comma here, is that expected or a bug in the formatter?

blegat: It seems to be used when it is multiline, it makes sense

odow: This is expected.

src/FileFormats/FileFormats.jl
odow: `do` is a function block, so there should be a `return` here. It clarifies that you can go
```Julia
x = foo() do arg
    return arg + 1
end
```
and `x` will have the value `arg + 1`.

blegat (approved):

PR #1212, opened 2020-12-11
Use JuliaFormatter to format /src/Bridges

odow:
Breaking these down into (semi-)manageable PRs, instead of one gigantic one that touches every line in MOI.

### DIFF
--- a/src/Bridges/Constraint/Constraint.jl
+++ b/src/Bridges/Constraint/Constraint.jl
@@ -23,7 +23,7 @@ include("single_bridge_optimizer.jl")
 # method seemed necessary to fix it.
 # See https://github.com/JuliaLang/julia/issues/32167 for more.
 function MOI.Bridges.Constraint.bridge_constraint(BridgeType, b, f, s)
-    throw(MOI.UnsupportedConstraint{typeof(f), typeof(s)}())
+    return throw(MOI.UnsupportedConstraint{typeof(f),typeof(s)}())
 end
 
 # Constraint bridges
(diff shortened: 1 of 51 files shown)


src/Bridges/Constraint/Constraint.jl
```diff
@@ -23,7 +23,7 @@ include("single_bridge_optimizer.jl")
 # method seemed necessary to fix it.
 # See https://github.com/JuliaLang/julia/issues/32167 for more.
 function MOI.Bridges.Constraint.bridge_constraint(BridgeType, b, f, s)
-    throw(MOI.UnsupportedConstraint{typeof(f), typeof(s)}())
+    return throw(MOI.UnsupportedConstraint{typeof(f),typeof(s)}())
```
blegat, line 26: This return is a bit weird but I guess it's difficult for the parser to distinguish it from other cases.

odow: The underlying issue is that we know by convention that `throw` will throw an error, and so the function will never return. 

But syntactically, `MOI.UnsupportedConstraint` could return any type, and if we overloaded `throw` for that new type, it might not throw an error :(

odow: Same goes for things like `error`. As manual fixes, we could add unreachable `return` statements, or we can just leave the `return error()`:
```Julia
function foo()
    error("a")
    return
end
```
As an automated approach, I think adding `return` is the correct thing to do. If you're working on a function and see a nearby `return throw` that needs fixing, you could do it then, but it's pretty low priority.

blegat: Yes, the unreachable return looks better but let's do that manually

blegat (approved):

PR #1215, opened 2020-12-13
[Utilities] Fix modifying an un-set objective sets the objective

### LINKED ISSUE #1214: Programmatically defined functions work but the same programmatically defined objective does not
Below is some runnable code with two methods. The expected behavior is for the methods to return the same result, but the second method does not use the objective function when it optimizes and thus returns an incorrect solution:

```julia
using JuMP
using GLPK
using MathOptInterface

function juliaWorking()
    model = Model(GLPK.Optimizer)
    @variable(model, v[1:10])
    for i in 1:10
    	set_lower_bound(v[i], 0)
    	set_upper_bound(v[i], 1)
    end
    @constraint(model, v[2] == .5)
    @constraint(model, v[1]<= .5)


    ex = AffExpr(0)
    for id in 1:10
        add_to_expression!(ex, 1.0, v[id])
    end
    @objective(model,MOI.MAX_SENSE,ex)
    optimize!(model)
    for i in 1:10
    	println("v[$i] => ", value.(v[i]))
    end
end

function juliaNotWorking()
    model = Model(GLPK.Optimizer)
    @variable(model, v[1:10])
    for i in 1:10
    	set_lower_bound(v[i], 0)
    	set_upper_bound(v[i], 1)
    end
    @constraint(model, v[2] == .5)
    @constraint(model, v[1]<= .5)


    for id in 1:10
        set_objective_coefficient(model, v[id], 1.0)
    end
    set_objective_sense(model, MOI.MAX_SENSE)
    optimize!(model)
	for i in 1:10
		println("v[$i] => ", value.(v[i]))
	end
end
```
juliaWorking() outputs
 ```
v[1] => 0.5
v[2] => 0.5
v[3] => 1.0
v[4] => 1.0
v[5] => 1.0
v[6] => 1.0
v[7] => 1.0
v[8] => 1.0
v[9] => 1.0
v[10] => 1.0
```

juliaNotWorking() outputs
```
v[1] => 0.0
v[2] => 0.5
v[3] => 0.0
v[4] => 0.0
v[5] => 0.0
v[6] => 0.0
v[7] => 0.0
v[8] => 0.0
v[9] => 0.0
v[10] => 0.0
```

But if you print the model, both methods print the same (correct) thing
```
Max v[1] + v[2] + v[3] + v[4] + v[5] + v[6] + v[7] + v[8] + v[9] + v[10]
Subject to
 v[2] = 0.5
 v[1] ≤ 0.5
 v[1] ≥ 0.0
 v[2] ≥ 0.0
 v[3] ≥ 0.0
 v[4] ≥ 0.0
 v[5] ≥ 0.0
 v[6] ≥ 0.0
 v[7] ≥ 0.0
 v[8] ≥ 0.0
 v[9] ≥ 0.0
 v[10] ≥ 0.0
 v[1] ≤ 1.0
 v[2] ≤ 1.0
 v[3] ≤ 1.0
 v[4] ≤ 1.0
 v[5] ≤ 1.0
 v[6] ≤ 1.0
 v[7] ≤ 1.0
 v[8] ≤ 1.0
 v[9] ≤ 1.0
 v[10] ≤ 1.0
```

The problem arises when needing to construct constraints and objectives programmatically. The first method returns the correct answer, the second does not. It is clear from the difference that juliaNotWorking() respects the constraints but not the objective function. juliaWorking() builds an AffineExpression and adds each variable programatically. juliaNotWorking() individually changes the objective coefficient of each variable and sets the sense of the objective function.

It is possible I am misunderstanding the purpose of set_objective_coefficient, but I can't think what the purpose is supposed to be if it updates the model, but is ignored by the solver. Same goes for set_objective_sense, what's the point if it changes the model but is ignored by the solver?

If I am wrong, I am very curious what the purpose of set_objective_coefficient and set_objective_sense is, and I'd throw out there that the documentation here https://jump.dev/JuMP.jl/stable/objective/#JuMP.set_objective_sense is very confusing if those methods are working as intended and are meant to affect the model but not the solver.

### PR DESCRIPTION
**Pull request by odow**
Closes #1214 

See the issue for background, but we have a decision.

Should `set_objective_coefficient` in the following:
```Julia
model = Model()
@variable(model, x)
set_objective_coefficient(model, x, 1.0)
```

1. Throw an error along the likes of "Unable to modify objective because no objective is set."
2. Assume an empty `ScalarAffineFunction` objective exists and modify the linear term?

Currently, this PR does 2, but I would also support the first option.

### DIFF
--- a/src/Utilities/model.jl
+++ b/src/Utilities/model.jl
@@ -510,10 +510,12 @@ end
 
 function MOI.modify(
     model::AbstractModel,
-    obj::MOI.ObjectiveFunction,
+    ::MOI.ObjectiveFunction,
     change::MOI.AbstractFunctionModification,
 )
-    return model.objective = modify_function(model.objective, change)
+    model.objective = modify_function(model.objective, change)
+    model.objectiveset = true
+    return
 end
 
 function MOI.get(::AbstractModel, ::MOI.ListOfOptimizerAttributesSet)
--- a/test/Utilities/model.jl
+++ b/test/Utilities/model.jl
@@ -331,3 +331,11 @@ end
     @test !haskey(dest.ext, :my_store)
     @test model.ext[:my_store] == 2
 end
+
+@testset "Objective set via modify" begin
+    model = MOIU.Model{Float64}()
+    x = MOI.add_variable(model)
+    attr = MOI.ObjectiveFunction{MOI.ScalarAffineFunction{Float64}}()
+    MOI.modify(model, attr, MOI.ScalarCoefficientChange(x, 1.0))
+    @test attr in MOI.get(model, MOI.ListOfModelAttributesSet())
+end


joehuchette:
I think behavior 2 makes the most intuitive sense.

blegat:
2. works here as we allow the sense to be FEASIBILITY_SENSE even if the objective function is set. However, if we go in the direction of objective that you defined with set and function given simultaneously as you suggested for multi-objective then 1. would be preferable.

odow:
Seems like we can stick with (2) for now.

matbesancon:
should this be merged? (maybe after checking that CI still passes?)

matbesancon:
Looks good here

PR #1219, opened 2020-12-15
Enable doctests and correct previously untested errors

odow:
Required for JuMP: https://github.com/jump-dev/JuMP.jl/pull/2393/files#r535004411. Should have been done anyway.


blegat:
We could add coverage from docs build so that we see a codecov drop if there is an issue with doctests
https://juliadocs.github.io/Documenter.jl/dev/man/hosting/#Add-code-coverage-from-documentation-builds

odow:
Adding coverage to the doctests seemed to lead to weird errors, like it picking up part of the docstring as code. I removed for now. We shouldn't rely on doctests to boost code coverage.

blegat (approved):

PR #1226, opened 2021-01-12
Use T <: Number for GreaterThan and LessThan

dourouc05:
This is required for constraint programming, as working with integers is the norm. This brings these two sets closer to `EqualTo`, which already uses `Number`.


odow:
Here was the original reasoning: https://github.com/jump-dev/MathOptInterface.jl/pull/229

Can't you have `LessThan{Int}` already since `Int <: Real`?

dourouc05:
Yes, you are right. I was mistaken by the error messages…

PR #1229, opened 2021-02-02
[Utilities] Drop MOI.RawParameters when resetting a CachingOptimizer

### LINKED ISSUE #1220: reset_optimizer attempts to pass RawParameters between models
When changing the optimizer, `CachingOptimizer` attempts to set `MOI.RawParameters`:
https://github.com/jump-dev/MathOptInterface.jl/blob/1937e27ae20c2c200468ab0a36fd06d3c29687cd/src/Utilities/cachingoptimizer.jl#L123-L127
We should probably just skip any `RawParameter`s, or not include them in the list of optimizer parameters set.

```Julia
julia> using JuMP, Clp, Cbc
[ Info: Precompiling JuMP [4076af6c-e467-56ae-b986-b466b2749572]

julia> model = Model(Cbc.Optimizer)
A JuMP Model
Feasibility problem with:
Variables: 0
Model mode: AUTOMATIC
Solver name: COIN Branch-and-Cut (Cbc)

julia> set_optimizer_attribute(model, "ratioGap", 0.001)
0.001

julia> set_optimizer(model, Clp.Optimizer)
ERROR: MathOptInterface.UnsupportedAttribute{MathOptInterface.RawParameter}: Attribute MathOptInterface.RawParameter("ratioGap") is not supported by the model.
Stacktrace:
 [1] set(::Clp.Optimizer, ::MathOptInterface.RawParameter, ::Float64) at /Users/oscar/.julia/packages/Clp/NUuKQ/src/MOI_wrapper/MOI_wrapper.jl:147
 [2] set(::MathOptInterface.Utilities.CachingOptimizer{Clp.Optimizer,MathOptInterface.Utilities.UniversalFallback{MathOptInterface.Utilities.Model{Float64}}}, ::MathOptInterface.RawParameter, ::Float64) at /Users/oscar/.julia/packages/MathOptInterface/ZJFKw/src/Utilities/cachingoptimizer.jl:665
 [3] set(::MathOptInterface.Bridges.LazyBridgeOptimizer{MathOptInterface.Utilities.CachingOptimizer{Clp.Optimizer,MathOptInterface.Utilities.UniversalFallback{MathOptInterface.Utilities.Model{Float64}}}}, ::MathOptInterface.RawParameter, ::Float64) at /Users/oscar/.julia/packages/MathOptInterface/ZJFKw/src/Bridges/bridge_optimizer.jl:621
 [4] reset_optimizer(::MathOptInterface.Utilities.CachingOptimizer{MathOptInterface.AbstractOptimizer,MathOptInterface.Utilities.UniversalFallback{MathOptInterface.Utilities.Model{Float64}}}, ::MathOptInterface.Bridges.LazyBridgeOptimizer{MathOptInterface.Utilities.CachingOptimizer{Clp.Optimizer,MathOptInterface.Utilities.UniversalFallback{MathOptInterface.Utilities.Model{Float64}}}}) at /Users/oscar/.julia/packages/MathOptInterface/ZJFKw/src/Utilities/cachingoptimizer.jl:105
 [5] reset_optimizer(::Model, ::MathOptInterface.Bridges.LazyBridgeOptimizer{MathOptInterface.Utilities.CachingOptimizer{Clp.Optimizer,MathOptInterface.Utilities.UniversalFallback{MathOptInterface.Utilities.Model{Float64}}}}, ::Bool) at /Users/oscar/.julia/packages/JuMP/qhoVb/src/optimizer_interface.jl:17
 [6] reset_optimizer(::Model, ::MathOptInterface.Bridges.LazyBridgeOptimizer{MathOptInterface.Utilities.CachingOptimizer{Clp.Optimizer,MathOptInterface.Utilities.UniversalFallback{MathOptInterface.Utilities.Model{Float64}}}}) at /Users/oscar/.julia/packages/JuMP/qhoVb/src/optimizer_interface.jl:16
 [7] set_optimizer(::Model, ::Type{T} where T; bridge_constraints::Bool) at /Users/oscar/.julia/packages/JuMP/qhoVb/src/optimizer_interface.jl:73
 [8] set_optimizer(::Model, ::Type{T} where T) at /Users/oscar/.julia/packages/JuMP/qhoVb/src/optimizer_interface.jl:61
 [9] top-level scope at REPL[6]:1
```

x-ref: https://discourse.julialang.org/t/solving-fixed-milp-problem/51749

odow:
Closes #1220.

@blegat: An alternative is to not pass any OptimizerAttributes. If you're passing a new empty optimizer, it seems weird that the parameters would follow.

What do we expect to happen in the following at the JuMP level?
```Julia
model = Model(Cbc.Optimizer)
set_silent(model)
set_optimizer(model, Clp.Optimizer)
MOI.get(model, MOI.Silent())  # true or false?
```

### DIFF
--- a/src/Utilities/cachingoptimizer.jl
+++ b/src/Utilities/cachingoptimizer.jl
@@ -113,14 +113,23 @@ mode(m::CachingOptimizer) = m.mode
 """
     reset_optimizer(m::CachingOptimizer, optimizer::MOI.AbstractOptimizer)
 
-Sets or resets `m` to have the given empty optimizer. Can be called
-from any state. The `CachingOptimizer` will be in state `EMPTY_OPTIMIZER` after the call.
+Sets or resets `m` to have the given empty optimizer `optimizer`.
+
+Can be called from any state. An assertion error will be thrown if `optimizer`
+is not empty.
+
+The `CachingOptimizer` `m` will be in state `EMPTY_OPTIMIZER` after the call.
 """
 function reset_optimizer(m::CachingOptimizer, optimizer::MOI.AbstractOptimizer)
     @assert MOI.is_empty(optimizer)
     m.optimizer = optimizer
     m.state = EMPTY_OPTIMIZER
     for attr in MOI.get(m.model_cache, MOI.ListOfOptimizerAttributesSet())
+        if attr isa MOI.RawParameter
+            # Skip RawParameters because they may not apply to the new
+            # optimizer.
+            continue
+        end
         value = MOI.get(m.model_cache, attr)
         optimizer_value = map_indices(m.model_to_optimizer_map, value)
         MOI.set(m.optimizer, attr, optimizer_value)
--- a/src/Utilities/cachingoptimizer.jl
+++ b/src/Utilities/cachingoptimizer.jl
@@ -127,6 +127,10 @@ function reset_optimizer(m::CachingOptimizer, optimizer::MOI.AbstractOptimizer)
     for attr in MOI.get(m.model_cache, MOI.ListOfOptimizerAttributesSet())
         # Skip attributes which don't apply to the new optimizer.
         if attr isa MOI.RawParameter
+            # Even if the optimizer claims to `supports` `attr`, the value 
+            # might have a different meaning (e.g., two solvers with `logLevel`
+            # as a RawParameter). To be on the safe side, just skip all raw 
+            # parameters.
             continue
         elseif !MOI.is_copyable(attr) || !MOI.supports(m.optimizer, attr)
             continue
--- a/src/Utilities/cachingoptimizer.jl
+++ b/src/Utilities/cachingoptimizer.jl
@@ -125,9 +125,10 @@ function reset_optimizer(m::CachingOptimizer, optimizer::MOI.AbstractOptimizer)
     m.optimizer = optimizer
     m.state = EMPTY_OPTIMIZER
     for attr in MOI.get(m.model_cache, MOI.ListOfOptimizerAttributesSet())
+        # Skip attributes which don't apply to the new optimizer.
         if attr isa MOI.RawParameter
-            # Skip RawParameters because they may not apply to the new
-            # optimizer.
+            continue
+        elseif !MOI.is_copyable(attr) || !MOI.supports(m.optimizer, attr)
             continue
         end
         value = MOI.get(m.model_cache, attr)
--- a/test/Utilities/cachingoptimizer.jl
+++ b/test/Utilities/cachingoptimizer.jl
@@ -662,6 +662,7 @@ function MOI.get(model::Issue1220, ::MOI.ListOfOptimizerAttributesSet)
     return collect(keys(model.optimizer_attributes))
 end
 MOI.supports(::Issue1220, ::MOI.AbstractOptimizerAttribute) = true
+MOI.supports(::Issue1220, ::MOI.NumberOfThreads) = false
 function MOI.get(model::Issue1220, attr::MOI.AbstractOptimizerAttribute)
     return model.optimizer_attributes[attr]
 end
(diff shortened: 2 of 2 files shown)


blegat:
This is currently a mess. We have a documented option:
https://github.com/jump-dev/MathOptInterface.jl/blob/f71ed8314da336a90f5713e2406ca4007c8c02e7/src/MathOptInterface.jl#L110-L114
which is not implemented as optimizer attributes are in fact not passed in `copy_to` but rather passed by the `CachingOptimizer` outside the `copy_to` function.
Note that we do have a warning for starting values though:
https://github.com/jump-dev/MathOptInterface.jl/blob/f71ed8314da336a90f5713e2406ca4007c8c02e7/src/Utilities/copy.jl#L284-L291
I would probably add this warning option to `CachingOptimizer` instead of `copy_to`. Then you check `supports` before calling `set` and throws a warning if it's not supported and the user did not give the option to ignore this.

So
```julia
model = Model(Cbc.Optimizer)
set_silent(model)
set_optimizer(model, Clp.Optimizer)
MOI.get(model, MOI.Silent())
```
returns `true` if `Clp` supports `MOI.Silent` and throws a warning otherwise.
```julia
model = Model(Cbc.Optimizer, warn_option_name=false) # do we give the option here ?
set_silent(model)
set_optimizer(model, Clp.Optimizer, warn_option_name=false) # or here ?
MOI.get(model, MOI.Silent())
```
returns `true` if `Clp` supports `MOI.Silent` and ignores it silently otherwise.

One issue with this is that many solvers simply do `MOI.supports(::Optimizer, ::MOI.RawParameter) = true` without checking the parameter string but I'm in favor of fixing the solvers `supports` function.

odow:
Passing `RawParameter` is also fraught though, even if it supports the parameter name. Because one solver might have `print_level` that takes an integer, and another might have `print_level` that takes a string or some solver-constant.

joaquimg:
I know it would be very breaking, but what if RawParameter takes name and value?

blegat:
We could do the following:
* If a `RawParameter` is set when no optimizer is attached: throw a warning or an error
* If a `RawParameter` is set when an optimizer is attached: set it to the optimizer but not to the cache.

odow:
We have 
https://github.com/jump-dev/MathOptInterface.jl/blob/1315e317da1d59c2cb172145a8699b3e0a6135b8/src/Utilities/cachingoptimizer.jl#L646-L657
which we haven't really utilized yet.

odow:
Discussed on the developer call. Consensus was that we should only copy `MOI.Silent` and `MOI.TimeLimitSec`, if they are set.

src/Utilities/cachingoptimizer.jl
```diff
@@ -113,14 +113,24 @@ mode(m::CachingOptimizer) = m.mode
...
+
+The `CachingOptimizer` `m` will be in state `EMPTY_OPTIMIZER` after the call.
 """
 function reset_optimizer(m::CachingOptimizer, optimizer::MOI.AbstractOptimizer)
     @assert MOI.is_empty(optimizer)
     m.optimizer = optimizer
     m.state = EMPTY_OPTIMIZER
     for attr in MOI.get(m.model_cache, MOI.ListOfOptimizerAttributesSet())
+        # Skip attributes which don't apply to the new optimizer.
+        if attr isa MOI.RawParameter
```
blegat, line 129: So if I undersand correctly, this is a workaround for solvers not implementing `supports` correctly for RawParameter? Otherwise, this would we caughr by the second condition

blegat: Or is it because RawParameter with the same name might have different meaning?

blegat: This should be clarified in a comment

odow: > because RawParameter with the same name might have different meaning?

This.

blegat (approved):

PR #1235, opened 2021-02-10
Add test with zero in off-diagonal entry

blegat:
This test would have caught the OSQP issue fixed by https://github.com/oxfordcontrol/OSQP.jl/pull/90

- [x] Add comments
- [x] Add tests

### DIFF
--- a/src/Test/UnitTests/objectives.jl
+++ b/src/Test/UnitTests/objectives.jl
@@ -314,6 +314,13 @@ function solve_qp_edge_cases(model::MOI.ModelLike, config::TestConfig)
 end
 unittests["solve_qp_edge_cases"] = solve_qp_edge_cases
 
+"""
+    solve_qp_zero_offdiag(model::MOI.ModelLike, config::TestConfig)
+
+Test quadratic program with a zero off-diagonal term.
+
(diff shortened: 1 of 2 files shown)


src/Test/UnitTests/objectives.jl
```diff
@@ -314,6 +314,50 @@ function solve_qp_edge_cases(model::MOI.ModelLike, config::TestConfig)
...
+            0.0,  # constant
+        ),
+    )
+    test_model_solution(
+        model,
+        config;
+        objective_value = 5.0,
+        variable_primal = [(x[1], 1.0), (x[2], 2.0)],
+    )
+end
```
odow, line 359: Needs
```
unittests["solve_qp_zero_offdiag"] = solve_qp_zero_offdiag
```

This could go in `solve_qp_edge_cases` so it doesn't break the tests of every solver who doesn't support QP? Or add a check for `supports`?

blegat: I tried adding it to `edge_case` but the bug in OSQP only occurs at `copy_to`, not if you do `MOI.set` so it didn't break OSQP.

PR #1237, opened 2021-02-14
[Utilities] add supports_shift_constant

odow:
In order to close https://github.com/jump-dev/JuMP.jl/issues/1943, we need a way of checking whether it is possible to push the function constant into the set. The easiest way is just to define this fallback, so the JuMP code will become:
```Julia
function build_constraint(
    ::Function,
    expr::Union{GenericAffExpr, GenericQuadExpr},
    set::MOI.AbstractScalarSet,
)
    offset = constant(expr)
    new_set = MOIU.shift_constant(set, -offset)
    if new_set === nothing
        return ScalarConstraint(expr, set)
    else
        add_to_expression!(expr, -offset)
        return ScalarConstraint(expr, new_set)
    end
end
```

This makes JuMP's heuristic of "I'll push the constant into the set" well-defined by "does it implement shift_constant".

As a follow-up question: should this be moved to MOI proper?

### DIFF
--- a/docs/src/apireference.md
+++ b/docs/src/apireference.md
@@ -1073,6 +1073,7 @@ The following utilities are available for moving the function constant to the
 set for scalar constraints:
 ```@docs
 Utilities.shift_constant
+Utilities.supports_shift_constant
 Utilities.normalize_and_add_constraint
 Utilities.normalize_constant
 ```
--- a/src/Utilities/sets.jl
+++ b/src/Utilities/sets.jl
@@ -1,15 +1,32 @@
 """
-    shift_constant(set::MOI.AbstractScalarSet,
-                   offset)
+    shift_constant(set::MOI.AbstractScalarSet, offset)
 
 Returns a new scalar set `new_set` such that `func`-in-`set` is equivalent to
 `func + offset`-in-`new_set`.
+
+Only define this function if it makes sense to!
+
+By default, this function returns `nothing`, which can be used as a flag to
+check if the set supports shifting:
+```Julia
+new_set = shift_constant(old_set, offset)
+if new_set === nothing
+    add_constraint(model, f, old_set)
+else
+    f.constant = 0
+    add_constraint(model, f, new_set)
+end
+```
 
 ## Examples
 
 The call `shift_constant(MOI.Interval(-2, 3), 1)` is equal to
 `MOI.Interval(-1, 4)`.
 """
+function shift_constant(::MOI.AbstractScalarSet, ::Any)
+    return nothing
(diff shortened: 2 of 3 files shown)


blegat:
I'm hesitant between this has adding a `supports_constant_shift(::Type{<:MOI.AbstractScalarSet}) = false`. The advantage of this PR is that it's not possible to implement `shift_constant` but forget to implement `supports_constant_shift`. The advantage of `supports_constant_shift` is that it can be checked when we only have the type of the set. A disadvantage of this PR is that if the user pass a constant of incorrect type, it might use the fallback even if there is a shift implemented.

odow:
> it's not possible to implement shift_constant but forget to implement supports_constant_shift.

I always get confused with out `supports_xxx`. 

> A disadvantage of this PR is that if the user pass a constant of incorrect type, it might use the fallback even if there is a shift implemented.

If they pass the incorrect type, then the shift implemented will not work anyway.

For example: `shift_constant(MOI.EqualTo{Int}(1), 0.5)`. We shouldn't just check that `MOI.EqualTo{Int}` supports the shift.

blegat:
> If they pass the incorrect type, then the shift implemented will not work anyway.

Yes but `shift_constant(MOI.EqualTo(1), 2.5)` should probably error or return `MOI.EqualTo(3.5)`, not `nothing`

odow:
Returning `nothing` says that it isn't okay to shift the constant into the set. Otherwise we need to define `supports_shift_constant(::Type{Set}, ::Type{T})`.

And `EqualTo` could define to implement `shift_contant(set::MOI.EqualTo, shift::T) = MOI.EqualTo(set.value + shift)` if it so chose.

It really, do you want to implement an extra `supports_shift_constant` method, or do you want a `nothing` fallback?

blegat:
Some user might also not see the error when `nothing` is returned, for instance if they store the result in a `Any[]`. Also, we might need to check whether the set can be shifted only with the set type (e.g. in `concrete_bridge_type` of a bridge).

odow:
> Some user might also not see the error when nothing is returned, for instance if they store the result in a Any[]

This is user-error.

> Also, we might need to check whether the set can be shifted only with the set type

We haven't needed to yet?

blegat:
> This is user-error.

Why do you mean ? The user should never have `Any[]` vectors ?

odow:
> Why do you mean?

If the user doesn't check for a return of `nothing`, they're using the function wrong. Or you mean this is breaking because currently it will error?

blegat (approved):

PR #1239, opened 2021-02-16
Benchmark last in test

matbesancon:
Alphabetical order is nice and tidy, but running benchmarks before running actual tests does not make sense, we would want to test first for correctness before running benchmarks for early termination.


odow (approved):

matbesancon:
I was tempted and push it to add an ENV variable to activate benchmarks but we can keep it for another day

PR #1244, opened 2021-02-19
Skip unit tests when problem not supported

blegat:
The contquadratic, contconic, etc... are sorted according to the problem class so it makes sense to fail the tests when a class is not supported.
In the case of unit tests, we assume that `supports...` are correctly implemented so it should be ok to rely on it to skip tests.


odow (approved):
Re-run the SolverTest CI job after this?

PR #1245, opened 2021-02-19
Refactor Model and UniversalFallback

blegat:
Closes https://github.com/jump-dev/MathOptInterface.jl/issues/1234

### DIFF
--- a/src/Utilities/CleverDicts.jl
+++ b/src/Utilities/CleverDicts.jl
@@ -62,22 +62,6 @@ mutable struct CleverDict{K,V,F<:Function,I<:Function} <: AbstractDict{K,V}
     set::BitSet
     vector::Vector{V}
     dict::OrderedCollections.OrderedDict{K,V}
-    function CleverDict{K,V}(n::Integer = 0) where {K,V}
-        set = BitSet()
-        sizehint!(set, n)
-        vec = Vector{K}(undef, n)
-        inverse_hash = x -> index_to_key(K, x)
-        hash = key_to_index
-        return new{K,V,typeof(hash),typeof(inverse_hash)}(
-            0,
-            hash,
-            inverse_hash,
-            true,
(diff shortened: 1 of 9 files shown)


src/Utilities/CleverDicts.jl
```diff
@@ -62,22 +62,6 @@ mutable struct CleverDict{K,V,F<:Function,I<:Function} <: AbstractDict{K,V}
     set::BitSet
     vector::Vector{V}
     dict::OrderedCollections.OrderedDict{K,V}
-    function CleverDict{K,V}(n::Integer = 0) where {K,V}
-        set = BitSet()
-        sizehint!(set, n)
-        vec = Vector{K}(undef, n)
-        inverse_hash = x -> index_to_key(K, x)
```
blegat, line 69: We need the type of `inverse_hash` to be easy to get to have the concrete type of the field of `VectorOfConstraints`. Here, you cannot get it with `typeof(some_function)` if the function is parametrized by `K`. Maybe we could do `typeof(Fix1(index_to_key, K))` if we use `Fix1` here.

odow:
So this is actually slower than before, when run on the JuMP examples (The `stress_precompile` function here: https://github.com/jump-dev/JuMP.jl/pull/2484).

Before:
```julia
 78.339035 seconds (204.57 M allocations: 11.833 GiB, 5.22% gc time, 6.89% compilation time)
```
After:
```julia
88.580525 seconds (220.25 M allocations: 12.709 GiB, 5.61% gc time, 14.41% compilation time)
```

However, it is more readable. I'd be in favor of merging so we can make other changes in future PRs. There are likely some performance things we can tweak.

blegat:
Agreed, I think there is an issue with `Fix1` and `CleverDict` as we have `Fix1{DataType, ...}` which does not play well with inference.

PR #1252, opened 2021-02-26
Add auto_bridge to CachingOptimizer

### LINKED ISSUE #1156: Change caching copy logic to allow incremental solvers to use a copy_to function
Currently, solvers such as GLPK skip `copy_to` via JuMP, due to the behavior of `CachingOptimizer`.
```Julia
model = Model(GLPK.Optimizer)
@variable(model, x)
optimize!(model)
```
Doesn't use `copy_to`, even though GLPK provides a fast one.

Rather than making an ad-hoc change, we should take the time to throughly document the caching optimizer system, choose a design that makes the most sense, and then implement that. There are quite a few points at which various layers are added, dropped, emptied, and reset.

### LINKED ISSUE #1249: Normal execution path of CLP involves exception throwing/catching
The normal execution path of using `Clp` to solve a linear program through `JuMP` involves throwing and catching an exception. In particular, an exception is *always* thrown here
https://github.com/jump-dev/MathOptInterface.jl/blob/ac22378a3dd310603ced6ccf0141f16d73cae091/src/variables.jl#L37
and is caught here (or in a similar place in the same file)
https://github.com/jump-dev/MathOptInterface.jl/blob/ac22378a3dd310603ced6ccf0141f16d73cae091/src/Utilities/cachingoptimizer.jl#L308
The exception handling can take up a significant portion of the overall optimization time, as shown by the benchmark below. Is there a way to avoid it?

```
using JuMP, Clp, MathOptInterface
using Profile, PProf

function test_clp(n::Int = 1000)
    for i = 1:n
        model = Model(Clp.Optimizer)
        set_optimizer_attribute(model, "LogLevel", 0)
        @variable(model, λ[1:3] ≥ 0.0)
        obj = @expression(model, sum(λ[jj] for jj in 1:3))
        @objective(model, Min, obj)
        @constraint(model, con[ii in 1:3],
                    sum(λ[jj] for jj in 1:3 if jj != ii) ≥ 1.0)
        optimize!(model)
        @assert termination_status(model) == MathOptInterface.OPTIMAL
        sol = value.(λ)
        obj_value = objective_value(model)
    end
end

test_clp(1000) # warmpup
Profile.init(n=Int(1e8), delay=0.002)
@pprof test_clp(1000)
```
Clp_profile

### LINKED ISSUE #1251: Skip the inner (inner) cache if bridges are not used
One option to improve performance is to skip the inner-most cache if no bridges are used.
```Julia
julia> model = Model(Clp.Optimizer)
A JuMP Model
Feasibility problem with:
Variables: 0
Model mode: AUTOMATIC
Solver name: Clp

julia> @variable(model, x)
x

julia> backend(model)
MOIU.CachingOptimizer{MOI.AbstractOptimizer,MOIU.UniversalFallback{MOIU.Model{Float64}}}
in state EMPTY_OPTIMIZER
in mode AUTOMATIC
with optimizer MOIB.LazyBridgeOptimizer{MOIU.CachingOptimizer{Clp.Optimizer,MOIU.UniversalFallback{MOIU.Model{Float64}}}}
  with 0 variable bridges
  with 0 constraint bridges
  with 0 objective bridges
  with inner model MOIU.CachingOptimizer{Clp.Optimizer,MOIU.UniversalFallback{MOIU.Model{Float64}}}
    in state ATTACHED_OPTIMIZER
    in mode AUTOMATIC
    with optimizer Clp.Optimizer
```

odow:
## What problem is this trying to solve

Consider this model in Clp
```Julia
model = Model(Clp.Optimizer)
@variable(model, x[1:2] >= 0)
optimize!(model)
```
**Pop quiz: how many caches are there?**
```Julia
julia> backend(model)
MOIU.CachingOptimizer{MOI.AbstractOptimizer,MOIU.UniversalFallback{MOIU.Model{Float64}}}
in state ATTACHED_OPTIMIZER
in mode AUTOMATIC
with optimizer MOIB.LazyBridgeOptimizer{MOIU.CachingOptimizer{Clp.Optimizer,MOIU.UniversalFallback{MOIU.Model{Float64}}}}
  with 0 variable bridges
  with 0 constraint bridges
  with 0 objective bridges
  with inner model MOIU.CachingOptimizer{Clp.Optimizer,MOIU.UniversalFallback{MOIU.Model{Float64}}}
    in state ATTACHED_OPTIMIZER
    in mode AUTOMATIC
    with optimizer Clp.Optimizer
```
There are two! There is the outer cache, then a bridging layer, and then the inner cache. 
But there are `0` bridges added to this problem! So our two caches are just copies of each other.

Why do we need two? Because a user might write this
```Julia
model = Model(Clp.Optimizer)
@variable(model, x[1:2])
@constraint(model, x in MOI.Nonnegatives(2))
optimize!(model)
Julia> backend(model)
MOIU.CachingOptimizer{MOI.AbstractOptimizer,MOIU.UniversalFallback{MOIU.Model{Float64}}}
in state ATTACHED_OPTIMIZER
in mode AUTOMATIC
with optimizer MOIB.LazyBridgeOptimizer{MOIU.CachingOptimizer{Clp.Optimizer,MOIU.UniversalFallback{MOIU.Model{Float64}}}}
  with 0 variable bridges
  with 1 constraint bridge
  with 0 objective bridges
  with inner model MOIU.CachingOptimizer{Clp.Optimizer,MOIU.UniversalFallback{MOIU.Model{Float64}}}
    in state ATTACHED_OPTIMIZER
    in mode AUTOMATIC
    with optimizer Clp.Optimizer
```

For the majority of users just trying to solve LPs and MIPs, this means that their first solve involves a double copy, and lots of extra inference sorting out the bridging functions.

## The alternative

If you knew there was going to be no bridging, you could write
```Julia
model = Model(Clp.Optimizer; bridge_constraints = false)
@variable(model, x[1:2] >= 0)
optimize!(model)
julia> backend(model)
MOIU.CachingOptimizer{MOI.AbstractOptimizer,MOIU.UniversalFallback{MOIU.Model{Float64}}}
in state ATTACHED_OPTIMIZER
in mode AUTOMATIC
with optimizer Clp.Optimizer
```
and if you get it wrong, you get a nice error message:
```Julia
model = Model(Clp.Optimizer; bridge_constraints = false)
@variable(model, x[1:2])
julia> @constraint(model, x in MOI.Nonnegatives(2))
ERROR: Constraints of type MathOptInterface.VectorOfVariables-in-MathOptInterface.Nonnegatives are not supported by the solver, try using `bridge_constraints=true` in the `JuMP.Model` constructor if you believe the constraint can be reformulated to constraints supported by the solver.
```

Thus, one option is to change the default in JuMP from `bridge_constraints = true` to `bridge_constraints = false`.

Pros: faster. Users are more aware when bridges are used. Most users don't use bridges
Cons: breaking. But it's a one-line change for users.

## Proposed approach

Start with the equivalent of `bridge_constraints = false`. If, when adding a constraint, the constraint is unsupported, add a `full_bridge_optimizer`.

Pros: Opt-in at MOI level. Better performance without breaking at JuMP level.
Cons: More complexity. Might be good for users to see when bridges are used.

See the companion PR in JuMP: https://github.com/jump-dev/JuMP.jl/pull/2513 which contains benchmarks proving the efficacy.

## TODOs

- [ ] Decide whether to do this, or change JuMP's default.
- [x] Tests
- [x] Benchmarks
- [ ] Docs
- [x] A way to get the `Number` type, rather than hard-coding `Float64`
- [ ] A way for JuMP to add extra bridges?
- [x] Bikeshed the argument name

Closes #1156 
Closes #1249
Closes #1251

### DIFF
--- a/src/Utilities/cachingoptimizer.jl
+++ b/src/Utilities/cachingoptimizer.jl
@@ -388,16 +388,25 @@ end
 
 
 """
-    _bridge_if_needed(f::Function, m::CachingOptimizer)
+    _bridge_if_needed(
+        f::Function,
+        m::CachingOptimizer;
+        add::Bool = false,
+    )
 
-Return `f(m)`, adding bridges if `f(m.optimizer)` is currently `false`, but
-after doing so, it will allow `m.optimizer` to return `f(m) == true`. If bridges
-are added, it will return `f(m) = true`, not the original `f(m) = false`.
+Return `f(m)`, under the assumption that the `.optimizer` field of `m` will be
+wrapped in a `LazyBridgeOptimizer` if `f(m)` is currently false, and that doing
+so would allow `f(m) == true`. However, only modify the `.optimizer` field if
+`add == true`.
 
 `f` is a function that takes `m` as a single argument. It is typically a call
 like `f(m) = MOI.supports_constraint(m, F, S)` for some `F` and `S`.
 """
-function _bridge_if_needed(f::Function, model::CachingOptimizer)
+function _bridge_if_needed(
+    f::Function,
+    model::CachingOptimizer;
+    add::Bool = false,
+)
     if !f(model.model_cache)
         # If the cache doesn't, we dont.
         return false
@@ -415,7 +424,9 @@ function _bridge_if_needed(f::Function, model::CachingOptimizer)
     T = MOI.get(model, MOI.CoefficientType())
     bridge = MOI.instantiate(model.optimizer; with_bridge_type = T)
     if f(bridge)
-        model.optimizer = bridge
+        if add
+            model.optimizer = bridge
+        end
         return true  # We bridged, and now we support.
     end
     return false  # Everything fails.
@@ -434,10 +445,11 @@ function MOI.add_constrained_variable(
     m::CachingOptimizer,
     set::S,
 ) where {S<:MOI.AbstractScalarSet}
-    if !MOI.supports_add_constrained_variable(m, S)
-        if state(m) == ATTACHED_OPTIMIZER
-            throw(MOI.UnsupportedConstraint{MOI.SingleVariable,S}())
-        end
+    supports = _bridge_if_needed(m; add = true) do model
+        return MOI.supports_add_constrained_variable(model, S)
+    end
+    if !supports && state(m) == ATTACHED_OPTIMIZER
+        throw(MOI.UnsupportedConstraint{MOI.SingleVariable,S}())
     end
     if m.state == MOIU.ATTACHED_OPTIMIZER
         if m.mode == MOIU.AUTOMATIC
@@ -505,12 +517,11 @@ function MOI.add_constrained_variables(
     m::CachingOptimizer,
     set::S,
 ) where {S<:MOI.AbstractVectorSet}
-    if !MOI.supports_add_constrained_variables(m, S)
-        if state(m) == ATTACHED_OPTIMIZER
-            throw(
-                MOI.UnsupportedConstraint{MOI.VectorOfVariables,S}()
-            )
-        end
+    supports = _bridge_if_needed(m; add = true) do model
+        return MOI.supports_add_constrained_variables(model, S)
+    end
+    if !supports && state(m) == ATTACHED_OPTIMIZER
+        throw(MOI.UnsupportedConstraint{MOI.VectorOfVariables,S}())
     end
     if m.state == ATTACHED_OPTIMIZER
         if m.mode == AUTOMATIC
@@ -568,10 +579,11 @@ function MOI.add_constraint(
     func::F,
     set::S,
 ) where {F<:MOI.AbstractFunction,S<:MOI.AbstractSet}
-    if !MOI.supports_constraint(m, F, S)
-        if state(m) == ATTACHED_OPTIMIZER
-            throw(MOI.UnsupportedConstraint{F,S}())
-        end
+    supports = _bridge_if_needed(m; add = true) do model
+        return MOI.supports_constraint(model, F, S)
+    end
+    if !supports && state(m) == ATTACHED_OPTIMIZER
+        throw(MOI.UnsupportedConstraint{F,S}())
     end
     if m.state == ATTACHED_OPTIMIZER
         if m.mode == AUTOMATIC
@@ -802,12 +814,33 @@ end
 # they are sent to the optimizer and when they are returned from the optimizer.
 # As a result, values of attributes must implement `map_indices`.
 
-function MOI.set(m::CachingOptimizer, attr::MOI.AbstractModelAttribute, value)
-    if !MOI.supports(m, attr)
-        if state(m) == ATTACHED_OPTIMIZER
-            throw(MOI.UnsupportedAttribute(attr))
+function MOI.set(m::CachingOptimizer, attr::MOI.ObjectiveFunction, value)
+    supports = _bridge_if_needed(m; add = true) do model
+        return MOI.supports(model, attr)
+    end
+    if !supports && state(m) == ATTACHED_OPTIMIZER
+        throw(MOI.UnsupportedAttribute(attr))
+    end
+    if m.state == ATTACHED_OPTIMIZER
+        optimizer_value = map_indices(m.model_to_optimizer_map, value)
+        if m.mode == AUTOMATIC
+            try
+                MOI.set(m.optimizer, attr, optimizer_value)
+            catch err
+                if err isa MOI.NotAllowedError
+                    reset_optimizer(m)
+                else
+                    rethrow(err)
+                end
+            end
+        else
+            MOI.set(m.optimizer, attr, optimizer_value)
         end
     end
+    return MOI.set(m.model_cache, attr, value)
+end
+
+function MOI.set(m::CachingOptimizer, attr::MOI.AbstractModelAttribute, value)
     if m.state == ATTACHED_OPTIMIZER
         optimizer_value = map_indices(m.model_to_optimizer_map, value)
         if m.mode == AUTOMATIC
--- a/src/Utilities/cachingoptimizer.jl
+++ b/src/Utilities/cachingoptimizer.jl
@@ -18,50 +18,111 @@
 and links it with an optimizer. It supports incremental model
 construction and modification even when the optimizer doesn't.
 
-A `CachingOptimizer` may be in one of three possible states (`CachingOptimizerState`):
+A `CachingOptimizer` may be in one of three possible states
+(`CachingOptimizerState`):
 
 * `NO_OPTIMIZER`: The CachingOptimizer does not have any optimizer.
 * `EMPTY_OPTIMIZER`: The CachingOptimizer has an empty optimizer.
   The optimizer is not synchronized with the cached model.
-* `ATTACHED_OPTIMIZER`: The CachingOptimizer has an optimizer, and it is synchronized with the cached model.
+* `ATTACHED_OPTIMIZER`: The CachingOptimizer has an optimizer, and it is
+  synchronized with the cached model.
 
 A `CachingOptimizer` has two modes of operation (`CachingOptimizerMode`):
 
 * `MANUAL`: The only methods that change the state of the `CachingOptimizer`
   are [`Utilities.reset_optimizer`](@ref), [`Utilities.drop_optimizer`](@ref),
-  and [`Utilities.attach_optimizer`](@ref).
-  Attempting to perform an operation in the incorrect state results in an error.
+  and [`Utilities.attach_optimizer`](@ref). Attempting to perform an operation
+  in the incorrect state results in an error.
 * `AUTOMATIC`: The `CachingOptimizer` changes its state when necessary. For
   example, `optimize!` will automatically call `attach_optimizer` (an
   optimizer must have been previously set). Attempting to add a constraint or
   perform a modification not supported by the optimizer results in a drop to
   `EMPTY_OPTIMIZER` mode.
 """
-mutable struct CachingOptimizer{OptimizerType,ModelType<:MOI.ModelLike} <:
-               MOI.AbstractOptimizer
+mutable struct CachingOptimizer{
+    OptimizerType,
+    ModelType<:MOI.ModelLike,
+} <: MOI.AbstractOptimizer
     optimizer::Union{Nothing,OptimizerType}
     model_cache::ModelType
     state::CachingOptimizerState
     mode::CachingOptimizerMode
     model_to_optimizer_map::IndexMap
     optimizer_to_model_map::IndexMap
-    # CachingOptimizer externally uses the same variable and constraint indices
-    # as the model_cache. model_to_optimizer_map maps from the model_cache indices to the
-    # optimizer indices.
+    auto_bridge::Bool
 end
 
+"""
+    CachingOptimizer(
+        model_cache::MOI.ModelLike,
+        optimizer::Union{Nothing,MOI.AbstractOptimizer} = nothing;
+        mode::CachingOptimizerMode = AUTOMATIC,
+        state::CachingOptimizerState =
+            optimizer === nothing ? NO_OPTIMIZER : EMPTY_OPTIMIZER,
+        auto_bridge::Bool = false,
+    )
+
+Creates a `CachingOptimizer` using `model_cache` and `optimizer`.
+
+## Notes
+
+ * If `auto_bridge == true`, when the caching optimizer encounters a constraint
+   or objective function that is not supported by `optimizer`, it automatically
+   adds a bridging layer to `optimizer`.
+ * If `auto_bridge == true`, and an optimizer is provided, the state is forced
+   to `EMPTY_OPTIMIZER`.
+ * If an `optimizer` is passed, the returned CachingOptimizer does not support
+   the function `reset_optimizer(model, new_optimizer)` if the type of
+   `new_optimizer` is different from the type of `optimizer`.
+
+## Examples
+
+```julia
+model = MOI.Utilities.CachingOptimizer(
+    MOI.Utilities.Model{Float64}(),
+    GLPK.Optimizer(),
+)
+```
+
+```julia
+model = MOI.Utilities.CachingOptimizer(
+    MOI.Utilities.Model{Float64}(),
+    auto_bridge = true,
+)
+MOI.Utilities.reset_optimizer(model, GLPK.Optimizer())
+```
+"""
 function CachingOptimizer(
     model_cache::MOI.ModelLike,
-    mode::CachingOptimizerMode,
+    optimizer::Union{Nothing,MOI.AbstractOptimizer} = nothing;
+    mode::CachingOptimizerMode = AUTOMATIC,
+    state::CachingOptimizerState =
+        optimizer === nothing ? NO_OPTIMIZER : EMPTY_OPTIMIZER,
+    auto_bridge::Bool = false,
 )
-    return CachingOptimizer{MOI.AbstractOptimizer,typeof(model_cache)}(
-        nothing,
+    T = optimizer !== nothing ? typeof(optimizer) : MOI.AbstractOptimizer
+    if optimizer !== nothing
+        @assert MOI.is_empty(model_cache)
+        @assert MOI.is_empty(optimizer)
+        if auto_bridge
+            state = EMPTY_OPTIMIZER
+            T = MOI.AbstractOptimizer
+        end
+    end
+    return CachingOptimizer{T,typeof(model_cache)}(
+        optimizer,
         model_cache,
-        NO_OPTIMIZER,
+        state,
         mode,
         IndexMap(),
         IndexMap(),
+        auto_bridge,
     )
+end
+
+# Added for compatibility with MOI 0.9.20
+function CachingOptimizer(cache::MOI.ModelLike, mode::CachingOptimizerMode)
+    return CachingOptimizer(cache; mode = mode)
 end
 
 function Base.show(io::IO, C::CachingOptimizer)
@@ -73,33 +134,6 @@ function Base.show(io::IO, C::CachingOptimizer)
     show(IOContext(io, :indent => get(io, :indent, 0) + 2), C.model_cache)
     print(io, "\n$(indent)with optimizer ")
     return show(IOContext(io, :indent => get(io, :indent, 0) + 2), C.optimizer)
-end
-
-"""
-    CachingOptimizer(model_cache::MOI.ModelLike, optimizer::AbstractOptimizer)
-
-Creates an `CachingOptimizer` in `AUTOMATIC` mode, with the optimizer
-`optimizer`.
-
-The type of the optimizer returned is `CachingOptimizer{typeof(optimizer),
-typeof(model_cache)}` so it does not support the function
-`reset_optimizer(::CachingOptimizer, new_optimizer)` if the type of
-`new_optimizer` is different from the type of `optimizer`.
-"""
-function CachingOptimizer(
-    model_cache::MOI.ModelLike,
-    optimizer::MOI.AbstractOptimizer,
-)
-    @assert MOI.is_empty(model_cache)
-    @assert MOI.is_empty(optimizer)
-    return CachingOptimizer{typeof(optimizer),typeof(model_cache)}(
-        optimizer,
-        model_cache,
-        EMPTY_OPTIMIZER,
-        AUTOMATIC,
-        IndexMap(),
-        IndexMap(),
-    )
 end
 
 function MOI.get(model::CachingOptimizer, attr::MOI.CoefficientType)
@@ -145,8 +179,8 @@ function reset_optimizer(m::CachingOptimizer, optimizer::MOI.AbstractOptimizer)
         if attr isa MOI.RawOptimizerAttribute
             # Even if the optimizer claims to `supports` `attr`, the value
             # might have a different meaning (e.g., two solvers with `logLevel`
-            # as a RawOptimizerAttribute). To be on the safe side, just skip all raw
-            # parameters.
+            # as a RawOptimizerAttribute). To be on the safe side, just skip all
+            # raw parameters.
             continue
         elseif !MOI.is_copyable(attr) || !MOI.supports(m.optimizer, attr)::Bool
             continue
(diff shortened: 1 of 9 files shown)


src/Utilities/cachingoptimizer.jl
```diff
@@ -80,11 +80,13 @@ typeof(model_cache)}` so it does not support the function
...
 function CachingOptimizer(
     model_cache::MOI.ModelLike,
-    optimizer::MOI.AbstractOptimizer,
+    optimizer::MOI.AbstractOptimizer;
+    auto_bridge::Bool = false,
 )
     @assert MOI.is_empty(model_cache)
     @assert MOI.is_empty(optimizer)
-    return CachingOptimizer{typeof(optimizer),typeof(model_cache)}(
+    T = auto_bridge ? MOI.AbstractOptimizer : typeof(optimizer)
```
blegat, line 88: JuMP does not call this constructor, it calls the one with the cache and the mode

src/Utilities/cachingoptimizer.jl
```diff
@@ -393,9 +395,21 @@ end
...
     m::CachingOptimizer,
-    func::MOI.AbstractFunction,
-    set::MOI.AbstractSet,
-)
+    func::F,
+    set::S,
+) where {F<:MOI.AbstractFunction,S<:MOI.AbstractSet}
+    if m.state != NO_OPTIMIZER && !MOI.supports_constraint(m.optimizer, F, S)
+        reset_optimizer(m)
+        new_opt = if supports_default_copy_to(m.optimizer, false)
```
blegat, line 403: We should create a function doing this to avoid having twice the same code: here and
https://github.com/jump-dev/MathOptInterface.jl/blob/ac22378a3dd310603ced6ccf0141f16d73cae091/src/instantiate.jl#L120-L125

src/Utilities/cachingoptimizer.jl
blegat, line 88: Having a non-concrete type for the optimizer might be quite bad for performance. As JuMP already has a non-concrete type, maybe we should change the type of the backend instead. Another option is to use a small union of the two possibilities here.

src/Utilities/cachingoptimizer.jl
```diff
@@ -393,9 +395,21 @@ end
 
 function MOI.add_constraint(
     m::CachingOptimizer,
-    func::MOI.AbstractFunction,
-    set::MOI.AbstractSet,
-)
+    func::F,
+    set::S,
+) where {F<:MOI.AbstractFunction,S<:MOI.AbstractSet}
+    if m.state != NO_OPTIMIZER && !MOI.supports_constraint(m.optimizer, F, S)
```
blegat, line 401: Check that the optimizer is not already a `LazyBridgeOptimizer`

blegat:
Solvers that benefit from this second case are the solvers that have an efficient `copy_to`. These solvers usually have a `copy_to` function that start by converting the model into a matrix form and then load it into the solver.
So here we really have 3 copies of the model:
User model -> Bridged model -> Matrix form
With your suggestion, it would be reduced to two:
User model -> Matrix form
So another way to do this would be to use the Matrix form for the Bridged model. That is, instead of using a `MOIU.Model` in the `CachingOptimizer`, we use a matrix format instead so that when we call `MOI.copy_to(::Optimtizer, ::MatrixForm)`, the solvers sees that it's already exactly the matrix form it needs, get the pointers of the vectors describing the sparse matrix and give that to the solver directly, no copy needed.

src/Utilities/cachingoptimizer.jl
odow: The reasoning behind this is you could use this constructor, but set `auto_bridge=true` to get the behavior. I've removed.

src/Utilities/cachingoptimizer.jl
odow: Done.

src/Utilities/cachingoptimizer.jl
odow: Removed.

src/Utilities/cachingoptimizer.jl
odow: Done

odow:
The matrix form is only used for loading though. It gets discarded and gc'd after `copy_to`.

There could be a way for the optimizer to say here is my desired cache. The default could be
```Julia
function desired_cache(model::AbstractOptimizer)
    return Utilities.UniversalFallback(Utilities.Model{Float64}())
end
```
and matrix solvers could specify something else.

src/Utilities/cachingoptimizer.jl
```diff
@@ -10,52 +10,114 @@
...
 )
-    return CachingOptimizer{MOI.AbstractOptimizer,typeof(model_cache)}(
-        nothing,
+    T = typeof(optimizer)
+    if optimizer !== nothing
+        @assert MOI.is_empty(model_cache)
+        @assert MOI.is_empty(optimizer)
+        if bridge_constraints
+            state = EMPTY_OPTIMIZER
+            T = MOI.AbstractOptimizer
```
blegat, line 100: The type is not concrete so it will be inefficient

blegat: We could use a union of the type with and without bridges

blegat:
One issue with the approach in the PR is that the bridge layer is added when the user calls `supports_...`. However, we should only add it when the user actually calls `add_...`. To make `supports_...` work, we should indeed check if it is supported if a bridge layer is added but we should not modify `model.optimizer`.

Another concern is that it fixes the use case of JuMP which has a cache before the bridge layer but for other use cases such as meta-solvers, you want to do basically
```julia
optimizer = MOI.instantiate(solver, with_bridge_type=Float64)
MOI.copy_to(optimizer, src)
```
In this case, we don't get any benefit from this PR.
To solve both use cases and avoid the issue with `supports` discussed above we could do the following:
1) Replace
https://github.com/jump-dev/MathOptInterface.jl/blob/9552c68316ddab8d89736fc84ca0ff20a28bdda2/src/Bridges/bridge_optimizer.jl#L323-L325
by
```julia
function MOI.copy_to(dest::AbstractBridgeOptimizer, src::MOI.ModelLike; kws...)
    if # Nothing is going to be bridged anyway
        MOI.copy_to(dest.model, src)
    else
        MOIU.default_copy_to(dest, , src; kws...)
    end
end
```
2) Add an option for the `CachingOptimizer` and do
```julia
function MOI.copy_to(dest::CachingOptimizer, ...)
    if state != NO_OPTIMIZER && # the option is true
        MOI.copy_to(dest.optimizer, ...)
    else
        # Same than what's currently done
    end
end
```
So with this option, we avoid using the cache in case `copy_to` is called.
In case modifications are done that are not supported by the solver, we simple do not catch the error.
JuMP could safely use it as there is a second layer of cache on top that would catch it and empty the optimizer (containing the bridges, the (unused) second cache and optimizer).

src/Utilities/cachingoptimizer.jl
```diff
@@ -10,52 +10,114 @@
...
+            optimizer === nothing ? NO_OPTIMIZER : ATTACHED_OPTIMIZER,
+        bridge_constraints::Bool = false,
+    )
+
+Creates an `CachingOptimizer` using `model_cache` and `optimizer`.
+
+## Notes
+
+ * If `bridge_constraints == true`, when the caching optimizer encounters a
+   constriant that is not supported by `optimizer`, it automatically adds a
```
joehuchette, line 62:
```suggestion
   constraint that is not supported by `optimizer`, it automatically adds a
```

src/Utilities/cachingoptimizer.jl
```diff
@@ -287,19 +324,52 @@ function MOI.add_variables(m::CachingOptimizer, n)
     return vindices
 end
 
+
+"""
+    _add_bridges_if_needed(f::Function, m::CachingOptimizer)
```
joehuchette, line 329: Update name

src/Utilities/cachingoptimizer.jl
```diff
@@ -287,19 +324,52 @@ function MOI.add_variables(m::CachingOptimizer, n)
...
+
+Adds bridges if it will allow `m.optimizer` to support `f(m)`.
+
+Returns `true` if bridges are added and `false` otherwise.
+"""
+function _auto_bridge_support(f::Function, model::CachingOptimizer)
+    if !f(model.model_cache)
```
joehuchette, line 336: It's unclear what `f` is supposed to be. Here, it looks like it's supposed to return a bool indicating something, but the docstring seems to suggest that it returns something to be bridged.

src/Utilities/cachingoptimizer.jl
odow: It's not sufficient to do this because the bridges may need a caching layer below. The abstract type has worked fine for JuMP, and having a complicated union may not actually help things.

odow:
The reason to add bridges on a `supports` call is two-fold:
  1. If you're asking `supports`, we don't want to return `false` when actually calling `add_constraint` will work.
  2. We don't want to initialize a bridge every time we call `supports`, only to reinitialize it when we call `add_constraint`.  

I'll have a play with your suggestion. But it seems like there are a lot of edge cases with modification and deletion.

blegat:
> The reason to add bridges on a `supports` call is two-fold:
>  1. If you're asking `supports`, we don't want to return `false` when actually calling `add_constraint` will work.
>  2. We don't want to initialize a bridge every time we call `supports`, only to reinitialize it when we call `add_constraint`.

Yes, for this reason, the only sensible way to make `supports` not have any effect is to have two fields `optimizer` and `bridged_optimizer` and a `Bool` indicating which one is active.
But that's going away from having MOI layers following the "Do one thing and do it well" principle.

odow:
More JuMP benchmarks using the script below:
```Julia
(base) oscar@Oscars-MBP JuMP % ~/julia --project=. example_diet.jl clp       
 17.319493 seconds (57.13 M allocations: 2.858 GiB, 8.30% gc time)
(base) oscar@Oscars-MBP JuMP % ~/julia --project=. example_diet.jl clp --auto
  6.071365 seconds (18.02 M allocations: 925.261 MiB, 5.73% gc time)
(base) oscar@Oscars-MBP JuMP % ~/julia --project=. example_diet.jl glpk      
 13.341120 seconds (44.97 M allocations: 2.257 GiB, 8.24% gc time)
(base) oscar@Oscars-MBP JuMP % ~/julia --project=. example_diet.jl glpk --auto
  8.593967 seconds (28.04 M allocations: 1.406 GiB, 8.13% gc time)
```
2 or 3x improvement in time-to-first-solve for models users care about. And this is not counting additional speedups we will get from precompilation.

## Code

```Julia
using JuMP, GLPK, Clp
function example_diet(optimizer, auto_bridge)
    categories = ["calories", "protein", "fat", "sodium"]
    category_data = Containers.DenseAxisArray([
        1800 2200;
        91   Inf;
        0    65;
        0    1779
        ], categories, ["min", "max"]
    )
    foods = [
        "hamburger", "chicken", "hot dog", "fries", "macaroni", "pizza",
        "salad", "milk", "ice cream",
    ]
    cost = Containers.DenseAxisArray(
        [2.49, 2.89, 1.50, 1.89, 2.09, 1.99, 2.49, 0.89, 1.59],
        foods
    )
    food_data = Containers.DenseAxisArray(
        [
            410 24 26 730;
            420 32 10 1190;
            560 20 32 1800;
            380  4 19 270;
            320 12 10 930;
            320 15 12 820;
            320 31 12 1230;
            100  8 2.5 125;
            330  8 10 180
        ], foods, categories
    )
    model = Model(optimizer, auto_bridge = auto_bridge)
    set_silent(model)
    @variables(model, begin
        category_data[c, "min"] <= nutrition[c = categories] <= category_data[c, "max"]
        buy[foods] >= 0
    end)
    @objective(model, Min, sum(cost[f] * buy[f] for f in foods))
    @constraint(model, [c in categories],
        sum(food_data[f, c] * buy[f] for f in foods) == nutrition[c]
    )
    optimize!(model)
    term_status = termination_status(model)
    @assert term_status == MOI.OPTIMAL
    @constraint(model, buy["milk"] + buy["ice cream"] <= 6)
    optimize!(model)
    @assert termination_status(model) == MOI.INFEASIBLE
    return
end

if length(ARGS) > 0
    auto = get(ARGS, 2, "") == "--auto"
    if ARGS[1] == "clp"
        @time example_diet(Clp.Optimizer, auto)
    else
        @assert ARGS[1] == "glpk"
        @time example_diet(GLPK.Optimizer, auto)
    end
end
```

src/Utilities/cachingoptimizer.jl
blegat: I don't agree, we get an allocation everytime we need to access the field, we have all these `_moi` function barriers in JuMP to avoid this. If we have the same issue in the `CachingOptimizer`, then it doesn't make sense.

odow: This is an opt-in feature for users of CachingOptimizer. JuMP will opt-in with no change to performance because it already has the abstract type. If others opt-in, they should check performance and/or implement function barriers.

Overall, this is a big win for JuMP with minimal impact on other users. It's simple to implement, and there are no edge cases.

blegat: I thing there will be a change in perf even for JuMP. Now there are two fields of abstract type.
We don't know what `backend` is, so have one hit, then you figure out it's `CachingOptimizer` so you make a call to it and then the `optimizer` field is abstract again so you get a second hit.

odow: There is no change in the JuMP behavior:
```Julia
julia> model = Model(Clp.Optimizer);

julia> backend(model)
MOIU.CachingOptimizer{MOI.AbstractOptimizer,MOIU.UniversalFallback{MOIU.Model{Float64}}}
in state EMPTY_OPTIMIZER
in mode AUTOMATIC
with optimizer MOIB.LazyBridgeOptimizer{MOIU.CachingOptimizer{Clp.Optimizer,MOIU.UniversalFallback{MOIU.Model{Float64}}}}
  with 0 variable bridges
  with 0 constraint bridges
  with 0 objective bridges
  with inner model MOIU.CachingOptimizer{Clp.Optimizer,MOIU.UniversalFallback{MOIU.Model{Float64}}}
    in state ATTACHED_OPTIMIZER
    in mode AUTOMATIC
    with optimizer Clp.Optimizer

julia> model2 = Model(Clp.Optimizer; auto_bridge = true);

julia> backend(model2)
MOIU.CachingOptimizer{MOI.AbstractOptimizer,MOIU.UniversalFallback{MOIU.Model{Float64}}}
in state EMPTY_OPTIMIZER
in mode AUTOMATIC
with optimizer Clp.Optimizer

julia> @variable(model2, x[1:2] in MOI.Nonnegatives(2))
2-element Array{VariableRef,1}:
 x[1]
 x[2]

julia> backend(model2)
MOIU.CachingOptimizer{MOI.AbstractOptimizer,MOIU.UniversalFallback{MOIU.Model{Float64}}}
in state EMPTY_OPTIMIZER
in mode AUTOMATIC
with optimizer MOIB.LazyBridgeOptimizer{MOIU.CachingOptimizer{Clp.Optimizer,MOIU.UniversalFallback{MOIU.Model{Float64}}}}
  with 0 variable bridges
  with 0 constraint bridges
  with 0 objective bridges
  with inner model MOIU.CachingOptimizer{Clp.Optimizer,MOIU.UniversalFallback{MOIU.Model{Float64}}}
    in state ATTACHED_OPTIMIZER
    in mode AUTOMATIC
    with optimizer Clp.Optimizer
```

odow: If bridges are applied, we get back to the current JuMP behavior. If bridges are not applied, then we skip all the issues with bridges, but the backend type that JuMP sees is still the same.

odow: Can we have a call to discuss instead of this back-and-forth?

blegat: Yes, I think we reached the point where messages with opposite timezone makes this process long and inefficient :D

odow:
Talked to @blegat offline. The alternative is to add a `DIRECT` state to `CachingOptimizer` which just ignores the cache, and throws errors if things are not supported.

odow:
@mlubin, I think we need your input on this one.

mlubin:
This PR claims to close #1156. #1156 says:
> Rather than making an ad-hoc change, we should take the time to throughly document the caching optimizer system, choose a design that makes the most sense, and then implement that.

Where's the thorough documentation? :)

Changing the external state of the `CachineOptimizer` in `supports_*` is a bad code smell. Even changing the internal state would be strange because it's implied that `supports_*` has no side effects.

Could you elaborate on what the `DIRECT` state would look like?

odow:
> Where's the thorough documentation? :)

Still to come
image

> is a bad code smell

I've made it so the optimizer is only modified during an actual `add_constraint` call.

> Could you elaborate on what the DIRECT state would look like?

It would function just like JuMP's current direct mode. Everything is forwarded straight to `.optimizer`, the `.model_cache` is not kept in sync (it stays empty), and anything that isn't supported throws an error. This is only useful if there are two caches though, so something like: Cache -> Bridge -> Cache(direct) -> Optimizer. Because then when the inner cache(direct) throws, we can repopulate the model from the outer cache. We would also have to use Benoit's suggestion for `copy_to`, so that the caching optimizer in direct would `copy_to` the optimizer, and the bridges would have to check if any bridging needed to happen, and if not, it would forward the `copy_to` to the optimizer.

Note that even if we merge this PR, we could still add a DIRECT state in future...

odow:
As for the documentation, the code in `cachingoptimizer.jl` is actually pretty straight forward, and less complicated than I had realized. (Most of the complexity is in `MOI.Utilities.Model` and in the bridges.)

One option I would be in favor of is to remove `MANUAL` mode. It doesn't get used, and there aren't many reasons to use it over `AUTOMATIC` mode. But that's a different topic.

odow:
Latest timings on Julia 1.5.3
```julia
(base) oscar@Oscars-MBP scripts % ~/julia --project=. example_diet.jl clp 
 13.592436 seconds (35.24 M allocations: 1.775 GiB, 6.56% gc time)
(base) oscar@Oscars-MBP scripts % ~/julia --project=. example_diet.jl clp --auto
  5.070573 seconds (11.74 M allocations: 606.745 MiB, 4.71% gc time)
(base) oscar@Oscars-MBP scripts % ~/julia --project=. example_diet.jl glpk      
 10.806421 seconds (29.22 M allocations: 1.471 GiB, 7.37% gc time)
(base) oscar@Oscars-MBP scripts % ~/julia --project=. example_diet.jl glpk --auto
  5.876158 seconds (12.85 M allocations: 666.154 MiB, 3.98% gc time)
```

And 1.6-RC1
```Julia
(base) oscar@Oscars-MBP scripts % '/Applications/Julia-1.6.app/Contents/Resources/julia/bin/julia' --project=. example_diet.jl clp
 12.588525 seconds (32.80 M allocations: 1.842 GiB, 6.04% gc time, 33.69% compilation time)
(base) oscar@Oscars-MBP scripts % '/Applications/Julia-1.6.app/Contents/Resources/julia/bin/julia' --project=. example_diet.jl clp --auto
  5.087086 seconds (10.93 M allocations: 647.489 MiB, 3.38% gc time, 92.28% compilation time)
(base) oscar@Oscars-MBP scripts % '/Applications/Julia-1.6.app/Contents/Resources/julia/bin/julia' --project=. example_diet.jl glpk     
  9.666504 seconds (27.98 M allocations: 1.566 GiB, 5.27% gc time, 38.85% compilation time)
(base) oscar@Oscars-MBP scripts % '/Applications/Julia-1.6.app/Contents/Resources/julia/bin/julia' --project=. example_diet.jl glpk --auto
  6.029965 seconds (12.95 M allocations: 772.171 MiB, 3.82% gc time, 99.95% compilation time)
```

Still lots of room to improve, but we've gone from 17s to 5s on Clp, and 13s to 6s on GLPK.

blegat:
As we discussed offline, the use case of this is to remove this second layer of cache that is required for solvers that to not implement incremental building such as Clp. However, there is currently another copy to a third model that is done internally in the MOI wrapper of Clp into a matrix format that can directly be passed (by pointers, without copy) to Clp.
As detailed in https://github.com/jump-dev/MathOptInterface.jl/issues/1261, we could merge the two last models by directly storing the bridged model in the right format needed by Clp in the cache of the CachingOptimizer.
So it seems https://github.com/jump-dev/MathOptInterface.jl/issues/1261 and this PR are solving the same use-case so we should probably keep only one of them.
As we discussed, this could be a short-term solution while we do https://github.com/jump-dev/MathOptInterface.jl/issues/1261 but I don't think it would take to much time to do https://github.com/jump-dev/MathOptInterface.jl/issues/1261 (given that https://github.com/jump-dev/MathOptInterface.jl/pull/1245 and https://github.com/jump-dev/MathOptInterface.jl/pull/1267 are already done) and we don't want to add an option (resp. an opt-in behavior) that we deprecate shortly after adding it (resp. that we change shortly after).
It's good to see the latency reduction that we get with this PR though and we should check that we can at least the same amount by the alternative approaches.

odow:
> we should check that we can at least the same amount whether we choose to do this change or another one.

Agreed. 

What if we release `0.9.21` soon with the current changes, then start on 0.10? There are quite a few breaking changes queued up, so we could merge this into a 0.10-dev, but hold off releasing 0.10 for a while.

blegat:
> There are quite a few breaking changes queued up

I don't think we need to make a breaking releases. We can merge https://github.com/jump-dev/MathOptInterface.jl/pull/1254 and see if it breaks anything but I would suspect that if it breaks a package then the package was using the `CachingOptimizer` incorrectly.

> so we could merge this into a 0.10-dev

I don't think this should be merged, we can leave it open as a reference but as discussed, I don't see how it fits in the long term plan. Also, we should aim at getting the advantages of this PR also if only some part of the model is bridged.

blegat:
> There are quite a few breaking changes queued up

I don't think we need to make a breaking releases. We can merge https://github.com/jump-dev/MathOptInterface.jl/pull/1254 and see if it breaks anything but I would suspect that if it breaks a package then the package was using the `CachingOptimizer` incorrectly.

> so we could merge this into a 0.10-dev

I don't think this should be merged, we can leave it open as a reference but as discussed, I don't see how it fits in the long term plan. Also, we should aim at getting the advantages of this PR also if only some part of the model is bridged.

blegat:
> There are quite a few breaking changes queued up

I don't think we need to make a breaking releases. We can merge https://github.com/jump-dev/MathOptInterface.jl/pull/1254 and see if it breaks anything but I would suspect that if it breaks a package then the package was using the `CachingOptimizer` incorrectly.

> so we could merge this into a 0.10-dev

I don't think this should be merged, we can leave it open as a reference but as discussed, I don't see how it fits in the long term plan. Also, we should aim at getting the advantages of this PR also if only some part of the model is bridged.

odow:
Adding this to the 0.10 milestone because we should decide whether or not to add it before releasing 0.10. I'm in favor of adding it, unless we can demonstrate that the alternatives are more performant.

blegat:
I think we should better understand what's going on with the benchmark. Is that only compile-time difference or does some difference persists in the run-time ? What happens if you remove the bridge layer but keep a second cache ? Intuitively, I would think that there is not much to compile at the bridge layers if all constraints are supported. As the bridge graph is built lazily, if all constraints are supported, nothing is built. If we notice there is still a lot of things compiled when nothing is computed then maybe there is a simple fix to the bridges module that could resolve it. This fix would be useful even when only part of the model is bridged.
For the second layer of cache, this should be resolved by https://github.com/jump-dev/MathOptInterface.jl/pull/1287. It has taken time but this PR should now be ready.

odow:
I don't think it was compilation, it was mainly an inference problem. But I'll redo the benchmarks in light of the new MOI release and Julia 1.6.

odow:
Things have gotten worse with MOI 0.9.21:
```Julia
(base) oscar@Oscars-MBP auto-cache % ~/julia --project=. --depwarn=error bench.jl clp
 24.237062 seconds (57.88 M allocations: 3.274 GiB, 6.30% gc time, 41.11% compilation time)
  0.001585 seconds (6.36 k allocations: 563.711 KiB)
(base) oscar@Oscars-MBP auto-cache % ~/julia --project=. --depwarn=error bench.jl clp --auto
 12.316551 seconds (20.23 M allocations: 1.157 GiB, 4.61% gc time, 96.23% compilation time)
  0.001926 seconds (3.29 k allocations: 329.508 KiB)
(base) oscar@Oscars-MBP auto-cache % ~/julia --project=. --depwarn=error bench.jl glpk      
 16.994045 seconds (36.90 M allocations: 2.125 GiB, 5.72% gc time, 55.21% compilation time)
  0.000760 seconds (3.46 k allocations: 270.109 KiB)
(base) oscar@Oscars-MBP auto-cache % ~/julia --project=. --depwarn=error bench.jl glpk --auto
 13.525074 seconds (22.89 M allocations: 1.315 GiB, 4.67% gc time, 99.97% compilation time)
  0.000798 seconds (2.80 k allocations: 254.281 KiB)
```

This is with
```Julia
(auto-cache) pkg> st
      Status `~/Documents/JuMP/performance/auto-cache/Project.toml`
  [e2554f3b] Clp v0.8.4 `https://github.com/jump-dev/Clp.jl.git#od/moi10`
  [60bf3e95] GLPK v0.14.8 `https://github.com/jump-dev/GLPK.jl.git#od/moi10`
  [4076af6c] JuMP v0.21.7 `https://github.com/jump-dev/JuMP.jl.git#od/autobridge`
  [b8f27783] MathOptInterface v0.9.21 `https://github.com/jump-dev/MathOptInterface.jl.git#od/cache_auto_bridge`
```

I'll take a deeper look.

odow:
Closing this because I talked to @blegat and we decided this could actually be done at the JuMP level, rather than as a complication of MathOptInterface.

PR #1254, opened 2021-03-01
Change the default mode of CachingOptimizer

odow:
Another way of fixing #1249.

The default of a `CachingOptimizer` is to initialize in `ATTACHED_OPTIMIZER` mode. This has two consequences:
* Solvers that do not support incremental building will throw an error at the first incremental call and reset the optimizer. 
* Solvers that do support incremental building will build incrementally, even if they have a fast `copy_to` available.

It seems like a much more sensible default is to initialize in `EMPTY_OPTIMIZER` mode. 
* Solvers that do not support incremental building no longer need to catch an error
* Solvers that do support incremental building will use the fast `copy_to` if available.

This is breaking (see the test change), so we may want to hold off until MOI 0.10, but it does need doing.


mlubin (approved):
Makes sense to me, pretty clean fix. Agreed it should be for MOI 0.10.

blegat (approved):
Looks good. I don't think it's breaking. We can release this in a patch release. It only affect the state in `AUTOMATIC` mode anyway

test/Utilities/cachingoptimizer.jl
```diff
@@ -108,7 +108,11 @@ struct DummyConstraintAttribute <: MOI.AbstractConstraintAttribute end
...
 @testset "Mapping of variables" begin
     mock = MOIU.MockOptimizer(MOIU.UniversalFallback(MOIU.Model{Float64}()))
-    model = MOIU.CachingOptimizer(MOIU.UniversalFallback(MOIU.Model{Float64}()), mock)
+    model = MOIU.CachingOptimizer(
+        MOIU.UniversalFallback(MOIU.Model{Float64}()),
+        mock,
+    )
+    MOIU.attach_optimizer(model)
     x = MOI.add_variable(model)
     y = first(MOI.get(mock, MOI.ListOfVariableIndices()))
```
odow, line 117: @blegat it's breaking here without an explicit `attach_optimizer`. People doing stuff with the backend of JuMP before `optimize!` may run into errors.

blegat: Which error ? In `AUTOMATIC` mode, we automatically attach instead of erroring.

odow: This example queries the inner optimizer of the cache directly.

blegat: You shouldn't expect anything on the state of the inner optimizer in `AUTOMATIC` mode.
The only good reason users were accessing the inner optimizer was for optimizer attributes but that is not changed by this PR.

blegat (approved):

PR #1256, opened 2021-03-02
Fix inference issue in MOIU.Model

odow:
Closes https://github.com/jump-dev/MathOptInterface.jl/issues/1255

## Before

The troublesome line was (note the `S1::Type`)
```julia
julia> @code_warntype MOI.Utilities.throw_if_lower_bound_set(f.variable, typeof(s), mask, T)
Variables
  #self#::Core.Compiler.Const(MathOptInterface.Utilities.throw_if_lower_bound_set, false)
  variable::MathOptInterface.VariableIndex
  S2::Core.Compiler.Const(MathOptInterface.GreaterThan{Float64}, false)
  mask::UInt8
  T::Core.Compiler.Const(Float64, false)
  flag::UInt8
  S1::Type
  @_8::Bool

Body::Nothing
1 ─       Core.NewvarNode(:(S1))
│         (flag = MathOptInterface.Utilities.single_variable_flag(S2))
│   %3  = (flag::Core.Compiler.Const(0x02, false) & MathOptInterface.Utilities.LOWER_BOUND_MASK)::Core.Compiler.Const(0x02, false)
│   %4  = MathOptInterface.Utilities.iszero(%3)::Core.Compiler.Const(false, false)
│   %5  = !%4::Core.Compiler.Const(true, true)
│         %5
│   %7  = (mask & MathOptInterface.Utilities.LOWER_BOUND_MASK)::UInt8
│   %8  = MathOptInterface.Utilities.iszero(%7)::Bool
│         (@_8 = !%8)
└──       goto #3
2 ─       Core.Compiler.Const(:(@_8 = false), false)
3 ┄       goto #5 if not @_8
4 ─ %13 = (mask & MathOptInterface.Utilities.LOWER_BOUND_MASK)::UInt8
│         (S1 = MathOptInterface.Utilities.flag_to_set_type(%13, T))
│   %15 = MathOptInterface.LowerBoundAlreadySet::Core.Compiler.Const(MathOptInterface.LowerBoundAlreadySet, false)
│   %16 = S1::Type
│   %17 = Core.apply_type(%15, %16, S2)::Type{MathOptInterface.LowerBoundAlreadySet{_A,MathOptInterface.GreaterThan{Float64}}} where _A
│   %18 = (%17)(variable)::MathOptInterface.LowerBoundAlreadySet{_A,MathOptInterface.GreaterThan{Float64}} where _A
└──       MathOptInterface.Utilities.throw(%18)
5 ┄       return
```

```Julia
julia> function plain(N)
           model = MOIU.Model{Float64}()
           x = MOI.add_variables(model, N)
           return x
       end
plain (generic function with 1 method)

julia> function with_bounds(N)
           model = MOIU.Model{Float64}()
           x = MOI.add_variables(model, N)
           for xi in x
               MOI.add_constraint(model, MOI.SingleVariable(xi), MOI.GreaterThan(0.0))
           end
           return x
       end
with_bounds (generic function with 1 method)

julia> @btime plain(1_000_000);
  37.242 ms (175 allocations: 27.64 MiB)

julia> @btime with_bounds(1_000_000);
  95.231 ms (175 allocations: 27.64 MiB)
```

## After

With these hints, type inference succeeds
```julia
julia> @code_warntype MOI.Utilities.throw_if_lower_bound_set(f.variable, typeof(s), mask, T)
Variables
  #self#::Core.Compiler.Const(MathOptInterface.Utilities.throw_if_lower_bound_set, false)
  variable::MathOptInterface.VariableIndex
  S2::Core.Compiler.Const(MathOptInterface.GreaterThan{Float64}, false)
  mask::UInt8
  T::Core.Compiler.Const(Float64, false)
  flag::UInt8
  S1::Union{}
  @_8::Bool

Body::Nothing
1 ─       Core.NewvarNode(:(S1))
│         (flag = MathOptInterface.Utilities.single_variable_flag(S2))
│   %3  = (flag::Core.Compiler.Const(0x02, false) & MathOptInterface.Utilities.LOWER_BOUND_MASK)::Core.Compiler.Const(0x02, false)
│   %4  = MathOptInterface.Utilities.iszero(%3)::Core.Compiler.Const(false, false)
│   %5  = !%4::Core.Compiler.Const(true, true)
│         %5
│   %7  = (mask & MathOptInterface.Utilities.LOWER_BOUND_MASK)::UInt8
│   %8  = MathOptInterface.Utilities.iszero(%7)::Bool
│         (@_8 = !%8)
└──       goto #3
2 ─       Core.Compiler.Const(:(@_8 = false), false)
3 ┄       goto #5 if not @_8
4 ─ %13 = (mask & MathOptInterface.Utilities.LOWER_BOUND_MASK)::UInt8
│         (S1 = MathOptInterface.Utilities.flag_to_set_type(%13, T))
│         Core.Compiler.Const(:(MathOptInterface.LowerBoundAlreadySet), false)
│         Core.Compiler.Const(:(S1), false)
│         Core.Compiler.Const(:(Core.apply_type(%15, %16, S2)), false)
│         Core.Compiler.Const(:((%17)(variable)), false)
└──       Core.Compiler.Const(:(MathOptInterface.Utilities.throw(%18)), false)
5 ┄       return
```

```julia
julia> @btime plain(1_000_000);
  37.303 ms (175 allocations: 27.64 MiB)

julia> @btime with_bounds(1_000_000);
  39.206 ms (175 allocations: 27.64 MiB)
```

### DIFF
--- a/src/Utilities/model.jl
+++ b/src/Utilities/model.jl
@@ -550,7 +550,7 @@ single_variable_flag(::Type{<:MOI.Semiinteger}) = 0x80
 # If a set is added here, a line should be added in
 # `MOI.delete(::AbstractModel, ::MOI.VariableIndex)`
 
-function flag_to_set_type(flag::UInt8, T::Type)
+function flag_to_set_type(flag::UInt8, ::Type{T}) where {T}
     if flag == 0x1
         return MOI.EqualTo{T}
     elseif flag == 0x2
@@ -565,28 +565,46 @@ function flag_to_set_type(flag::UInt8, T::Type)
         return MOI.ZeroOne
     elseif flag == 0x40
         return MOI.Semicontinuous{T}
-    elseif flag == 0x80
-        return MOI.Semiinteger{T}
     else
-        # $flag would print it in decimal
-        error("Invalid flag `$(sprint(show, flag))`.")
+        @assert flag == 0x80
+        return MOI.Semiinteger{T}
     end
+end
+
+# Julia doesn't infer `S1` correctly, so we use a function barrier to improve
+# inference.
+function _throw_if_lower_bound_set(variable, S2, mask, T)
+    S1 = flag_to_set_type(mask, T)
+    throw(MOI.LowerBoundAlreadySet{S1,S2}(variable))
+    return
 end
 
 function throw_if_lower_bound_set(variable, S2, mask, T)
-    flag = single_variable_flag(S2)
-    if !iszero(flag & LOWER_BOUND_MASK) && !iszero(mask & LOWER_BOUND_MASK)
-        S1 = flag_to_set_type(mask & LOWER_BOUND_MASK, T)
-        throw(MOI.LowerBoundAlreadySet{S1,S2}(variable))
+    lower_mask = mask & LOWER_BOUND_MASK
+    if iszero(lower_mask)
+        return  # No lower bound set.
+    elseif iszero(single_variable_flag(S2) & LOWER_BOUND_MASK)
+        return  # S2 isn't related to the lower bound.
     end
+    return _throw_if_lower_bound_set(variable, S2, lower_mask, T)
+end
+
+# Julia doesn't infer `S1` correctly, so we use a function barrier to improve
+# inference.
+function _throw_if_upper_bound_set(variable, S2, mask, T)
+    S1 = flag_to_set_type(mask, T)
+    throw(MOI.UpperBoundAlreadySet{S1,S2}(variable))
+    return
 end
 
 function throw_if_upper_bound_set(variable, S2, mask, T)
-    flag = single_variable_flag(S2)
-    if !iszero(flag & UPPER_BOUND_MASK) && !iszero(mask & UPPER_BOUND_MASK)
-        S1 = flag_to_set_type(mask & UPPER_BOUND_MASK, T)
-        throw(MOI.UpperBoundAlreadySet{S1,S2}(variable))
+    upper_mask = mask & UPPER_BOUND_MASK
+    if iszero(upper_mask)
+        return  # No upper bound set.
+    elseif iszero(single_variable_flag(S2) & UPPER_BOUND_MASK)
+        return  # S2 isn't related to the upper bound.
     end
+    return _throw_if_upper_bound_set(variable, S2, upper_mask, T)
 end
 
 # Sets setting lower bound:
--- a/test/Utilities/model.jl
+++ b/test/Utilities/model.jl
@@ -102,9 +102,8 @@ end
 
 @testset "Setting lower/upper bound twice" begin
     @testset "flag_to_set_type" begin
-        err = ErrorException("Invalid flag `0x11`.")
         T = Int
-        @test_throws err MOIU.flag_to_set_type(0x11, T)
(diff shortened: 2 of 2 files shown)


odow:
And proof from the original JuMP issue: https://github.com/jump-dev/MathOptInterface.jl/issues/1255.
```julia
julia> @btime plain(1_000_000);
  99.679 ms (2000244 allocations: 142.09 MiB)

julia> @btime with_int_and_bounds(1_000_000);
  117.368 ms (2000244 allocations: 142.09 MiB)
```

mlubin (approved):
Nice!

blegat:
This is awesome! However I don't quite understand what's happening. In both case there is the type unstability but in both cases we do not run the type unstable code. In the first case, the type unstable code is inside an `if` statement and in the second case, it's inside a separate function. However, looking at the output of `@code_warntype`, it seems to be inlined so why is that different ?

> With these hints, type inference succeeds

The inference of which type ? The type of `S1` is still not inferrable, the type of which variable is now inferrable ?

odow:
oops. I didn't update the head of the PR. My original solution (annotating with `::Type{<:AbstractSet}`) caused a weird error when it actually got thrown, so I used a function barrier and it seemed to do the same job.

odow:
Here's what happens now:
```Julia
julia> model = MOIU.Model{Float64}()
MOIU.Model{Float64}

julia> @code_warntype MOIU.throw_if_upper_bound_set(x, MOI.LessThan{Float64}, 0x0, Float64)
Variables
  #self#::Core.Compiler.Const(MathOptInterface.Utilities.throw_if_upper_bound_set, false)
  variable::MathOptInterface.VariableIndex
  S2::Core.Compiler.Const(MathOptInterface.LessThan{Float64}, false)
  mask::UInt8
  T::Core.Compiler.Const(Float64, false)
  upper_mask::UInt8

Body::Nothing
1 ─      (upper_mask = mask & MathOptInterface.Utilities.UPPER_BOUND_MASK)
│   %2 = MathOptInterface.Utilities.iszero(upper_mask)::Bool
└──      goto #3 if not %2
2 ─      return
3 ─ %5 = MathOptInterface.Utilities.single_variable_flag(S2)::Core.Compiler.Const(0x04, false)
│   %6 = (%5 & MathOptInterface.Utilities.UPPER_BOUND_MASK)::Core.Compiler.Const(0x04, false)
│   %7 = MathOptInterface.Utilities.iszero(%6)::Core.Compiler.Const(false, false)
└──      goto #5 if not %7
4 ─      Core.Compiler.Const(:(return), false)
5 ┄      MathOptInterface.Utilities._throw_if_upper_bound_set(variable, S2, upper_mask, T)
└──      Core.Compiler.Const(:(return %10), false)
```

The `_throw` method still infers badly, but we only hit that if an error is thrown, so we don't really care.

blegat:
That makes more sense. Is there a way to avoid the function being inlined?

odow:
My guess is that Julia would only inline if it infers nicely. I don't think we need to do anything explicitly here. It's something we can check in a future release.

blegat:
If it stops inferring we won't notice as we have no tests. We could use `@noinline`

PR #1257, opened 2021-03-02
Fix complexity of getindex(::ScalarFunctionIterator, i)

### LINKED ISSUE #418: ScalarFunctionIterator has quadratic complexity

### LINKED ISSUE #1137: Performance issue with Quadratic objective to SOC constraint
Hi,

I am trying to solve some large QPs and I noticed quite a big performance issue when I use MOI to pass the problem to solvers that don't support quadratic objectives. For very large problems the reformulation takes 10 - 100x the solve time.

Here is a MWE of moderate size:
```julia
using MathOptInterface, SCS
const MOI = MathOptInterface
const MOIU = MOI.Utilities
const MOIB = MOI.Bridges
using Random, SparseArrays, LinearAlgebra

rng = Random.MersenneTwister(123123)

# create random problem data
n = 300
P = spdiagm(0 => ones(n^2))
q = rand(rng, n^2)

# make MOI model
x = MOI.add_variables(model, n^2);

# define the quadratic objective 1/2 x'x + q'x
affine_terms = MOI.ScalarAffineTerm{Float64}.(q, x)
quadratic_terms = MOI.ScalarQuadraticTerm{Float64}.(1., x, x)
quad_func = MOI.ScalarQuadraticFunction{Float64}(affine_terms, quadratic_terms, 0.)
MOI.set(model, MOI.ObjectiveSense(), MOI.MIN_SENSE);
MOI.set(model, MOI.ObjectiveFunction{MOI.ScalarQuadraticFunction{Float64}}(), quad_func);

# add inequality constraints (not important here)
MOI.add_constraint(model, MOI.VectorOfVariables(x), MOI.Nonnegatives(n));

# copy to a solver that doesn't support quadratic objectives to invoke the bridging
cache = MOI.Utilities.CachingOptimizer(
        MOI.Utilities.Model{Float64}(), SCS.Optimizer())
bridged = MOI.Bridges.full_bridge_optimizer(cache, Float64)
@time MOI.copy_to(bridged, model)

#@profile MOI.copy_to(bridged, model)
```
For `n=300` this takes about `39s`. I am trying to solve problems with up to `n=1200` in which case this step takes up to an hour. (Notice, that I am actually using n^2)

Looking at the profile results, I get:
flamegraph

it seems like all the time is spent to map indices in 
https://github.com/jump-dev/MathOptInterface.jl/blob/c602677d9a61a05328dd93aa772074b5ca38d9a0/src/Utilities/functions.jl#L274-L277
in the `findall()` function. I am wondering if anything could be done to make this more efficient.

odow:
**Updated times for most recent commit**
Before
```Julia
julia> @time main(300)
 10.678643 seconds (632.16 k allocations: 132.960 MiB, 1.65% gc time)
```
After
```julia
julia> @time main(300);
  0.167274 seconds (182.72 k allocations: 137.331 MiB, 17.05% gc time)
```

## Code

```Julia
function main(n)
    model = MOIU.Model{Float64}()
    x = MOI.add_variables(model, n^2)
    MOI.set(model, MOI.ObjectiveSense(), MOI.MIN_SENSE)
    f = MOI.ScalarQuadraticFunction{Float64}(
        MOI.ScalarAffineTerm{Float64}.(rand(n^2), x),
        MOI.ScalarQuadraticTerm{Float64}.(1.0, x, x),
        0.0,
    )
    MOI.set(
        model,
        MOI.ObjectiveFunction{MOI.ScalarQuadraticFunction{Float64}}(),
        f,
    )
    MOI.add_constraint(model, MOI.VectorOfVariables(x), MOI.Nonnegatives(n))
    bridged = MOI.Bridges.full_bridge_optimizer(
        MOI.Utilities.CachingOptimizer(
            MOI.Utilities.Model{Float64}(),
            SCS.Optimizer(),
        ),
        Float64,
    )
    MOI.copy_to(bridged, model)
    return bridged
end
```

Closes #1137 
Closes #418

### DIFF
--- a/src/Utilities/functions.jl
+++ b/src/Utilities/functions.jl
@@ -320,9 +320,47 @@ function scalar_type(::Type{MOI.VectorQuadraticFunction{T}}) where {T}
     return MOI.ScalarQuadraticFunction{T}
 end
 
+"""
+    ScalarFunctionIterator{F<:MOI.AbstractVectorFunction}
+
+A type that allows iterating over the scalar-functions that comprise an
+`AbstractVectorFunction`.
+"""
 struct ScalarFunctionIterator{F<:MOI.AbstractVectorFunction}
     f::F
+    # Dictionaries which map output indices to their terms.
+    affine::Dict{Int,Vector{Int}}
+    quadratic::Dict{Int,Vector{Int}}
 end
+
+function ScalarFunctionIterator(f::MOI.VectorOfVariables)
+    return ScalarFunctionIterator(
+        f,
+        Dict{Int,Vector{Int}}(),
+        Dict{Int,Vector{Int}}(),
+    )
+end
+
+function ScalarFunctionIterator(f::MOI.VectorAffineFunction)
+    d = Dict(i => Int[] for i = 1:MOI.output_dimension(f))
+    for (i, term) in enumerate(f.terms)
+        push!(d[term.output_index], i)
+    end
+    return ScalarFunctionIterator(f, d, Dict{Int,Vector{Int}}())
+end
+
+function ScalarFunctionIterator(f::MOI.VectorQuadraticFunction)
+    aff = Dict(i => Int[] for i = 1:MOI.output_dimension(f))
+    quad = Dict(i => Int[] for i = 1:MOI.output_dimension(f))
+    for (i, term) in enumerate(f.affine_terms)
+        push!(aff[term.output_index], i)
+    end
+    for (i, term) in enumerate(f.quadratic_terms)
+        push!(quad[term.output_index], i)
+    end
+    return ScalarFunctionIterator(f, aff, quad)
+end
+
 eachscalar(f::MOI.AbstractVectorFunction) = ScalarFunctionIterator(f)
 eachscalar(f::AbstractVector) = f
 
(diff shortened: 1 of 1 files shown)


blegat:
This does not fix #418 completely.
If you use getindex several time which is what happens if you do `collect` the it's still slow.
We should precompute an array `next::Vector{Int} ` that has the same size than the number of terms and `next[i]` is the index of the smallest `j > i` such that `terms[i].output_index == terms[i].output_index`.
This can be computed by going through the vector of terms just once. 
Then every `getindex` is efficient

odow:
Okay. I've rewritten all the `getindex` functions to use a cache of the output index to the index of the term in the array. This should be fast now, at the expense of some more memory.

src/Utilities/functions.jl
```diff
@@ -320,9 +320,47 @@ function scalar_type(::Type{MOI.VectorQuadraticFunction{T}}) where {T}
...
+function ScalarFunctionIterator(f::MOI.VectorOfVariables)
+    return ScalarFunctionIterator(
+        f,
+        Dict{Int,Vector{Int}}(),
+        Dict{Int,Vector{Int}}(),
+    )
+end
+
+function ScalarFunctionIterator(f::MOI.VectorAffineFunction)
+    d = Dict(i => Int[] for i = 1:MOI.output_dimension(f))
```
blegat, line 345: Why a `Dict` and not a `Vector` ?

blegat:
Creating a vector of vector will be much slower than a single vector as suggested in https://github.com/jump-dev/MathOptInterface.jl/pull/1257#issuecomment-788833578.
You can do it with a backward lookup
```julia
next = Vector{Int}(undef, length(terms))
last = zeros(Int, output_dimension)
for i in eachindex(terms)
    j = term[i].output_index
    if !iszero(last[j])
        next[last[j]] = i
    end
    last[j] = i
end
for j in eachindex(last)
    if !iszero(last[j])
        next[last[j]] = 0
    end
end
```
then you store `next` in the struct and `last`, you can just drop it.

odow:
This assumes that the function is canonicalized?

blegat:
> This assumes that the function is canonicalized?

No

odow:
@blegat's way is faster:
```Julia
julia> @time main(300);
  0.129429 seconds (2.69 k allocations: 123.049 MiB, 19.75% gc time)
```

blegat (approved):

PR #1264, opened 2021-03-05
Force Julia to specialize on ::Function arguments.

odow:
See the section in Performance Tips:
https://docs.julialang.org/en/v1/manual/performance-tips/index.html#Be-aware-of-when-Julia-avoids-specializing

This should help simplify (and improve) the precompilation files.

## Benchmark

Using the script from https://github.com/jump-dev/JuMP.jl/pull/2484

#### Before

```Julia
(base) oscar@Oscars-MBP scripts % '/Applications/Julia-1.6.app/Contents/Resources/julia/bin/julia' --project=. precompile.jl
 13.389848 seconds (43.71 M allocations: 2.494 GiB, 7.98% gc time, 24.12% compilation time)
```

#### After

```julia
(base) oscar@Oscars-MBP scripts % '/Applications/Julia-1.6.app/Contents/Resources/julia/bin/julia' --project=. precompile.jl
 11.452046 seconds (37.67 M allocations: 2.103 GiB, 8.23% gc time, 26.85% compilation time)
```

### DIFF
--- a/src/Utilities/copy.jl
+++ b/src/Utilities/copy.jl
@@ -2,7 +2,7 @@
 
 """
     automatic_copy_to(dest::MOI.ModelLike, src::MOI.ModelLike;
-                      copy_names::Bool=true, 
+                      copy_names::Bool=true,
                       filter_constraints::Union{Nothing, Function}=nothing)
 
 Use [`Utilities.supports_default_copy_to`](@ref) and
@@ -11,8 +11,8 @@ Use [`Utilities.supports_default_copy_to`](@ref) and
 apply the copy operation.
 
 If the `filter_constraints` arguments is given, only the constraints for which
-this function returns `true` will be copied. This function is given a 
-constraint index as argument. 
+this function returns `true` will be copied. This function is given a
+constraint index as argument.
 """
 function automatic_copy_to(
     dest::MOI.ModelLike,
@@ -402,8 +402,8 @@ Copy the constraints `cis_src` from the model `src` to the model `dest` and fill
 [`pass_attributes`] to copy the constraint attributes.
 
 If the `filter_constraints` arguments is given, only the constraints for which
-this function returns `true` will be copied. This function is given a 
-constraint index as argument. 
+this function returns `true` will be copied. This function is given a
+constraint index as argument.
 """
 function copy_constraints(
     dest::MOI.ModelLike,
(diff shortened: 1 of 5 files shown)


odow:
With the precompile statements
```Julia
(base) oscar@Oscars-MBP scripts % '/Applications/Julia-1.6.app/Contents/Resources/julia/bin/julia' --project=. precompile.jl
  8.106619 seconds (22.41 M allocations: 1.234 GiB, 5.04% gc time, 27.18% compilation time)
```

odow:
There are two commits here. I can either split them in two PRs, or we can use "Rebase and merge" to add them as separate commits to `master`.

src/Utilities/precompile.jl
```diff
@@ -0,0 +1,44 @@
...
+    vector_functions = (
+        MOI.VectorOfVariables,
+        MOI.VectorAffineFunction{T},
+        MOI.VectorQuadraticFunction{T},
+    )
+    constraints = vcat(
+        [(F, S) for F in scalar_functions, S in scalar_sets],
+        [(F, S) for F in vector_functions, S in vector_sets],
+    )
+    MOI.precompile_model(UniversalFallback{Model{T}}, constraints)
```
blegat, line 38: Shouldn't this be on a separate PR ?

src/precompile.jl
```diff
@@ -0,0 +1,68 @@
+function precompile_constraint(model, F, S)
+    Base.precompile(get, (model, ListOfConstraintIndices{F,S}))
```
blegat, line 2: This too ?

blegat:
If `::F, ...) where {F<:Function}` is now the default unless we have a good reason to do `::Function, ..)` in which case we should add a comment explaining why then we should probably add something saying this in the JuMP style guide

odow:
So it's not the case that we want this as the default. The heuristic is there for a good reason: most times it is more expensive to compile the method than the added runtime benefit. 

I've gone through and removed some of the cases that didn't do anything.

odow:
```
(base) oscar@Oscars-MBP scripts % '/Applications/Julia-1.6.app/Contents/Resources/julia/bin/julia' --project=. precompile.jl
 11.481412 seconds (37.60 M allocations: 2.098 GiB, 8.62% gc time, 26.81% compilation time)
```

odow:
Rebased onto `master` which contains the precompile statements:
```
(base) oscar@Oscars-MBP scripts % '/Applications/Julia-1.6.app/Contents/Resources/julia/bin/julia' --project=. precompile.jl
  8.253007 seconds (23.77 M allocations: 1.312 GiB, 5.16% gc time, 28.03% compilation time)
```

blegat (approved):

PR #1267, opened 2021-03-05
More flexible Model with StructOfConstraints

blegat:
This refactors the code of `AbstractModel` to make it more versatile.
Instead of having one field for each function type which contains a struct with one field for each set type, it now has a field `constraints` that contains one field for each function type, ...
As discussed in https://github.com/jump-dev/MathOptInterface.jl/issues/1261, by changing the type of the `constraints` field, we can now target many more types of models.

The base is set to https://github.com/jump-dev/MathOptInterface.jl/pull/1245 so that it shows the diff with that PR but this is going to be merged to master once https://github.com/jump-dev/MathOptInterface.jl/pull/1245 is merged.

- [x] Add an option to hardcode the type of the fields in `StructOfConstraints` instead of adding them as type parameter as otherwise the type can be huge. This option will be used by `@model`.
- [x] If there is only one function or only one set, drop the corresponding layer of `StructOfConstraint`
- [x] Docs
- [x] Tests

### DIFF
--- a/docs/src/submodules/Utilities/overview.md
+++ b/docs/src/submodules/Utilities/overview.md
@@ -82,9 +82,10 @@ julia> MOI.Utilities.@model(
            (MOI.VectorAffineFunction,),     # Typed vector functions
            true,                            # <:MOI.AbstractOptimizer?
        )
+MathOptInterface.Utilities.GenericOptimizer{T,MyNewModelFunctionConstraints{T}} where T
 
 julia> model = MyNewModel{Float64}()
-MOIU.GenericModel{Float64,MyNewModelFunctionConstraints{Float64}}
+MOIU.GenericOptimizer{Float64,MyNewModelFunctionConstraints{Float64}}
 ```
 
 !!! warning
@@ -110,6 +111,7 @@ julia> MOI.Utilities.@model(
            (MOI.VectorAffineFunction,),  # Typed vector functions
            true,  # is_optimizer
        )
+MathOptInterface.Utilities.GenericOptimizer{T,MathOptInterface.Utilities.VectorOfConstraints{MathOptInterface.VectorAffineFunction{T},MathOptInterface.Complements}} where T
 ```
 However, `PathOptimizer` does not support some `SingleVariable`-in-Set
 constraints, so we must explicitly define:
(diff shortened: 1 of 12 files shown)


docs/src/submodules/Utilities/overview.md
```diff
@@ -149,11 +151,11 @@ julia> model = MOI.Utilities.CachingOptimizer(
            MOI.Utilities.Model{Float64}(),
            PathOptimizer{Float64}(),
        )
-MOIU.CachingOptimizer{PathOptimizer{Float64},MOIU.Model{Float64}}
+MOIU.CachingOptimizer{MOIU.GenericOptimizer{Float64,MOIU.VectorOfConstraints{MOI.VectorAffineFunction{Float64},MOI.Complements}},MOIU.GenericModel{Float64,MOIU.ModelFunctionConstraints{Float64}}}
```
odow, line 154: Can we overload `show` for this model so that it prints `PathOptimizer{Float64}`? this issue going to cause the JuMP model printing to be horrible.

src/Utilities/model.jl
```diff
@@ -959,50 +839,46 @@ The model describing an linear program would be:
     )
 ```
 
-Let `MOI` denote `MathOptInterface`, `MOIU` denote `MOI.Utilities` and
-`MOIU.ConstraintEntry{F, S}` be defined as `MOI.Tuple{MOI.ConstraintIndex{F, S}, F, S}`.
-The macro would create the types:
+Let `MOI` denote `MathOptInterface`, `MOIU` denote `MOI.Utilities`:x.
```
odow, line 842: `:x`?

I think you just mean
```
Let `MOI` denote `MathOptInterface` and `MOIU` denote `MOI.Utilities`
```

src/Utilities/model.jl
```diff
@@ -1077,130 +996,37 @@ macro model(
...
-        field = _field(f)
-        code = quote
-            $code
-            function $MOIU.constraints(
-                model::$esc_model_name,
-                ::Type{<:$fun},
-                ::Type{S}
-            ) where S
-                return $MOIU.constraints(model.$field, S)
+            function $name{T,C}() where {T,C}
```
odow, line 1008: So maybe something like:
```julia
Base.show(io::IO, model::$(name){T,C}) where {T,C} = print(io, "$(name){T}")
```

odow (commented):
Do we have some benchmarks for obvious performance regressions?

docs/src/submodules/Utilities/overview.md
blegat: Isn't it printing the type `PathOptimizer`, not an object of type `PathOptimizer` So we should overload `show(:IO, ::Type{PathOptimizer{T}})`?

odow: Oh, yeah.
```Julia
Base.show(io::IO, ::Type{$(name){T,C}}) where {T,C} = print(io, "$(name){T}")
```

blegat: > this issue going to cause the JuMP model printing to be horrible.

Which printing ? JuMP does not print the type of the backend

odow: Anyone interacting with the backend:
https://jump.dev/JuMP.jl/dev/manual/models/#Backends
Or anyone interacting with MOI models like https://github.com/chkwon/PATHSolver.jl which are based on this.

blegat: If we define this show method then how can the user see what's the actual type?
We also have `const AffExpr =...` and we don't define printing for this type

odow: I withdraw my objection. We can always revisit if it becomes a pain, or down-stream solvers can define their own `show`.

src/Utilities/model.jl
blegat: What about this show? Do we implement or not?

odow: Not. I meant it to be the `Type{}` one.

odow (approved):
I ran the benchmarks. They're slightly slower, but I can tidy up after you've finished making large changes. It's probably just the precompile statements that need tweaking.

PR #1276, opened 2021-03-09
Remove union in IndexMap

blegat:
Follows what the comment says.

### DIFF
--- a/src/Utilities/CleverDicts.jl
+++ b/src/Utilities/CleverDicts.jl
@@ -384,11 +384,17 @@ function map_values!(f::Function, d::CleverDict)
     return
 end
 
-function Base.convert(::Type{CleverDict{K,V,F,I}}, d::CleverDict{K,V,F,I}) where {K,V,F,I}
+function Base.convert(
+    ::Type{CleverDict{K,V,typeof(key_to_index),typeof(index_to_key)}},
+    d::CleverDict{K,V,typeof(key_to_index),typeof(index_to_key)},
+) where {K,V}
     return d
 end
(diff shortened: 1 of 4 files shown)


blegat:
As briefly mentioned in https://github.com/jump-dev/MathOptInterface.jl/pull/1245#issuecomment-796557880,
there is an issue with the use of `CleverDicts` for constraint indices.
By default, `CleverDicts` uses an anonymous function for `inverse_map` but that is not an option if you want
to know the concrete type of the `CleverDict`.
What Gurobi, etc... do is to define a function that maps Int to variable indices and use the type of that function.
For `VectorOfConstraints` you cannot do that as it depends on the type of the constraints.
Currently, we use `Fix1` for that but it does not play well with inference because the constraint type is stored as a `DataType`:
https://github.com/jump-dev/MathOptInterface.jl/blob/e0f8b05682d98ae0d45812e1ca7bb6183053c148/src/Utilities/vector_of_constraints.jl#L29
This PR provides a non-breaking change that do not create the anonymous function by default and is non-breaking as if the user provides a custom function, the behavior is unchanged.

odow (approved):
Nice. Does this improve the `stress_precompile` benchmark?

blegat:
Before:
```julia
julia> @time Foo.example_diet()
 18.208597 seconds (26.08 M allocations: 1.303 GiB, 4.28% gc time)
```
After:
```julia
julia> @time Foo.example_diet()
 17.757552 seconds (25.81 M allocations: 1.291 GiB, 4.34% gc time)
```

PR #1283, opened 2021-03-11
Add MOI.attribute_value_type

odow:
Originally proposed all the way back in #31.

## Problem

Lots of `MOI.get` fail to infer. For example, getting the termination status
of a caching optimizer is `Any` instead of `TerminationStatusCode`.

```julia
julia> b = backend(model)
MOIU.CachingOptimizer{MOI.AbstractOptimizer, MOIU.UniversalFallback{MOIU.Model{Float64}}}
in state ATTACHED_OPTIMIZER
in mode AUTOMATIC
  with 1 optimizer attribute
with optimizer MOIB.LazyBridgeOptimizer{GLPK.Optimizer}
  with 0 variable bridges
  with 0 constraint bridges
  with 0 objective bridges
  with inner model A GLPK model

julia> @code_warntype MOI.get(b, MOI.TerminationStatus())
Variables
  #self#::Core.Const(MathOptInterface.get)
  model::MathOptInterface.Utilities.CachingOptimizer{MathOptInterface.AbstractOptimizer, MathOptInterface.Utilities.UniversalFallback{MathOptInterface.Utilities.Model{Float64}}}
  attr::Core.Const(MathOptInterface.TerminationStatus())

Body::Any
1 ─ %1  = MathOptInterface.is_set_by_optimize::Core.Const(MathOptInterface.is_set_by_optimize)
│   %2  = (%1)(attr)::Core.Const(true)
│         %2
│   %4  = MathOptInterface.Utilities.state(model)::MathOptInterface.Utilities.CachingOptimizerState
│   %5  = (%4 == MathOptInterface.Utilities.NO_OPTIMIZER)::Bool
└──       goto #4 if not %5
2 ─ %7  = MathOptInterface.TerminationStatus::Core.Const(MathOptInterface.TerminationStatus)
│   %8  = (%7)()::Core.Const(MathOptInterface.TerminationStatus())
│   %9  = (attr == %8)::Core.Const(true)
│         %9
│   %11 = MathOptInterface.OPTIMIZE_NOT_CALLED::Core.Const(MathOptInterface.OPTIMIZE_NOT_CALLED)
└──       return %11
3 ─       Core.Const(:(MathOptInterface.PrimalStatus))
│         Core.Const(:((%13)()))
│         Core.Const(:(attr == %14))
│         Core.Const(:(%15))
│         Core.Const(:(MathOptInterface.NO_SOLUTION))
│         Core.Const(:(return %17))
│         Core.Const(:(MathOptInterface.DualStatus))
│         Core.Const(:((%19)()))
│         Core.Const(:(attr == %20))
│         Core.Const(:(%21))
│         Core.Const(:(MathOptInterface.NO_SOLUTION))
│         Core.Const(:(return %23))
│         Core.Const(:(Base.string("Cannot query ", attr, " from caching optimizer because no")))
│         Core.Const(:(%25 * " optimizer is attached."))
└──       Core.Const(:(MathOptInterface.Utilities.error(%26)))
4 ┄ %28 = Base.getproperty(model, :optimizer_to_model_map)::MathOptInterface.Utilities.IndexMap
│   %29 = MathOptInterface.get::Core.Const(MathOptInterface.get)
│   %30 = Base.getproperty(model, :optimizer)::Union{Nothing, MathOptInterface.AbstractOptimizer}
│   %31 = (%29)(%30, attr)::Any
│   %32 = MathOptInterface.Utilities.map_indices(%28, %31)::Any
└──       return %32
5 ─       Core.Const(:(MathOptInterface.get))
│         Core.Const(:(Base.getproperty(model, :model_cache)))
│         Core.Const(:((%34)(%35, attr)))
└──       Core.Const(:(return %36))
```

Whereas with this PR we have
```julia
julia> @code_warntype MOI.get(b, MOI.TerminationStatus())
Variables
  #self#::Core.Const(MathOptInterface.get)
  model::MathOptInterface.Utilities.CachingOptimizer{MathOptInterface.AbstractOptimizer, MathOptInterface.Utilities.UniversalFallback{MathOptInterface.Utilities.Model{Float64}}}
  attr::Core.Const(MathOptInterface.TerminationStatus())

Body::MathOptInterface.TerminationStatusCode
1 ─ %1  = MathOptInterface.is_set_by_optimize::Core.Const(MathOptInterface.is_set_by_optimize)
│   %2  = (%1)(attr)::Core.Const(true)
│         %2
│   %4  = MathOptInterface.Utilities.state(model)::MathOptInterface.Utilities.CachingOptimizerState
│   %5  = (%4 == MathOptInterface.Utilities.NO_OPTIMIZER)::Bool
└──       goto #4 if not %5
2 ─ %7  = MathOptInterface.TerminationStatus::Core.Const(MathOptInterface.TerminationStatus)
│   %8  = (%7)()::Core.Const(MathOptInterface.TerminationStatus())
│   %9  = (attr == %8)::Core.Const(true)
│         %9
│   %11 = MathOptInterface.OPTIMIZE_NOT_CALLED::Core.Const(MathOptInterface.OPTIMIZE_NOT_CALLED)
└──       return %11
3 ─       Core.Const(:(MathOptInterface.PrimalStatus))
│         Core.Const(:((%13)()))
│         Core.Const(:(attr == %14))
│         Core.Const(:(%15))
│         Core.Const(:(MathOptInterface.NO_SOLUTION))
│         Core.Const(:(return %17))
│         Core.Const(:(MathOptInterface.DualStatus))
│         Core.Const(:((%19)()))
│         Core.Const(:(attr == %20))
│         Core.Const(:(%21))
│         Core.Const(:(MathOptInterface.NO_SOLUTION))
│         Core.Const(:(return %23))
│         Core.Const(:(Base.string("Cannot query ", attr, " from caching optimizer because no")))
│         Core.Const(:(%25 * " optimizer is attached."))
└──       Core.Const(:(MathOptInterface.Utilities.error(%26)))
4 ┄ %28 = Base.getproperty(model, :optimizer_to_model_map)::MathOptInterface.Utilities.IndexMap
│   %29 = MathOptInterface.get::Core.Const(MathOptInterface.get)
│   %30 = Base.getproperty(model, :optimizer)::Union{Nothing, MathOptInterface.AbstractOptimizer}
│   %31 = (%29)(%30, attr)::Any
│   %32 = MathOptInterface.return_type::Core.Const(MathOptInterface.return_type)
│   %33 = (%32)(attr)::Core.Const(MathOptInterface.TerminationStatusCode)
│   %34 = Core.typeassert(%31, %33)::MathOptInterface.TerminationStatusCode
│   %35 = MathOptInterface.Utilities.map_indices(%28, %34)::MathOptInterface.TerminationStatusCode
└──       return %35
5 ─       Core.Const(:(MathOptInterface.get))
│         Core.Const(:(Base.getproperty(model, :model_cache)))
│         Core.Const(:((%37)(%38, attr)))
└──       Core.Const(:(return %39))
```

Unfortunately, attributes like `MOI.VariablePrimal` fail to infer, because we don't know if they are `Float64` or not.

## Alternatives

If we won't want to add `MOI.return_type`, I can move it into CachingOptimizer.


src/attributes.jl
```diff
@@ -610,14 +618,15 @@ An optimizer attribute for the `Vector{AbstractOptimizerAttribute}` of all optim
 struct ListOfOptimizerAttributesSet <: AbstractOptimizerAttribute end
 
 """
-    SolverName()
+    SolverName()::String
```
joehuchette, line 621: This way of documenting the return value of `return_type(::SolverName)` is confusing; it reads like the constructor for `SolverName` returns a String.

joehuchette (commented):
Is this a general problem that a number of optimizers run into? Or is it mostly just a problem with the CachingOptimizer, because of it's extra layer or redirection?

odow:
This is mainly `CachingOptimizer` with this problem, because it has a backend of `Union{Nothing,AbstractOptimizer}` which needs dynamic dispatch.

But more widely, standardizing the return types of `MOI.xxx` functions would be a good thing. There are lots of things like 
https://github.com/jump-dev/MathOptInterface.jl/pull/1282 which cause issues if the return types are not explicitly stated.

blegat:
Alternatively, we could avoid having a non-concrete type for the optimizer as suggested in https://github.com/jump-dev/JuMP.jl/pull/2520.
We have a non-concrete type for the JuMP backend so we need to annotate all JuMP function to avoid these issues.
Now we need to workaround the same issues at the level of the CachingOptimizer.
Only one non-concrete type is needed to be able to implement `set_optimizer` so why have two ?

mlubin:
> we need to annotate all JuMP function to avoid these issues

If these types are properly defined in MOI, then we can add unit tests that solvers return the correct types. This prevents surprises when using the solver from JuMP.

In terms of bikeshedding, I'd suggest `attribute_value_type` instead of `return_type`. "Return type" applies to functions, and attributes aren't functions.

blegat:
Independently of the alternative, this could indeed be useful, `attribute_value_type` seems more appropriate indeed.
We could have
```julia
attribute_value_type(::ModelLike, attr::AnyAttribute) = attribute_value_type(attr)
attribute_value_type(::ModelLike, attr::Union{MOI.VariablePrimal, MOI.VariablePrimalStart, ...}) = MOI.get(model, CoefficientType())
```

joehuchette:
I'm a +1 to standardizing the "attribute value types" of MOI attributes, and having MOIT unit tests for this expected behavior.

src/Test/UnitTests/attributes.jl
```diff
@@ -128,3 +128,103 @@ function solve_time(model::MOI.ModelLike, config::TestConfig)
...
+    :SolveTime,
+    :TerminationStatus,
+    :TimeLimitSec,
+)
+    f = Symbol("test_attribute_$(attr)")
+    @eval begin
+        function $(f)(model::MOI.ModelLike, config::TestConfig)
+            MOI.empty!(model)
+            MOI.optimize!(model)
+            attribute = MOI.$(attr)()
```
blegat, line 158: Maybe skip the test is `supports` is false

odow: For now I've made these tests opt-in individually (They aren't in the `unittest` testset). We probably want to wait a few releases for solvers to adapt before enforcing it and breaking every solver.

blegat: What are the changes needed in the solver ? I assume they already have objects of the correct type returned, don't they ?

odow: They will return the correct object, but they may not infer correctly.

src/attributes.jl
```diff
@@ -879,6 +901,7 @@ struct ListOfConstraintIndices{F,S} <: AbstractModelAttribute end
 A model attribute for the number of constraints of the type `F`-in-`S` present in the model.
 """
 struct NumberOfConstraints{F,S} <: AbstractModelAttribute end
+attribute_value_type(::NumberOfConstraints) = Int
```
blegat, line 904: Why do we use `Int64` in the indices if the type of `NumberOfConstraints` is `Int`?

odow: I think the argument was just for consistency. I can make `NumberOfConstraints` and `NumberOfVariables` `Int64` if we want. 

I don't know if anyone has actually solved a model with more than `typemax(Int32)` variables or constraints in JuMP, because CPLEX and Gurobi only wrap the 32-bit API of the solvers, and no one has complained...

blegat: We could also use `Union{Int64, Int}`, on a 64-bit computer, we have `Union{Int64, Int} == Int64` anyway.
If we don't allow returning a number of variables of `Int64`, it's no use to allow the indices to be `Int64`.

odow: Are there any solvers which actually return `Int64`? We could always change it to `Int64` later, if we have evidence that people are solving problems on 32-bit machines with more than `typemax(Int32)` variables...

joaquimg: I know that people have complained to FICO Xpress about the Int32 limit on constraints and they added Int64 methods.
Ex. https://www.fico.com/fico-xpress-optimization/docs/latest/solver/optimizer/HTML/XPRSaddcols.html

blegat: Anyway, we should use the same type than the indices. If we were to use `Int` instead of `Int64` then we should change indices as well

odow:
@blegat how is this. I changed to `Int64`.

odow:
@blegat bump.

src/Utilities/model.jl
```diff
@@ -1027,7 +1028,7 @@ for (loop_name, loop_super_type) in [
                 MOI.ScalarAffineFunction{T},
                 MOI.ScalarQuadraticFunction{T},
             }
-            num_variables_created::Int64
+            num_variables_created::Int
```
blegat, line 1031: Why not keep `Int` here ?

src/Test/UnitTests/attributes.jl
```diff
@@ -128,3 +128,144 @@ function solve_time(model::MOI.ModelLike, config::Config)
...
+                    return
+                end
+                MOI.optimize!(model)
+            end
+            T = MOI.attribute_value_type(attribute)
+            try
+                MOI.get(model, attribute)
+            catch err
+                if err isa ArgumentError
+                    return  # Solver does not support accessing the attribute.
```
blegat, line 169: `@test !MOI.supports(...)` ?

blegat (approved):

PR #1285, opened 2021-03-13
Update CleverDicts.jl

amburosesekar:
version issue  not working when we use v"1.4.0-rc2.0". its true but consider as false and create/misguide to shows error , when we use optimization. so we update simply
(VersionNumber)(split(string(VERSION),"-")[1]) >= v"1.4.0".

### DIFF
--- a/src/Utilities/CleverDicts.jl
+++ b/src/Utilities/CleverDicts.jl
@@ -285,7 +285,7 @@ function Base.iterate(
         else
             el, i = itr
             new_el = c.inverse_hash(Int64(el))::K => c.vector[el]::V
-            @static if VERSION >= v"1.4.0"
+            @static if (VersionNumber)(split(string(VERSION),"-")[1]) >= v"1.4.0"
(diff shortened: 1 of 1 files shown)


joaquimg:
Why would one use 1.4.0rc2 instead of 1.4.0 itself?

blegat:
Can't we do something like `VERSION >= v"1.4.0-"

amburosesekar:
> Why would one use 1.4.0rc2 instead of 1.4.0 itself?

The problem is , i have downloaded prerelease version and installed in pc. but this problem happens when i used JuMP optimization for simple optimization task, its shows error "No Close match found --mathOptInterface.Utilities.CleverDicts.State(::Tuple{UInt64,Int64})". so totally confused where the bug present after decode the package , i understood where the problem is persist. in my suggestion used to all julia users, may be syntax simple as @blegat suggestion

blegat:
v"1.4.0-" will not work for versions before the commit changing the behavior in Julia. You should find which commit this is in Julia

PR #1286, opened 2021-03-13
Pass copy operation for non-variable constraints

blegat:
If the `constraints` field of the model has a more efficient copy_to than adding constraints one by one (e.g. SparseMatrixCSC) then even if it's behind layers of bridges, CachingOptimizer and UniversalFallback, it should be copied (if not bridged), not added one by one.
This PR addresses that.

- [x] Bridge layer
- [x] Tests
- [x] Find out what to do with unsupported attributes

See https://github.com/jump-dev/MathOptInterface.jl/issues/1261

Requires https://github.com/jump-dev/MathOptInterface.jl/pull/1267

### DIFF
--- a/src/Bridges/bridge_optimizer.jl
+++ b/src/Bridges/bridge_optimizer.jl
@@ -320,6 +320,47 @@ function MOI.supports(
     return MOI.supports(b.model, attr)
 end
 
+function MOIU.pass_nonvariable_constraints(
+    dest::AbstractBridgeOptimizer,
+    src::MOI.ModelLike,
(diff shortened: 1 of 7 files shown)


odow:
Any benchmarks?

odow (approved):

blegat:
I don't believe this will provide any gain yet but once we have https://github.com/jump-dev/MathOptInterface.jl/pull/1287, it will allow efficient `copy_to` to go thrown to the matrix cache if the affine constraints are not bridged.

Issue #1261, opened 2021-03-03
Versatile model type


blegat:
Currently, MOI only provides one implementation of model: `MOIU.AbstractModel` that are created with `MOIU.@model` with the common choice of `MOIU.Model`.
This represents models in the following way:
We have a struct containing:
* The objective which can be `SingleVariable`, `ScalarAffineFunction` or `ScalarQuadraticFunction`
* A datastructure for `SingleVariable` constraints
* One field per supported function type containing
  - One field per supported set type containing
    * a datastructure for storing each constraint of that type.
This is useful to use as the cache used by JuMP to store the model as it gives good performance for all MOI operations. However, when copying this model to a solver, some work should be done by the solver to transfrom it into its canonical form.
If we look at it more closely, the form required by the solvers are a combination of the following:
* An objective type
* A datastructure for `SingleVariable` constraints
* One or more separate datastructures for subsets of constraint types. For example, some conic solvers have one datastructure for `VectorAffineFunction`-in-`Zeros` and one for `VectorAffineFunction`-in-`K` in other cones `K`.

LP solvers such as Clp, GLPK, HiGHS need all the constraints mixed up in a single matrix stored either row-wise or column-wise.

If we refactor `AbstractModel`, it could cover all these use case, simply by being parametrized by:
* The type of objective
* The type of the datastructure for `SingleVariable` constraints
* The type of the datastructure for the other constraints.

This third type is either
* A dispatching type that has subcontainers and depending on the type of the set and function, decide in which subcontainer the constraint is stored.
* A container storing the constraints as a `VectorOfConstraints`: https://github.com/jump-dev/MathOptInterface.jl/pull/1245
* A container storing the constraints as a `MatrixOfConstraints` in either a `MutableSparseMatrixCSC` or a `Transpose{<:MutableSparseMatrixCSC}`. For `MutableSparseMatrixCSC`, incrementally adding the constraints is not supported. See https://github.com/jump-dev/MatrixOptInterface.jl/blob/master/src/sparse_matrix.jl for what it would look like

Then, each solver would be able to significantly simplify it's copy_to function by doing, e.g. LP solvers could do
```julia
const MODEL = MOI.Utilities.GenericModel{MOI.ScalarAffineFunction{Float64}, ContinuousVariables, MatrixOfConstraints{..., SparseMatrixCSC{..., MOI.ZeroBasedIndexing}}
function MOI.copy_to(dest::Optimizer, src::MODEL)
[...]

PR #1287, opened 2021-03-14
MatrixOfConstraints

blegat:
Current benchmark:
```julia
import MathOptInterface
const MOI = MathOptInterface
const MOIU = MOI.Utilities
import Clp
function bench(n, m)
    model = MOIU.Model{Float64}()
    x = MOI.add_variables(model, n)
    fx = MOI.SingleVariable.(x)
    ofx = 1.0 .* fx
    for i in 1:n
        MOI.add_constraint(model, ofx[i], MOI.LessThan(1.0))
        MOI.add_constraint(model, ofx[i], MOI.GreaterThan(1.0))
    end
    s = sum(ofx)
    for i in 1:n
        MOI.add_constraint(model, s, MOI.EqualTo(1.0))
    end
    optimizer = MOI.Bridges.full_bridge_optimizer(
        MOIU.CachingOptimizer(
            M(),
            Clp.Optimizer(),
        ),
        Float64,
    )
    @time MOI.copy_to(optimizer, model)
    @time MOIU.attach_optimizer(optimizer.model)
    optimizer.model
end
```

With this branch, Cbc#od/moi10 and `M=MOIU.Model{Float64}`, I get:
```
julia> bench(1000, 1000);
  0.035744 seconds (18.71 k allocations: 32.124 MiB)
  0.075348 seconds (429 allocations: 42.859 MiB)
```

With this branch, Cbc#bl/matrix and `M=Cbc.Model`, I get:
```
julia> bench(1000, 1000);
  0.011251 seconds (698 allocations: 551.617 KiB)
  0.000164 seconds (105 allocations: 146.023 KiB)
```

What really matters is the sum of the two. The speedup is approximately 10x.
~~We already have a significant decrease but I think we could have a number of allocation which does not scale linearly with `n` and `m` so I'm trying to figure it out.~~
The number of allocations is a constant + some logarithmic on `n` and `m` which is expected (e.g. because of `push` in `caches`).

See #1261

Requires #1267, #1286


odow:
What's the status of this? Can we make a new MOI release without it?

blegat:
There is still some work needed, we don't need to wait for it for a new MOI release

src/Test/config.jl
```diff
@@ -2,6 +2,7 @@ struct TestConfig{T<:Real}
     atol::T # absolute tolerance for ...
     rtol::T # relative tolerance for ...
     solve::Bool # optimize and test result
+    query_number_of_constraints::Bool # can get `MOI.NumberOfConstraints` attribute
```
odow, line 5: This should be a separate PR.

odow:
As another benchmark: how is the first solve? Compilation times are our big killer for this.

src/Test/config.jl
blegat: Done in https://github.com/jump-dev/MathOptInterface.jl/pull/1327

src/Utilities/copy.jl
```diff
@@ -69,6 +69,25 @@ function IndexMap(n = 0)
         DoubleDicts.IndexDoubleDict(),
     )
 end
+function _identity_constraints_map(
+    index_map::MOIU.DoubleDicts.IndexWithType{F,S},
```
odow, line 73: Same comment about argument order. Model first?

src/Utilities/copy.jl
```diff
@@ -608,6 +627,8 @@ function try_constrain_variables_on_creation(
     single_variable_not_added
 end
 
+function final_touch(::MOI.ModelLike, idxmap) end
```
odow, line 630: This needs a docstring.

src/Utilities/matrix_of_constraints.jl
```diff
@@ -0,0 +1,614 @@
+abstract type ProductOfSets{T} end
```
odow, line 1: This needs a docstring explaining how it stores the constraints.

src/Utilities/model.jl
```diff
@@ -14,14 +14,21 @@ function MOI.get(model::AbstractModel, ::MOI.NumberOfVariables)::Int64
...
     vi = VI(model.num_variables_created += 1)
     push!(model.single_variable_mask, 0x0)
-    push!(model.lower_bound, zero(T))
-    push!(model.upper_bound, zero(T))
+    push!(model.lower_bound, _no_lower_bound(T))
+    push!(model.upper_bound, _no_upper_bound(T))
     if model.variable_indices !== nothing
         push!(model.variable_indices, vi)
     end
+    _add_variable(model.constraints)
```
odow, line 31: It seems weird to force other types to overload `_add_variable`. Why not just overload `MOI.add_variable(model.constraints)`?

odow (commented):
This is quite hard to review because there is a lot of new types being added
- ProductOfSets
- MixOfScalarSets
- OrderedProductOfSets
- OrderedProductOfScalarSets
- UnevenIterator
- Box
- MatrixOfConstraints
- IdentityMap
- MutableSparseMatrixCSC

Can we break it down into smaller pieces? It's not obvious why we need so many different structures, or how they inter-relate.

src/Utilities/model.jl
blegat: Then `VectorOfConstraints` should return a `VariableIndex` but currently it does not keep track of this information.

src/Utilities/copy.jl
blegat: Let me know what you think for the name. Another option would be `end_of_copy`

test/Utilities/copy.jl
```diff
@@ -653,6 +653,7 @@ struct OnlyCopyConstraints{F,S} <: MOI.ModelLike
         return new{F,S}(MOIU.VectorOfConstraints{F,S}())
     end
 end
+function MOI.Utilities._add_variable(::OnlyCopyConstraints) end
```
odow, line 656: If this is a public API that people need to implement, it shouldn't begin with `_`.

src/Utilities/matrix_of_constraints.jl
```diff
@@ -0,0 +1,357 @@
+# Constants: stored in a `Vector` or `Box` or any other type implementing:
+# `empty!`, `resize!` and `load_constants`.
+
+"""
+    load_constants(constants, offset, func_or_set)
+
+This function loads the constants of `func_or_set` in `constants` at an offset
+of `offset`. The storage should be preallocated with `resize!` before calling
+this function.
```
odow, line 9: What is `offset`? "Where `offset` is the number of constants that have previously been loaded"?

src/Utilities/matrix_of_constraints.jl
```diff
@@ -0,0 +1,357 @@
...
+    empty!(b.lower)
+    empty!(b.upper)
+    return b
+end
+function Base.resize!(b::Box, n)
+    resize!(b.lower, n)
+    resize!(b.upper, n)
+    return
+end
+function load_constants(b::Box{T}, offset, set) where {T}
```
odow, line 47: Should `set` have some type restrictions here?

src/Utilities/matrix_of_constraints.jl
```diff
@@ -0,0 +1,357 @@
...
+
+"""
+    rows(model::MatrixOfConstraints, ci::MOI.ConstraintIndex)
+
+Return the rows corresponding to the constraint of index `ci`. If it is a
+vector constraint, this is a `UnitRange`, otherwise, this is an integer.
+"""
+rows(model::MatrixOfConstraints, ci::MOI.ConstraintIndex) = rows(model.sets, ci)
+
+function affine_function_type(
```
odow, line 137: Is this a public function? If not, it should start with `_`.

src/Utilities/matrix_of_constraints.jl
```diff
@@ -0,0 +1,357 @@
...
+    ::Type{T},
+    ::Type{<:MOI.AbstractScalarSet},
+) where {T}
+    return MOI.ScalarAffineFunction{T}
+end
+function affine_function_type(
```
odow, line 143: A pet peeve of mine: new lines between methods?

src/Utilities/matrix_of_constraints.jl
```diff
@@ -0,0 +1,357 @@
...
+a `src::MOI.Utilities.Model` to a bridged model with a `MatrixOfConstraints`.
+For all the types that are not bridged, the constraints will be copied
+with `pass_nonvariable_constraints` hence the functions stored in
+`caches` are exactly the same as the ones stored in `src`.
+This is ok since this is only during the `copy_to` operation during which `src`
+cannot be modified.
+On the other hand, for the types that are bridged, the functions added
+may contain duplicates even if the functions did not contain duplicates in
+`src` so duplicates are removed with `MOI.Utilities.canonical`.
+"""
```
odow, line 106: I started making inline notes, but I'll stop. The API needed to implement MatrixOfConstraints is quite large, and needs to be documented.

This docstring, plus a page in the documentation, should mention what functions need to be implemented for the different `coefficients`, `constants`, and `sets` (e.g., `set_types`, `rows`, ...) fields.


See the comment at the top of this file, and the notes in `ProductOfSets.jl`

src/Utilities/matrix_of_constraints.jl
```diff
@@ -0,0 +1,357 @@
...
+    filter_constraints::Union{Nothing,Function} = nothing,
+)
+    set_number_of_columns(
+        dest.coefficients,
+        MOI.get(src, MOI.NumberOfVariables()),
+    )
+
+    for (F, S) in constraint_types
+        _allocate_constraints(dest, src, index_map, F, S, filter_constraints)
+    end
```
odow, line 340: `return`

src/Utilities/matrix_of_constraints.jl
```diff
@@ -0,0 +1,357 @@
...
+    )
+    if filter_constraints !== nothing
+        filter!(filter_constraints, cis_src)
+    end
+    for ci_src in cis_src
+        func = MOI.get(src, MOI.CanonicalConstraintFunction(), ci_src)
+        set = MOI.get(src, MOI.ConstraintSet(), ci_src)
+        push!(model.are_indices_mapped, length(model.caches) + 1)
+        index_map[ci_src] = _add_constraint(model, i, index_map, func, set)
+    end
```
odow, line 282: `return`

src/Utilities/product_of_sets.jl
```diff
@@ -0,0 +1,308 @@
...
+by [`set_types`](@ref)..
+"""
+set_index(sets::ProductOfSets, ::Type{S}) where {S<:MOI.AbstractSet} = nothing
+
+"""
+    set_types(sets::ProductOfSets)
+
+Return the list of the types of the sets allowed in the cartesian product.
+"""
+function set_types end
```
odow, line 21: This really belong in `MatrixOfConstraints`.

src/Utilities/matrix_of_constraints.jl
```diff
@@ -0,0 +1,357 @@
...
+        MOI.get(src, MOI.NumberOfVariables()),
+    )
+
+    for (F, S) in constraint_types
+        _allocate_constraints(dest, src, index_map, F, S, filter_constraints)
+    end
+end
+
+function final_touch(model::MatrixOfConstraints, index_map)
+    num_rows = number_of_rows(model.sets)
```
odow, line 344: This needs documenting

src/Utilities/matrix_of_constraints.jl
```diff
@@ -0,0 +1,357 @@
...
+    resize!(model.constants, num_rows)
+    set_number_of_rows(model.coefficients, num_rows)
```
odow, line 346: This needs documenting

src/Utilities/matrix_of_constraints.jl
```diff
@@ -0,0 +1,357 @@
+# Constants: stored in a `Vector` or `Box` or any other type implementing:
+# `empty!`, `resize!` and `load_constants`.
```
odow, line 2: This needs to be documented better.

src/Utilities/product_of_sets.jl
```diff
@@ -0,0 +1,308 @@
...
+"""
+abstract type ProductOfSets{T} end
+
+"""
+    set_index(sets::ProductOfSets, ::Type{S}) where {S<:MOI.AbstractSet}
+
+Return an integer corresponding to the index of the set type in the list given
+by [`set_types`](@ref)..
+"""
+set_index(sets::ProductOfSets, ::Type{S}) where {S<:MOI.AbstractSet} = nothing
```
odow, line 14: This needs to be documented in MatrixOfConstraints

src/Utilities/sparse_matrix.jl
```diff
@@ -0,0 +1,118 @@
+import SparseArrays
+
+abstract type AbstractIndexing end
+struct ZeroBasedIndexing <: AbstractIndexing end
+struct OneBasedIndexing <: AbstractIndexing end
+
+first_index(::ZeroBasedIndexing) = 0
```
odow, line 7: This should all be private

odow (commented):
This needs documentation added to the `Utilities` part of the manual.

I wonder about making a subdirectory in `src/Utilities` to contain these. That would isolate the code a little more, and better document what goes with what. 

I'm also keen to see what impact this has on our time-to-first-solve issue.

test/Utilities/copy.jl
blegat: This needs to be implemented by the type of the `constraints` field but it's currently not public API.
It's best we don't let user create new types at the moment until we converge. This is similar to `_delete_variables`, `_deleted_constraints` and `_throw_if_cannot_delete`.

odow (approved):
I have some reservations about adding yet more complexity to MOI. And it would be better if the documentation clearly identified which functions needed to be implemented for which bits, but it may be easier to merge and keep iterating in new PRs. We still have time to break things, so we should merge and then update solvers in branches to figure out what tweaks need to be made.

docs/src/submodules/Utilities/reference.md
```diff
@@ -79,6 +79,47 @@ Utilities.load
...
+Utilities.set_number_of_rows
+Utilities.load_terms
+Utilities.final_touch
+```
+
+### Product of sets
+
+```@docs
+Utilities.ProductOfSets
+Utilities.set_index
```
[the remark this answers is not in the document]
blegat: I made this into a well-defined API to make the code more understandable. We will need yet another one for list of sets that are not known at compile time, e.g., DiffOpt but we might have enough with 3, can't say for now.

src/Utilities/product_of_sets.jl
```diff
@@ -0,0 +1,331 @@
...
+which allows a more efficient implementation.
+"""
+abstract type OrderedProductOfScalarSets{T} <: OrderedProductOfSets{T} end
+macro product_of_scalar_sets(name, set_types...)
+    esc_name = esc(name)
+    T = esc(:T)
+    type_def = :(struct $esc_name{$T} <: $MOIU.OrderedProductOfScalarSets{$T}
+        num_rows::Vector{Int}
+        function $esc_name{$T}() where {$T}
+            return new(zeros(Int, $(1 + length(set_types))))
```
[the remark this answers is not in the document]
blegat: Yes, that not so much. I used to store the dimension for the scalar sets for `OrderedProductOfSets` so `OrderedProductOfScalarSets` was a big win but now even `OrderedProductOfSets` don't store the dimension of scalar sets so we could indeed get rid of `OrderedProductOfScalarSets` (although it's just a few lines of code).

blegat:
> And it would be better if the documentation clearly identified which functions needed to be implemented for which bits, but it may be easier to merge and keep iterating in new PRs. We still have time to break things, so we should merge and then update solvers in branches to figure out what tweaks need to be made.

Yes I agree. It's ready so that people can start play with it and we can iterate a bit before releasing.

src/Utilities/product_of_sets.jl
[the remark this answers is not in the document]
odow: Part of this is that it's just a few lines of code, but much more code to compile.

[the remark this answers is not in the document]
blegat: Yes, I think we can remove it

PR #1295, opened 2021-04-08
Add a way to overload MOI.empty for '@model's

odow:
I need this for https://github.com/jump-dev/AmplNLWriter.jl/pull/125


odow:
@blegat, MutableArithmetics is failing on nightly.

odow:
Closing because I did something else in AmplNLWriter. This may still be needed, but not yet.

PR #1296, opened 2021-04-16
Add LinearAlgebra fallbacks for adjoint and transpose

odow:
Closes https://github.com/jump-dev/MutableArithmetics.jl/issues/96


test/Utilities/mutable_arithmetics.jl
```diff
@@ -121,6 +121,7 @@ function all_tests(T::Type, a, b, c, d, e, f, g)
         "matrix_uniform_scaling",
         "symmetric_matrix_uniform_scaling",
         "matrix_vector_division",
+        "transpose",
```
blegat, line 124: Should be fixed by https://github.com/jump-dev/MutableArithmetics.jl/pull/97

odow:
Yeah I got a bit confused with what this was doing.

PR #1298, opened 2021-04-17
doc typo

matbesancon:
There was a slight error in the "implementing a solver" page, with a ref broken. I didn't find the equivalent section so the ref is now removed


odow (approved):
I'll tag a new MutableArithmetics release so the tests pass before merging

PR #1299, opened 2021-04-19
Implement print for MOI.ModelLike

odow:
## TODO
 - [x] Documentation
 - [x] Nonlinear printing
 - [x] Decide whether to move to MOI.Utilities. This isn't really part of the interface, so it should probably go in `Utilities`.
 
## Example

Open to suggestions for improvements. Here is what it looks like at present. Since it's MOI, I wanted to explicitly state which function-set pairs were in the model. I'd also be in favor of adding these to JuMP to make it more explicit.

```Julia
MOI.Utilities.loadfromstring!(
    model,
    """
    variables: x, y, z
    minobjective: x + 2 + 3.1*y + -1.2*z
    c1: x >= 0.1
    c2: y in ZeroOne()
    c2: z in Integer()
    c3: [x, y] in SecondOrderCone(2)
    c4: [1, x, y] in SecondOrderCone(2)
    c4: [1.0 * x * x, y, 1] in ExponentialCone()
    c4: [1, 1.0 * x * x, y] in ExponentialCone()
    c2: x in ZeroOne()
    c5: 2.0 * x * x + y + -1 * z <= 1.0
    c5: x + x >= 1.0
    c5: x + x in Interval(1.0, 2.0)
    c5: x + -1 * y == 0.0
    """,
)

julia> print(model)
Minimize ScalarAffineFunction{Float64}:
 2.0 + 1.0 x + 3.1 y - 1.2 z

Subject to:

ScalarAffineFunction{Float64}-in-EqualTo{Float64}
 0.0 + 1.0 x - 1.0 y == 0.0

ScalarAffineFunction{Float64}-in-GreaterThan{Float64}
 0.0 + 2.0 x >= 1.0

ScalarAffineFunction{Float64}-in-Interval{Float64}
 0.0 + 2.0 x ∈ [1.0, 2.0]

ScalarQuadraticFunction{Float64}-in-LessThan{Float64}
 0.0 + 1.0 y - 1.0 z + 2.0 x² <= 1.0

VectorOfVariables-in-SecondOrderCone
 ┌ ┐
 │x│
 │y│
 └ ┘ ∈ SecondOrderCone(2)

VectorAffineFunction{Float64}-in-SecondOrderCone
 ┌           ┐
 │1.0        │
 │0.0 + 1.0 x│
 │0.0 + 1.0 y│
 └           ┘ ∈ SecondOrderCone(2)

VectorQuadraticFunction{Float64}-in-ExponentialCone
 ┌            ┐
 │0.0 + 1.0 x²│
 │0.0 + 1.0 y │
 │1.0         │
 └            ┘ ∈ ExponentialCone()
 ┌            ┐
 │1.0         │
 │0.0 + 1.0 x²│
 │0.0 + 1.0 y │
 └            ┘ ∈ ExponentialCone()

SingleVariable-in-GreaterThan{Float64}
 x >= 0.1

SingleVariable-in-Integer
 z ∈ ℤ

SingleVariable-in-ZeroOne
 x ∈ {0, 1}
 y ∈ {0, 1}
```

Closes https://github.com/jump-dev/MathOptInterface.jl/issues/1144
Closes https://github.com/jump-dev/MathOptInterface.jl/pull/1180


src/Utilities/print.jl
```diff
@@ -0,0 +1,454 @@
...
+    name_2 = _to_string(mime, model, term.variable_index_2)
+    name = if term.variable_index_1 == term.variable_index_2
+        string(" ", name_1, _to_string(mime, :sq))
+    else
+        string(" ", name_1, _to_string(mime, :times), name_2)
+    end
+    if term.coefficient < 0
+        return string(" - ", string(-term.coefficient), name)
+    else
+        return string(" + ", string(term.coefficient), name)
```
odow, line 121: These coefficients aren't correct, because they don't account for `0.5 x'Qx`

blegat:
Do you agree with the plan to redirect JuMP printing to printing of the backend (possibly setting some parameters so that the printing is the same) ? Currently, the printing of the MOI model is different from what we currently have in JuMP so we might at least need an option so that the printing is the same, or do you want the printing on JuMP to change ?

src/Utilities/print.jl
```diff
@@ -0,0 +1,458 @@
...
+
+#------------------------------------------------------------------------
+# Nonlinear constraints
+#------------------------------------------------------------------------
+
+"""
+    _VariableNode
+
+A type used to work-around the default printing of Julia expressions. If we
+subsititued the variable names into the expression and the converted to a
```
blegat, line 240: subsititued -> substitute

odow:
There are a couple of differences we could discuss:

**MOI printing**

* We print the constants in a function, even if they are `0.0`
* No fancy printing of `2.0` as `2`
* No dropping of constants if they are 1
* Function-Set types are explicit

**JuMP printing**

* Nonlinear expressions and parameters
* Matrices are printed for Matrix sets (PSDCone)
* JuMP sets are printed differently (SOS1 instead of MOI.SOS1{Float64})

I'm in favor of this printing style, because it makes everything more explicit. (One problem is that I have seen questions on discourse asking why `print(model)` doesn't produce a valid LP file.) JuMP should probably make heavy use of the machinery, but might not want to call `print(backend(model))`.

blegat:
I understand that the default printing for a JuMP model and an MOI model might be different but there given that the JuMP model is completely represented by an MOI model and that someone could also want the JuMP printing style given an MOI model, we could have a printing function with option such as
* `print_zero_constant`
* `print_one_coefficient`
* `print_zero_decimal`
* an optional preprocessing for constraint (to handle shapes of JuMP constraints)
* what default name should be used for variables with no name, e.g. `v[$(vi.value)]` instead of `noname`.

I'm not saying this PR should implement all of these but it's best if we're on the same page. I always find myself wanting to print an MOI model and wanting such options and I think it's best if we don't keep twice almost the same code.

odow:
Okay, adding options might be a good thing. Let's hold-off merging, and I'll have a play.

> I always find myself wanting to print an MOI model

Yes. Particularly if you could print a model, and then the inner bridged model to see the difference.

odow:
@blegat, this now allows:
```Julia
julia> print(model; simplify_coefficients = true, print_constraint_types = false)
Minimize: 2 + x + 3.1 y - 1.2 z

Subject to:
 x - y == 0
 2 x >= 1
 2 x ∈ [1, 2]
 y - z + 2 x² <= 1
 ┌ ┐
 │x│
 │y│
 └ ┘ ∈ SecondOrderCone(2)
 ┌ ┐
 │1│
 │x│
 │y│
 └ ┘ ∈ SecondOrderCone(2)
 ┌  ┐
 │x²│
 │y │
 │1 │
 └  ┘ ∈ ExponentialCone()
 ┌  ┐
 │1 │
 │x²│
 │y │
 └  ┘ ∈ ExponentialCone()
 x >= 0.1
 z ∈ ℤ
 x ∈ {0, 1}
 y ∈ {0, 1}

julia> MOIU.latex_formulation(model; simplify_coefficients = true, print_constraint_types = false)
$$ \begin{aligned}
\min\quad & 2 + x + 3.1 y - 1.2 z \\
\text{Subject to}\\
 & x - y = 0 \\
 & 2 x \ge 1 \\
 & 2 x \in \[1, 2\] \\
 & y - z + 2 x^2 \le 1 \\
 & \begin{bmatrix}
x\\
y\end{bmatrix} \in \text{SecondOrderCone(2)} \\
 & \begin{bmatrix}
1\\
x\\
y\end{bmatrix} \in \text{SecondOrderCone(2)} \\
 & \begin{bmatrix}
x^2\\
y\\
1\end{bmatrix} \in \text{ExponentialCone()} \\
 & \begin{bmatrix}
1\\
x^2\\
y\end{bmatrix} \in \text{ExponentialCone()} \\
 & x \ge 0.1 \\
 & z \in \mathbb{Z} \\
 & x \in \{0, 1\} \\
 & y \in \{0, 1\} \\
\end{aligned} $$
```

blegat (approved):
So great we finally have this!

PR #1300, opened 2021-04-19
Document how to implement a solver interface

odow:
I have a feeling this could get rather long...

Issues that this PR should cover
 - [x] https://github.com/jump-dev/MathOptInterface.jl/issues/608
 - [x] https://github.com/jump-dev/MathOptInterface.jl/issues/662


odow:
Okay this is getting a bit long, so I might stop there. It's at least a big improvement on what we had before. 

The next time someone writes an interface, we should ask them to take notes on all the things that aren't obvious, and we can go from there.

odow:
Preview link: https://jump.dev/MathOptInterface.jl/previews/PR1300/manual/implementing

odow:
If you have comments, please take notes and open a new issue. It'll be easier if we can review smaller PRs.

PR #1302, opened 2021-04-22
Reminder to exclude tests for unsupported features

rschwarz:
Initially, I had assumed that if my solver implements `supports(feature) = false`, the relevant tests would be skipped automatically. Instead, they are typically failing, since the tests contain `@assert supports(feature)`.


rschwarz:
@odow: this is something I remembered while looking at the new [implementing a solver interface](https://jump.dev/MathOptInterface.jl/dev/manual/implementing/) docs.

odow:
Thanks

PR #1305, opened 2021-04-30
[breaking] Rename ListOfConstraints -> ListOfConstraintTypesPresent

### LINKED ISSUE #214: rename ListOfConstraints?

odow:
This has always been a weirdly named attribute. 

Closes #214

### DIFF
--- a/docs/src/manual/implementing.md
+++ b/docs/src/manual/implementing.md
@@ -420,7 +420,7 @@ Constraint-related attributes:
 | ---------------------- | --------------| ------------- | ------------------ |
 | [`ListOfConstraintAttributesSet`](@ref) | Yes          | No            | No |
 | [`NumberOfConstraints`](@ref)           | Yes          | No            | No |
-| [`ListOfConstraints`](@ref)             | Yes          | No            | No |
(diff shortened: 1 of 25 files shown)


blegat (approved):
LGTM but let's wait for other inputs about the name

joaquimg:
This is already much better than before.

Why not just `ListOfConstraintTypes`?
Do we really need `Present`? We don't do `NumberOfConstraintsPresent`. Found no discussion on the links, sorry if I missed (double sorry if I was in the previous discussion).
If we really need, then
I like that the `Present` qualification is at the end, however, `ListOfCurrentConstraintTypes` sounded better to me.
Not sure...

odow:
The argument was that `ListOfConstraintTypes` could be interpreted as the constraint type supported by the model, not just the ones for which `NumberOfConstraints > 0`.

We could have `ListOfPresentConstraintTypes` so allow `ListOfSupportedConstraintTypes` in future.

joaquimg (approved):

PR #1306, opened 2021-04-30
[breaking] Rename .variable_index fields to .variable

### LINKED ISSUE #641: Naming consistency
I came across an annoying inconsistency developing [MathOptFormat](https://github.com/odow/MathOptFormat.jl). Some structs use `variable::VariableIndex`:
https://github.com/JuliaOpt/MathOptInterface.jl/blob/97c92db2abb05ac8a53cedab0c4a68cc72d2fcf7/src/functions.jl#L40-L42
and other structs use `variable_index::VariableIndex`
https://github.com/JuliaOpt/MathOptInterface.jl/blob/97c92db2abb05ac8a53cedab0c4a68cc72d2fcf7/src/functions.jl#L64-L67

Thoughts on standardizing to `variable`? Because I can see people saying "Why?"

As a con, this is a horribly breaking change...

odow:
I'm not necessarily sold on this one. 

The other option is to rename `SingleVariable` to `.variable_index` and `VectorOfVariables` to `.variable_indices`. Either way, we should standardize.

Closes #641

### DIFF
--- a/src/Bridges/Constraint/indicator_activate_on_zero.jl
+++ b/src/Bridges/Constraint/indicator_activate_on_zero.jl
@@ -3,7 +3,7 @@
 
 The `IndicatorActiveOnFalseBridge` replaces an indicator constraint activated
 on 0 with a variable ``z_0`` with the constraint activated on 1, with a variable ``z_1``.
-It stores the added `variable_index` and added constraints:
+It stores the added `variable` and added constraints:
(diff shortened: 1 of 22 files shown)


blegat (approved):
I agree, I never remember for which one it is `variable` and which one it is `variable_index`.
I like the rule: no need to add this in the name if it's already in the signature. Here, the fact that it's an index is already in the type so we can just call it `variable`.

PR #1307, opened 2021-04-30
[breaking] Rename MOIU.supports_default_copy_to to MOI.supports_incremental_interface

### LINKED ISSUE #757: Move supports_default_copy_to to MOI?
`supports_default_copy_to` is obligatory to implement according to http://www.juliaopt.org/MathOptInterface.jl/dev/apimanual/#Testing-guideline-1 and http://www.juliaopt.org/MathOptInterface.jl/dev/apimanual/#Implementing-copy-1. Should we move it from MOIU to MOI? MOIU should be for helper functions, not for core methods of the API that solvers need to implement.

odow:
`supports_default_copy_to` was a bit of a misnomer, and it confused the separation between `MOI` and `MOI.Utilities`. This is really what it's aiming at, and the language now lines up with our implementation guide: https://jump.dev/MathOptInterface.jl/dev/manual/implementing/#The-big-decision:-copy-to-or-incremental-modifications?

Closes #757


joaquimg (approved):

src/MathOptInterface.jl
```diff
@@ -94,6 +94,36 @@ Empty the model, that is, remove all variables, constraints and model attributes
...
+```
+
+The main purpose of this function is to determine whether a model can be loaded
+into `model` incrementally or whether it should be cached and copied at once
+instead.
+
+This is used by [`instantiate`](@ref) to determine whether to add a cache in two
+situations:
+1. A first cache can be used to store the model as entered by the user as well
+   as the names of variables and constraints. This cache is created if this
```
blegat, line 118: This cache is not added by `instantiate`, it is added by JuMP independently of `supports_incremental_interface`

odow: https://github.com/jump-dev/MathOptInterface.jl/blob/db801f5981d61c01f05ba2fce249bc55d4750e86/src/instantiate.jl#L115-L130

blegat: That's the cache added by 2., it stores the bridged model, not the model entered by the user. Instantiate never creates 2 caches. JuMP does but the first cache is not done by `instantiate`

blegat (approved):

PR #1309, opened 2021-04-30
Add NL file writer.

### LINKED ISSUE #1293: Add support for .nl files
There is a .nl writer as part of https://github.com/jump-dev/AmplNLWriter.jl.

We should refactor this writer out into a `FileFormats` submodule. Then, AmplNLWriter can be a thin wrapper around `MOI.FileFormats.NL` that writes the file, sends it to the executable, and parses the sol file. 

That should make it possible to revive https://github.com/odow/NEOS.jl, which would be a similarly thin wrapper around the MPS and NL writers, and handle sending the problem to NEOS via XML-RPC.

The NL file writer is also the focus of external benchmarks, so we should improve the performance of the writer. https://informatica.vu.lt/journal/INFORMATICA/article/1216/text

odow:
This is a single commit that copies the NL file writer from
AmplNLWriter.jl that I recently wrote.

It's a little different from the other FileFormats models in that it
only supports copy_to, and it doesn't support reading.

TODO
 - [x] More tests. Seems that the MOI tests covered functionality not included in these tests.

Closes #1293


src/FileFormats/NL/gen.jl
```diff
@@ -0,0 +1,16 @@
+# This script builds the list of recognized ASL opcodes using the header files
+# in ASL_jll. Only re-run it if AMPL adds new opcodes (which is unlikely).
```
blegat, line 2: This file is not included right ?  Maybe it could be outside `src` ? Or make it clear that it's not included in this header ? I was a bit confused.

odow: Yeah it's just a helper script to rebuild the OPCODEs.

src/FileFormats/NL/NL.jl
```diff
@@ -0,0 +1,788 @@
...
+    upper::Float64
+    # Whether variable is binary or integer.
+    type::_VariableType
+    # Primal start of the variable.
+    start::Union{Float64,Nothing}
+    # Number of constraints that the variable appears in.
+    jacobian_count::Int
+    # If the variable appears in the objective.
+    in_nonlinear_objective::Bool
+    # If the objetive appears in a nonlinear constraint.
```
blegat, line 106: objetive -> objective (or variable ?)

src/FileFormats/NL/NL.jl
```diff
@@ -0,0 +1,788 @@
...
+    for (i, bound) in enumerate(nlp_block.constraint_bounds)
+        push!(
+            dest.g,
+            _NLConstraint(MOI.constraint_expr(nlp_block.evaluator, i), bound),
+        )
+    end
+    dest.nlpblock_dim = length(dest.g)
+    starts = MOI.supports(model, MOI.VariablePrimalStart(), MOI.VariableIndex)
+    for x in MOI.get(model, MOI.ListOfVariableIndices())
+        dest.x[x] = _VariableInfo()
```
blegat, line 290: Since `x` is build like this, if `model` has variables from 1 to n, `x` will have it too. It therefore make sense to use `CleverDict`

src/FileFormats/NL/NL.jl
```diff
@@ -0,0 +1,788 @@
...
+    else
+        [1, 2, 5, 6, 3, 4, 7, 8, 9]
+    end
+    # Now we can order the variables.
+    n = 0
+    for i in order_i
+        # Since variables come from a dictionary, there may be differences in
+        # the order depending on platform and Julia version. Sort by creation
+        # time for consistency.
+        for x in sort!(types[i]; by = y -> y.value)
```
blegat, line 433: Not needed of we use `CleverDict` as it uses an `OrderedDict`

src/FileFormats/NL/NL.jl
```diff
@@ -0,0 +1,788 @@
...
+            start = MOI.get(model, MOI.VariablePrimalStart(), x)
+            MOI.set(dest, MOI.VariablePrimalStart(), x, start)
+        end
+        mapping[x] = x
+    end
+    dest.sense = MOI.get(model, MOI.ObjectiveSense())
+    resize!(dest.order, length(dest.x))
+    # Now deal with the normal MOI constraints.
+    for (F, S) in MOI.get(model, MOI.ListOfConstraintTypesPresent())
+        _process_constraint(dest, model, F, S, mapping)
```
blegat, line 301: `mapping` -> `mapping.conmap[F, S]`

src/FileFormats/NL/NL.jl
```diff
@@ -0,0 +1,788 @@
...
+    end
+    return mapping
+end
+
+_set_to_bounds(set::MOI.Interval) = (0, set.lower, set.upper)
+_set_to_bounds(set::MOI.LessThan) = (1, -Inf, set.upper)
+_set_to_bounds(set::MOI.GreaterThan) = (2, set.lower, Inf)
+_set_to_bounds(set::MOI.EqualTo) = (4, set.value, set.value)
+
+function _process_constraint(dest::Model, model, F, S, mapping)
```
blegat, line 447: This should be `mapping::MOIU.DoubleDicts.IndexWithType{F, S}` so no need to pass `F` and `S` along, you can do `_process_constraint(dest::Model, model, mapping::MOIU.DoubleDicts.IndexWithType{F, S})`

blegat: See https://github.com/jump-dev/DiffOpt.jl/blob/cc0914012a1868552fda53f00221e5680fd3f294/src/MOI_wrapper.jl#L1164-L1201 for instance

src/FileFormats/NL/NL.jl
```diff
@@ -0,0 +1,788 @@
...
+        s = MOI.get(model, MOI.ConstraintSet(), ci)
+        op, l, u = _set_to_bounds(s)
+        con = _NLConstraint(l, u, op, _NLExpr(f))
+        if isempty(con.expr.linear_terms) && isempty(con.expr.nonlinear_terms)
+            if l <= con.expr.constant <= u
+                continue
+            else
+                error(
+                    "Malformed constraint. There are no variables and the " *
+                    "bounds don't make sense.",
```
blegat, line 459: "Malformed constraint $ci ... : the function constant `$con.expr.constant` is not in the interval [`$l`, `$u`]"

src/FileFormats/NL/NL.jl
odow: No, this type ordering is very specific to NL files. It isn't the same as creation order.

src/FileFormats/NL/NL.jl
odow: I think it's fine as-is. There isn't much to be gained from a clever dict, and we don't need the ordered nature of the original variables.

PR #1311, opened 2021-05-02
Update NEWS.md

odow:
We should probably keep this up-to-date with each PR, rather than waiting until just before release.


blegat:
The disadvantage is that it creates conflicts between each PRs

odow:
It's not necessarily a bad thing to have to fix the conflict and re-run CI after merging a breaking change.

PR #1312, opened 2021-05-02
[breaking] error on SingleVariable ConstraintNames

### LINKED ISSUE #832: Names for variable bounds
@mlubin suggested that https://github.com/JuliaOpt/GLPK.jl/pull/101#discussion_r295583456

> I'm proposing to make setting ConstraintName on SingleVariable constraint into an error. This wouldn't change lookups otherwise.
>
> To look up the bound on a variable, you first look up the variable. The index of the variable gives you the index of the bound constraint.

Is this something for MOI 0.10?

odow:
Started work on https://github.com/jump-dev/MathOptInterface.jl/issues/832, but I'm not sure it's worth it. It just adds lots of breakage and a special case for limited upside.

What might be more useful is a utility to manage the lookup rules constraint names for solvers: https://github.com/jump-dev/MathOptInterface.jl/issues/674

Closes #832

### DIFF
--- a/docs/src/manual/implementing.md
+++ b/docs/src/manual/implementing.md
@@ -511,6 +511,29 @@ These methods have the following rules:
   variable or constraint
 * If there are multiple variables or constraints with the name, throw an error.
 
+!!! warning
+    You should _not_ implement [`ConstraintName`](@ref) for
+    [`SingleVariable`](@ref) constraints. If you implement [`ConstraintName`](@ref)
+    for other constraints, you may need to add the following two methods to
+    avoid method ambiguities:
+    ```julia
+    function MOI.supports(
+        ::Optimizer,
+        ::MOI.ConstraintName,
+        ::Type{MOI.ConstraintIndex{MOI.SingleVariable,<:MOI.AbstractScalarSet}},
+    )
+        return throw(MOI.SingleVariableConstraintNameError())
+    end
+    function MOI.set(
+        ::Optimizer,
+        ::MOI.ConstraintName,
+        ::MOI.ConstraintIndex{MOI.SingleVariable,<:MOI.AbstractScalarSet},
+        ::String,
+    )
+        return throw(MOI.SingleVariableConstraintNameError())
+    end
+    ```
+
 ## Solutions
 
 Implement [`optimize!`](@ref) to solve the model:
--- a/docs/src/manual/implementing.md
+++ b/docs/src/manual/implementing.md
@@ -514,13 +514,13 @@ These methods have the following rules:
 !!! warning
     You should _not_ implement [`ConstraintName`](@ref) for
     [`SingleVariable`](@ref) constraints. If you implement [`ConstraintName`](@ref)
-    for other constraints, you may need to add the following two methods to
-    avoid method ambiguities:
+    for other constraints, you can add the following two methods to disable
+    [`ConstraintName`](@ref) for [`SingleVariable`](@ref) constraints.
     ```julia
     function MOI.supports(
         ::Optimizer,
         ::MOI.ConstraintName,
-        ::Type{MOI.ConstraintIndex{MOI.SingleVariable,<:MOI.AbstractScalarSet}},
+        ::Type{<:MOI.ConstraintIndex{MOI.SingleVariable,<:MOI.AbstractScalarSet}},
     )
         return throw(MOI.SingleVariableConstraintNameError())
     end
--- a/src/Bridges/bridge_optimizer.jl
+++ b/src/Bridges/bridge_optimizer.jl
@@ -1324,7 +1324,7 @@ function MOI.supports(
     ::MOI.ConstraintName,
     ::Type{MOI.ConstraintIndex{MOI.SingleVariable,S}},
 ) where {S}
-    return false
+    return throw(MOI.SingleVariableConstraintNameError())
 end
 
 function MOI.set(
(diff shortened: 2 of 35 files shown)


src/Test/UnitTests/basic_constraint_tests.jl
```diff
@@ -333,7 +333,7 @@ function basic_constraint_test_helper(
         c = MOI.add_constraint(model, constraint_function, set)
         @test MOI.get(model, MOI.NumberOfConstraints{F,S}()) == 1
 
-        if name
+        if name && F != MOI.SingleVariable
```
blegat, line 336: We  should check that the error is thrown

src/attributes.jl
```diff
@@ -1133,13 +1133,33 @@ struct ListOfConstraintAttributesSet{F,S} <: AbstractModelAttribute end
...
+"""
+    SingleVariableConstraintNameError()
+
+An error to be thrown when the user tries to set `ConstraintName` on a
+`SingleVariable` constraint.
+"""
+function SingleVariableConstraintNameError()
+    return UnsupportedAttribute(
+        ConstraintName(),
+        "ConstraintNames are not supported for SingleVariable constraints",
```
blegat, line 1159: "`ConstraintName`s" ?

src/Bridges/bridge_optimizer.jl
```diff
@@ -1230,6 +1230,23 @@ function MOI.set(
     end
 end
 
+function MOI.supports(
+    ::AbstractBridgeOptimizer,
+    ::MOI.ConstraintName,
+    ::Type{MOI.ConstraintIndex{MOI.SingleVariable,S}},
+) where {S}
+    return false
```
blegat, line 1238: Shouldn't we throw the error here as well, I'm hesitant

odow: Why would we throw an error on `supports`?

blegat: This is not a valid call, it's like calling `supports` with a non-copyable attribute which is throwing an error.
Otherwise, the user would think that it's an issue with the model not supporting it but in fact, it's not valid to set a constraint name for such constraint.

odow: How's this now.

blegat:
What happens if a solver does not implement `supports` and `set` for `SingleVariable` ? Does it throw the right error ?
Calling `set` redirects to `supports`:
https://github.com/jump-dev/MathOptInterface.jl/blob/2c1666d3e112032bac79502c4b230980c9370e87/src/attributes.jl#L432
but then `supports` return false:
https://github.com/jump-dev/MathOptInterface.jl/blob/2c1666d3e112032bac79502c4b230980c9370e87/src/attributes.jl#L217
Having `supports` redirect to `supports_fallback` (to avoid ambiguities) which has a method catching `ConstraintName` for `SingleVariable` throwing the right error should make it work for both `set` and `supports`.

odow:
Added the fallback and some documentation.

odow:
Huh. This passed with Revise doing it's magic. Interesting.

src/attributes.jl
```diff
@@ -1133,13 +1133,41 @@ struct ListOfConstraintAttributesSet{F,S} <: AbstractModelAttribute end
...
+        "`ConstraintName`s are not supported for `SingleVariable` constraints.",
+    )
+end
+
+function supports(
+    ::ModelLike,
+    ::ConstraintName,
+    ::Type{ConstraintIndex{SingleVariable,S}},
+) where {S}
+    return throw(SingleVariableConstraintNameError())
```
blegat, line 1168: This is going to cause ambiguities because it is both broad in the first argument and wide in the second and last one. We should redirect `supports` to a `supports_fallback` to avoid it.

odow: > This is going to cause ambiguities

Not really? Only if someone defines `set(::Optimizer, ::ConstraintName, ::ConstraintIndex)` which is invalid because it doesn't handle the SingleVariable case.

The documentation also gives two functions to implement that would resolve any ambiguities.

blegat: If you are an MOI layer, you would want to pass the call along and trust the inner optimizer to throw the error.
That would cause ambiguity

odow: Then you can just implement the functions listed in the documentation to resolve it?

blegat: Yes but having to implement this function for all MOI layers is a lot more work than implementing one fallback. That's also consistent with what we already do with `set_fallback`

docs/src/manual/implementing.md
```diff
@@ -511,6 +511,29 @@ These methods have the following rules:
   variable or constraint
 * If there are multiple variables or constraints with the name, throw an error.
 
+!!! warning
+    You should _not_ implement [`ConstraintName`](@ref) for
+    [`SingleVariable`](@ref) constraints. If you implement [`ConstraintName`](@ref)
+    for other constraints, you may need to add the following two methods to
+    avoid method ambiguities:
```
blegat, line 518: I don't think there is any ambiguity issue anymore, is it?

odow: I've rephrased. It may still be needed if they implement a `::MOI.ConstraintIndex` method. Ideally, Julia would have a "not-subtype" relation:
```julia
where {F !<: MOI.SingleVariable}
```

odow:
How's this now?

docs/src/manual/implementing.md
blegat: That would be futuristic

blegat (approved):

PR #1317, opened 2021-05-05
Minor fixes to CachingOptimizer

odow:
Turned out this had no effect on the benchmarks, but was useful to rule out as a cause.

Unifying the return types and providing some extra annotations can't hurt.


blegat:
Added the breaking label and removing it from backport v0.9 as it broke JuMP tests which currently assumes that `set` return the value `set` which this PR changes it to `nothing`.

odow:
I think this was an artifact of `CachingOptimizer`. There is no guarantee that solvers will return anything, so it makes sense to stardardize on `nothing`.

blegat:
Yes, we should probably make a decision and add something in the docstring.

PR #1319, opened 2021-05-07
[breaking] modify type of RawParameter

### LINKED ISSUE #1314: Make RawParameter a String
Lots of solvers mixed between Symbols and Strings. However, JuMP assumes that it is a `String` to use `set_optimizer_attribute`. We could even rename `RawParameter` to `RawOptimizerAttribute`?

https://github.com/jump-dev/MathOptInterface.jl/blob/2e99e255d435f556360ac9df3762b6026f7c5447/src/attributes.jl#L664-L672

odow:
Closes #1314

Since JuMP now uses `set_optimizer_attribute` that takes a string, it makes sense to force all solvers to accept `RawParameter` as a `::String`. They can define other attributes for enums etc.

TODOs:

- [x] Should we rename it to `RawOptimizerAttribute` instead of `RawParameter`?


blegat (approved):
Not sure about the renaming, renaming would make sense but we could gather more opinion

joaquimg:
I like RawAttribute and RawOptimizerAttribute better than RawParameter. (On a first glance. I havent thought about consequences).

mlubin (approved):
Makes sense for uniformity.

> They can define other attributes for enums etc.

Can they though? `RawOptimizerAttribute` gets special treatment in `reset_optimizer`.

odow:
Solver-specific attributes should be fine, because other solves won't declare `supports` for them. It's only this one that is the issue.

PR #1320, opened 2021-05-07
Add VariableBasisStatus

### LINKED ISSUE #1297: Add VariableBasisStatus
`VariableBasisStatus` was removed here: https://github.com/jump-dev/MathOptInterface.jl/pull/679

However, we over-looked an edge case: without it, we can't tell if variables with free bounds with a `VariablePrimal` of `0.0` are in the (degenerate) basis or not.

This became a problem here: https://github.com/jump-dev/JuMP.jl/issues/2560, https://github.com/jump-dev/JuMP.jl/pull/2562

Given the redundancy with `ConstraintBasisStatus` on variable bounds, perhaps we just need 
```Julia
"""
    MOI.VariableInBasis

A variable attribute which indicates `true` or `false` if some variable is in 
the current optimal solution basis.
"""
struct MOI.VariableInBasis <: MOI.AbstractVariableAttribute end
```

cc @geoffleyland

odow:
I started to implement this, but it needs some discussion.

- [x] Do we need to worry about mapping it through bridges? I guess not. This is likely only a problem for models created using variable bridges, which won't really be MILP solvers.
- [x] Do we leave `ConstraintBasisStatus{SingleVariable,S}` in place? It seems like we should remove it, but then this is another special case of "an attribute that isn't supported for bounds" (c.f. https://github.com/jump-dev/MathOptInterface.jl/pull/1312). That's starting to feel like our abstraction is wrong/incomplete.

Really, the only thing we're missing is "is a variable with no constraints in the basis?" The only answers can be `BASIC` or `SUPERBASIC`.

I'd vote for my previous suggestion of:
```julia
"""
    VariableInBasis(result_index::Int = 1)

A variable attribute which indicates `true` or `false` if some variable in
result `result_index` is in the corresponding optimal solution basis.
"""
struct VariableInBasis <: AbstractVariableAttribute 
    result_index::Int
    VariableInBasis(result_index::Int = 1) = new(result_index0
end
```

Closes #1297

### DIFF
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -1286,6 +1286,18 @@ struct ConstraintBasisStatus <: AbstractConstraintAttribute
     ConstraintBasisStatus(result_index::Int = 1) = new(result_index)
 end
 
+function get_fallback(
+    ::ModelLike,
+    ::ConstraintBasisStatus,
+    ::ConstraintIndex{SingleVariable,<:AbstractScalarSet},
+)
+    error(
+        "Querying the basis status of a `SingleVariable` constraint is not ",
+        "supported. Use [`VariableBasisStatus`](@ref) instead.",
+    )
+end
+
+
 """
     CanonicalConstraintFunction()
 
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -1114,8 +1114,9 @@ is_set_by_optimize(::CallbackVariablePrimal) = true
 """
     BasisStatusCode
 
-An Enum of possible values for the `ConstraintBasisStatus` attribute, explaining
-the status of a given element with respect to an optimal solution basis.
+An Enum of possible values for the [`ConstraintBasisStatus`](@ref) and
+[`VariableBasisStatus`](@ref) attributes, explaining the status of a given
+element with respect to an optimal solution basis.
 
 Possible values are:
 
@@ -1126,7 +1127,7 @@ Possible values are:
 * `SUPER_BASIC`: element is not in the basis but is also not at one of its
   bounds
 
-Notes
+## Notes
 
 * `NONBASIC_AT_LOWER` and `NONBASIC_AT_UPPER` should be used only for
   constraints with the `Interval` set. In this case, they are necessary to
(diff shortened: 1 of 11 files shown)


joehuchette:
If we add this, then +1 to removing ConstraintBasisStatus{SingleVariable,S}. An alternative (that I don’t necessarily endorse) might be to stick with only ConstraintBasis and add a scalar Reals set to MOI to cover the missing case.

odow:
> An alternative (that I don’t necessarily endorse) might be to stick with only ConstraintBasis and add a scalar Reals set to MOI to cover the missing case.

I would endorse this. It's the only part of the abstraction that is missing.

odow:
Here's the `ConstraintIndex{SingleVariable,Reals}` version: https://github.com/jump-dev/MathOptInterface.jl/pull/1351. It's much simpler, for what is a niche attribute.

odow:
Okay, how's this. I have GLPK passing tests locally. Much simpler.

src/attributes.jl
```diff
@@ -1260,8 +1273,7 @@ A constraint attribute for the `BasisStatusCode` of some constraint in result
 
 See [`ResultCount`](@ref) for information on how the results are ordered.
 
-**For the basis status of a variable, query the corresponding `SingleVariable`
-constraint that enforces the variable's bounds.**
+For the basis status of a variable, query [`VariableBasisStatus`](@ref).
```
blegat, line 1276: Maybe also say explicitly that `ConstraintBasisStatus` for `SingleVariable` is not defined: solver should not implement it and user should not call it.

blegat:
We should have a `MOI.get_fallback` (Not `MOI.Utilities.get_fallback`!) for ConstraintBasisStatus` with `SingleVariable` that shows a warning and redirects to a `get` with `VariableBasisStatus`

blegat (approved):

odow:
Huh. I ran `include("test/attributes.jl")` locally and it passed. Not sure what happened. I guess the file wasn't saved.

PR #1325, opened 2021-05-09
inbounds for is_strictly_increasing

blegat:
As this is taking a non-negligeable time in https://github.com/jump-dev/MathOptInterface.jl/pull/1287, it is worth optimizing.
Consider the benchmark:
```julia
function bench(n)
    x = MOI.VariableIndex.(1:n)
    f = MOI.ScalarAffineFunction(MOI.ScalarAffineTerm.(ones(length(x)), x), 0.0)
    @btime MOI.Utilities.is_canonical($f)
end
```
Before:
```
julia> bench(10000)
  8.769 μs (0 allocations: 0 bytes)
```
Initial suggestion:
```
julia> bench(10000)
  6.477 μs (0 allocations: 0 bytes)
```
@odow improved implementation:
```
julia> bench(10000)
  5.808 μs (0 allocations: 0 bytes)
```

### DIFF
--- a/src/Utilities/functions.jl
+++ b/src/Utilities/functions.jl
@@ -669,18 +669,21 @@ function is_strictly_sorted(x::Vector, by, filter)
     if isempty(x)
         return true
     end
-    if !filter(first(x))
+    current_x = first(x)
+    if !filter(current_x)
         return false
     end
-    for i in eachindex(x)[2:end]
-        prev = @inbounds(x[i-1])
-        cur = @inbounds(x[i])
-        if by(cur) <= by(prev)
+    current_fx = by(current_x)
+    @inbounds for i in eachindex(x)[2:end]
+        next_x = x[i]
+        if !filter(next_x)
             return false
         end
-        if !filter(cur)
+        next_fx = by(next_x)
+        if next_fx <= current_fx
             return false
         end
+        current_x, current_fx = next_x, next_fx
     end
     return true
 end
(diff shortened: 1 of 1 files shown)


odow (approved):
Does it really make that much of a difference for #1287? It's not often we have functions with 10k terms.

joaquimg (approved):

joaquimg:
This seems to be proportional to the overall number of non-zeros, not just in one function.

odow:
Before
```
julia> bench(10_000)
  10.609 μs (0 allocations: 0 bytes)
```
After
```
julia> bench(10_000)
  5.383 μs (0 allocations: 0 bytes)
```

```Julia
function MathOptInterface.Utilities.is_strictly_sorted(x::Vector, by, filter)
    if isempty(x)
        return true
    end
    current_x = first(x)
    if !filter(current_x)
        return false
    end
    current_fx = by(current_x)
    @inbounds for i in eachindex(x)[2:end]
        next_x = x[i]
        if !filter(next_x)
            return false
        end
        next_fx = by(next_x)
        if next_fx <= current_fx
            return false
        end
        current_x, current_fx = next_x, next_fx
    end
    return true
end
```

blegat:
This is the part of copying to `Clp.Model` for the benchmark of https://github.com/jump-dev/MathOptInterface.jl/pull/1287:
strictly_sorted

PR #1329, opened 2021-05-10
Add identity_index_map and fix DoubleDict bug

blegat:
Needed for https://github.com/jump-dev/Clp.jl/pull/111


src/Utilities/DoubleDicts.jl
```diff
@@ -417,7 +417,9 @@ end
 end
 
 function initialize_inner!(d::AbstractWithType{F,S,V,D}) where {F,S,V,D}
-    return d.inner = D()
+    d.inner = D()
+    d.dict.dict[(F, S)] = d.inner
```
odow, line 421: Ah. I ran into this with the NL writer, but I assumed it was expected behavior.

PR #1330, opened 2021-05-10
Add throw_if_scalar_and_constant_not_zero

blegat:
Extracted from https://github.com/jump-dev/MathOptInterface.jl/pull/1287/

### DIFF
--- a/src/Bridges/bridge_optimizer.jl
+++ b/src/Bridges/bridge_optimizer.jl
@@ -1742,23 +1742,10 @@ function bridged_constraint_function(
     if !Variable.has_bridges(Variable.bridges(b))
         return func, set
     end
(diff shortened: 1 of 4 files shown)


blegat:
It could but then an invalid call would return `nothing` instead of a `MethodError`

odow:
How could it get a method error? And when would a method error be preferable?

odow:
That should be caught at the original caller? We shouldn't rely on this method to throw a method error. It would violate the MethodError principle.

blegat:
Since this is a documented function, the user might call this directly in which case it would get the `MethodError` at the first method called so it agrees with the principle.
I don't think the places where it is called from the MOI source code will ever never hit the `MethodError`.

odow (approved):

PR #1333, opened 2021-05-11
[breaking] move JSONSchema to a test dependency

odow:
A big cause of method invalidation in JuMP is HTTP (https://github.com/jump-dev/JuMP.jl/issues/2273). One option is to fix them in HTTP, but that requires modifications to Base Julia (https://github.com/JuliaWeb/HTTP.jl/issues/712). 

The better question is: why do we even have HTTP as a dependency? Turns out it is a transitive dependency of JSONSchema, which may need to access schemas from the web. We only use JSONSchema to validate MOF files, and then not even by default.

And if we move JSONSchema to a test and doc dependency, we save 0.4 seconds on `using`, and 1000 method invalidations!

Before 

```julia
julia> @time @eval using MathOptInterface
  2.258680 seconds (3.82 M allocations: 274.937 MiB, 7.42% gc time, 0.15% compilation time)

julia> exit()
(base) oscar@Oscars-MBP MathOptInterface % ~/julia --project=. --banner=no
julia> using SnoopCompileCore

julia> invalidations = @snoopr using MathOptInterface
1339-element Vector{Any}:
  MethodInstance for show(::IOBuffer, ::Any)
  "invalidate_mt_cache"
```

After 

```julia
julia> @time @eval using MathOptInterface
  1.858869 seconds (2.80 M allocations: 213.619 MiB, 8.45% gc time, 0.20% compilation time)

julia> exit()
(base) oscar@Oscars-MBP MathOptInterface % ~/julia --project=. --banner=no
julia> using SnoopCompileCore

julia> invalidations = @snoopr using MathOptInterface
374-element Vector{Any}:
  MethodInstance for show(::IOBuffer, ::Any)
  "invalidate_mt_cache"
```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -7,7 +7,6 @@ BenchmarkTools = "6e4b80f9-dd63-53aa-95a3-0cdb28fa8baf"
 CodecBzip2 = "523fee87-0ab8-5b00-afb7-3ecf72e48cfd"
 CodecZlib = "944b1d66-785c-5afd-91f1-9de20f533193"
 JSON = "682c06a0-de6a-54ab-a142-c8b1cf79cde6"
-JSONSchema = "7d188eb4-7ad8-530c-ae41-71a32a6d4692"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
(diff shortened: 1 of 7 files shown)


odow:
Any objections to merging?

blegat (approved):

PR #1315, opened 2021-05-05
Lazily create storage in StructOfConstraints

odow:
Part of https://github.com/jump-dev/MathOptInterface.jl/issues/1313

Calling `MOIU.Model{Float64}()` does a lot of work. It has to create 97 different `CleverDict`s to store all the constraints in! Even though most users only use a small fraction of the possible constraint types. 

This PR uses a small `Union{Nothing,X}` to avoid creating the dictionaries, instead it lazily creates them on the first call to `add_constraint`.

Before
```
(base) oscar@Oscars-MBP auto-cache % ~/julia --project=. bench.jl clp
 21.929698 seconds (54.74 M allocations: 3.082 GiB, 6.57% gc time, 34.33% compilation time)
  0.001764 seconds (6.02 k allocations: 538.508 KiB)
(base) oscar@Oscars-MBP auto-cache % ~/julia --project=. bench.jl clp --no-bridge
 11.345386 seconds (19.50 M allocations: 1.112 GiB, 4.68% gc time, 98.00% compilation time)
  0.001040 seconds (3.05 k allocations: 309.961 KiB)
(base) oscar@Oscars-MBP auto-cache % ~/julia --project=. bench.jl glpk           
 16.459349 seconds (34.09 M allocations: 1.955 GiB, 6.23% gc time, 52.33% compilation time)
  0.000730 seconds (3.27 k allocations: 251.531 KiB)
(base) oscar@Oscars-MBP auto-cache % ~/julia --project=. bench.jl glpk --no-bridge
 12.485400 seconds (20.88 M allocations: 1.194 GiB, 4.43% gc time, 99.97% compilation time)
  0.000660 seconds (2.55 k allocations: 234.750 KiB)
```

After
```
Running: clp 
 18.193627 seconds (42.68 M allocations: 2.395 GiB, 7.23% gc time, 19.00% compilation time)
  0.001554 seconds (3.85 k allocations: 396.117 KiB)
Running: clp --no-bridge
  5.388487 seconds (7.22 M allocations: 424.453 MiB, 3.97% gc time, 92.45% compilation time)
  0.000915 seconds (1.90 k allocations: 238.125 KiB)
Running: glpk 
  9.581772 seconds (21.91 M allocations: 1.263 GiB, 5.77% gc time, 34.16% compilation time)
  0.000766 seconds (2.15 k allocations: 180.867 KiB)
Running: glpk --no-bridge
[...]

PR #1344, opened 2021-05-13
Style fixes for Bridges

odow:
Style fixes and unification of return types for a small win. This was mainly an excuse for me to read slowly through the code to better understand it.

Before
```
Running: clp 
 10.690129 seconds (25.79 M allocations: 1.501 GiB, 6.46% gc time, 21.90% compilation time)
  0.001109 seconds (1.58 k allocations: 114.789 KiB)
```

After
```
Running: clp 
 10.350866 seconds (25.08 M allocations: 1.463 GiB, 6.01% gc time, 21.59% compilation time)
  0.000871 seconds (1.58 k allocations: 114.789 KiB)
```

Edit: I'll add that functions who didn't explicitly `return` could have type unstable return types at inference. At run time these were obviously never used, which explains the identical second-time results. It's only an issue on the first run when inference has to attempt to figure out what the return is. A blank `return` lets it figure this out easily.

### DIFF
--- a/docs/src/submodules/Bridges/overview.md
+++ b/docs/src/submodules/Bridges/overview.md
@@ -167,10 +167,8 @@ with inner model MOIU.GenericModel{Float64,MOIU.ModelFunctionConstraints{Float64
 Then use [`Bridges.add_bridge`](@ref) to add individual bridges:
 ```jldoctest lazy_bridge_optimizer
 julia> MOI.Bridges.add_bridge(optimizer, MOI.Bridges.Constraint.SplitIntervalBridge{Float64})
-Dict{Any,DataType} with 0 entries
 
 julia> MOI.Bridges.add_bridge(optimizer, MOI.Bridges.Objective.FunctionizeBridge{Float64})
-Dict{Any,DataType} with 0 entries
 ```
 
 Now the constraints will be bridged only if needed:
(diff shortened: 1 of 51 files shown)


odow:
Okay, I now understand the bridging code a lot better. 

This PR is only style and returns. No logic or bigger code changes, so it should be pretty safe to merge.

 * I added some extra TODOs to the code, and I noticed a few scattered around. Once this is merged, I'll open an issue to track them all.
* Some bridges return a vector for things like `ListOfVariableIndices` that they use internally. Since we have the (unstated???) contract that results returned from `MOI.get` are now the property of the caller, these should switch to returning copies. Anything else just seems liable to introduce bugs.
* There is a bit of duplication of code and types, e.g., `IndexInVector` and `trimap`. We should simplify these, where possible.

The core reason for the time-to-first-plot problem is this:
 * We have a lot of code in the bridges. This code, by design is not type-stable, since it uses things like a list of added bridges. A lot of the core reasoning for doing this was that Julia could figure out which bridges were used "at compile time", but since we produce code that doesn't compile well, a lot of this happens at run-time. I wonder if we need something similar to #1315, where the `.variable_map`, `.constraint_map`, and `.objetive_map` fields are `Union{Nothing` to give the compiler a better chance and cut the amount of code it tries to infer. This might be the equivalent of https://github.com/jump-dev/MathOptInterface.jl/pull/1252.

We at least have some threads to pull on going forward.

PR #1348, opened 2021-05-14
Ensure bridges return vectors that can be modified by the user

### LINKED ISSUE #1346: Bridges should return get by value, not reference
Check all `get` return values not references that the user can modify without fear.

As one example:
https://github.com/jump-dev/MathOptInterface.jl/blob/7923c601506cba2edb44f25eca830fc74d7cfe70/src/Bridges/Constraint/geomean.jl#L173
If the user were to modify the returned value, this would modify the vector inside the bridge.

The fix would be
```julia
MOI.get(b::GeoMeanBridge, ::MOI.ListOfVariableIndices) = copy(b.xij)
```

odow:
Closes #1346 

The current behavior was ripe for introducing bugs. 

In addition, the `copy(list)` in `bridge_optimizer` was unnecessary for most bridges, which just return `[variable]`, so removing it avoids an unnecessary copy in those cases, and maintains the same number of copies for the ones modified here.

### DIFF
--- a/src/Bridges/bridge_optimizer.jl
+++ b/src/Bridges/bridge_optimizer.jl
@@ -632,7 +632,8 @@ function get_all_including_bridged(
 )
     list = MOI.get(b.model, attr)
     if !isempty(Variable.bridges(b))
-        return vcat(list, keys(Variable.bridges(b)))
+        # The conversion from `keys` into a `Vector` happens inside `append!`.
+        append!(list, keys(Variable.bridges(b)))
     end
     return list
 end
--- a/src/Bridges/bridge_optimizer.jl
+++ b/src/Bridges/bridge_optimizer.jl
@@ -632,8 +632,7 @@ function get_all_including_bridged(
 )
     list = MOI.get(b.model, attr)
     if !isempty(Variable.bridges(b))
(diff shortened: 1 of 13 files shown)


src/Bridges/bridge_optimizer.jl
```diff
@@ -632,8 +632,8 @@ function get_all_including_bridged(
 )
     list = MOI.get(b.model, attr)
     if !isempty(Variable.bridges(b))
-        # TODO(odow): Fix this!
-        list = append!(copy(list), keys(Variable.bridges(b)))
+        # The conversion from `keys` into a `Vector` happens inside `append!`.
+        append!(list, keys(Variable.bridges(b)))
```
blegat, line 636: This might modify the `list`. Is it clearly stated in `get` that what is returned can be modified ?

odow: This does modify `list`. The `get` docstring doesn't mention it, but anything else is just asking for trouble.

odow: The `get_all_including_bridged` function below also modifies `list`, so any model violating this principle is already wrong.

blegat (approved):

PR #1349, opened 2021-05-14
Add error checks for set dimensions

### LINKED ISSUE #967: Add a new error for situation where cone is not defined for the specified dimension
Suggested in https://github.com/JuliaOpt/MathOptInterface.jl/pull/962#issuecomment-558726806.

There are several cones that are only defined for dimension >= x in the docs, and there is no error for trying to build a cone with dimension < x.

Another example is `RelativeEntropyCone` which should have odd dimension ~(also needs documentation)~.

odow:
This fixes a TODO in Bridges and closes #967.

One point is up for debate: can you initialize a zero-dimensional set? What do we do in cases like
 - `Nonnegatives(0)` => ??
 - `SecondOrderCone(1)` => `t <= 0`?

### DIFF
--- a/src/FileFormats/SDPA/SDPA.jl
+++ b/src/FileFormats/SDPA/SDPA.jl
@@ -164,11 +164,8 @@ function Base.write(io::IO, model::Model{T}) where {T}
     end
     println(io)
 
-    max_index_dim = if max_dim > 0
+    max_index_dim =
         MOI.dimension(MOI.PositiveSemidefiniteConeTriangle(max_dim))
-    else
-        0
-    end
     index_map = Vector{Tuple{Int,Int}}(undef, max_index_dim)
     k = 0
     for col in 1:max_dim
--- a/src/FileFormats/SDPA/SDPA.jl
+++ b/src/FileFormats/SDPA/SDPA.jl
@@ -164,10 +164,12 @@ function Base.write(io::IO, model::Model{T}) where {T}
     end
     println(io)
 
-    index_map = Vector{Tuple{Int,Int}}(
-        undef,
-        MOI.dimension(MOI.PositiveSemidefiniteConeTriangle(max_dim)),
-    )
+    max_index_dim = if max_dim > 0
+        MOI.dimension(MOI.PositiveSemidefiniteConeTriangle(max_dim))
+    else
+        0
+    end
+    index_map = Vector{Tuple{Int,Int}}(undef, max_index_dim)
     k = 0
     for col in 1:max_dim
         for row in 1:col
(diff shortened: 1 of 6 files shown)


odow:
Tests are failing because of the following, which neatly summarizes my question:
https://github.com/jump-dev/MathOptInterface.jl/blob/7923c601506cba2edb44f25eca830fc74d7cfe70/test/Bridges/Variable/rsoc_to_psd.jl#L16-L112

odow:
I've relaxed the restriction on the sets where it makes sense. So you can have `RotatedSecondOrderCone(2)` and `NormOneCone(1)`. But you can't have matrix cones with zero dimensional matrices. And you can't have a 1-dimensional GeometricMeanCone (0th root of zero?). 

Thoughts?

odow:
@blegat how's this? Do they make sense?

odow:
Bump @blegat

src/sets.jl
```diff
@@ -342,7 +443,9 @@ end
 """
     RelativeEntropyCone(dimension)
 
-The relative entropy cone ``\\{ (u, v, w) \\in \\mathbb{R}^{1+2n} : u \\ge \\sum_{i=1}^n w_i \\log (\\frac{w_i}{v_i}), v_i \\ge 0, w_i \\ge 0 \\}`` of dimension `dimension```{}=2n+1``.
+The relative entropy cone
+``\\{ (u, v, w) \\in \\mathbb{R}^{1+2n} : u \\ge \\sum_{i=1}^n w_i \\log(\\frac{w_i}{v_i}), v_i \\ge 0, w_i \\ge 0 \\}``,
+where `dimension = 2n + 1 >= 3`.
```
blegat, line 448: 3 -> 1 to match the code

test/Bridges/Variable/map.jl
```diff
@@ -86,13 +86,15 @@ v2, c2 = MOIB.Variable.add_keys_for_bridge(map, () -> b2, set2)
     @test collect(values(map)) == [b1, b2]
 end
 
+struct _ZeroDimSet <: MOI.AbstractVectorSet
+    dimension::Int
+end
 b3 = VariableDummyBridge(3)
-set3 = MOI.Zeros(0)
+set3 = _ZeroDimSet(0)
```
blegat, line 93: `MOI.Zeros` should work

src/FileFormats/SDPA/SDPA.jl
```diff
@@ -164,10 +164,12 @@ function Base.write(io::IO, model::Model{T}) where {T}
     end
     println(io)
 
-    index_map = Vector{Tuple{Int,Int}}(
-        undef,
-        MOI.dimension(MOI.PositiveSemidefiniteConeTriangle(max_dim)),
-    )
+    max_index_dim = if max_dim > 0
+        MOI.dimension(MOI.PositiveSemidefiniteConeTriangle(max_dim))
```
blegat, line 168: I think this is not needed anymore, `MOI.PositiveSemidefiniteConeTriangle(0)` should work.

blegat (approved):

PR #1350, opened 2021-05-14
[bridges] unify IndexInVector

odow:
I don't think we need to do this. I tried it out and it actually makes things slightly less readable, because the Variable and Constraint modules need to have `MOIB.IndexInVector`. It's a trivial type that isn't the source of our problems, and the difference isn't even really measurable.

### DIFF
--- a/src/Bridges/Constraint/bridge.jl
+++ b/src/Bridges/Constraint/bridge.jl
@@ -7,15 +7,6 @@ bridges.
 abstract type AbstractBridge <: MOIB.AbstractBridge end
 
 """
-    IndexInVector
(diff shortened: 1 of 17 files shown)


blegat:
From withing the module, you need `MOIB.IndexInVector` instead of `IndexInVector` but from outside MOI, it replaces `MOIB.Constraint.IndexInVector` so it improves readability.
Some user might be confused using the wrong type and getting errors. Merging them should avoid this confusion.

odow:
Okay. Done.

blegat (approved):

PR #1351, opened 2021-05-14
Add ConstraintBasisStatus for free variables

### LINKED ISSUE #1297: Add VariableBasisStatus

odow:
An alternative to #1320

Closes #1320
Closes #1297

### DIFF
--- a/src/Test/UnitTests/solve.jl
+++ b/src/Test/UnitTests/solve.jl
@@ -462,3 +462,68 @@ function solve_twice(model::MOI.ModelLike, config::TestConfig)
     end
 end
 unittests["solve_twice"] = solve_twice
+
+"""
+The problem:
+```
+ min 1x
+s.t. x >= 1
+     y free
+```
+This has the solution `(x, y) = (1, 0)`. So the lower-bound constraint on `x` is
+non-basic, y is superbasic, and querying the status of x as a free variable
+throws `MOI.InvalidIndex`
+"""
+function solve_basis_lower_bound(model::MOI.ModelLike, config::TestConfig)
+    if !config.solve || !config.basis
+        return
+    end
+    MOI.empty!(model)
+    x = MOI.add_variable(model)
+    y = MOI.add_variable(model)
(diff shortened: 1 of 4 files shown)


joehuchette:
The “scalar function”-in-“vector set” is a bit confusing. I can see that this a bit niche and may not justify adding a scalar Real set on its own, but I do find that alternative more consistent.

odow:
You could argue that the set is `Reals(1)`, so the dimensions are consistent. 

I think we want to avoid the situation where `ConstraintIndex{SingleVariable,Real}` is more widely used. This is an attribute where we need the concept of an absence of a constraint. 

Another option is `ConstraintIndex{SingleVariable,AbstractScalarSet}`.

To summarize, the three main options are

1) Add `VariableBasisStatus` and keep `ConstraintBasisStatus` for `SingleVariable` constraints
2) Add `VariableBasisStatus` and remove `ConstraintBasisStatus` for `SingleVariable` constraints
3) Define `ConstraintBasisStatus` for `ConstraintIndex{SingleVariable,Reals}` as a special case

I'm okay with 1 or 3. But 2) just feels like we're adding a special case that requires a lot of deep code changes, whereas 3) is a special case with minimal impact.

odow:
Decision from the developer call was to go with #1320

PR #1358, opened 2021-05-19
Correct type in ProductOfConstraints

odow:
Working on writing the model for Clp.

@blegat I thought you had Clp working?

We also need a way of getting the row from the `MatrixOfConstraints`. Ideally, the `ConstraintIndex` would correspond to the row, and the `VariableIndex` would correspond to the column.


blegat:
Clp was working enough to have the benchmark working but not enough to have its tests pass

odow:
Yeah I'm almost there. Seems like there are a few bugs.

PR #1361, opened 2021-05-21
Remove is_empty assertion in CachingOptimizer

odow:
There is no reason that the `model_cache` field has to be empty when creating a `CachingOptimizer`. (And it isn't checked in the other method.)

This is also needed for an upcoming PR to JuMP.


blegat (approved):
This `@assert` made sense before https://github.com/jump-dev/MathOptInterface.jl/pull/1254/ as we were setting the mode to `ATTACHED_OPTIMIZER` but not anymore.

PR #1362, opened 2021-05-21
Fix index mapping of MatrixOfConstraints

blegat:
`are_indices_mapped` was storing the index of the set, not the index of the constraint in the cache so that wasn't working.
It was passing the tests because the mapping was always the identity so I now use `MockOptimizer` to make the mapping nontrivial.

Closes https://github.com/jump-dev/MathOptInterface.jl/pull/1360

### DIFF
--- a/src/Utilities/matrix_of_constraints.jl
+++ b/src/Utilities/matrix_of_constraints.jl
@@ -80,11 +80,8 @@ end
         coefficients::AT
         constants::BT
         sets::ST
-        caches::Union{Nothing,Vector}
-        are_indices_mapped::Union{Nothing,Vector{BitSet}}
-        function MatrixOfConstraints{T,AT,BT,ST}() where {T,AT,BT,ST}
-            return new{T,AT,BT,ST}(AT(), BT(), ST(), BitSet(), nothing)
-        end
+        caches::Vector
+        are_indices_mapped::Vector{BitSet}
     end
 
 Represents affine constraints in a matrix form where the linear coefficients
@@ -124,10 +121,10 @@ mutable struct MatrixOfConstraints{T,AT,BT,ST} <: MOI.ModelLike
     coefficients::AT
     constants::BT
     sets::ST
-    caches::Union{Nothing,Vector}
-    are_indices_mapped::Union{Nothing,Vector{BitSet}}
+    caches::Vector
+    are_indices_mapped::Vector{BitSet}
     function MatrixOfConstraints{T,AT,BT,ST}() where {T,AT,BT,ST}
-        return new{T,AT,BT,ST}(AT(), BT(), ST(), nothing, nothing)
+        return new{T,AT,BT,ST}(AT(), BT(), ST(), [], BitSet[])
     end
 end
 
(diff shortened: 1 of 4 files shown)


src/Utilities/matrix_of_constraints.jl
```diff
@@ -80,8 +80,8 @@ end
         coefficients::AT
         constants::BT
         sets::ST
-        are_indices_mapped::BitSet
-        caches::Union{Nothing, Vector}
+        caches::Union{Nothing,Vector}
+        are_indices_mapped::Union{Nothing,Vector{BitSet}}
```
odow, line 84: What is the point of making these `Union{Nothing` if `empty!` sets them to empty vectors?

src/Utilities/matrix_of_constraints.jl
```diff
@@ -240,7 +240,7 @@ function _allocate_constraints(
     for ci_src in cis_src
         func = MOI.get(src, MOI.CanonicalConstraintFunction(), ci_src)
         set = MOI.get(src, MOI.ConstraintSet(), ci_src)
-        push!(model.are_indices_mapped, length(model.caches) + 1)
+        push!(model.are_indices_mapped[i], length(model.caches[i]) + 1)
```
odow, line 243: Ah. This was the bit that really confused me. Makes sense now.

src/Utilities/matrix_of_constraints.jl
blegat: If you add or remove set types to the `sets` field then the length of `caches` and `are_indices_mapped` should change so creating new ones at `empty!` is more robust to that.

odow: To clarify: do we ever use the fact that these are `nothing`? I can't see so. So the caches could be initialized as `Any[]` and `BitSet[]`?

blegat: The `nothing` is what's modeling best the situation I feel. When you are not in `copy_to`, `caches` and `are_indices_mapped` don't make sense so it's appropriate that it's `nothing` so that it's explicit.

odow: But this assumes you call `empty!` at the start of `copy_to`? Did you notice any compilation issues with the union in the functions? The compiler might not like we iterate over a `Union{Nothing,Vector}`. The code is simpler if we make this `caches::Vector{Any}` and `::Vector{BitSet}`, and add some comments to the docstring.

src/Utilities/matrix_of_constraints.jl
```diff
@@ -124,10 +121,10 @@ mutable struct MatrixOfConstraints{T,AT,BT,ST} <: MOI.ModelLike
     coefficients::AT
     constants::BT
     sets::ST
-    are_indices_mapped::BitSet
-    caches::Union{Nothing,Vector}
+    caches::Vector
```
odow, line 124: `Vector{Any}`? Or `Vector{Vector{Tuple{DataType,DataType}}}`?

blegat: Each entry of `caches` is a `Vector` of concrete `eltype`. If we write `Vector{Vector{...}}` where `...` is an abstract type then it will convert these vectors to vector with non-concrete eltypes.

odow (approved):

PR #1367, opened 2021-05-23
Fix pass_nonvariable_constraints with bridged variables bis

blegat:
Forgot the `return` in https://github.com/jump-dev/MathOptInterface.jl/pull/1354/ so things were copied twice :facepalm:


odow:
Yeah I don't fully understand the role of `pass_nonvariable_constraints`. Once the changes to matrixofconstraints are done, I'm going to go through and we what we can do.

blegat:
Meanwhile, any objection to merge this ?

odow:
Nope. You're good to merge.

PR #1368, opened 2021-05-23
Various changes and refactoring of ProductOfSets

odow:
@blegat this is still WIP. I wrote proper tests for ProductOfSets and found some bugs (see `@test_broken`).

What is the point of returning the `LazyMap`?

I also changed a few types to private. We can probably split this PR up once things are working, or merge it as a mix of updates.

### DIFF
--- a/src/Utilities/product_of_sets.jl
+++ b/src/Utilities/product_of_sets.jl
@@ -12,7 +12,7 @@ Return an integer corresponding to the index of the set type in the list given
 by [`set_types`](@ref). If this set is not part of the list then it returns
 `nothing`.
 """
-set_index(sets::ProductOfSets, ::Type{S}) where {S<:MOI.AbstractSet} = nothing
+set_index(::ProductOfSets, ::Type{S}) where {S<:MOI.AbstractSet} = nothing
 
 """
     set_types(sets::ProductOfSets)
@@ -73,6 +73,7 @@ Product of scalar sets in the order the constraints are added, mixing the
 constraints of different types.
 """
 abstract type MixOfScalarSets{T} <: ProductOfSets{T} end
+
 macro mix_of_scalar_sets(name, set_types...)
     esc_name = esc(name)
     T = esc(:T)
@@ -86,13 +87,18 @@ macro mix_of_scalar_sets(name, set_types...)
 end
 
 MOI.is_empty(sets::MixOfScalarSets) = isempty(sets.set_ids)
+
 MOI.empty!(sets::MixOfScalarSets) = empty!(sets.set_ids)
+
 MOI.dimension(sets::MixOfScalarSets) = length(sets.set_ids)
+
 indices(::MixOfScalarSets, ci::MOI.ConstraintIndex) = ci.value
+
 function add_set(sets::MixOfScalarSets, i)
     push!(sets.set_ids, i)
     return length(sets.set_ids)
 end
+
 function MOI.get(
     sets::MixOfScalarSets{T},
     ::MOI.ListOfConstraintTypesPresent,
@@ -103,6 +109,7 @@ function MOI.get(
         S in set_types(sets) if set_index(sets, S) in present
     ]
 end
+
 function MOI.get(
     sets::MixOfScalarSets,
     ::MOI.NumberOfConstraints{F,S},
@@ -110,19 +117,18 @@ function MOI.get(
     i = set_index(sets, S)
     return count(isequal(i), sets.set_ids)
 end
+
 function MOI.get(
     sets::MixOfScalarSets,
     ::MOI.ListOfConstraintIndices{F,S},
 ) where {F,S}
     i = set_index(sets, S)
-    return LazyMap{MOI.ConstraintIndex{F,S}}(
-        j -> MOI.ConstraintIndex{F,S}(j),
-        Base.Iterators.Filter(
-            j -> sets.set_ids[j] == i,
-            eachindex(sets.set_ids),
-        ),
-    )
+    return MOI.ConstraintIndex{F,S}[
+        MOI.ConstraintIndex{F,S}(j)
+        for j in eachindex(sets.set_ids) if sets.set_ids[j] == i
+    ]
 end
+
 function MOI.is_valid(
     sets::MixOfScalarSets,
     ci::MOI.ConstraintIndex{F,S},
@@ -140,6 +146,7 @@ Product of sets in the order the constraints are added, grouping the
 constraints of the same types contiguously.
 """
 abstract type OrderedProductOfSets{T} <: ProductOfSets{T} end
+
 macro product_of_sets(name, set_types...)
     esc_name = esc(name)
     T = esc(:T)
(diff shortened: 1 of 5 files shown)


src/Utilities/product_of_sets.jl
```diff
@@ -140,86 +146,85 @@ Product of sets in the order the constraints are added, grouping the
...
     return _sets_code(esc_name, T, type_def, set_types...)
 end
 
-"""
-    abstract type OrderedProductOfScalarSets{T} <: OrderedProductOfSets{T} end
-
-Same as [`OrderedProductOfSets`](@ref) except that all types are scalar sets,
-which allows a more efficient implementation.
-"""
-abstract type OrderedProductOfScalarSets{T} <: OrderedProductOfSets{T} end
```
odow, line 164: @blegat I removed this.

src/Utilities/product_of_sets.jl
```diff
@@ -140,86 +146,85 @@ Product of sets in the order the constraints are added, grouping the
...
+            A dictionary which maps the `set_index` and `offset` of a set to the
+            dimension, i.e., `dimension[(set_index,offset)] → dim`.
+            """
             dimension::Dict{Tuple{Int,Int},Int}
-            function $esc_name{$T}() where {$T}
-                return new(zeros(Int, $(1 + length(set_types))), Dict{Int,Int}())
+
+            function $(esc_name){$(T)}() where {$(T)}
+                return new(
+                    zeros(Int, $(length(set_types))),
```
odow, line 168: @blegat I changed `num_rows` to store the number of elements, rather than the cumulative list. It makes the code simpler, and the extra summations required are unlikely to cause a bottleneck.

src/Utilities/product_of_sets.jl
```diff
@@ -140,86 +146,85 @@ Product of sets in the order the constraints are added, grouping the
...
     return
 end
-function MOI.empty!(sets::OrderedProductOfScalarSets)
-    fill!(sets.num_rows, 0)
-    return
-end
 
-function MOI.dimension(sets::OrderedProductOfSets)
-    for i in 3:length(sets.num_rows)
-        sets.num_rows[i] += sets.num_rows[i-1]
```
odow, line 190: @blegat this was the cause of much confusion for me before I found the cause.

src/Utilities/product_of_sets.jl
```diff
@@ -103,26 +109,26 @@ function MOI.get(
...
 ) where {F,S}
     i = set_index(sets, S)
-    return LazyMap{MOI.ConstraintIndex{F,S}}(
-        j -> MOI.ConstraintIndex{F,S}(j),
-        Base.Iterators.Filter(
-            j -> sets.set_ids[j] == i,
-            eachindex(sets.set_ids),
-        ),
-    )
+    return MOI.ConstraintIndex{F,S}[
```
blegat, line 126: What's the point of collecting ? Being able to iterate through the list of indices without needing to collect them can be useful

src/Utilities/product_of_sets.jl
blegat: When there is only scalar set, `rows` is trivial with `num_rows` being the cumulative sum. Having to take a cumulative sum should make it much slower. I don't think the simplification is worth it.

src/Utilities/product_of_sets.jl
odow: Simplifying the number of types we are using. Unless we identify something as a bottleneck, we should use the simplest approach. Adding the LazyMap here just complicates things.

src/Utilities/product_of_sets.jl
odow: > When there is only scalar set

Then they should use `MixOfScalarSets`.

> Having to take a cumulative sum should make it much slower

I don't think so, because this vector is usually pretty small (e.g., length 4). Having to maintain the cumulative vector every time we add a new set also takes time.

blegat: > Then they should use `MixOfScalarSets`.

Not so sure, `MixOfScalarSets` is slower since it needs one push per constraints so if you don't mind your constraints being reordered, `ProductOfSets` is faster.

> Having to maintain the cumulative vector every time we add a new set also takes time.

We don't have to maintain it every time, we update it at the end of `copy_to`. You cannot add a constraint outside of `copy_to`. During `copy_to` it represents the number for each type of set and at the end of `copy_to`, it represents the cumulative.

src/Utilities/product_of_sets.jl
blegat: Simple is a bit subjective. `LazyMap` is already used a lot in bridges and so on so it's not like we introduce another type.
Writing things slower until we have a benchmark isn't the right approach either. We should at least write it the fastest way if it does not decrease readability too much. In this case, I don't think it affects readability that much.

odow: We don't have any benchmarks showing that `LazyMap` is faster.

src/Utilities/product_of_sets.jl
odow: Why do we have `MixOfScalarSets` then? We should just be aiming to minimize the surface area of this code.

The previous code was mutating `.sets` when it called `MOI.dimension`, and so was implicitly assuming that `MOI.dimension` was called once and only once, and at the beginning of `final_touch` before everything else. That seems very brittle. 

Once we have things working, with solvers hooked up, we can return with benchmarks. Until then, I'm leaving as-is.

src/Utilities/product_of_sets.jl
odow: Repeating my point below, we can return to this once everything is hooked up and we can run benchmarks on solvers. Until then, I'm leaving as-is.

src/Utilities/product_of_sets.jl
blegat: Then I suggest having a `final_touch` for the sets API as well. This is consistent with `MatrixOfConstraints` and `MutableSparseMatrixCSC`. In `final_touch` you do the cumulative sum and in `dimension` you just return `sets.num_rows[end]`.
If you look at `colptr` of `MutableSparseMatrixCSC`, the meaning of the stored values also changes throughout the API calls. These are not easy to use API that you can safely use in any order. It's only meant to be used by `MatrixOfConstraints` or adventurous users.

> Why do we have `MixOfScalarSets` then?

Not sure yet where we are headed but I wrote them to compare. We need to have that at least even here or in MatrixOptInterface since it's useful for some formats there where the constraints have mixed order.

odow: Okay, I added `final_touch` and did the ordering.

src/Utilities/product_of_sets.jl
```diff
@@ -34,28 +35,60 @@ end
 
 Product of scalar sets in the order the constraints are added, mixing the
 constraints of different types.
+
+Use [`@mix_of_scalar_sets`](@ref) to generate a new subtype.
+
+If the sets are all scalar, use this type instead of [`OrderedProductOfSets`](@ref)
```
blegat, line 41: I wouldn't say it's more efficient. In fact, for Clp I ended up using `OrderedProductOfSets` as it was faster.

src/Utilities/product_of_sets.jl
odow: > useful for some formats there where the constraints have mixed order

Which formats? We can always add things in later, but we don't want to be carrying code and tests just for the sake of it.

odow:
Merging. Let's address any remaining issues in other PRs.

PR #1369, opened 2021-05-24
Add tests for Utilities/sparse_matrix.jl

odow:
No bugs. Just some tidying and adding stand-alone tests.

### DIFF
--- a/src/Utilities/sparse_matrix.jl
+++ b/src/Utilities/sparse_matrix.jl
@@ -128,7 +128,7 @@ _variable(term::MOI.VectorAffineTerm) = _variable(term.scalar_term)
 """
     allocate_terms(
         A::MutableSparseMatrixCSC,
-        index_map::MOI.Utilities.IndexMap,
+        index_map,
         func::Union{MOI.ScalarAffineFunction,MOI.VectorAffineFunction},
     )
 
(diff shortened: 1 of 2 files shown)


src/Utilities/sparse_matrix.jl
```diff
@@ -116,61 +118,71 @@ function final_touch(A::MutableSparseMatrixCSC)
...
+    return
 end
 
 _variable(term::MOI.ScalarAffineTerm) = term.variable
 _variable(term::MOI.VectorAffineTerm) = _variable(term.scalar_term)
-function _allocate_terms(colptr, index_map, terms)
-    for term in terms
-        colptr[index_map[_variable(term)].value+1] += 1
-    end
-end
```
odow, line 128: @blegat what was the motivation to have these `_allocate_terms` function barriers?

blegat: It dates back to https://github.com/jump-dev/SCS.jl/pull/192 where I benchmarked the two approaches and the function barrier seemed to be faster even if I didn't really get why.
It's maybe not the case anymore in Julia v1.6 or maybe I was mistaken when writing this.
We can revisit later. An easier speedup would be obtained by using `@inbounds` I guess.
The good thing with MatrixOfConstraints is that we will be able to spend time optimizing these pieces at once in MOI and it will affect all solver's `copy_to` performance so it might start being worth playing with `@inbounds`, etc...

odow: Given the massive amount of new complexity we're introducing, I'd rather that we merged a simplified version first, got it working and debugged, then wrote the various solvers, and only then revisited performance implications with benchmarks.

blegat: Yes, definitely

PR #1371, opened 2021-05-25
Remove MixOfScalarSets

odow:
There's no point having two versions if we can simplify to only `OrderedProductOfSets` (now renamed `ProductOfSets`).

The `MixOfScalarSets` doesn't add anything w.r.t. constraint ordering, because the way we cache the constraints first means that they will get added in this `OrderedProductOfSets` order regardless.

### DIFF
--- a/docs/src/submodules/Utilities/reference.md
+++ b/docs/src/submodules/Utilities/reference.md
@@ -128,9 +128,6 @@ Utilities.rows
 
 ```@docs
 Utilities.ProductOfSets
-Utilities.MixOfScalarSets
-Utilities.@mix_of_scalar_sets
-Utilities.OrderedProductOfSets
 Utilities.@product_of_sets
 ```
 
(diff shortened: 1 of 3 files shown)


blegat:
I agree that we should make this decision before MOI v0.10 but we're not there yet. Can we delay it until then ? I haven't experimented enough yet to make a call.

odow:
Yes, we can delay. But we should aim to delete, unless there is clear evidence to the contrary.

odow:
Bump. Did you find any reason to keep this?

blegat:
It will be used by MatrixOptInterface. If the users gives you an LP where the sets are mixed up, you can either reorder the rows of the matrix or use this set. I would argue that the second option is best. We could move the implementation to MatrixOptInterface but the advantage of having it here is that it allows to make small changes in the MatrixOfConstraints interfaces with the product of sets without breaking any package

odow:
Okay. I'm happy to close.

PR #1374, opened 2021-05-26
Add conic test with starting values

blegat:
Add a test combining a conic constraint and starting values.
This would have caught https://github.com/jump-dev/MosekTools.jl/issues/66


odow (approved):
See the format failure.

PR #1375, opened 2021-05-27
Document order of multiple solutions

### LINKED ISSUE #1372: Order of solutions
When the termination status is `OPTIMAL`, do the solver need to guarantee that the optimal solution is the first one ?
If this is the case, then this is a bug: https://github.com/jump-dev/MosekTools.jl/issues/66
In fact, we kind of assume it the tests already since we only ever test with the first solution.
Is this in the documentation somewhere ?
I remember we discussed about that, see e.g. https://github.com/jump-dev/JuMP.jl/pull/2512#discussion_r592657852

odow:
Closes #1372

### DIFF
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -928,9 +928,7 @@ struct ObjectiveFunctionType <: AbstractModelAttribute end
 
 A model attribute for the objective value of the primal solution `result_index`.
 
-If multiple results are present, the first result (i.e., `result_index=1`) must
-be an optimal solution. Other results may be alternative optimal solutions; use
-`ObjectiveValue` to compare their objective values.
+See [`ResultCount`](@ref) for information on how the results are ordered.
 """
 struct ObjectiveValue <: AbstractModelAttribute
     result_index::Int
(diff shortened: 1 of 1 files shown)


blegat:
> If multiple results are present, the first result (i.e., `result_index=1`) must be an optimal solution.

This disallows having multiple results in case the termination status is not `OPTIMAL`.
I would rather say something like

> If the termination status is `OPTIMAL`, and multiple results are present, the first result (i.e., `result_index=1`) must be an optimal solution.
> If the termination status is `INFEASIBLE` and one of the solutions is an infeasibility certificate, it should be first result.
> If the termination status is `DUAL_INFEASIBLE` and one of the solutions is a dual infeasibility certificate, it should be the first result.

At least that's what we assume in the tests so it makes sense to make it a requirements.

odow:
How's this now.

src/attributes.jl
```diff
@@ -1000,6 +1004,29 @@ struct RawSolver <: AbstractModelAttribute end
     ResultCount()
 
 A model attribute for the number of results available.
+
+## Order of solutions
+
+A number of attributes contain an index, `result_index`, which is used to refer
+to on of the available results. `result_index` must be an integer between `1`
```
blegat, line 1011: on -> one

odow: Fixed.

blegat (approved):

PR #1381, opened 2021-06-03
Allow changing default cache for instantiate

blegat:
The main use case for this is that once a solver defines a fast `copy_to` method with a `MatrixOfConstraint` model, then it can use it as cache directly to avoid having to store an extra cache. As `instantiate` is used by JuMP, this will be what's used by default in JuMP.

### DIFF
--- a/docs/src/reference/models.md
+++ b/docs/src/reference/models.md
@@ -60,6 +60,7 @@ AbstractOptimizer
 OptimizerWithAttributes
 optimize!
 instantiate
+default_cache
 ```
 
 ## Optimizer attributes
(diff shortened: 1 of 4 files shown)


odow (commented):
This is really only useful for the cache inside a bridge, correct? 

But this seems like you want JuMP to go with Option 2 as discussed here: https://github.com/jump-dev/Clp.jl/pull/114#issuecomment-848312487.

This has some pretty big implications that I don't fully understand. Like it's now not possible to modify a cache after `copy_to` has been called.

blegat:
> Like it's now not possible to modify a cache after `copy_to` has been called.

Yes, it is breaking. However, this cache is mostly there because the bridges need the underlying model to support incremental loading of the model, not to support modification if the solver don't support it.

odow:
Before merging, I'd like to see this implemented by a solver.

blegat:
I was doing it for Clp and then I got hit by issues which lead me to https://github.com/jump-dev/MathOptInterface.jl/pull/1386 and https://github.com/jump-dev/MathOptInterface.jl/pull/1387. I'll continue when these two are done.

PR #1386, opened 2021-06-15
EMPTY_OPTIMIZER mode after empty!

blegat:
Follow up of https://github.com/jump-dev/MathOptInterface.jl/pull/1254

- [x] Add tests
- [x] Update doc


odow (approved):
This seems reasonable.

PR #1387, opened 2021-06-15
Various improvements to MatrixOfConstraints

blegat:
Various improvements to `MatrixOfConstraints`. It can now pass the relevant tests in `Test/modellike.jl`.

- Add macros to create subtypes of `StructOfConstraints` and improve `StructOfConstraints` so that sub-fields can be `MatrixOfConstraints`s.
- [breaking] Use `Box` in `MOIU.Model` for storing variable bounds.
- Implement some of the getters for MOI attributes.

### DIFF
--- a/src/Test/modellike.jl
+++ b/src/Test/modellike.jl
@@ -25,7 +25,7 @@ function default_status_test(model::MOI.ModelLike)
     @test MOI.get(model, MOI.DualStatus()) == MOI.NO_SOLUTION
 end
 
-function nametest(model::MOI.ModelLike)
+function nametest(model::MOI.ModelLike; delete::Bool=true)
     @testset "Variables" begin
         MOI.empty!(model)
         x = MOI.add_variables(model, 2)
@@ -69,7 +69,7 @@ function nametest(model::MOI.ModelLike)
         MOI.set(model, MOI.ConstraintName(), c1, "c1")
         @test_throws ErrorException MOI.get(model, MOI.ConstraintIndex, "c1")
     end
-    @testset "Name test with $(typeof(model))" begin
+    @testset "Name" begin
         MOI.empty!(model)
         @test MOI.supports_incremental_interface(model, true) #=copy_names=#
         @test MOI.supports(model, MOI.Name())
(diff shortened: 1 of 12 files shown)


src/Test/modellike.jl
```diff
@@ -25,7 +25,7 @@ function default_status_test(model::MOI.ModelLike)
     @test MOI.get(model, MOI.DualStatus()) == MOI.NO_SOLUTION
 end
 
-function nametest(model::MOI.ModelLike)
+function nametest(model::MOI.ModelLike; delete::Bool=true)
```
odow, line 28: Shouldn't this take a `Test.Config`?

odow:
What is the motivation for this?

src/Test/modellike.jl
blegat: I was hesitant because then we also need to take into account other fields such as `query`, etc... and that will be a lot of `if`-`else`. We could do this but we should first agree that's the direction we want to go since that might be a few changes

blegat:
> What is the motivation for this?

The StructOfConstraints macro is needed, e.g. for solvers that have a separate matrix for equality constraints and conic constraints, e.g., ECOS.
Using Box in MOIU.Model allows to share code between MatrixOfConstraints and AbstractModel, notably the getter for AbstractSet which has a few methods.
For the getters, I get we could choose not to implement them but they are quite straightforward and if they are not supported, then the hacks we will need in the tests of the solver wrappers using a MatrixOfConstraints as cache would be more work than this PR (and maintaining these features). Moreover, having them allows to use more modellike tests like the one added in this PR which is quite useful to ensure that MatrixOfConstraints behaves consistenctly with how a good ModelLike should behave.

src/Utilities/modif_matrix.jl
```diff
@@ -0,0 +1,63 @@
+function MOI.delete(v::MatrixOfConstraints, ci::MOI.ConstraintIndex)
```
odow, line 1: Why are these functions in a different file? They aren't tested.

src/Utilities/struct_of_constraints.jl
```diff
@@ -274,16 +312,21 @@ function struct_of_constraint_code(struct_name, types, field_types = nothing)
...
+        return $typed_struct(0, $([:(nothing) for _ in field_types]...))
+    end)
+    if type_parametrized
+        append!(constructor_code.args[1].args, field_types)
     end
+    push!(code.args, constructor_code)
     return code
 end
+
+macro struct_of_constraints_by_function_types(name, func_types...)
```
odow, line 325: These need docstrings and added to the docs

odow (approved):
These changes make sense, modulo some more documentation and testing.

src/Utilities/modif_matrix.jl
blegat: Oops, this file shouldn't have been pushed

PR #1393, opened 2021-06-18
Fix and add test for throwing UnsupportedSubmittable

### LINKED ISSUE #1357: bug with error throwing
[This throw](https://github.com/jump-dev/MathOptInterface.jl/blob/ce19dce9147a8040d8582c79179b541b64e1ce31/src/attributes.jl#L472) does not provide an error message, where `UnsupportedSubmittable` requires one.

odow:
Closes #1357

### DIFF
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -472,7 +472,7 @@ function submit(model::ModelLike, sub::AbstractSubmittable, args...)
         throw(
             UnsupportedSubmittable(
                 sub,
-                "submit(::$(typeof(model)),::$(typeof(sub))) is not supported.",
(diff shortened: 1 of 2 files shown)


src/attributes.jl
```diff
@@ -469,7 +469,12 @@ function submit(model::ModelLike, sub::AbstractSubmittable, args...)
             ),
         )
     else
-        throw(UnsupportedSubmittable(sub))
+        throw(
+            UnsupportedSubmittable(
+                sub,
+                "submit(::$(typeof(model)),::$(typeof(sub))) is not supported.",
```
blegat, line 475: Missing a space after the comma

PR #1394, opened 2021-06-18
Update error message of get_fallback

### LINKED ISSUE #799: Improve get fallback
The message is incorrect if the signature is incorrect:
https://github.com/JuliaOpt/MathOptInterface.jl/blob/e458dfd55b581d2c0470a69095ab0001df823d82/src/attributes.jl#L248-L250
See https://github.com/JuliaOpt/MathOptInterface.jl/pull/797#issuecomment-513065543

odow:
Closes #799

### DIFF
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -295,10 +295,13 @@ end
 function get(model::ModelLike, attr::AnyAttribute, args...)
     return get_fallback(model, attr, args...)
 end
+
 function get_fallback(model::ModelLike, attr::AnyAttribute, args...)
     return throw(
         ArgumentError(
-            "ModelLike of type $(typeof(model)) does not support accessing the attribute $attr",
+            "Unable to get `$(attr)` with the provided arguments. Either " *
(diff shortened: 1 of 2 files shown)


src/attributes.jl
```diff
@@ -295,25 +295,33 @@ end
...
     return get_fallback(model, attr, args...)
 end
+
 function get_fallback(model::ModelLike, attr::AnyAttribute, args...)
     return throw(
         ArgumentError(
-            "ModelLike of type $(typeof(model)) does not support accessing the attribute $attr",
+            "Unable to get `$(attr)` with the provided arguments. Either " *
+            "$(typeof(model)) does not support the attribute, or the " *
+            "provided arguments are incorrect.",
```
blegat, line 304: It would be better to distinguish the cases. We just have to write 3 methods: one for model/solver attributes, one for variables attributes and one for constraint attibutes to make sure we don't catch any incorrect arguments. I agree that the current is confusing and I get it a lot. Having it say it can be 2 different things is less helpful.

blegat (approved):

PR #1395, opened 2021-06-18
Fix isapprox when the first function is missing terms

blegat:


odow:
Ah. This is https://github.com/jump-dev/MathOptInterface.jl/issues/1483 :)

PR #1400, opened 2021-06-20
[breaking] remove operate(::typeof(sum)

### LINKED ISSUE #1388: sum with VariableIndex
`MOI.VariableIndex` only represents an index, not a variable that can we can operate over with arithmetic operations to get affine, quadratic expressions, etc... `MOI.SingleVariable` plays that role. For that reason,
https://github.com/jump-dev/MathOptInterface.jl/blob/master/src/Utilities/functions.jl#L2496-L2502
should be removed.

odow:
I didn't run all the tests locally yet, so let's see what, if anything, this breaks.

Closes #1388

### DIFF
--- a/src/Bridges/Constraint/det.jl
+++ b/src/Bridges/Constraint/det.jl
@@ -199,8 +199,8 @@ function subsum(
     l::Vector{MOI.VariableIndex},
     ::Type{T},
 ) where {T}
-    n = length(l)
-    f = MOIU.operate!(-, T, t, MOIU.operate(sum, T, l))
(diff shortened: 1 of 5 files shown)


odow:
@blegat looks like you use it here:
https://github.com/jump-dev/MathOptInterface.jl/blob/9289e6e428b9bc752a356cf3eebb5ebe4232afad/src/Bridges/Constraint/det.jl#L203

Should I replace? Or can we just leave the definition and close the issue.

blegat:
We could rename it into something like `sum_of_variables(::Type{T}, ::Vector{MOI.VariableIndex})`

odow:
@blegat I just inlined the call.

blegat (approved):

PR #1401, opened 2021-06-21
[FileFormats.CBF] Refactor CBF module

### LINKED ISSUE #1072: Incorrect CBF file output
Using the same example in #1071, now I am writing CBF file:
```julia
JuMP.write_to_file(m, "ex.cbf", format=MOI.FileFormats.FORMAT_CBF)
```
The contents of `ex.cbf` should be
```
VER
3

OBJSENSE
MIN

VAR
8 1
F 8

OBJACOORD
5
0 1.0
1 6.0
3 9.0
2 9.0
5 7.0

CON
6 6
L= 1
L= 1
L+ 1
L+ 1
L- 1
L- 1

ACOORD
15
0 0 1.0
0 3 2.0
0 2 3.0
0 4 14.0
0 5 5.0
0 6 1.0
1 1 4.0
1 3 16.0
1 2 6.0
1 5 4.0
1 7 1.0
2 6 1.0
3 7 1.0
4 6 1.0
5 7 1.0

BCOORD
4
0 1.3399999999999999
1 -6.66
4 -111.06
5 -111.06

PSDCON
1
3

HCOORD
6
0 0 0 0 1.0
0 1 1 0 1.0
0 2 1 1 1.0
0 3 2 0 1.0
0 4 2 1 1.0
0 5 2 2 1.0
```

According to the documentation (http://cblib.zib.de/doc/format2.pdf), the CBF format should follow certain orders. For example, the version should come first, the problem structure is the next, and the problem data is the last. In the example output above, `OBJACOORD` is of the problem data, and `CON` and `PSDCON` are of the problem structure.

Moreover, `PSDCON` should be defined after `PSDVAR`, which is missing in the output. It looks like the writer considered everything as scalar variable.

Again, like in #1071, JuMP can still read this file for Mosek to solve. But, Mosek cannot read this file.

odow:
Part of #1072 

- I've split the reader and writer in to separate files
- Refactored the writer into functions, as opposed to one monolithic function
- Fixed the ordering of sections when writing

This doesn't close #1072 because it doesn't support PSD variables.


odow:
I opened a separate issue https://github.com/jump-dev/MathOptInterface.jl/issues/1402

Issue #1201, opened 2020-11-27
Add tests for DualObjectiveValue when infeasible


odow:
See discussion here: https://github.com/jump-dev/MathOptInterface.jl/pull/1190

Issue #470, opened 2018-08-15
add an NLP test with NaN returned from the callbacks


mlubin:
Ref https://github.com/JuliaOpt/Ipopt.jl/pull/133

shadiakiki1986:
I'm trying to tackle this issue by writing tests for it in my forks of MOI and Ipopt. My results don't match with my expectations, so I have a few questions. This is very draft-ish with a lot of coding malpractices like function and variable names as well as unnecessary code from copy-paste.

1. Is my approach of implementing tests in MOI and calling them from Ipopt in line with how you're building the packages?
2. My first test `max_x_is_inf` is implemented in MOI [here](https://github.com/shadiakiki1986/MathOptInterface.jl/blob/sa/isue470_nlptest_nancallback/src/Test/nlp_nancallback.jl#L8), but it errors when called from Ipopt [here](https://github.com/shadiakiki1986/Ipopt.jl/blob/sa/nan_callbacks/test/MOI_nlp_nancallback.jl#L31) with `Attribute MathOptInterface.VariableName() is not supported by the model`. What I was aiming for with this test is to set up an optimization that goes to infinity and see if it starts popping up NaN's on the way to Inf. I ended up re-implementing this idea in my next test `nancb_test` (details below) but without the `MOIU.loadfromstring`, which I was hoping would work and be a very simple and compact implementation of tests for this issue. I think I'm missing something very obvious about the syntax of the string model here.
3. The next test `nancb_test` is defined in MOI [here](https://github.com/shadiakiki1986/MathOptInterface.jl/blob/sa/isue470_nlptest_nancallback/src/Test/nlp_nancallback.jl#L164) and called in Ipopt [here](https://github.com/shadiakiki1986/Ipopt.jl/blob/sa/nan_callbacks/test/MOI_nlp_nancallback.jl#L36). It is a simple `max x st x>=1`. It's mostly copied from the NLP HS071 test with modifications for the simpler model here. The termination status that ends up here is `MOI.NORM_LIMIT`, with primal status = `UNKNOWN_RESULT_STATUS`. The objective value comes out at 4.94, but I get it that whatever number here is invalid given the unknown status flag. I don't understand why it doesn't stop at something much larger, but maybe this has to do with some configurations for Ipopt that I missed.
[...]

PR #1404, opened 2021-06-22
[breaking] [Tests] Large refactoring of Test

odow:
This PR was motivated by issue https://github.com/jump-dev/MathOptInterface.jl/issues/1398

## Where we are 

Our current testing regime is comprehensive, but a bit all over the place.

There's a mix of things like `MOI.Test.unittest` that wrap a whole lot of tests, and others like `MOI.Test.default_status_test` that you just need to add. Even the documentation for how to test a solver is complicated (https://github.com/jump-dev/MathOptInterface.jl/issues/224)! https://jump.dev/MathOptInterface.jl/dev/submodules/Test/overview/#How-to-test-a-solver

The current design is also bad because it's hard to add new tests. 

As evidenced by the documentation: https://jump.dev/MathOptInterface.jl/dev/submodules/Test/overview/#How-to-add-a-test, you need to write a test, then write a test for the test, and then make sure that everything works. It's hard to run a small contingent of tests, and it's hard to decide where to put a new test, and where to put the test of the test.

This also evidenced by the large number of open test issues that aren't getting addressed. These range from #470 (August 2018!) to #1201 (November 2020). If they were easier to add, it would happen quicker!

Naming and visibility of each test is also a problem: https://github.com/jump-dev/MathOptInterface.jl/issues/1029. If `linear10btest` fails, what does that mean? 

## New design

The basic design is `runtests`, a single entry point to _all_ tests in MOI.

Instead of breaking down tests by files or dictionaries, tests are normal Julia functions with descriptive names that can be excluded or included by the user.

Here's the code to test the `MockOptimizer`:
```Julia
MOI.Test.runtests(
    MOIU.MockOptimizer(MOIU.UniversalFallback(MOIU.Model{Float64}())),
    MOI.Test.Config(basis = true),
    exclude = ["test_nonlinear_"],
)
```
Much better. 

There is also a need for certain tests to modify the model prior to running the test (changing solver parameters/tolerances, for example). That can be achieved by overloading `setup_test(::typeof(f), ::MOI.ModelLike, ::MOI.Test.Config)` for the particular function `f`.

## Decisions and TODOs

This is horribly breaking, but I think we're okay with that. Sorting out the tests is a high priority.

- [x] Update documentation
- Finish migrating tests
  - [x] attributes
  - [ ] basic_constriant_tests
  - [x] constraints
  - [x] modifications
  - [x] objectives
  - [x] solve
  - [x] variables
  - [ ] contconic
  - [x] contlinear
  - [x] contquadratic
  - [x] intconic
  - [x] intlinear
  - [ ] modellike
  - [x] nlp
- [ ] It's also apparent that we would need some style guide for the test names. How descriptive do you get? These integration ones are difficult because they cover lots of topics. The shorter functions would be easier.
- [x] We need an easier way of writing tests for the tests. Perhaps we can create a callback in `runtests` that sets up the model (e.g., modifying parameters, etc) and the config? Then we could make sure that every test is actually tested, and it would be simpler to add new tests.
- [ ] Split into smaller PRs? Now that I've worked out the design and migrated the majority of the tests, it seems to work well.


odow:
I think I'm making progress on the design. 

The main premise is that you write a short test, and then immediately afterwards you define `setup_test` which handles the overload for `Utilities.MockOptimizer`:
https://github.com/jump-dev/MathOptInterface.jl/blob/ca415c070b44a820478949eae41bda3eeda3bd36/src/Test/UnitTests/variables.jl#L202-L258
That means you only need to write the test in one place, and you don't have to worry about hunting through the tests of the tests to figure out how to test the test!

Then, there's just a single entry point where you can run all of the tests in MOI for your solver:
https://github.com/jump-dev/MathOptInterface.jl/blob/ca415c070b44a820478949eae41bda3eeda3bd36/test/Test/Test.jl#L14-L22

src/Test/UnitTests/attributes.jl
```diff
@@ -1,21 +1,21 @@
 """
-    solver_name(model::MOI.ModelLike, config::Config)
+    test_solver_name(model::MOI.ModelLike, config::Config)
 
 Test that the [`MOI.SolverName`](@ref) attribute is implemented for `model`.
 """
-function solver_name(model::MOI.ModelLike, config::Config)
+function test_SolverName(model::MOI.ModelLike, config::Config)
```
dourouc05, line 6: The name of the function is not terribly consistent with the other function names (the same holds for all attributes). Maybe something like `test_attribute(model, ::MOI.SolverName, config)`, with the same change for `setup_test`? 

It could also be easier for solvers to implement, with a simple loop over attributes that are supported instead of a longer list of function calls.

dourouc05 (commented):
This would indeed make the Test module much easier to understand, in my opinion!

src/Test/UnitTests/attributes.jl
odow: Yeah this is my naming problem. I don't have a good solution.

Specific attributes:
```Julia
test_attribute_SolverName
test_modification_MultirowChange
test_modification_ScalarCoefficientChange_objective
test_modification_ScalarCoefficientChange_constraint
```

Problem classes?
```julia
test_lp_
test_milp_
test_qp_
test_qcp_
test_nlp_
test_soc_
test_conic_
```

Ideally, we need a formulaic way of generating test names:

`test_` + class + feature + unique identifier?

 - `test_lp_TerminationStatus_INFEASIBLE`
 - `test_lp_TerminationStatus_DUAL_INFEASIBLE`
 - `test_lp_PrimalStatus_INFEASIBILITY_CERTIFICATE`
 - `test_milp_integration_knapsack`
 - `test_soc_VectorOfVariables_extra_terms`
 - `test_soc_VectorAffineFunction_empty_row`

You want to be able to say things like 
"run all `test_soc_` problems excluding `INFEAS`"

blegat:
I feel that it's easier to maintain the tests we the solver don't need to include all tests involving things they don't support.
The set of things a solver supports is fixed while what it does not supports may grow with the growth of MOI.
So at the beginning of each test, a test will do
```julia
function test_...(model, config, force=false)
    if !(MOI.supports(...) && MOI.supports_constraint(...))
        @test !force
        return
    end
end
```
So by default, calling `runtests` with no exclude should work for all solvers and the excluded tests will actually be bugs in the MOI wrappers or the solvers.
If a solver want to check that `supports` is implemented correctly, it can add `force = [...]` with the name of some tests so that these tests will be called with `force=true`.
To make writing the tests easier and make the tests more readable, we can create a macro like `@require force MOI.supports_constraint(...)` which is rewritten into
```julia
if force
   @test MOI.supports_constraint(...)
elseif !MOI.supports_constraint(...)
    return
end
```

odow:
:+1: to this. I'm not planning on merging this before I update some of the solvers to check logistics.

I envisaged something like:
```Julia
if !MOI.supports_constraint(model, F, S)
    if force
        @warn("Skipping test xxx because you don't support F-in-S")
    end
    return
end
```

My plans is 
 - first to convert all of the existing tests.
- Then to bring some naming structure.
- And then to add these skips.

docs/src/submodules/Test/overview.md
```diff
@@ -57,196 +55,184 @@ const CONFIG = MOI.Test.Config(
...
 
-function test_contconic()
-    MOI.Test.contlineartest(BRIDGED, CONFIG)
-end
+end # module TestFooBar
 
-function test_intconic()
-    MOI.Test.intconictest(BRIDGED, CONFIG)
-end
+# This line at tne end of the file runs all the tests!
```
joaquimg, line 105: tne -> the

joaquimg:
This is very breaking.
Should we do it in two steps since MOI 0.10 will already break many methods?
So 0.10 breaks all methods and 0.11 only update tests. They could even be released in the same day (?).
Do both at once might be a bit cumbersome for wrappers. Developer will have to figure if they are supposed or not to pass a test that was renamed and at the same time fix many changes.

odow:
Closing for now.

I have a plan for progress that doesn't involve this multi-thousand line diff that breaks every existing solver :)

Issue https://github.com/jump-dev/MathOptInterface.jl/issues/1398 is to track progress.

PR #1407, opened 2021-06-24
[breaking] Deprecate the MOI.Test submodule

odow:
This is the first step in a series of pull requests to rewrite our
existing testing infrastructure. Rather than rewriting it in-place,
we're starting afresh to simplify the transition process for solvers.

Issue https://github.com/jump-dev/MathOptInterface.jl/issues/1398 is going to track my progress.


odow:
Gotta love debugging a 50,000 line segfault! 
https://github.com/jump-dev/MathOptInterface.jl/pull/1407/checks?check_run_id=2909086923#step:6:286

odow:
Discussed with @blegat offline.

PR #1408, opened 2021-06-24
Add the skeleton of the new MOI.Test submodule

odow:
Adding the skeleton for now, base against https://github.com/jump-dev/MathOptInterface.jl/pull/1407 until it is merged.

Part of https://github.com/jump-dev/MathOptInterface.jl/issues/1398


odow:
Merging this for now. We can make additional changes to `Config` in future PRs.

PR #1412, opened 2021-06-26
Throw error when adding constraint to MatrixOfConstraints

blegat:
Who knows whether this will be supported in the future but not today

### DIFF
--- a/src/Utilities/matrix_of_constraints.jl
+++ b/src/Utilities/matrix_of_constraints.jl
@@ -316,7 +316,7 @@ function _add_set(sets, i, func::MOI.AbstractVectorFunction)
     return add_set(sets, i, MOI.output_dimension(func))
 end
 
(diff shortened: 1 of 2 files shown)


src/Utilities/matrix_of_constraints.jl
```diff
@@ -316,6 +316,18 @@ function _add_set(sets, i, func::MOI.AbstractVectorFunction)
     return add_set(sets, i, MOI.output_dimension(func))
 end
 
+const _MODIF_NOT_ALLOWED = """
```
odow, line 319: I've been trying to remove all abbreviations from the source. Can we make this `_MATRIXOFCONSTRAINTS_MODIFY_NOT_ALLOWED_ERROR_MESSAGE`

odow (approved):

PR #1435, opened 2021-07-03
Fix Bridges.Constraint.ScalarizeBridge with unset starting values

blegat:
I like these new tests.

Closes https://github.com/jump-dev/MathOptInterface.jl/issues/1433

### DIFF
--- a/src/Bridges/Constraint/scalarize.jl
+++ b/src/Bridges/Constraint/scalarize.jl
@@ -128,10 +128,32 @@ end
 
 function MOI.get(
     model::MOI.ModelLike,
-    attr::Union{MOI.ConstraintPrimal,MOI.ConstraintPrimalStart},
+    attr::MOI.ConstraintPrimal,
     bridge::ScalarizeBridge,
 )
     return MOI.get.(model, attr, bridge.scalar_constraints) .+ bridge.constants
+end
+
+function MOI.get(
+    model::MOI.ModelLike,
+    attr::MOI.ConstraintPrimalStart,
+    bridge::ScalarizeBridge,
+)
+    values = MOI.get.(model, attr, bridge.scalar_constraints)
+    any(value -> value === nothing, values) && return nothing
+    return values .+ bridge.constants
+end
+
+function MOI.set(
+    model::MOI.ModelLike,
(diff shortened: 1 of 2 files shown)


src/Bridges/Constraint/scalarize.jl
```diff
@@ -128,12 +128,34 @@ end
...
     return MOI.get.(model, attr, bridge.scalar_constraints) .+ bridge.constants
 end
 
+function MOI.get(
+    model::MOI.ModelLike,
+    attr::MOI.ConstraintPrimalStart,
+    bridge::ScalarizeBridge,
+)
+    values = MOI.get.(model, attr, bridge.scalar_constraints)
+    any(value -> value === nothing, values) && return nothing
```
odow, line 143: This short-circuit returns make code coverage less accurate. Prefer:
```Julia
if any(value -> value === nothing, values)
    return
end

File: src/Bridges/Constraint/scalarize.jl [hunk 2]
```diff
@@ -147,12 +169,22 @@ end
...
     return MOI.get.(model, attr, bridge.scalar_constraints)
 end
 
+function MOI.get(
+    model::MOI.ModelLike,
+    attr::MOI.ConstraintDualStart,
+    bridge::ScalarizeBridge,
+)
+    values = MOI.get.(model, attr, bridge.scalar_constraints)
+    any(value -> value === nothing, values) && return nothing
```
odow, line 184: Ditto here

test/Bridges/Constraint/scalarize.jl
```diff
@@ -198,4 +198,15 @@ config = MOIT.Config()
...
+
+    @testset "constraint_ConstraintPrimalStart" begin
+        MOI.Test.test_constraint_ConstraintPrimalStart(
+            bridged_mock,
+            MOI.Test.Config(),
+        )
+        MOI.Test.test_constraint_ConstraintDualStart(
+            bridged_mock,
+            MOI.Test.Config(),
+        )
```
odow, line 210: Yeah my next step here is to migrate the Utilities and Bridges modules to the new test sets.

odow (approved):

Issue #1398, opened 2021-06-19
Rewrite MOI.Test


blegat:
<s>The test names do not follow the style guide, should we rename `contlineartest` to `continuous_linear_test` ?</s>
_Edit: @odow: hijacking this issue to lay out my plan for the new MOI.Test_

## Purpose

The purpose of this issue is to track the items that need to be completed in order to rewrite our MOI.Test submodule. 

Significant progress and design was started in https://github.com/jump-dev/MathOptInterface.jl/pull/1404. But rather than continue that, I'm going to break it into smaller pieces.

## Where we are 

Our current testing regime is comprehensive, but a bit all over the place.

There's a mix of things like `MOI.Test.unittest` that wrap a whole lot of tests, and others like `MOI.Test.default_status_test` that you just need to add. Even the documentation for how to test a solver is complicated (https://github.com/jump-dev/MathOptInterface.jl/issues/224)! https://jump.dev/MathOptInterface.jl/dev/submodules/Test/overview/#How-to-test-a-solver

The current design is also bad because it's hard to add new tests. 

As evidenced by the documentation: https://jump.dev/MathOptInterface.jl/dev/submodules/Test/overview/#How-to-add-a-test, you need to write a test, then write a test for the test, and then make sure that everything works. It's hard to run a small contingent of tests, and it's hard to decide where to put a new test, and where to put the test of the test.

This also evidenced by the large number of open test issues that aren't getting addressed. These range from #470 (August 2018!) to #1201 (November 2020). If they were easier to add, it would happen quicker!

Naming and visibility of each test is also a problem: https://github.com/jump-dev/MathOptInterface.jl/issues/1029. If `linear10btest` fails, what does that mean? 

## New design

The basic design is `runtests`, a single entry point to _all_ tests in MOI.

Instead of breaking down tests by files or dictionaries, tests are normal Julia functions with descriptive names that can be excluded or included by the user.

Here's the code to test the `MockOptimizer`:
```Julia
MOI.Test.runtests(
    MOIU.MockOptimizer(MOIU.UniversalFallback(MOIU.Model{Float64}())),
    MOI.Test.Config(basis = true),
    exclude = ["test_nonlinear_"],
)
```
Much better. 

There is also a need for certain tests to modify the model prior to running the test (changing solver parameters/tolerances, for example). That can be achieved by overloading `setup_test(::typeof(f), ::MOI.ModelLike, ::MOI.Test.Config)` for the particular function `f`.

[...]

PR #1438, opened 2021-07-04
[Test] Updates to Test/test_model.jl

odow:
Part of #1398


odow:
Merging for now. Codecov will get hit by these when I migrate the universal fallback tests.

PR #1441, opened 2021-07-05
Update the tests of UniversalFallback to new MOI.Test

odow:
Part of #1398


odow:
Merging because codecov reduction is in the DeprecatedTest module.

PR #1442, opened 2021-07-05
[Test] migrate Utilities/model.jl to new MOI.Test

odow:
Part of #1398


odow:
Merging because coverage decrease is in `DeprecatedTest` (which shows how mixed up our testing was!).

PR #1443, opened 2021-07-05
[Utilities] Migrate Utilities/cachingoptimizer to new MOI.Test and fix bugs

odow:
Part of #1398 

@blegat I didn't expect to find two bugs in CachingOptimizer! These new tests are great.

### DIFF
--- a/src/Utilities/cachingoptimizer.jl
+++ b/src/Utilities/cachingoptimizer.jl
@@ -545,14 +545,15 @@ end
 # ConstraintSet and ConstraintFunction methods, but allows us to strongly type
 # the third and fourth arguments of the set methods so that we only support
 # setting the same type of set or function.
-function replace_constraint_function_or_set(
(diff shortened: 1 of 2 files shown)


src/Utilities/cachingoptimizer.jl
```diff
@@ -587,25 +588,34 @@ function MOI.set(
...
-    m::CachingOptimizer,
+    model::CachingOptimizer,
     ::MOI.ConstraintFunction,
-    cindex::CI{F},
+    cindex::MOI.ConstraintIndex{F,S},
     func::F,
-) where {F}
-    replace_constraint_function_or_set(
-        m,
+) where {F,S}
```
blegat, line 600: What's the point of having `{F,S}` here ?

blegat:
Wow, it's surprising these have not been reported earlier. These new tests will provide us brighter days ahead

PR #1446, opened 2021-07-05
Refactor DoubleDicts.jl

### LINKED ISSUE #1230: Missing docstring for lazy_get
https://github.com/jump-dev/MathOptInterface.jl/blob/8d8beb387551d45138857bd472328718ba98ca8e/src/Utilities/DoubleDicts.jl#L176

odow:
I took a look through DoubleDicts for the first time in a long time (ever?).

Why are there so many types and implementations? This is very complicated (600 lines!) for something that should be simple.

From what I can tell, the only type in DoubleDicts that is used is
https://github.com/jump-dev/MathOptInterface.jl/blob/667ad660b7766c526c7526eacbf256d0b9ec9104/src/Utilities/copy/index_map.jl#L8
With this constructor
https://github.com/jump-dev/MathOptInterface.jl/blob/667ad660b7766c526c7526eacbf256d0b9ec9104/src/Utilities/copy/index_map.jl#L24
And this related type which is used to access the inner dictionaries
https://github.com/jump-dev/MathOptInterface.jl/blob/667ad660b7766c526c7526eacbf256d0b9ec9104/src/Utilities/copy/index_map.jl#L49
called from here
https://github.com/jump-dev/MathOptInterface.jl/blob/667ad660b7766c526c7526eacbf256d0b9ec9104/src/Utilities/copy/index_map.jl#L70

Does that mean I can remove anything unused? Are other packages using this somewhere? If so, how?

Closes #1230

### DIFF
--- a/docs/src/submodules/Utilities/overview.md
+++ b/docs/src/submodules/Utilities/overview.md
@@ -495,13 +495,14 @@ end
 ## DoubleDicts
 
 When writing MOI interfaces, we often need to handle situations in which we map
-[`MOI.ConstraintIndex`](@ref)s to different values. For example, to a string
-for [`MOI.ConstraintName`](@ref).
+[`ConstraintIndex`](@ref)s to different values. For example, to a string
+for [`ConstraintName`](@ref).
 
 One option is to use a dictionary like `Dict{MOI.ConstraintIndex,String}`.
 However, this incurs a performance cost because the key is not a concrete type.
 
 The DoubleDicts submodule helps this situation by providing two types main
-types [`DoubleDict`](@ref) and [`IndexDoubleDict`](@ref). These types act like
-normal dictionaries, but internally they use more efficient dictionaries
-specialized to the type of the function-set pair.
+types [`Utilities.DoubleDicts.DoubleDict`](@ref) and
+[`Utilities.DoubleDicts.IndexDoubleDict`](@ref). These types act like normal
+dictionaries, but internally they use more efficient dictionaries specialized to
+the type of the function-set pair.
--- a/docs/src/submodules/Utilities/overview.md
+++ b/docs/src/submodules/Utilities/overview.md
@@ -491,3 +491,17 @@ function MOI.get(model::Optimizer, attr::MOI.ObjectiveFunction)
     return MOI.Utilities.get_fallback(model, attr)
 end
 ```
+
+## DoubleDicts
+
+When writing MOI interfaces, we often need to handle situations in which we map
+[`MOI.ConstraintIndex`](@ref)s to different values. For example, to a string
+for [`MOI.ConstraintName`](@ref).
+
+One option is to use a dictionary like `Dict{MOI.ConstraintIndex,String}`.
+However, this incurs a performance cost because the key is not a concrete type.
+
+The DoubleDicts submodule helps this situation by providing two types main
+types [`DoubleDict`](@ref) and [`IndexDoubleDict`](@ref). These types act like
+normal dictionaries, but internally they use more efficient dictionaries
+specialized to the type of the function-set pair.
--- a/docs/src/submodules/Utilities/overview.md
+++ b/docs/src/submodules/Utilities/overview.md
@@ -506,3 +506,38 @@ types [`Utilities.DoubleDicts.DoubleDict`](@ref) and
 [`Utilities.DoubleDicts.IndexDoubleDict`](@ref). These types act like normal
 dictionaries, but internally they use more efficient dictionaries specialized to
 the type of the function-set pair.
+
+The most common usage of a `DoubleDict` is in the `index_map` returned by
+[`copy_to`](@ref). Performance can be improved, by using a function barrier.
+That is, instead of code like:
+```julia
+index_map = MOI.copy_to(dest, src)
+for (F, S) in MOI.get(src, MOI.ListOfConstraintTypesPresent())
+    for ci in MOI.get(src, MOI.ListOfConstraintIndices{F,S}())
+        dest_ci = index_map[ci]
+        # ...
+    end
+end
+```
+use instead:
+```julia
+function function_barrier(
+    dest,
+    src,
+    index_map,
+    ::Type{F},
+    ::Type{S},
+) where {F,S}
+    type_stable_map = index_map[F, S]
+    for ci in MOI.get(src, MOI.ListOfConstraintIndices{F,S}())
+        dest_ci = type_stable_map[ci]
+        # ...
+    end
+    return
+end
+
+index_map = MOI.copy_to(dest, src)
+for (F, S) in MOI.get(src, MOI.ListOfConstraintTypesPresent())
+    function_barrier(dest, src, index_map, F, S)
+end
+```
--- a/docs/src/submodules/Utilities/overview.md
+++ b/docs/src/submodules/Utilities/overview.md
@@ -524,13 +524,10 @@ use instead:
 function function_barrier(
     dest,
     src,
-    index_map,
-    ::Type{F},
-    ::Type{S},
+    index_map::MOI.Utilities.DoubleDicts.IndexDoubleDictInner{F,S},
 ) where {F,S}
-    type_stable_map = index_map[F, S]
     for ci in MOI.get(src, MOI.ListOfConstraintIndices{F,S}())
-        dest_ci = type_stable_map[ci]
+        dest_ci = index_map[ci]
         # ...
     end
     return
@@ -538,6 +535,6 @@ end
 
 index_map = MOI.copy_to(dest, src)
 for (F, S) in MOI.get(src, MOI.ListOfConstraintTypesPresent())
-    function_barrier(dest, src, index_map, F, S)
+    function_barrier(dest, src, index_map[F, S])
 end
 ```
--- a/src/Utilities/DoubleDicts.jl
+++ b/src/Utilities/DoubleDicts.jl
@@ -16,9 +16,7 @@ Convert the `value` stored inside `dict` to the equivalent on the outer
 `DoubleDict`. This is useful when the value type `V` of the inner dict is
 different to the outer dict. (See, e.g., [`IndexDoubleDict`](@ref).)
 """
-function typed_value(::AbstractDoubleDictInner{F,S,V}, value::V) where {F,S,V}
-    return value
-end
+typed_value(::AbstractDoubleDictInner, value) = value
 
 """
     DoubleDict{V}
@@ -164,10 +162,7 @@ function Base.get(
     key::MOI.ConstraintIndex{F,S},
     default,
 ) where {F,S}
-    if !haskey(d, key)
-        return default
-    end
-    return typed_value(d, d.inner[key.value])
+    return typed_value(d, get(d.inner, key.value, default))
 end
 
 # Base.getindex
@@ -184,10 +179,10 @@ function Base.getindex(
     d::AbstractDoubleDictInner{F,S},
     key::MOI.ConstraintIndex{F,S},
 ) where {F,S}
-    if !haskey(d, key)
+    x = get(d.inner, key.value) do
         throw(KeyError(key))
     end
-    return typed_value(d, d.inner[key.value])
+    return typed_value(d, x)
 end
 
 # Base.setindex!
--- a/src/Utilities/DoubleDicts.jl
+++ b/src/Utilities/DoubleDicts.jl
@@ -9,6 +9,8 @@ abstract type AbstractDoubleDict{V} <: AbstractDict{MOI.ConstraintIndex,V} end
 abstract type AbstractDoubleDictInner{F,S,V} <:
               AbstractDict{MOI.ConstraintIndex{F,S},V} end
 
+_inner(d::AbstractDoubleDictInner) = d.inner
+
 """
     DoubleDict{V}
 
@@ -42,21 +44,10 @@ A type stable inner dictionary of [`DoubleDict`](@ref).
 """
 mutable struct DoubleDictInner{F,S,V} <: AbstractDoubleDictInner{F,S,V}
     dict::DoubleDict{V}
-    inner::Union{Dict{Int64,V},Nothing}
+    inner::Dict{Int64,V}
     function DoubleDictInner{F,S}(d::DoubleDict{V}) where {F,S,V}
-        return new{F,S,V}(d, get(d.dict, (F, S), nothing))
+        return new{F,S,V}(d, get!(d.dict, (F, S), Dict{Int64,V}()))
     end
-end
-
-function _inner(
-    d::DoubleDictInner{F,S,V};
-    initialize::Bool = false,
-) where {F,S,V}
-    if initialize && d.inner === nothing
-        d.inner = Dict{Int64,V}()
-        d.dict.dict[(F, S)] = d.inner
-    end
-    return d.inner::Dict{Int64,V}
 end
 
 """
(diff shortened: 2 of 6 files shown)


blegat:
> Does that mean I can remove anything unused? Are other packages using this somewhere? If so, how?

No there are plan to us it in more places than the IndexMap, e.g. UniversalFallback, DiffOpt but it hasn't been done yet.
MOI brings a generic and extensible interface but supporting any function and set types should not mean that all the code become unstable. We should educate the user to store things in DoubleDicts, use function barriers taking type-stable versions of the double dicts through the barrier.
The current DoubleDicts cover many possible use cases but I've encountered cases that are not covered, e.g., where what you store depends on F and S but is not exactly Tuple{F,S}

odow:
> but it hasn't been done yet

My problem with this is that we have a lot of code in MOI that is sitting there un-used on the off-chance that it might be useful in the future. We should tear it out prior to 1.0. We can always add it back later...

odow:
Okay. This needs a pretty thorough review by @joaquimg and @blegat.

The changes are actually less scary than the diff seems. It may be best to review in side-by-side mode. The interleaved diff isn't very helpful.

- Removed `FunctionSetDoubleDict`. If we need this later, it can get added back then.
- Massively simplify the type signatures of `AbstractDoubleDict` and `AbstractWIthType` (now `AbstractDoubleDictInner`)
- Rename `WithType` to `DoubleDictInner`
- Rename `IndexWithType` to `IndexDoubleDictInner`

odow:
Bump @blegat

joaquimg:
Will read tomorrow

src/Utilities/DoubleDicts.jl
```diff
@@ -3,116 +3,100 @@ module DoubleDicts
...
+When looping through multiple keys of the same Function-in-Set type, use
+```julia
+inner = dict[F, S]
+```
+to return a type-stable [`DoubleDictInner`](@ref).
 """
-struct DoubleDict{V,DI,DO} <: AbstractDoubleDict{DataTypePair,V,Int64,DI,DO}
-    dict::DO
+struct DoubleDict{V} <: AbstractDoubleDict{V}
+    dict::Dict{Tuple{DataType,DataType},Dict{Int64,V}}
```
joaquimg, line 43: This way we wont be able to use clever dicts inside double dicts.
Which might be ok for now. But if variables need that for performance, chances are that constraints will need to.

joaquimg: Same thing on the keys side... if we want to generalize that later we might be in trouble.
I completely understand the point of cleaning up and adding thing back later. Wouldnt this way break if we bring back functionality?
If it wont break, it is less of an issue.
Other option is to make the fully parametric dd as a underscore type. While the Most used IndexDD is no undersocored.

src/Utilities/DoubleDicts.jl
```diff
@@ -132,445 +122,249 @@ function Base.length(d::AbstractDoubleDict)
...
+# Base.get
+
+function Base.get(
+    d::AbstractDoubleDict,
+    key::MOI.ConstraintIndex{F,S},
+    default,
+) where {F,S}
+    inner = getindex(d, F, S)
+    if haskey(inner, key)
+        return inner[key]
```
joaquimg, line 152: Cant we avoid haskey+getindex?
This double access pattern always comes back to us with penalties...

src/Utilities/DoubleDicts.jl
```diff
@@ -132,445 +122,249 @@ function Base.length(d::AbstractDoubleDict)
...
+    return out
 end
 
-function Base.haskey(d::AbstractWithType{F,S}, key::CI{F,S}) where {F,S}
-    if inner_is_empty(d)
-        return false
-    end
-    return haskey(inner(d), key.value)
+function Base.values(d::DoubleDictInner{F,S,V})::Vector{V} where {F,S,V}
+    return V[v for v in values(d.inner)]
```
joaquimg, line 274: Why not just values?

joaquimg (commented):
The use case for ParametricOptInterface would be ci as keys and arrays of Linear/QuadraricTerms as values.

DiffOpt already uses CIs as keys and functions (SAF) as values.

I am planning to test DDs in xpress in the short term. CleverDicts inside double dicta might be nice to consider.

joaquimg:
Here is one usage of DoubleDicts in DiffOpt:
https://github.com/jump-dev/DiffOpt.jl/blob/1f289b0e7e1a735bfb76f795bbc3ddfeb31da1ce/src/diff_opt.jl#L107

Here we are not using in POI, but we should be:
https://github.com/jump-dev/ParametricOptInterface.jl/blob/b1629f9f93c299e170f9bbf606f6e3e699ebdeba/src/ParametricOptInterface.jl#L33

I should use it here also:
https://github.com/joaquimg/QuadraticToBinary.jl/blob/462e241de47ff40724c056a193cfba2d0fbd07fb/src/moi.jl#L118

src/Utilities/DoubleDicts.jl
odow: I'll consider making the dict a type parameter if there is a clear real-world benchmark showing that this is a bottleneck. 

If we need some uber-DoubleDict-CleverDict, we can always add that in future as a new subtype of `AbstractDoubleDict`. 

For now, DoubleDicts solves a clear use-case: a ConstraintIndex->ConstaintIndex map in IndexMap. Let's keep that simplicity.

src/Utilities/DoubleDicts.jl
odow: I've re-organized things, but in this case we do need it.

If we used `get(d.inner, key.value, default)`, we'd then try to `typed_value` map the result back, which wouldn't be correct if the default was returned. And we wouldn't know whether the default was returned because the key was missing, or if the value was the same as the default.

odow:
These use-cases are all fine for `DoubleDict`. I don't think anything needs changing.

joaquimg:
As long as values have a single type they fall into the simpler version of DoubleDict.

The odd one is DiffOpt, which has two DD's, one for scalar and another for vector. It is not clear to me if they should be merged.

src/Utilities/DoubleDicts.jl
```diff
@@ -3,116 +3,108 @@ module DoubleDicts
 import MathOptInterface
 
 const MOI = MathOptInterface
-const MOIU = MOI.Utilities
-const CI = MOI.ConstraintIndex
-const AD = AbstractDict
 
-const DataTypePair = Tuple{DataType,DataType}
+abstract type AbstractDoubleDict{V} <: AbstractDict{MOI.ConstraintIndex,V} end
```
joaquimg, line 7: If we end up with this simple implementation...
We should rename this to ConstraintIndexedDict and reserve DoubleDict for future flexible implementation. Or ConstraintIndexDict or something along those lines.

DoubleDictInner could be SingleTypeConstraintIndexedDict.

IndexDoubleDict would become ConstraintIndexMap...

src/Utilities/DoubleDicts.jl
```diff
@@ -147,11 +155,19 @@ function Base.get(
...
+    inner = d[F, S]
+    return get(inner, key, default)
+end
+
+function Base.get(
+    d::AbstractDoubleDictInner,
+    key::MOI.ConstraintIndex{F,S},
+    default,
+) where {F,S}
+    if !haskey(d, key)
```
joaquimg, line 167: why do we need haskey? cant you do it with get?

AbstractDoubleDictInner already has parameters F and S, they should be in the function signature.
You can have a second method for non-matching F,S

src/Utilities/DoubleDicts.jl
```diff
@@ -160,42 +176,34 @@ function Base.getindex(
...
-function Base.getindex(d::DoubleDict, ::Type{F}, ::Type{S}) where {F,S}
-    return DoubleDictInner{F,S}(d)
-end
-
 function Base.getindex(
     d::AbstractDoubleDictInner{F,S},
     key::MOI.ConstraintIndex{F,S},
 ) where {F,S}
-    if !haskey(d.inner, key.value)
+    if !haskey(d, key)
```
joaquimg, line 187: get? with a default to nothing?

joaquimg (commented):
We can get rid of a few haskey's

blegat:
Yes with DiffOpt, I could just split it in two but in general, it would be nice to have any value type parametrized by `F, S`. For DiffOpt, it would be F. The user could give a function mapping F, S to the value. So for constraint indices, it is `(F, S)-> CI{F, S}` and for DiffOpt, it is `(F, S)-> F`.

src/Utilities/DoubleDicts.jl
odow: If we used
```julia
x = get(d.inner, key.value, default)
```
we wouldn't know whether to return `typed_value(x)` or `x`.

src/Utilities/DoubleDicts.jl
odow: The value might be `nothing`. And we need to throw `key` as the error, not `key.value`.

odow:
> For DiffOpt, it would be F

The right way to do this is for DiffOpt to implement this locally as a subtype of `AbstractDoubleDict`, show it has a performance benefit, debug, and them make a PR to MOI. 

I don't think we should add in this PR.

src/Utilities/DoubleDicts.jl
joaquimg: then
```julia
function Base.getindex(
    d::AbstractDoubleDictInner{F,S},
    key::MOI.ConstraintIndex{F,S},
) where {F,S}
    x = get(d.inner, key.value) do
        throw(KeyError(key))
    end
    return typed_value(d, x)
end
```

src/Utilities/DoubleDicts.jl
joaquimg: cant you define `typed_value` to be a no-op if the return type is correct?

src/Utilities/DoubleDicts.jl
odow: Oooo. I didn't consider this pattern! Nice.

odow:
Okay. I've addressed the `haskey` issue. But we still have our naming problem. I think it's fine to keep `DoubleDict` as the name. It's a submodule within the Utilities submodule, so it's obviously not for wider consumption. 

It solves the major issue, and provides a decent fallback for other cases. I'm in favor of merging as-is.

docs/src/submodules/Utilities/overview.md
```diff
@@ -491,3 +491,18 @@ function MOI.get(model::Optimizer, attr::MOI.ObjectiveFunction)
...
 ```
+
+## DoubleDicts
+
+When writing MOI interfaces, we often need to handle situations in which we map
+[`ConstraintIndex`](@ref)s to different values. For example, to a string
+for [`ConstraintName`](@ref).
+
+One option is to use a dictionary like `Dict{MOI.ConstraintIndex,String}`.
+However, this incurs a performance cost because the key is not a concrete type.
```
blegat, line 502: It's not clear what's the gain is here. The gain occurs only when you need to use the dictionary many times for the same set-function pair. Then DoubleDict is useful as `dict[F, S]` will give a dictionary with concrete key type. Bu this should be coupled with a function barrier.

src/Utilities/DoubleDicts.jl
```diff
@@ -3,116 +3,106 @@ module DoubleDicts
...
-`CI{MOI.SingleVariable, MOI.Integers}` from the dictionary `dict`, for instance:
+A specialized version of [`DoubleDict`] in which the values are of type
+`MOI.ConstraintIndex`
 
-    ``inner = dict[MOI.SingleVariable, MOI.Integers]``
+When looping through multiple keys of the same Function-in-Set type, use
+```julia
+inner = dict[F, S]
+```
+to return a type-stable [`IndexDoubleDictInner`](@ref).
```
blegat, line 74: Say that it should be used with a function barrier to get any performance benefit

src/Utilities/DoubleDicts.jl
```diff
@@ -3,116 +3,106 @@ module DoubleDicts
...
 """
-struct IndexDoubleDict{DI,DO} <: AbstractDoubleDict{DataTypePair,CI,Int64,DI,DO}
-    dict::DO
+struct IndexDoubleDict <: AbstractDoubleDict{MOI.ConstraintIndex}
+    dict::Dict{Tuple{DataType,DataType},Dict{Int64,Int64}}
```
blegat, line 77: It can be done in a separate PR but I think we should use a `CleverDict` here. With VectorOfConstraints and MatrixOfConstraints, the constraint indices are now 1:n for scalar constraints so using CleverDict here would speed up copy time.

src/Utilities/DoubleDicts.jl
```diff
@@ -3,116 +3,106 @@ module DoubleDicts
...
-            Dict{DataTypePair,Dict{Int64}}(),
-        )
+"""
+    IndexDoubleDictInner{F,S}
+
+A type stable inner dictionary of [`IndexDoubleDict`](@ref).
+"""
+mutable struct IndexDoubleDictInner{F,S} <:
+               AbstractDoubleDictInner{F,S,MOI.ConstraintIndex{F,S}}
+    dict::IndexDoubleDict
```
blegat, line 90: Why do we need to have the parent dictionary as a field ?

src/Utilities/DoubleDicts.jl
joaquimg: Yup, that is why it was previously parametrized by dict.
That is for the second Dict....
And.... we might consider LittleDict for the outer Dict.... (it is from OrderedCollections) leave it as comment?

src/Utilities/DoubleDicts.jl
joaquimg: Good point... +1 to remove it, we might get some stack allocation bonus without it

src/Utilities/DoubleDicts.jl
odow: > so using CleverDict here would speed up copy time for large LP's

Benchmark? I struggle to believe that the time looking up the index translation is significant. I'm against adding complexity, even if it is a minor speed improvement.

blegat: When doing the benchmarks for MatrixOfConstraints, I noticed creating the IndexMap was taking a significant time.

docs/src/submodules/Utilities/overview.md
```diff
@@ -491,3 +491,53 @@ function MOI.get(model::Optimizer, attr::MOI.ObjectiveFunction)
...
+for (F, S) in MOI.get(src, MOI.ListOfConstraintTypesPresent())
+    for ci in MOI.get(src, MOI.ListOfConstraintIndices{F,S}())
+        dest_ci = index_map[ci]
+        # ...
+    end
+end
+```
+use instead:
+```julia
+function function_barrier(
```
blegat, line 524: I find it simpler to do
```julia
function function_barrier(dest, src, index_map::Inner{F,S}) where {F,S}
end
index_map = MOI.copy_to(dest, src)
for (F, S) in MOI.get(src, MOI.ListOfConstraintTypesPresent())
    function_barrier(dest, src, index_map[F, S])
end
```
and you don't need to rely on the fact that `index_map[F, S]` is type stable which looks like a leap of faith in the compiler (even if I think it is indeed the case).

src/Utilities/DoubleDicts.jl
odow: In a proper model? Or in a microbenchmark?

src/Utilities/DoubleDicts.jl
```diff
@@ -3,116 +3,104 @@ module DoubleDicts
...
+When looping through multiple keys of the same Function-in-Set type, use
+```julia
+inner = dict[F, S]
+```
+to return a type-stable [`DoubleDictInner`](@ref).
 """
-struct DoubleDict{V,DI,DO} <: AbstractDoubleDict{DataTypePair,V,Int64,DI,DO}
-    dict::DO
+struct DoubleDict{V} <: AbstractDoubleDict{V}
+    dict::Dict{Tuple{DataType,DataType},Dict{Int64,V}}
```
odow, line 41: Add a function `_inner_type(F, S) = V` and have 
```suggestion
    dict::Dict{Tuple{DataType,DataType},Dict{Int64}}
```

src/Utilities/DoubleDicts.jl
```diff
@@ -3,116 +3,104 @@ module DoubleDicts
...
+
+When looping through multiple keys of the same Function-in-Set type, use
+```julia
+inner = dict[F, S]
+```
+to return a type-stable [`DoubleDictInner`](@ref).
 """
-struct DoubleDict{V,DI,DO} <: AbstractDoubleDict{DataTypePair,V,Int64,DI,DO}
-    dict::DO
+struct DoubleDict{V} <: AbstractDoubleDict{V}
```
odow, line 40: Potentially call it ConstraintIndexDict

odow:
@blegat shall I merge this for now, and we can address any other issues in a follow-up PR?

blegat (approved):

PR #1457, opened 2021-07-14
Migrate tests of matrix_of_constraints.jl

odow:
Part of #1398

Interested to see what this does to the code coverage. The previous tests in this file were pretty haphazard, and it's not obvious which tests were testing what. 

The individual components (ProductOfSets, MutableSparseMatrixCSC, etc.) are independently tested, so I don't see the need for wide-scale testing on lots of different types.

### DIFF
--- a/src/Utilities/matrix_of_constraints.jl
+++ b/src/Utilities/matrix_of_constraints.jl
@@ -450,6 +450,7 @@ function _load_constraints(
 end
 
 _add_variable(model::MatrixOfConstraints) = add_column(model.coefficients)
+
 function _add_variables(model::MatrixOfConstraints, n)
     return add_columns(model.coefficients, n)
 end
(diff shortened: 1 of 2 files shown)


odow:
That's quite a large coverage drop, so I'll hold off merging until that is resolved. This is a good example that refactoring the tests is useful! The files should be independently tested. We shouldn't have a big drop in coverage for sparse_matrix.jl by changing these tests.

Of course, it makes my earlier statement incorrect at present:

> The individual components (ProductOfSets, MutableSparseMatrixCSC, etc.) are independently tested, so I don't see the need for wide-scale testing on lots of different types.

😢 

image

odow:
This is now in a better state:
image
Coverage increases for this file. Large drop is due to `DeprecatedTests`. We can fix the other regressions in a different PR.

PR #1461, opened 2021-07-14
Fix StackOverflow with self-loop bridges

blegat:
This is breaking for a user relying on the fact that SingleBridgeOptimizer was creating the constraint to the same model recursively.
It would have been used in a non-stackoverflow situation in case a bridge would create a constraint of the same time as the constraint being transformed twice and then not. I would guess no user was doing that but we never know.

Closes https://github.com/jump-dev/MathOptInterface.jl/issues/1015

### DIFF
--- a/src/Bridges/Constraint/set_map.jl
+++ b/src/Bridges/Constraint/set_map.jl
@@ -98,7 +98,12 @@ function MOI.set(
     bridge::SetMapBridge{T,S2,S1,F,G},
     func::G,
 ) where {T,S2,S1,F,G}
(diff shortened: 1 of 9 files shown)


odow:
Potentially split the documentation updates and StackOverflow into two PRs?

blegat:
Yes, I move it to https://github.com/jump-dev/MathOptInterface.jl/pull/1465, I will also create a separate PR for the Variable.SetMap as it will be useful for creating the variable identity bridges used for testing

PR #1464, opened 2021-07-14
[Test] document how to debug a failing test

### LINKED ISSUE #224: Document MOIT in manual

odow:
Closes #224 

This may be okay for now. In future, we could look at adding every test docstring to the manual? That seems a lot, and I don't know how useful it would be.


blegat (approved):
Having a docstring for each test and saying in the doc that you can do `?test_name` is enough for now to fix that issue, I agree

PR #1467, opened 2021-07-15
[Utilities] updates to sparse_matrix.jl

odow:
Added by https://github.com/jump-dev/MathOptInterface.jl/pull/1387, but missing some documentation and the testing was a bit convoluted (which made #1457 look bad).


blegat:
I see so this breaks the lines to make coverage more accurate and then it adds a test to actually cover these lines, correct ?

odow:
> I see so this breaks the lines to make coverage more accurate and then it adds a test to actually cover these lines, correct ?

Yes. Single-line short-circuiting is just a recipe for bugs.

PR #1468, opened 2021-07-15
Refactor box.jl

odow:
This started off by adding some tests to `box.jl`, but I got confused. There's a muddle between `box.jl` and `model.jl`, and who owns what and does what isn't clear. Making this commit now, and then I'm going to try and encapsulate all of the mask stuff into box.jl.

Part of #1457

### DIFF
--- a/src/Utilities/box.jl
+++ b/src/Utilities/box.jl
@@ -1,23 +1,9 @@
-# Sets setting lower bound:
-extract_lower_bound(set::MOI.EqualTo) = set.value
-function extract_lower_bound(
-    set::Union{MOI.GreaterThan,MOI.Interval,MOI.Semicontinuous,MOI.Semiinteger},
-)
-    return set.lower
-end
-# 0xcb = 0x80 | 0x40 | 0x8 | 0x2 | 0x1
-const LOWER_BOUND_MASK = 0xcb
+"""
+    SUPPORTED_VARIABLE_SCALAR_SETS{T}
 
-# Sets setting upper bound:
-extract_upper_bound(set::MOI.EqualTo) = set.value
-function extract_upper_bound(
-    set::Union{MOI.LessThan,MOI.Interval,MOI.Semicontinuous,MOI.Semiinteger},
-)
-    return set.upper
-end
-# 0xcd = 0x80 | 0x40 | 0x8 | 0x4 | 0x1
-const UPPER_BOUND_MASK = 0xcd
-
+The union of scalar sets for `SingleVariable` constraints supported by
+`Utilities.Box` (and therefore in `Utilities.Model`).
+"""
 const SUPPORTED_VARIABLE_SCALAR_SETS{T} = Union{
     MOI.EqualTo{T},
     MOI.GreaterThan{T},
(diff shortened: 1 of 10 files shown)


blegat:
Yes, it's a bit sketchy. The issue is that it's quite performance critical for MOIU.Model as we built this data-structure so to fix the performance regression from JuMP v0.18 to v0.19 with SingleVariable constraint

odow:
I had a WIP that's coming along nicely. Just finishing up the tests.

src/Utilities/box.jl
```diff
@@ -39,72 +8,265 @@ Stores the constants of scalar constraints with the lower bound of the set in
 `lower` and the upper bound in `upper`.
 """
 struct Box{T}
+    set_mask::Vector{UInt8}
```
blegat, line 11: It seems weird that MatrixOfConstraints would have this `set_mask` because it is part the info normally set in the `set` field.
However, this `set_mask` is exactly `MixOfScalarSets`. If we were using `MixOfScalarSets` here (that we improve to use bitwise operations), it would allow `GenericModel` to be also generic on the `SingleVariable` constraints.
What would you think of having a VariableConstraints struct containing a Box and a MixOfScalarSets ?

odow: What's weird is that we have this `Box` type for MatrixOfConstraint, but we also used it as the variable constraint field in Model.

The abstractions already broken, so what's the harm in the extra `set_mask` field?

blegat: > The abstractions already broken

Why is it weird or broken ? A `Box` is a hyperrectangle, you can have both a vector of variables or a vector of affine constraints belong to a hyperrectangle, there is nothing weird about that

src/Utilities/vector_bounds.jl
```diff
@@ -0,0 +1,411 @@
...
+        push!(b.variable_indices, x)
+    end
+    return x
+end
+
+function MOI.get(b::SingleVariableConstraints, ::MOI.ListOfVariableIndices)
+    if b.variable_indices === nothing
+        return MOI.VariableIndex.(1:length(b.set_mask))
+    end
+    x = collect(b.variable_indices)
```
blegat, line 218: I'm wondering why we don't just put a flag in the mask for deleted variables. What's the point of this `Set` ?

src/Utilities/model.jl
```diff
@@ -526,18 +425,11 @@ function MOI.get(
...
     MOI.throw_if_not_valid(model, ci)
-    flag = single_variable_flag(S)
-    model.single_variable_mask[ci.value] &= ~flag
-    if !iszero(flag & LOWER_BOUND_MASK)
-        model.variable_bounds.lower[ci.value] = _no_lower_bound(T)
-    end
-    if !iszero(flag & UPPER_BOUND_MASK)
-        model.variable_bounds.upper[ci.value] = _no_upper_bound(T)
-    end
+    MOI.delete(model.variable_bounds, ci)
```
blegat, line 432: `MOI.throw_if_not_valid(model, ci)` should be moved to inside `MOI.delete`

src/Utilities/vector_bounds.jl
```diff
@@ -0,0 +1,411 @@
...
+    end
+    return list
+end
+
+###
+### MatrixBounds
+###
+
+"""
+    struct MatrixBounds{T} <: AbstractVectorBounds
```
blegat, line 362: This is confusing, it seems in the name that it is the bound on each entry of a matrix. I find it weird to name something with what is using it.

odow: Open to bike shed. Suggestions for a better name?

blegat: Not really :/

odow: Renamed to `Hyperrectangle`

blegat (approved):

blegat:
I like how simpler AbstractModel is

odow:
Yeah, less state to carry around, easier to test, etc.

PR #1469, opened 2021-07-15
Add support for Union on StructOfConstraints

blegat:
Needed by ECOS since it needs a `StructOfConstraints` with `MOI.Zeros` on one side and `Union{MOI.Nonnegatives,MOI.SecondOrderCone,MOI.ExponentialCone}` on the other side.

### DIFF
--- a/test/Utilities/matrix_of_constraints.jl
+++ b/test/Utilities/matrix_of_constraints.jl
@@ -526,6 +526,65 @@ function test_modif()
     @test_throws err MOI.add_constraint(model, func, set)
 end
 
+MOIU.@struct_of_constraints_by_set_types(
+    ZerosOrNot,
+    MOI.Zeros,
+    Union{MOI.Nonnegatives,MOI.Nonpositives},
+)
+
+function test_multicone()
+    T = Int
+    Indexing = MOIU.OneBasedIndexing
+    model = MOIU.GenericOptimizer{
+        T,
+        ZerosOrNot{T}{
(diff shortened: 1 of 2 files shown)


test/Utilities/matrix_of_constraints.jl
```diff
@@ -526,6 +526,65 @@ function test_modif()
     @test_throws err MOI.add_constraint(model, func, set)
 end
 
+MOIU.@struct_of_constraints_by_set_types(
+    ZerosOrNot,
+    MOI.Zeros,
+    Union{MOI.Nonnegatives,MOI.Nonpositives},
```
odow, line 532: Can you clarify why this is needed in more detail?!? Why not just two separate fields.

blegat: This allows to have one sparse matrix with the two instead of two separate sparse matrices. ECOS needs one sparse matrix with the three non-Zeros cones

odow: Sure, but you could concatenate the sparse matrices? It seems weird that we go to all this effort and then allow `Union`

blegat: I'd argue that this PR is simpler than implementing concatenation in the solver wrapper though. DiffOpt also needs to split the matrix onto equality and non-equality. This is quite common so having this feature is not just for ECOS

odow (approved):

Issue #1459, opened 2021-07-14
Consistent naming for sets


dourouc05:
No concrete set in MOI has the word `set` in it, with the exception of `IndicatorSet`. Moreover, all the abstract sets have the word `set` in their name. 

By the way, indicators are not covered in https://jump.dev/MathOptInterface.jl/dev/manual/constraints/, I think they should be. 

It would also be useful to document a naming guide for sets, as it's probably one of the first things that are extended in MOI. Currently, the best place looks like https://jump.dev/JuMP.jl/stable/developers/extensions/#Define-a-new-set, but I think it would rather belong to MOI's doc. 

- [ ] MOI Indicator instead of IndicatorSet: https://github.com/jump-dev/MathOptInterface.jl/pull/1475
- [ ] MOF Indicator instead of IndicatorSet: https://github.com/jump-dev/MathOptFormat/pull/18
- [x] MOI Zeros, Reals to ZeroCone, RealCone: https://github.com/jump-dev/MathOptInterface.jl/pull/1477
- [x] MOF Zeros, Reals to ZeroCone, RealCone: abandoned
- [ ] MOI Complements to something else
- [ ] MOF Complements to something else
- [ ] Naming conventions: https://github.com/jump-dev/MathOptInterface.jl/pull/1476
- [ ] Order of arguments

odow:
So you're suggestion it be renamed to `Indicator`? I'd support this.

> It would also be useful to document a naming guide for sets

:+1: to this. 

Some thoughts: 

* Conic sets should end in `Cone`
* Matrix-valued sets should provide both `XXXConeSquare` and `XXXConeTriangle` versions
* Follow established naming conventions if at all possible:
  - Conic cone from CVX
  - Constraint programming sets from MiniZinc
* Prefer singular `Integer` instead of `Integers`

There are a couple of other violations of these rules:
 - Nonnegatives -> NonnegativeCone
 - Nonpositives -> NonpositivesCone
 - Zeros -> ZeroCone
 - Reals -> RealCone

dourouc05:
Another note about sets: we should settle for an order of arguments, mostly where the dimension should be put (before or after the other arguments?). I don't have any preference for any order (and I have used both for CP sets, I reckon).

dourouc05:
[...]

PR #1475, opened 2021-07-18
Rename IndicatorSet to Indicator.

dourouc05:
This is more consistent with the naming of the other sets.

Follows #1459.

I also renamed through MOF, I don't know if this should happen (or maybe tag a new version of MOF JSON schema?). 

There is only one occurrence of `IndicatorSet` present, in `NEWS.md`.

### DIFF
--- a/src/FileFormats/MOF/mof.0.6.schema.json
+++ b/src/FileFormats/MOF/mof.0.6.schema.json
@@ -769,11 +769,11 @@
                 }
             }, {
                 "description": "If `activate_on=one`: (y, x) ∈ {0,1}×Rᴺ: y = 0 ⟹ x ∈ S, otherwise when `activate_on=zero`: (y, x) ∈ {0,1}×Rᴺ: y = 1 ⟹ x ∈ S.",
-                "examples": ["{\"type\": \"IndicatorSet\", \"set\": {\"type\": \"LessThan\", \"upper\": 2.0}, \"activate_on\": \"one\"}"],
+                "examples": ["{\"type\": \"Indicator\", \"set\": {\"type\": \"LessThan\", \"upper\": 2.0}, \"activate_on\": \"one\"}"],
                 "required": ["set", "activate_on"],
                 "properties": {
                     "type": {
-                        "const": "IndicatorSet"
+                        "const": "Indicator"
                     },
                     "set": {
                         "oneOf": [{
--- a/src/FileFormats/MOF/mof.0.6.schema.json
+++ b/src/FileFormats/MOF/mof.0.6.schema.json
@@ -769,11 +769,11 @@
                 }
             }, {
                 "description": "If `activate_on=one`: (y, x) ∈ {0,1}×Rᴺ: y = 0 ⟹ x ∈ S, otherwise when `activate_on=zero`: (y, x) ∈ {0,1}×Rᴺ: y = 1 ⟹ x ∈ S.",
-                "examples": ["{\"type\": \"Indicator\", \"set\": {\"type\": \"LessThan\", \"upper\": 2.0}, \"activate_on\": \"one\"}"],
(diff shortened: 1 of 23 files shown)


src/FileFormats/MOF/mof.0.6.schema.json
```diff
@@ -769,11 +769,11 @@
                 }
             }, {
                 "description": "If `activate_on=one`: (y, x) ∈ {0,1}×Rᴺ: y = 0 ⟹ x ∈ S, otherwise when `activate_on=zero`: (y, x) ∈ {0,1}×Rᴺ: y = 1 ⟹ x ∈ S.",
-                "examples": ["{\"type\": \"IndicatorSet\", \"set\": {\"type\": \"LessThan\", \"upper\": 2.0}, \"activate_on\": \"one\"}"],
+                "examples": ["{\"type\": \"Indicator\", \"set\": {\"type\": \"LessThan\", \"upper\": 2.0}, \"activate_on\": \"one\"}"],
                 "required": ["set", "activate_on"],
                 "properties": {
                     "type": {
-                        "const": "IndicatorSet"
+                        "const": "Indicator"
```
odow, line 776: Yeah, it's not that easy. Ideally we wouldn't do this because it's breaking. Leave this unchanged for now, and maybe open a PR at https://github.com/jump-dev/MathOptFormat?

src/FileFormats/MOF/write.jl
```diff
@@ -295,12 +295,12 @@ head_name(::Type{<:MOI.SOS1}) = "SOS1"
...
 
 function moi_to_object(
-    set::MOI.IndicatorSet{I,S},
+    set::MOI.Indicator{I,S},
     name_map::Dict{MOI.VariableIndex,String},
 ) where {I,S}
     @assert I == MOI.ACTIVATE_ON_ONE || I == MOI.ACTIVATE_ON_ZERO
     return OrderedObject(
-        "type" => "IndicatorSet",
+        "type" => "Indicator",
```
odow, line 303: Leave this un-changed for now

src/FileFormats/MOF/read.jl
```diff
@@ -469,13 +469,13 @@ function set_to_moi(::Val{:SOS2}, object::Object)
     return MOI.SOS2(Float64.(object["weights"]))
 end
 
-function set_to_moi(::Val{:IndicatorSet}, object::Object)
+function set_to_moi(::Val{:Indicator}, object::Object)
```
odow, line 472: leave this unchanged for now

dourouc05:
It makes sense to first do the MOF part, then import it into MOI. I've added a commit to revert those changes.

odow (commented):
Developer call says yes to this. Just need to fix the test failures.

dourouc05:
The tests now pass; there are still format checks that fail, but not for the altered part of the code.

odow:
> there are still format checks that fail, but not for the altered part of the code.

That's weird, it's passing on master: https://github.com/jump-dev/MathOptInterface.jl/runs/3147720574

dourouc05:
Do I need to do anything more in this PR? CI accepted a previous version of the patch, and there were only cosmetic changes in between.

odow:
I just cycled the CI. If it doesn't work, we probably just need to rebase the branch.

src/Utilities/model.jl
```diff
@@ -887,9 +887,9 @@ for (loop_name, loop_super_type) in [
 end
 
 const LessThanIndicatorSetOne{T} =
```
odow, line 889: A nit: did you search for "IndicatorSet" that didn't match the entire word? seems not? How many are there

dourouc05: Good find! I renamed the four occurrences there. There should be none remaining, except for MOF: 

image

PR #1477, opened 2021-07-18
Rename plural conic sets

dourouc05:
Follows #1459 and #1475.

As in #1475, I altered MOF files.


odow:
I'll add this one to the agenda for the monthly meeting. It's pretty widespread breakage.

dourouc05:
I also added two other issues to the Google Docs: it would be good to discuss also the naming conventions at the same time.

odow:
Consensus from today's call is that we don't want to do this. I'll add some comments in the naming convention PR.

PR #1480, opened 2021-07-20
Migrate tests and fix bugs in objective bridges

odow:
Part of #1398 

@blegat I found a bug in getting a slack objective bridge with a constant. Are you okay if I redo the tests a bit more systematically (e.g., like this PR) rather than picking a few MOI.Test functions to run?


blegat:
I think it's good to have both. We can keep the tests that are calling MOI.Test, are they difficult to convert from DeprecatedTest?

odow:
I guess it's not obvious to me why or how some of the tests were chosen, or what they are testing. I don't think we need to run the full battery of MOI.Test for every bridge? They already take quite a long time to run...

blegat:
> I guess it's not obvious to me why or how some of the tests were chosen, or what they are testing.

If you write a custom test for the bridge then you don't exclude the possibility that both the implementation and the tests are wrong and that cancels out so that the test pass. At least if you reuse an existing test, you exclude that possibility :)

> I don't think we need to run the full battery of MOI.Test for every bridge? They already take quite a long time to run...

No, we can just keep it the way it is. Run it with one or two tests as is currently done is good enough.

odow:
I added them back.

test/Bridges/Objective/slack.jl
```diff
@@ -201,6 +203,244 @@ function test_SlackBridge_ObjectiveFunctionValue()
...
+            [4 / 7, 3 / 7, 6 / 7, -2 * 13 / 7],
+            (MOI.ScalarAffineFunction{Float64}, MOI.GreaterThan{Float64}) =>
+                [10 / 7, 12 / 7],
+        ),
+    )
+    MOI.Test.test_quadratic_duplicate_terms(bridged_mock, MOI.Test.Config())
+    return
+end
+
+function test_()
```
blegat, line 418: Is the name of this function intentional ?

odow: Oops.

PR #1481, opened 2021-07-21
Migrate tests of Variable bridges

odow:
Part of #1398 

I haven't gone through the bridges in any detail. I just migrated the tests. I may take another pass after this PR is merged.


odow:
Coverage decrease is in DoubleDicts.jl, which will be addressed by https://github.com/jump-dev/MathOptInterface.jl/pull/1446

PR #1486, opened 2021-07-22
Migrate Bridge/Constraint tests to new Test

odow:
Part of #1398


odow:
Coverage decrease was due to DeprecatedTests

PR #1491, opened 2021-07-26
[breaking] remove index_value function

odow:
These functions were not called anywhere. We should remove unused things. People can just use `.value`.

### DIFF
--- a/src/indextypes.jl
+++ b/src/indextypes.jl
@@ -19,7 +19,6 @@ index.value == MOI.get(model, MOI.ConstraintFunction(), index).variable.value
 struct ConstraintIndex{F,S}
     value::Int64
 end
-index_value(ci::ConstraintIndex) = ci.value
 
 """
     VariableIndex
@@ -30,7 +29,6 @@ To allow for deletion, indices need not be consecutive.
 struct VariableIndex
     value::Int64
 end
-index_value(vi::VariableIndex) = vi.value
 
 # The default hash is slow. It's important for the performance of dictionaries
 # of VariableIndices to define our own.


blegat:
It's quite useful for solver wrappers to get the vector of index values from the vector of indices using broadcast.

odow:
If solvers need it they can define it themselves. There's also `getfield.(x, :value)`. When in doubt, I vote remove. We want to minimize the surface area of MOI if at all possible.

blegat:
It would be more readable if solvers use all a same documented way for doing so `getfield.(x, :value)` seems a bit hacky.

> We want to minimize the surface area of MOI if at all possible.

I agree but it does not seem that we would need to ever break this function.

OTOH, it's true that we may not want to have two ways to do the same thing so it all depends whether we would consider `getfield.(x, :value)` to be a recommended way or not.

odow:
It can't be that useful, because it isn't used anywhere in the whole of MOI. (Including in the tests...)

odow:
I have the majority of solvers `dev`'d to my machine

image

odow:
Bump.

odow:
There is only one instance of someone using this:
https://github.com/search?q=%22index_value%22+language%3Ajulia&type=code

and that's @Azzaare:
https://github.com/JuliaConstraints/CBLS.jl/blob/8a8764fbbc78ddbc121127acab4e42be602d31e7/src/variables.jl#L46

Azzaare:
I am on my phone atm so it will difficult to make the changes, but I don't mind updating the code to not use this function.

If one of you has a suggestion, I will update once I'm back home.

odow:
You never use broadcast. You only use `index_value(x)`. So it's just `x.value` instead. or you can define `index_value` locally.

I think it's safe to standardize the API on `.value`, so we shouldn't have two ways of doing this.

odow:
I found where it used to be used:
https://github.com/jump-dev/MathOptInterface.jl/blob/f559840765587598255e647dc144db8b74566d21/src/Utilities/copy.jl#L95
thanks to the JuliaHub search engine:
https://juliahub.com/ui/Search?q=index_value&type=code&w=true

Azzaare:
Since it is a breaking change, I suppose you will tag a new minor version.

In that case, you can merge this PR independently of CBLS.jl and I will make the change at the same I change the compat entry in the Project.toml

If for some reason the version will still be 0.9.x, let me know, and I will update before MOI.

odow:
> I suppose you will tag a new minor version.

Yes.  We're well into breaking change territory.  The next release will be 0.10 and require lots of changes.

PR #1492, opened 2021-07-26
[breaking] remove isapprox for SOS sets

odow:
This was added in https://github.com/jump-dev/MathOptInterface.jl/pull/711 because quote

> [it] was defined in Test for no apparent reason

Seems like we can safely remove. It isn't hit by code coverage, and we don't offer `isapprox` for other sets.


blegat (approved):
Yes, distance between sets is not well defined, distance from points to sets is being investigated in MathOptSetDistances and we don't yet want to have any interfaces for distances between sets

PR #1493, opened 2021-07-26
[breaking] remove fallback for supporting free variables

odow:
Introduced by https://github.com/jump-dev/MathOptInterface.jl/pull/1004, which seems far enough ago to remove. We don't want it in 1.0 anyway.


blegat (approved):
Thanks! I forgot this should be done for MOI v0.10 but it should definitely be done!

PR #1495, opened 2021-07-26
[breaking] Remove the Allocate-Load API

### LINKED ISSUE #1341: Deprecate Allocate-Load
The purpose of Allocate-Load is to be able to communicate the model in two-pass across layers without the need to create intermediate `CachingOptimizer`.
However, it's not yet implemented by the bridge optimizers and it's not clear to me whether there is a way to do this without complicating the job of creating new bridges for users.
This new features https://github.com/jump-dev/MathOptInterface.jl/pull/1286 allows to pass non-bridges constraints across layers which allows to avoid copying the function as the destination model of the copy can get the `CanonicalConstraintFunction` directly from the source model.
So a model needed Allocate-Load could implement `MOIU.pass_nonvariable_constraints` instead and it would work if no constraint if bridges.
To also work with bridging constraints, it should also implement `MOI.add_constraint` but only allow it to be called during the copy operation.
This is what is done for MatrixOfConstraints: https://github.com/jump-dev/MathOptInterface.jl/pull/1287.
In fact, I expect most solver needing Allocate-Load to be rewritten with MatrixOfConstraints like Clp in https://github.com/jump-dev/Clp.jl/pull/111.
The only two exceptions are CSDP and SDPA but there is no point in keeping the Allocate-Load API just for these two.
I'll keep this open until we're close to releasing v0.10 to evaluate if things happens as expected and then we can decide.

odow:
We should decide if we want to do this before merging, but just seeing what the diff would be.

Closes #1341


docs/src/submodules/Utilities/overview.md
```diff
@@ -318,76 +318,6 @@ $$ \begin{aligned}
     In IJulia, calling `print` or ending a cell with
     [`Utilities.latex_formulation`](@ref) will render the model in LaTex.
```
blegat, line 319: LaTex -> LaTeX

odow: https://github.com/jump-dev/MathOptInterface.jl/pull/1496

odow:
So yay/nay on this?

blegat (approved):
I'd say yay :)

src/Utilities/copy.jl
```diff
@@ -24,17 +24,14 @@ function automatic_copy_to(
     copy_names::Bool = true,
     filter_constraints::Union{Nothing,Function} = nothing,
 )
-    if MOI.supports_incremental_interface(dest, copy_names)
-        return default_copy_to(dest, src, copy_names, filter_constraints)
-    elseif supports_allocate_load(dest, copy_names)
-        return allocate_load(dest, src, copy_names, filter_constraints)
-    else
+    if !MOI.supports_incremental_interface(dest, copy_names)
```
blegat, line 27: One could argue whether we want to simplify this now and just having the fallback
```julia
copy_to(...) = default_copy_to(...)
```

odow: Yeah. Let's merge this first and then re-assess all the copy stuff.

PR #1499, opened 2021-07-27
[breaking] re-order fields of QuadraticFunction

### LINKED ISSUE #1406: Reorder fields of ScalarQuadraticFunction
I always felt having quadratic terms first and affine terms second would make more sense so that the degree of the terms is decreasing. Here, the degrees are 1, 2, 0 which is weird.
https://github.com/jump-dev/MathOptInterface.jl/blob/6bbdf82ec65381ea27f86b2e12de8c1204e23372/src/functions.jl#L172-L176

odow:
Closes #1406

@blegat pros and cons of merging this?


blegat (approved):
We could still consider having an affine function as a subfield of the quadratic but merging this won't prevent that so I don't see any cons.

PR #1501, opened 2021-07-28
[breaking] s/DataType/Type

odow:
Closes https://github.com/jump-dev/MathOptInterface.jl/issues/783

We should use `Type` instead of `DataType`.

The issue is that `DataType` is a concrete type of Type (confusing right) that causes the compiler problems (to quote Jeff "this sort of stuff drives the compiler crazy" and "if you're doing that, please stop").

```julia
julia> isconcretetype(DataType)
true

julia> subtypes(Type)
4-element Vector{Any}:
 Core.TypeofBottom
 DataType
 Union
 UnionAll
```

I tried a few other things, such as `Tuple{Type{<:MOI.AbstractFunction},Type{<:MOI.AbstractSet}}` but couldn't get them to work. It seems for a lot of these things, we want simpler type signatures with some run-time asserts to reduce our latency. These paths aren't hot and time-critical.

### DIFF
--- a/docs/src/reference/standard_form.md
+++ b/docs/src/reference/standard_form.md
@@ -32,8 +32,8 @@ VectorQuadraticFunction
 output_dimension
 constant(f::Union{ScalarAffineFunction, ScalarQuadraticFunction})
 constant(f::Union{VectorAffineFunction, VectorQuadraticFunction})
-constant(f::SingleVariable, ::DataType)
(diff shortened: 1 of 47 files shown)


blegat (approved):
Looks good to me. I'm wondering whether I tried `Tuple{Type,Type}` at the time but if it works and the compiler does not like `DataType` then let's use that.

PR #1505, opened 2021-08-03
[docs] Add tutorial on transitioning from MPB

### LINKED ISSUE #654: Documentation or tutorial on transitioning from MathProgBase
I am working on transitioning a fairly simple Linear Programming problem (```linprog(c, A, sense, b, l, u, solver)``` ) from MathProgBase to MathOptInterface.   I am finding it hard to find a good one to one example that could help with that.   
It would be great that if for a few of most common use case there was a tutorial that would show how people well versed in MathOptInterface would do it.

odow:
Closes #654

### DIFF
--- a/docs/src/tutorials/mathprogbase.md
+++ b/docs/src/tutorials/mathprogbase.md
@@ -1,0 +1,96 @@
+# Transitioning from MathProgBase
+
+MathOptInterface is a replacement for [MathProgBase.jl](https://github.com/JuliaOpt/MathProgBase.jl).
+However, it is not a direct replacement.
+
+## Transitioning a solver interface
+
+Writing a solver interface in MathOptInterface is more work than the equivalent
+interface in MathProgBase.
+
+For more information, read [Implementing a solver interface](@ref).
+
+## Transitioning the high-level functions
+
+MathOptInterface doesn't provide replacements for the high-level interfaces in
+MathProgBase. We recommend you use JuMP instead.
+
+!!! tip
+    If you haven't used JuMP before, start with the tutorial
+    [Getting started with JuMP](https://jump.dev/JuMP.jl/stable/tutorials/Getting%20started/getting_started_with_JuMP/)
+
+### linprog
+
+Here is one way of transitioning from `linprog`:
+
+```julia
+using JuMP
+
+function linprog(c, A, sense, b, l, u, solver)
+    N = length(c)
+    model = Model(solver)
+    @variable(model, l[i] <= x[i=1:N] <= u[i])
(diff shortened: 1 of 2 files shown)


docs/src/tutorials/mathprogbase.md
```diff
@@ -0,0 +1,96 @@
...
+
+Writing a solver interface in MathOptInterface is more work than the equivalent
+interface in MathProgBase.
+
+For more information, read [Implementing a solver interface](@ref).
+
+## Transitioning the high-level functions
+
+MathOptInterface doesn't provide replacements for the high-level interfaces in
+MathProgBase. We recommend you use JuMP instead.
```
dourouc05, line 16: Maybe "We recommend you use JuMP instead as a modelling layer on top of MOI"? Otherwise, it sounds a bit like you're telling people to use completely different packages.

docs/src/tutorials/mathprogbase.md
```diff
@@ -0,0 +1,96 @@
+# Transitioning from MathProgBase
+
+MathOptInterface is a replacement for [MathProgBase.jl](https://github.com/JuliaOpt/MathProgBase.jl).
+However, it is not a direct replacement.
```
dourouc05, line 4: Isn't https://github.com/jump-dev/MatrixOptInterface.jl providing a closer interface to what MPB looked like?

odow: For the modeling side, yes. But it still needs some development to integrate properly, so I held off mentioning it.

docs/src/tutorials/mathprogbase.md
```diff
@@ -0,0 +1,97 @@
...
+
+## Transitioning a solver interface
+
+Writing a solver interface in MathOptInterface is more work than the equivalent
```
blegat, line 8: I wouldn't say this. It may potentially be much closer to the solver. MPB was tailored to 3 specific format. If you didn't fit then it was way more work. CSDP MPB wrapper was so complicated I had to create SemidefiniteModels.jl. Now it's quite simple.

blegat: For conic solver, it will be much simpler than with MPB with MatrixOfConstraints.

odow: Make a PR with the suggested wording. I didn't think it was controversial that it was more complicated! We now have lots of constraint types and attributes and copy-to and caching and layers. It might be simpler for you, but it certainly isn't for most people.

blegat: It depends what you mean by simpler I guess :)

PR #1510, opened 2021-08-05
Nuance MPB vs MOI implementation

blegat:
Closes https://github.com/jump-dev/MathOptInterface.jl/issues/1508

### DIFF
--- a/docs/src/tutorials/mathprogbase.md
+++ b/docs/src/tutorials/mathprogbase.md
@@ -5,8 +5,10 @@ However, it is not a direct replacement.
 
 ## Transitioning a solver interface
 
-Writing a solver interface in MathOptInterface is more work than the equivalent
-interface in MathProgBase.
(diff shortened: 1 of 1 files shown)


odow:
I'm not sure I agree, but anyway...

blegat:
> I'm not sure I agree, but anyway...

I'm open to rephrase, for LPQP and NLP solvers, MPB was probably simpler. I'm not sure how to phrase it either. The previous phrasing needed to be nuanced though. It really depends on which kind of solvers. Solvers that fitted well in one of the three classes of MPB were simple in MPB but for others, MOI is simpler to implement because there is no bridging to do.
Even for conic solver, it seems more complicated than with MatrixOfConstraints: https://github.com/jump-dev/SCS.jl/blob/master/src/MPB_wrapper.jl#L135-L407
And with that implementation, it does not support rotated SOC.
MOI is maybe more complicated but it's not more work.

PR #1512, opened 2021-08-08
[breaking] remove automatic_copy_to

### LINKED ISSUE #1503: Default implementation for copy
https://github.com/jump-dev/MathOptInterface.jl/pull/1495#discussion_r678979361

odow:
Closes #1503

### DIFF
--- a/src/Utilities/copy.jl
+++ b/src/Utilities/copy.jl
@@ -562,14 +562,13 @@ end
         filter_constraints::Union{Nothing,Function} = nothing,
     )
 
-Implements `MOI.copy_to(dest, src)` by adding the variables and then the
-constraints and attributes incrementally. The function
-[`MathOptInterface.supports_incremental_interface`](@ref) can be used to check
-whether `dest` supports the copying a model incrementally.
+A default implementation of `MOI.copy_to(dest, src)` for models that implement
+the incremental interface, i.e., [`MOI.supports_incremental_interface`](@ref)
+returns `true`.
 
-If the `filter_constraints` arguments is given, only the constraints for which
-this function returns `true` will be copied. This function is given a
-constraint index as argument.
+If `filter_constraints` is a `Function`, only constraints for which
+`filter_constraints(ci)` returns `true` will be copied, where `ci` is the
+[`MOI.ConstraintIndex`](@ref) of the constraint.
 """
 function default_copy_to(
     dest::MOI.ModelLike,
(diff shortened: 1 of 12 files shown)


joaquimg:
docs could be a bit more direct:

```
... The function
[`MathOptInterface.supports_incremental_interface`](@ref) can be used to check
whether `dest` supports the copying a model incrementally.
```

```
... The function
[`MathOptInterface.supports_incremental_interface`](@ref) can be used to check
whether this function will work since it will check wether `dest` supports the copying of a model incrementally.
```

src/Utilities/copy.jl
```diff
@@ -486,14 +454,6 @@ function copy_free_variables(
     return
 end
 
-function default_copy_to(dest::MOI.ModelLike, src::MOI.ModelLike)
-    Base.depwarn(
-        "default_copy_to(dest, src) is deprecated, use default_copy_to(dest, src, true) instead or default_copy_to(dest, src, false) if you do not want to copy names.",
-        :default_copy_to,
-    )
-    return default_copy_to(dest, src, true)
-end
```
odow, line 495: It seems crazy, but this has been sitting here since MOI 0.2.0! https://github.com/jump-dev/MathOptInterface.jl/commit/1ae8661f8a0151013c4d92ea1aeb66568e361e1e

src/Utilities/copy.jl
```diff
@@ -575,29 +535,54 @@ only once all the model information is gathered.
...
 function default_copy_to(
     dest::MOI.ModelLike,
-    src::MOI.ModelLike,
-    copy_names::Bool,
+    src::MOI.ModelLike;
+    copy_names::Bool = true,
     filter_constraints::Union{Nothing,Function} = nothing,
 )
+    if !MOI.supports_incremental_interface(dest, copy_names)
+        error(
```
blegat, line 580: I'm wondering whether it makes sense to remove this error.
`MOI.supports_incremental_interface` is what is used by `instantiate`.
Would there be a case where you want instantiate to add a cache but you would still like to be able to call default_copy_to?
Removing the error is nonbreaking so it might be fine for now

odow: This check is trivial and will be in-lined. It also prevents a later error like "add variable not allowed" in some cases. I'd suggest we leave it in, at least until all the solvers are updated. We could remove it for 1.0 if it's not needed.

blegat: > This check is trivial and will be in-lined

Yes, I was not worried about peformance

> I'd suggest we leave it in, at least until all the solvers are updated. We could remove it for 1.0 if it's not needed.

Sounds good

blegat (approved):

PR #1519, opened 2021-08-10
Remove redundant optimize call

blegat:
`optimize!` is already called by `test_model_solution`. I was a bit confused when this test was causing ECOS to segfault because it's not obvious that it's calling `optimize!` twice.


odow:
You removed this from `DeprecatedTest`. Did you mean to change
https://github.com/jump-dev/MathOptInterface.jl/blob/592297973aa1ddf26079be33b5dbb536dfb4b6e9/src/Test/test_constraint.jl#L683
instead?

blegat:
We can change both. ECOS.jl is still using DeprecatedTest so the fact that it was changed there is intentional.

PR #1524, opened 2021-08-11
WIP: move var_to_name into SingleVariableBounds

odow:
Thoughts on this @blegat? I don't know if it makes sense or not, since every other implementation would then have to worry about names. 

If we don't do this, there are still stuff I should clean up, like the remaining single variable constraint names.

### DIFF
--- a/src/Utilities/model.jl
+++ b/src/Utilities/model.jl
@@ -69,51 +69,9 @@ function filter_variables(
     return g, t
 end
 
-function _delete_variable(
-    model::AbstractModel{T},
-    vi::MOI.VariableIndex,
-) where {T}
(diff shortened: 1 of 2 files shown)


blegat:
I'm hesitant as well. As the handling of the names can be decoupled, it's best to leave it decoupled so that we don't have to implement names in all implementations.
OTOH, the `variables` field could be layers, one layer handling the names and the inner layer handling the rest so that we don't have to duplicate name implementations.
If we go that path, maybe we should also allow the `variables` field to implement custom variable attributes like starting values ? It's maybe already the case if `GenericModel` is just forwarding the calls to `.variables`.

odow:
I'll leave it for now. I have some other plans to tidy things up.

PR #1525, opened 2021-08-11
Remove unneeded deletions

odow:


blegat (approved):
This is removed as variable constraints don't have names anymore: https://github.com/jump-dev/MathOptInterface.jl/pull/1312/

PR #1529, opened 2021-08-12
[Bridges] move lazy iterators to Bridges

### LINKED ISSUE #1528: Move lazy_iterators.jl to Bridges
I don't know what this is doing there, since it is only used by the bridges: https://github.com/jump-dev/MathOptInterface.jl/blob/master/src/Utilities/lazy_iterators.jl

Is it used anywhere else @blegat?

odow:
Closes #1528


blegat:
If it starts being used by another module we would need to move it again ?
I'd rather leave it in Utilities because it's well defined independently out of the concept of bridges.

odow:
Do we even gain anything by having `EmptyVector` and `LazyMap`? It's just more stuff to compile for minimal benefit.

odow:
I'll close this for now. But I may look at removing `EmptyVector` and `LazyMap`

PR #1530, opened 2021-08-12
Improve documentation of IndexMap

### LINKED ISSUE #1527: Should IndexMap be in MOI
`copy_to` doesn't define what type of index map should be returned. 

https://github.com/jump-dev/MathOptInterface.jl/blob/9e79136ce46e288cf0eeb4cc7fabeb895cb09cb2/src/MathOptInterface.jl#L126-L160

Should we let implementations return different types? Or should we standardize on `Utilities.IndexMap` and move it to MOI proper? 

I would vote moving to MOI proper, since this is a pretty key piece of the API, and it's effectively just this file: https://github.com/jump-dev/MathOptInterface.jl/blob/master/src/Utilities/copy/index_map.jl

odow:
Closes #1527 

Just a look at what this would look like.

### DIFF
--- a/src/MathOptInterface.jl
+++ b/src/MathOptInterface.jl
@@ -195,11 +195,8 @@ An optimizer can decide to implement this function instead of implementing
 !!! warning
     This is an experimental new feature of MOI v0.10 that may break in MOI v1.0.
 """
-function copy_to_and_optimize!(
-    dest::AbstractOptimizer,
-    src::ModelLike;
-    kwargs...,
-)
+function copy_to_and_optimize!(dest, src; kwargs...)
+    # The arguments above are untyped to avoid ambiguities.
     index_map = copy_to(dest, src; kwargs...)
     optimize!(dest)
     return index_map
--- a/src/MathOptInterface.jl
+++ b/src/MathOptInterface.jl
@@ -230,7 +230,21 @@ include("instantiate.jl")
 include("deprecate.jl")
 include("DeprecatedTest/DeprecatedTest.jl")
 
-# Work around a boostrapping issue. See the comment above Utilities.IndexMap.
+"""
+    IndexMap(number_of_variables::Int = 0)
+
+The dictionary-like object returned by [`copy_to`](@ref).
+
+If known in advance, pass `number_of_variables` to preallocate the necessary
+space for the variables.
+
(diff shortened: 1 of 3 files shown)


odow:
I guess if I'd tried this locally I would have found the problem.

We have a bootstrap issue: IndexMap needs Utilities.DoubleDicts,  but Utilities needs IndexMap. this was why it was in MOI.Utilities in the first place.

I may just add a `const IndexMap = Utilities.IndexMap` to the bottom of MOI.

src/MathOptInterface.jl
```diff
@@ -124,23 +124,41 @@ This is used in two places to determine whether to add a cache:
...
-then an [`UnsupportedAttribute`](@ref) error is thrown. Unsupported *optimizer*
-attributes are treated differently:
-
-* If `warn_attributes` is `true`, a warning is displayed, otherwise,
-* the attribute is silently ignored.
+    copy_to(
+        dest::ModelLike,
+        src::ModelLike;
+        copy_names::Bool = true,
+        warn_attributes::Bool = true,
```
odow, line 131: @blegat do we really use this `warn_attributes` thing? Perhaps we should just remove? We don't really copy optimizer attributes.

src/MathOptInterface.jl
```diff
@@ -124,23 +124,41 @@ This is used in two places to determine whether to add a cache:
...
-or constraint attribute that is copied from `src` is not supported by `dest`
-then an [`UnsupportedAttribute`](@ref) error is thrown. Unsupported *optimizer*
-attributes are treated differently:
-
-* If `warn_attributes` is `true`, a warning is displayed, otherwise,
-* the attribute is silently ignored.
+    copy_to(
+        dest::ModelLike,
+        src::ModelLike;
+        copy_names::Bool = true,
```
odow, line 130: Remind me why we need `copy_names`? Why can't we just try by default, and skip if they aren't supported?

src/MathOptInterface.jl
blegat: No we should remove in MOI v0.10! It's a good thing we didn't forget.

blegat (approved):
Let's address warn_attributes and copy_names in separate issues/PRs

PR #1532, opened 2021-08-12
WIP: [Bridges] refactor graph

odow:
Not sure if we should merge this.

In the spirit of "use fewer distinct types to improve latency and improve inferrability" I had a go at refactoring the graph to use a single `Node` and `Edge` type. There's essentially no change on the benchmarks.

Precompilation really breaks slightly above this layer in the vector of bridge types :(


blegat:
Yes, the graph part isn't the issue. The issue is building the graph from the vector of bridge types.

PR #1534, opened 2021-08-14
Remove warn_attributes

odow:
Technically breaking, but I think this only ever existed in the documentation.

Looks like Hypatia has it as an argument (cc @chriscoey), but didn't do anything with it:

image


chriscoey:
thanks for the notification, I'll remove from Hypatia. (btw that's a nice search function that you screenshat, how did you access it?)

odow:
It's the JuliaHub code search over all registered packages: https://juliahub.com/ui/Search?q=warn_attributes&type=code

Super useful for seeing what is used where.

PR #1535, opened 2021-08-14
Remove copy_names argument

### LINKED ISSUE #1533: The need for copy_names
`copy_to` has a `copy_names` argument:

https://github.com/jump-dev/MathOptInterface.jl/blob/b391731104a08630e4e4215c990458457a05680c/src/MathOptInterface.jl#L127-L135

This also leads to things like 

https://github.com/jump-dev/MathOptInterface.jl/blob/b391731104a08630e4e4215c990458457a05680c/src/MathOptInterface.jl#L98-L109

The conceptual understanding of when and why these are used is a bit confusing.

It was added here, without much explanation of why:
https://github.com/jump-dev/MathOptInterface.jl/pull/297

As far as I can tell, the reason this was added was because we want to add names to the cache of a CachingOptimizer but not the optimizer:
https://github.com/jump-dev/MathOptInterface.jl/blob/b391731104a08630e4e4215c990458457a05680c/src/Utilities/cachingoptimizer.jl#L192-L203
This leads to confusion:
https://github.com/jump-dev/MathOptInterface.jl/issues/1225
and supporting names is hard:
https://github.com/jump-dev/MathOptInterface.jl/issues/674

Why do we need `copy_names`? Can we just ask the optimizer if it supports names? And if so, shouldn't we set them? This would make optimizer errors much nicer instead of `C21 was infeasible`.

We already skip some attributes if they aren't supported, so why not copy?
https://github.com/jump-dev/MathOptInterface.jl/blob/f08c78cc776e36e92270d42912128543d131c025/src/Utilities/copy.jl#L151-L162

We should either:

 - Remove `copy_names`
 - Document why it is needed

odow:
A look at what this would look like. I still need to benchmark to cost of always trying to copy names.

Closes #1533

### DIFF
--- a/docs/src/submodules/Utilities/overview.md
+++ b/docs/src/submodules/Utilities/overview.md
@@ -374,19 +374,14 @@ function _copy_to(
     return
 end
 
-function MOI.copy_to(
-    dest::Optimizer,
-    src::Model;
-    copy_names::Bool = false
-)
+function MOI.copy_to(dest::Optimizer, src::Model)
     _copy_to(dest, src)
     return MOI.Utilities.identity_index_map(src)
 end
 
 function MOI.copy_to(
     dest::Optimizer,
-    src::MOI.Utilities.UniversalFallback{Model};
-    copy_names::Bool = false
+    src::MOI.Utilities.UniversalFallback{Model},
 )
     # Copy attributes from `src` to `dest` and error in case any unsupported
     # constraints or attributes are set in `UniversalFallback`.
@@ -395,8 +390,7 @@ end
 
 function MOI.copy_to(
     dest::Optimizer,
-    src::MOI.ModelLike;
-    copy_names::Bool = false
+    src::MOI.ModelLike,
 )
     model = Model()
     index_map = MOI.copy_to(model, src)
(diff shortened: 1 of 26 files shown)


odow:
Not sure if/how to benchmark this. It's going to depend on the solver. In most cases, this should be minimal. If it is, it needs to be fixed in the solver, or an option in the solver provided to skip supporting names.

blegat (approved):

blegat:
If it turns out to be noticeable in performance, we can add a filter layer and an option to JuMP/CachingOptimizer to apply this layer to filter out the names. The question is : what would be the default ?
If the default would be to copy names then making this change later is non-breaking.
It seems to make sense to me that names would be copied by default. We can have a performance tips section in JuMP that recommends things like "unset the copy of names, use direct mode, ...". We could even have an option to not even set the names to the `moi_backend` from JuMP but these will not be the default and we will wait to have a confincing benchmark first and can be done post v1.0 as non-breaking.

odow:
Yeah there are multiple ways to improve performance later, whether that's adding a new layer, adding an option to JuMP, or adding an option to the solver. Either way, I don't think a `copy_names` argument is the answer.

blegat:
Yes, we should aim at having a `copy_to` with only 2 arguments. Otherwise we are pushing additional complexity to all the solver wrappers.

PR #1540, opened 2021-08-20
[Bridges] tidy quad_to_soc.jl

### LINKED ISSUE #538: Add QuadtoSOC test with permutation
The QuadtoSOC bridge has been added in https://github.com/JuliaOpt/MathOptInterface.jl/pull/483 using sparse cholesky.
For the sparse cholesky of A, we have `A = P' * L * L' * P`. The implementation takes into account the permutation but in the tests, the permutation matrix is always the identity so we wouldn't catch if we were using `P'` were it should be `P` for instance.

odow:
Closes #538

In addition, I better documented the function and fixed some inference issues. Julia doesn't like splatting, so let's avoid it if at all possible.

### DIFF
--- a/src/Bridges/Constraint/quad_to_soc.jl
+++ b/src/Bridges/Constraint/quad_to_soc.jl
@@ -97,8 +97,7 @@ function bridge_constraint(
         MOI.VectorAffineTerm(
             2,
             MOI.ScalarAffineTerm(scale * term.coefficient, term.variable),
-        )
-        for term in func.affine_terms
+        ) for term in func.affine_terms
     ]
     # Add the Lx terms...
     L, p = SparseArrays.sparse(F.L), F.p
--- a/src/Bridges/Constraint/quad_to_soc.jl
+++ b/src/Bridges/Constraint/quad_to_soc.jl
@@ -75,7 +75,6 @@ function bridge_constraint(
     if !less_than
         LinearAlgebra.rmul!(Q, -1)
     end
-    # We have L × L' ≈ Q[p, p]
     F = try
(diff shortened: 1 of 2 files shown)


src/Bridges/Constraint/quad_to_soc.jl
```diff
@@ -136,23 +156,7 @@ function matrix_from_quadratic_terms(
...
-    index_to_variable_map::Dict{Int,MOI.VariableIndex},
-) where {T}
-    # We know that L × L' ≈ Q[p, p] hence (L × L')[i, :] ≈ Q[p[i], p]
-    # We precompute the map to avoid having to do a dictionary lookup for every
-    # term
-    variable = map(i -> index_to_variable_map[p[i]], 1:size(L, 1))
-    function term(i::Integer, j::Integer, v::T)
-        return MOI.VectorAffineTerm(j, MOI.ScalarAffineTerm(v, variable[i]))
-    end
-    return map(ijv -> term(ijv...), zip(findnz(L)...))
```
blegat, line 155: Is this what Julia couldn't infer ? I wouldn't have guessed. It can infer the result of `findnz(L)` so the type of the elements of the `zip`, etc...

blegat (approved):

src/Bridges/Constraint/quad_to_soc.jl
odow: Hmm. I just tried in a new REPL and it seemed to work. So maybe I had something weird with Revise. The new way is quicker anyway because we avoid the multiple copies of the list of terms.

Issue #1345, opened 2021-05-13
TODO list for Bridges


odow:
## Fix TODO items in code

- [x] #1543 https://github.com/jump-dev/MathOptInterface.jl/blob/7923c601506cba2edb44f25eca830fc74d7cfe70/src/Bridges/bridge_optimizer.jl#L601
- [x] https://github.com/jump-dev/MathOptInterface.jl/pull/1348 <s>https://github.com/jump-dev/MathOptInterface.jl/blob/7923c601506cba2edb44f25eca830fc74d7cfe70/src/Bridges/bridge_optimizer.jl#L635-L636</s>
- [x] #1543 https://github.com/jump-dev/MathOptInterface.jl/blob/7923c601506cba2edb44f25eca830fc74d7cfe70/src/Bridges/bridge_optimizer.jl#L668-L670
- [ ] https://github.com/jump-dev/MathOptInterface.jl/blob/7923c601506cba2edb44f25eca830fc74d7cfe70/src/Bridges/bridge_optimizer.jl#L1905
- [x] https://github.com/jump-dev/MathOptInterface.jl/pull/1350 <s>https://github.com/jump-dev/MathOptInterface.jl/blob/7923c601506cba2edb44f25eca830fc74d7cfe70/src/Bridges/Constraint/bridge.jl#L9</s>
- [x] #1548 https://github.com/jump-dev/MathOptInterface.jl/blob/7923c601506cba2edb44f25eca830fc74d7cfe70/src/Bridges/Constraint/Constraint.jl#L99
- [x] #1542 https://github.com/jump-dev/MathOptInterface.jl/blob/7923c601506cba2edb44f25eca830fc74d7cfe70/src/Bridges/Constraint/det.jl#L1
- [x] #1544 https://github.com/jump-dev/MathOptInterface.jl/blob/7923c601506cba2edb44f25eca830fc74d7cfe70/src/Bridges/Constraint/geomean.jl#L49
- [x] https://github.com/jump-dev/MathOptInterface.jl/pull/1349 <s>https://github.com/jump-dev/MathOptInterface.jl/blob/7923c601506cba2edb44f25eca830fc74d7cfe70/src/Bridges/Constraint/geomean.jl#L68-L69</s>
- [x] #433 https://github.com/jump-dev/MathOptInterface.jl/blob/7923c601506cba2edb44f25eca830fc74d7cfe70/src/Bridges/Constraint/scalarize.jl#L143

[...]

PR #1542, opened 2021-08-23
[Utilities] Move trimap from bridges

odow:
`trimap` was defined in two files, but used all-over...

Part of #1345

### DIFF
--- a/src/Utilities/sets.jl
+++ b/src/Utilities/sets.jl
@@ -122,3 +122,18 @@ Return the dimension `d` such that
 function side_dimension_for_vectorized_dimension(n::Base.Integer)
     return div(isqrt(1 + 8n), 2)
 end
+
+"""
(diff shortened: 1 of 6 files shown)


src/Utilities/sets.jl
```diff
@@ -122,3 +122,18 @@ Return the dimension `d` such that
...
+Convert between the row and column indices of a matrix, to the linear index of
+the corresponding element in the triangular representation.
+
+This is most useful when mapping between `ConeSquare` and `ConeTriangle` sets.
+"""
+function trimap(row::Integer, column::Integer)
+    if row < column
+        return trimap(column, row)
+    end
+    return div((row - 1) * row, 2) + column
```
blegat, line 138: Maybe use a ref to AbstractSymmetricMatrixSetTriangle

PR #1544, opened 2021-08-23
[Bridges] Maintenance of GeoMeanBridge

odow:
I was trying to address this TODO:
https://github.com/jump-dev/MathOptInterface.jl/blob/5fb36a46fe1de6c4b23abb515e6a5af50e4f23f8/src/Bridges/Constraint/geomean.jl#L48-L52
and I got very confused with the GeoMean implementation.

I started with some renaming, but then I gave up. It could use some much better documentation.

### DIFF
--- a/src/Bridges/Constraint/geomean.jl
+++ b/src/Bridges/Constraint/geomean.jl
@@ -1,16 +1,3 @@
-function _ilog2(n, i)
-    if n <= (one(n) << i)
-        i
-    else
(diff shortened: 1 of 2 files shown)


blegat (approved):

odow:
@blegat do you have the pdf of the book? (Or want to expand the documentation?)

odow:
Merging this for now. I'll come back with more documentation.

PR #1547, opened 2021-08-23
[Utilities] Remove .set field of CleverDict

### LINKED ISSUE #1326: set field of CleverDict
Looking at the code of `CleverDict`, `set` is simply `eachindex(vector)` isn't it ?
I thought that when an index was deleted, `set` was updated but it seems we switch to Dict instead so we should either remove this `set` vector or decide to stay dense when an element is deleted.

odow:
Closes #1326

This simplifies things by removing `.set` field. It also fixes an inference issue in `keys`.

The downside is that it breaks the iteration order for items in a `CleverDict` initialized with storage but not yet filled. 

To be honest, I don't really understand the need for pre-initializing the CleverDict. I've left the option, but it now calls `size hint` instead of pre-allocating the space.

### DIFF
--- a/src/Utilities/CleverDicts.jl
+++ b/src/Utilities/CleverDicts.jl
@@ -80,7 +80,6 @@ mutable struct CleverDict{K,V,F<:Function,I<:Function} <: AbstractDict{K,V}
     hash::F
     inverse_hash::I
     is_dense::Bool
-    set::BitSet
     vector::Vector{V}
     dict::OrderedCollections.OrderedDict{K,V}
     function CleverDict{K,V}(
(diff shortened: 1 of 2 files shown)


blegat:
> The downside is that it breaks the iteration order for items in a `CleverDict` initialized with storage but not yet filled. 

Can you elaborate on that ? I don't understand the change of iteration order

odow:
```julia
d = CleverDicts.CleverDict{Int,Float64}(hash, inverse_hash, 3)
d[4] = 0.25
d[2] = 1.5
```
used to iterate `2`, `4` because it assumed there were three elements, then we added `d[4]`, then set `d[2]`. That leaves `1` and `3` as `undef`.

Now preallocating with `3` will `sizehint!(d.vector, 3)`, but not actually preallocate the storage.

blegat:
Ok I see, the `OrderedDict` iterates by order of creation while the vector storage iterates by order `1:n`. There order match if we only allow the vector storage to be filled in the order 1, 2, 3, ....
In your example, we were using the vector storage even if it wasn't filled in the right order. That makes the order of iteration not match the order of creation which wasn't great. I would say this changes is an improvement.

blegat (approved):

odow:
I'll wait for @joaquimg to take a look before merging.

PR #1548, opened 2021-08-23
[Bridges] Fix indicator bridges

odow:
@blegat is this correct?

### DIFF
--- a/src/Bridges/Constraint/Constraint.jl
+++ b/src/Bridges/Constraint/Constraint.jl
@@ -90,11 +90,12 @@ const LogDet{T,OT<:MOI.ModelLike} = SingleBridgeOptimizer{LogDetBridge{T},OT}
 const RootDet{T,OT<:MOI.ModelLike} = SingleBridgeOptimizer{RootDetBridge{T},OT}
 
 include("indicator_activate_on_zero.jl")
-# TODO(odow): should something go here?
+const IndicatorActiveOnFalse{T,F,S,OT<:MOI.ModelLike} =
+    SingleBridgeOptimizer{IndicatorActiveOnFalseBridge{T,F,S},OT}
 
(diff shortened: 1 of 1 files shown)


src/Bridges/Constraint/Constraint.jl
```diff
@@ -90,11 +90,12 @@ const LogDet{T,OT<:MOI.ModelLike} = SingleBridgeOptimizer{LogDetBridge{T},OT}
 const RootDet{T,OT<:MOI.ModelLike} = SingleBridgeOptimizer{RootDetBridge{T},OT}
 
 include("indicator_activate_on_zero.jl")
-# TODO(odow): should something go here?
+const IndicatorActiveOnFalse{T,F,S,OT<:MOI.ModelLike} =
```
blegat, line 93: It should be `IndicatorActiveOnFalse{T,OT<:MOI.ModelLike}`

src/Bridges/Constraint/Constraint.jl
```diff
@@ -90,11 +90,12 @@ const LogDet{T,OT<:MOI.ModelLike} = SingleBridgeOptimizer{LogDetBridge{T},OT}
...
+    SingleBridgeOptimizer{IndicatorActiveOnFalseBridge{T,F,S},OT}
 
 include("indicator_sos.jl")
-const IndicatortoSOS1{T,BC<:MOI.AbstractScalarSet,MaybeBC} =
-    SingleBridgeOptimizer{IndicatorSOS1Bridge{T,BC,MaybeBC}}
+const IndicatortoSOS1{T,BC<:MOI.AbstractScalarSet,MaybeBC,OT<:MOI.ModelLike} =
```
blegat, line 97: Same here, what was previously here was incorrect. Use the other ones as examples

PR #1549, opened 2021-08-24
[breaking] Move test_models_equal to Test

odow:
It was a bit weird having `test_models_equal` as part of `Utilities` instead of `Test`. This also clarifies that it is not a function that works in all cases. It's intended for small models with everything named.

### DIFF
--- a/src/Test/Test.jl
+++ b/src/Test/Test.jl
@@ -440,6 +440,107 @@ function _test_attribute_value_type(
     return
 end
 
+function _test_variablenames_equal(model, variable_names)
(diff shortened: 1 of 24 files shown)


src/Test/Test.jl
```diff
@@ -176,6 +176,8 @@ function runtests(
             continue
         elseif !isempty(exclude) && any(s -> occursin(s, name), exclude)
             continue
+        elseif name == "test_models_equal"
```
blegat, line 179: Maybe we can rename the function to avoid confusion with the convention?

PR #1550, opened 2021-08-24
Clarify ConstraintPrimal for conic solvers

### LINKED ISSUE #564: test that ConstraintPrimal returns exact value of function at result point (not set constant)
see http://www.juliaopt.org/MathOptInterface.jl/dev/apireference/#MathOptInterface.ConstraintPrimal

this would help detect issues like https://github.com/JuliaOpt/ECOS.jl/issues/81

the test should check (exact) equality, not approximate equality

odow:
Closes #564

### DIFF
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -1320,14 +1320,22 @@ end
 """
     ConstraintPrimalStart()
 
-A constraint attribute for the initial assignment to some constraint's primal value(s) that the optimizer may use to warm-start the solve. May be a number or `nothing` (unset).
+A constraint attribute for the initial assignment to some constraint's
+[`ConstraintPrimal`](@ref) that the optimizer may use to warm-start the solve.
+
+May be `nothing` (unset), a number for [`AbstractScalarFunction`](@ref), or a
+vector for [`AbstractVectorFunction`](@ref).
 """
 struct ConstraintPrimalStart <: AbstractConstraintAttribute end
 
 """
     ConstraintDualStart()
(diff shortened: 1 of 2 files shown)


src/attributes.jl
```diff
@@ -1337,17 +1337,17 @@ struct ConstraintDualStart <: AbstractConstraintAttribute end
...
-Given a constraint `function-in-set`, the `ConstraintPrimal` is the value of the
-function evaluated at the primal solution of the variables.
+## Conic solvers
 
-For example, given the constraint
-`ScalarAffineFunction([x,y], [1, 2], 3)`-in-`Interval(0, 20)` and a primal
-solution of `(x,y) = (4,5)`, the `ConstraintPrimal` solution of the constraint
-is `1 * 4 + 2 * 5 + 3 = 17`.
+Some conic solvers reformulate `b - Ax in K` to `s = b - Ax, s in K`. These
+solvers should return the value of `s` for `ConstraintPrimal`, rather than
```
blegat, line 1348: I think it should be formulated as a new rule but rather as an example that the definition is flexible, e.g. that the solver may return something that is not exactly `f(x)` if it serves as a slack.
So the only think you know is that `ConstraintPrimal` and `ConstraintPrimalStart` correspond to the same thing and that `ConstraintPrimal` should be close to `b - Ax` for a feasible solution

odow: How's this now

blegat (approved):

PR #1553, opened 2021-08-24
Tutorial about expressions

dourouc05:
I wished I had this when I started with CP stuff, so here it is :). It's mostly agnostic to #1474, only two sentences would need to be changed. 

What is the preferred way of exploring the terms of a function? For now, I've been mostly using the `terms` member; should it be advertised more as a recommended practice?


docs/src/tutorials/manipulating_expressions.md
```diff
@@ -0,0 +1,138 @@
...
+
+When working with MathOptInterface, the major use of functions is creating 
+them, which is reviewed in this section.
+
+### Creating scalar affine functions
+
+The simplest scalar function is simply a variable: 
+
+```julia
+var_idx = add_variable(model) # Create the variable x
```
odow, line 26: Why not just
```suggestion
x = MOI.add_variable(model)
```

odow (commented):
This is great! One major comment: should we consider formatting this as a literate tutorial so we make sure it runs? The alternative is to have everything as doctests.

Currently people can't copy-paste the code and have it run.

docs/src/tutorials/manipulating_expressions.md
```diff
@@ -0,0 +1,138 @@
...
+merged without changing the semantics of the function: `2 * x + 1`. 
+
+Working with these objects might be cumbersome. Canonicalization helps maintain 
+redundancy to zero. 
+
+[`Utilities.is_canonical`](@ref) checks whether a function is already in its 
+canonical form:
+
+```julia
+Utilities.is_canonical(f2 + f2) # (x + 2) + (x + 2) is stored as x + x + 2
```
blegat, line 105: `x^2 +x +x+4` no?

docs/src/tutorials/manipulating_expressions.md
```diff
@@ -0,0 +1,138 @@
...
+```
+
+[`Utilities.canonical`](@ref) returns the equivalent canonical version of the 
+function:
+
+```julia
+Utilities.canonical(f2 + f2) # Returns 2 * x + 2
```
blegat, line 112: `2x + 4`

docs/src/tutorials/manipulating_expressions.md
dourouc05: Indeed, thanks!

dourouc05:
I have no idea how to make a literate tutorial or doctests, that's the main reason why I started in this way. I saw inhttps://github.com/jump-dev/MathOptInterface.jl/blob/master/docs/src/tutorials/example.md that there are a lot of `jldoctest`: I guess this is related to https://juliadocs.github.io/Documenter.jl/stable/man/doctests/#REPL-Examples?

odow:
> that there are a lot of jldoctest

Yeah. These blocks get run on build to check that they reproduce the output. That way you can make sure you don't have typos in the documentation.

The alternative is Literate.jl (https://github.com/fredrikekre/Literate.jl)

Here's one from JuMP:

https://github.com/jump-dev/JuMP.jl/blob/master/docs/src/tutorials/Getting%20started/an_introduction_to_julia.jl

Here's the modification to make to the `docs/make.jl`

https://github.com/jump-dev/JuMP.jl/blob/5a739673e27997cb783ef3a7f827de974a18d858/docs/make.jl#L39-L66

dourouc05:
I've converted the tutorial into doctests. I checked them in the REPL while working on it, I hope the tests will pass…

PR #1557, opened 2021-08-26
Replace filter_constraints with ModelFilter

### LINKED ISSUE #1554: Remove filter_constraints
We should consider removing `filter_constraints` keyword and make a function that does a filtered copy instead.
https://github.com/jump-dev/MathOptInterface.jl/blob/6f166a98f0bcc4141273135a77b6f642e3511b25/src/Utilities/copy.jl#L257
(and everywhere around there)

Or a layer that filters out things.

@blegat had something similar
https://github.com/lanl-ansi/Juniper.jl/blob/3a30ee5298d0dae426348da840c994eda901b985/src/filter.jl#L1-L76

Some use-cases:
 - filter out VariableName and ConstraintName
 - filter out Integer and ZeroOne constraints
 - filter out constraints not in the IISS

odow:
The `filter_constraints` argument was always a bit weird. This goes to a model-wrapper approach that @blegat created instead.

Closes #1554

### DIFF
--- a/src/Utilities/copy.jl
+++ b/src/Utilities/copy.jl
@@ -467,10 +467,7 @@ Use the `do` syntax to provide a single function.
 
 ```julia
 filtered_src = MOI.Utilities.ModelFilter(src) do item
-    if item isa Tuple{Type,Type}
-        return item != (MOI.SingleVariable, MOI.Integer)
-    end
-    return attr != MOI.ListOfConstraintIndices{MOI.SingleVariable,MOI.Integer}()
+    return item != (MOI.VariableIndex, MOI.Integer)
 end
 MOI.copy_to(dest, filtered_src)
 ```
--- a/src/Utilities/copy.jl
+++ b/src/Utilities/copy.jl
@@ -556,7 +556,7 @@ end
 function MOI.is_empty(model::ModelFilter)
     if MOI.is_empty(model.inner)
         return true
-    elseif MOI.get(model, MOI.NumberOfVariables()) > 0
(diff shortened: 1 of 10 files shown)


odow:
How's this now?

src/Utilities/copy.jl
```diff
@@ -459,13 +425,142 @@ function default_copy_to(
...
     final_touch(dest, index_map)
     return index_map
 end
+
+"""
+    ModelFilter(filter::Function, model::MOI.ModelLike)
+
+A layer to filter out various components of `model`. The filter function takes
+two arguments: an attribute (see below) and an element from the list returned by
+that attribute. It returns `true` if the element should be included in `dest`,
```
blegat, line 438: `dest` is not defined in this context. I think `ModelFilter` can be defined outside the context of a `copy_to`. It could be also useful for printing a subset of a model for instance.

src/Utilities/copy.jl
```diff
@@ -459,13 +425,142 @@ function default_copy_to(
...
+   * `MOI.ListOfConstraintIndices{F,S}`
+ * Specific attributes via:
+   * `MOI.ListOfModelAttributesSet`
+   * `MOI.ListOfConstraintAttributesSet`
+   * `MOI.ListOfVariableAttributesSet`
+
+!!! warning
+    The list of attributes filtered may change in a future release. You should
+    write functions that are generic and not limited to the five types listed
+    above.
```
blegat, line 455: By generic, you mean having a method `my_filter(::Any, ::Any) = true` ?

blegat (approved):

PR #1561, opened 2021-08-26
Document interface expected by subtypes

odow:
We have never explicitly documented which attributes we expect to be implemented. To date it has been "whatever is needed to pass the tests."

### DIFF
--- a/docs/src/manual/models.md
+++ b/docs/src/manual/models.md
@@ -64,15 +64,7 @@ Consult the docstsrings of each attribute for information on what it represents.
 
 ## ModelLike API
 
-Objects subtyping [`ModelLike`](ref) must implement:
-
- * [`is_empty`](@ref)
- * [`empty!`](@ref)
-
-The following attributes are available. Some are required by all
-implementations, and others are optional.
-
-**Required**
+The following attributes are available:
 
  * [`ListOfConstraintAttributesSet`](@ref)
  * [`ListOfConstraintIndices`](@ref)
@@ -82,41 +74,19 @@ implementations, and others are optional.
  * [`ListOfVariableIndices`](@ref)
  * [`NumberOfConstraints`](@ref)
  * [`NumberOfVariables`](@ref)
-
-**Optional**
-
  * [`Name`](@ref)
  * [`ObjectiveFunction`](@ref)
  * [`ObjectiveFunctionType`](@ref)
  * [`ObjectiveSense`](@ref)
-
-!!! note
-    To be useful, a [`ModelLike`](@ref) should, in most cases, also implement
-    [`copy_to`](@ref) or the incremental interface ([`add_variable`](@ref),
-    [`add_constraint`](@ref), etc.) See [Implementing a solver interface](@ref)
-    for more details.
-
 ## AbstractOptimizer API
 
-There are two options for [`AbstractOptimizer`](@ref)s:
-
- 1. Implement the required portion of the [ModelLike API](@ref), plus
-    [`optimize!`](@ref)
- 2. Implement [`copy_to_and_optimize!`](@ref)
-
-The following attributes are available. Some are required by all
-implementations, and others are optional.
-
-**Required**
+The following attributes are available:
 
  * [`DualStatus`](@ref)
  * [`PrimalStatus`](@ref)
  * [`RawStatusString`](@ref)
  * [`ResultCount`](@ref)
  * [`TerminationStatus`](@ref)
-
-**Optional**
-
  * [`BarrierIterations`](@ref)
  * [`DualObjectiveValue`](@ref)
  * [`NodeCount`](@ref)
--- a/docs/src/manual/models.md
+++ b/docs/src/manual/models.md
@@ -66,7 +66,7 @@ Consult the docstsrings of each attribute for information on what it represents.
 
 Objects subtyping [`ModelLike`](ref) must implement:
 
- * [`isempty`](@ref)
+ * [`is_empty`](@ref)
  * [`empty!`](@ref)
 
 The following attributes are available. Some are required by all
@@ -92,7 +92,8 @@ implementations, and others are optional.
 
 ## AbstractOptimizer API
 
-In addition to implementing the [ModelLike API](@ref), objects subtyping [`AbstractOptimizer`](ref) must implement:
+In addition to implementing the [ModelLike API](@ref), objects subtyping
+[`AbstractOptimizer`](ref) must implement:
 
  * [`optimize!`](@ref)
 
--- a/docs/src/manual/models.md
+++ b/docs/src/manual/models.md
@@ -90,12 +90,19 @@ implementations, and others are optional.
  * [`ObjectiveFunctionType`](@ref)
  * [`ObjectiveSense`](@ref)
 
+!!! note
+    To be useful, a [`ModelLike`](@ref) should, in most cases, also implement
+    [`copy_to`](@ref) or the incremental interface ([`add_variable`](@ref),
+    [`add_constraint`](@ref), etc.) See [Implementing a solver interface](@ref)
+    for more details.
+
 ## AbstractOptimizer API
 
-In addition to implementing the [ModelLike API](@ref), objects subtyping
-[`AbstractOptimizer`](ref) must implement:
+There are two options for [`AbstractOptimizer`](@ref)s:
 
- * [`optimize!`](@ref)
+ 1. Implement the required portion of the [ModelLike API](@ref), plus
+    [`optimize!`](@ref)
+ 2. Implement [`copy_to_and_optimize!`](@ref)
 
 The following attributes are available. Some are required by all
 implementations, and others are optional.
@@ -103,7 +110,6 @@ implementations, and others are optional.
 **Required**
 
  * [`DualStatus`](@ref)
- * [`ListOfOptimizerAttributesSet`](@ref)
  * [`PrimalStatus`](@ref)
  * [`RawStatusString`](@ref)
  * [`ResultCount`](@ref)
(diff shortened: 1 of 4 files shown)


docs/src/manual/models.md
```diff
@@ -62,50 +62,66 @@ functions.
...
+
+ * [`optimize!`](@ref)
+
+The following attributes are available. Some are required by all
+implementations, and others are optional.
+
+**Required**
+
+ * [`DualStatus`](@ref)
+ * [`ListOfOptimizerAttributesSet`](@ref)
```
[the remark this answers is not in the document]
odow: Shouldn't it be? I'm not actually sure if anyone implements this correctly. It's hard to track `RawOptimizerParameter`.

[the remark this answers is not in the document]
blegat: Yes, no optimizer implement it. It's only implemented by UniversalFallback and I don't think solvers need to implement it

docs/src/manual/models.md
```diff
@@ -62,50 +62,66 @@ functions.
...
- - [`DualObjectiveValue`](@ref)
- - [`ObjectiveBound`](@ref)
- - [`RelativeGap`](@ref)
- - [`SolveTimeSec`](@ref)
- - [`SimplexIterations`](@ref)
- - [`BarrierIterations`](@ref)
- - [`NodeCount`](@ref)
+## AbstractOptimizer API
+
+In addition to implementing the [ModelLike API](@ref), objects subtyping
```
blegat, line 95: I woulf not say that solvers have to implement the required attributes above. One may choose that the solver cannot be copied from for instance. It makes sense to what need to be implemented in order to be copied from but a solver is allowed to only implement `copy_to_and_optimize!`, the empty functions and getters for the results and nothing else

docs/src/manual/models.md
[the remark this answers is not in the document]
odow: Do we even need it as an official attribute then? Can't it just be something local within UniversalFallback?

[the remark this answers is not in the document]
blegat: Not if `is_set_by_optimize` of if it's `SolverName`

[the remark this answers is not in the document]
blegat: `SolverName` is just for printing so it's optional. Required is if it's needed to work with CachingOptimizer to solve and return the solution

docs/src/manual/models.md
```diff
@@ -62,50 +62,72 @@ functions.
...
- - [`ObjectiveValue`](@ref)
- - [`DualObjectiveValue`](@ref)
- - [`ObjectiveBound`](@ref)
- - [`RelativeGap`](@ref)
- - [`SolveTimeSec`](@ref)
- - [`SimplexIterations`](@ref)
- - [`BarrierIterations`](@ref)
- - [`NodeCount`](@ref)
+!!! note
+    To be useful, a [`ModelLike`](@ref) should, in most cases, also implement
```
blegat, line 94: What does "useful" means ? I'd rather say that the above required attributes are the attributes required for a model to be copied from and this below is the required functions to be implemented to be copied to. A read-only model is useful (e.g. it can be created from files  or from matrices), and a write-only as well (it can be stored on files or as matrices for instance).

docs/src/manual/models.md
```diff
@@ -62,50 +62,72 @@ functions.
...
+    [`copy_to`](@ref) or the incremental interface ([`add_variable`](@ref),
+    [`add_constraint`](@ref), etc.) See [Implementing a solver interface](@ref)
+    for more details.
+
+## AbstractOptimizer API
+
+There are two options for [`AbstractOptimizer`](@ref)s:
+
+ 1. Implement the required portion of the [ModelLike API](@ref), plus
```
blegat, line 103: If you rephrase the above section as suggested, you can here say that an `AbstractOptimizer` should either 1) be writable (e.g., implement `copy_to` or incremental interface) + implement `optimize!` or 2) implement `copy_to_and_optimize!`

docs/src/tutorials/implementing.md
```diff
@@ -241,9 +241,14 @@ Base.unsafe_convert(::Type{Ptr{Cvoid}}, model::Optimizer) = model.ptr
 
 ### Implement methods for `Optimizer`
 
-Now that we have an `Optimizer`, we need to implement a few basic methods.
+All `Optimizer`s must implement the following methods:
 
-* [`empty!`](@ref) and [`is_empty`](@ref)
+ * [`empty!`](@ref)
+ * [`is_empty`](@ref)
+ * [`optimize!`](@ref)
```
blegat, line 248: Or `copy_to_and_optimize!`

docs/src/tutorials/implementing.md
```diff
@@ -276,36 +280,16 @@ For each attribute
     attribute. You should make sure that your [`get`](@ref) function correctly
     infers to this type (or a subtype of it).
 
-Each column in the table indicates whether you need to implement the particular
-method for each attribute.
+All `Optimizer`s must implement [`get`](@ref) for the the following attributes:
```
blegat, line 283: This is the readable list, isn't it ? I would just say that the optimizers that want to be readable should implement the readable list detailed in the other sections ?

src/MathOptInterface.jl
```diff
@@ -5,6 +5,29 @@ module MathOptInterface
 
 Abstract supertype for objects that implement the "Model" interface for defining
 an optimization problem.
+
+## Interface
+
+The API of MathOptInterface is large. At a minimum, objects subtyping
+`ModelLike` must implement [`get`](@ref) for the following attributes:
```
blegat, line 12: Again, only if you want to be readable

docs/src/manual/models.md
odow: This discussion is difficult because the use-cases are so varied. There is no standard API required by all solvers. I think it's fair to say "in most cases" and leave it at that. We can always revise in future if confusion arises.

docs/src/tutorials/implementing.md
odow: I'm still not overly sold on this method.

src/MathOptInterface.jl
odow: No. I think it's _all_. If they aren't readable, they won't implement `get` for the specific attributes. I think we can ask all models "how many constraints do you have" regardless of whether we can access them.

blegat: I'm not going to implement `NumberOfConstraints` in solvers that only implement `copy_to_and_optimize!` since it's not even possible. ECOS now only stores the results so there is no way to implement it. These solvers will be used with a CachingOptimizer most of the time so there is no need to implement model-reading features.

odow: Then we don't really have a minimal API that every solver needs to implement. It's really "it depends."

mlubin: If we want to give up on defining a required API, then we at least should document that we intentionally use duck typing, and provide a list of recommended minimal attributes to support and a guide to the effect of, if you want your solver to support X then you need to implement Y.

blegat: I think the main use cases are:
1) Be useable from JuMP
2) Be useable from JuMP in direct mode
3) Being able to copy from it
4) Being able to add things incrementally

and 2) = 1) + 4) (+ 3) if you want printing, etc... to work)

docs/src/tutorials/implementing.md
```diff
@@ -20,6 +20,32 @@ MathOptInterface for a new solver.
...
+   after a solve, and have native support for things like variable names.
+ * Other solvers are "one-shot" solvers that require all of the problem data to
+   construct and solve the problem in a single function call. They do not
+   support modification or things like variable names.
+ * Other "solvers" are not solvers at all, but things like file readers. These
+   may only support functions like [`read_from_file`](@ref), and may not even
+   support the ability to add variables or constraints directly!
+ * Finally, some "solvers" are layers which take a problem as input, transform
+   it according to some rules, and pass the transformed problem to an inner
+   solver.
```
odow, line 47: How's this paragraph now

docs/src/tutorials/implementing.md
```diff
@@ -241,9 +267,14 @@ Base.unsafe_convert(::Type{Ptr{Cvoid}}, model::Optimizer) = model.ptr
...
-Now that we have an `Optimizer`, we need to implement a few basic methods.
+All `Optimizer`s must implement the following methods:
 
-* [`empty!`](@ref) and [`is_empty`](@ref)
+ * [`empty!`](@ref)
+ * [`is_empty`](@ref)
+ * [`optimize!`](@ref) (or [`copy_to_and_optimize!`](@ref))
+
+Other methods, detailed below, are optional or depend on how you implement the
+itnerface.
```
blegat, line 277:
```suggestion
interface.
```

docs/src/tutorials/implementing.md
blegat: It's accurate I think. New users feel the interface overwhelming when they need to implement it. This paragraph might improve things

blegat (approved):

docs/src/tutorials/implementing.md
```diff
@@ -20,6 +20,32 @@ MathOptInterface for a new solver.
...
+## A note on the API
+
+The API of MathOptInterface is large and varied. In order to support the
+diversity of solvers and use-cases, we make heavy use of
+[duck-typing](https://en.wikipedia.org/wiki/Duck_typing). That is, solvers are
+not expected to implement the full API, nor is there a well-defined minimal
+subset of what must be implemented. Instead, you should implement the API as
+necessary in order to make the solver function as you require.
+
+The main reason for using duck-typing over a well-defined API is that solvers
```
mlubin, line 32: Not clear what "over a well-defined API" means, I think this phrase can be dropped.

mlubin (approved):

PR #1562, opened 2021-08-27
Fixing docs following breaking change

matbesancon:


odow:
You can run with `doctests=:fix` to update the docs. We could do that in a separate PR.

test/Utilities/model.jl
```diff
@@ -182,7 +182,7 @@ function test_ObjectiveFunction()
     return
 end
 
-function _test_T
+function _test_T()
```
odow, line 185: Haha. I didn't consider this edge case.

matbesancon: Was it caused by one of the regexes I assume?

PR #1559, opened 2021-08-26
[breaking] remove SingleVariable

odow:
I think I lost the battle on SingleVariable, so here goes. Tests are broken because there are lots of manual changes still needed, but overall, the script covered most changes.

Closes #1472 
Closes #1474 

## TODOs

- [x] Formatting
- [x] Deprecations
- [x] Proof-read doc changes (@matbesancon)

## Script

```julia
create_regex(regex) = regex => s -> match(regex, s)[1]

regexes = [
    # Some type renames
    # TODO(odow): deprecate these
    "SingleVariableConstraintNameError" => "VariableIndexConstraintNameError",
    "SettingSingleVariableFunctionNotAllowed" => "SettingVariableIndexFunctionNotAllowed",
    "_ParsedSingleVariable" => "_ParsedVariableIndex",
    # Single find-and-replace for variables inside SingleVariable
    create_regex(r"MOI\.SingleVariable\(([a-zA-Z\.\[\]0-9]+?)\)"),
    create_regex(r"MOI\.SingleVariable\.\(([a-zA-Z\.\[\]0-9]+?)\)"),
    create_regex(r"SingleVariable\(([a-zA-Z\.\[\]0-9]+?)\)"),
    create_regex(r"SingleVariable\.\(([a-zA-Z\.\[\]0-9]+?)\)"),
    # Some safe changes, where there are now variables involved
    ", SingleVariable," => ", VariableIndex,",
    "isa SingleVariable`" => "isa VariableIndex`",
    "= SingleVariable`" => "= VariableIndex`",
    " SingleVariable " => " VariableIndex ",
    "SingleVariable functions" => "VariableIndex functions",
    "SingleVariable-in-" => "VariableIndex-in-",
    " SingleVariable-" => " VariableIndex-",
    "\"SingleVariable\"" => "\"VariableIndex\"",
    "SingleVariable's" => "VariableIndex's",
    "`SingleVariable`" => "`VariableIndex`",
    "`MOI.SingleVariable`" => "`MOI.VariableIndex`",
    "`MathOptInterface.SingleVariable`" => "`MathOptInterface.VariableIndex`",
    "= MOI.SingleVariable\n" => "= MOI.VariableIndex\n",
    "isa MOI.SingleVariable" => "isa MOI.VariableIndex",
    ", MOI.SingleVariable)" => ", MOI.VariableIndex)",
    "(MOI.SingleVariable, " => "(MOI.VariableIndex, ",
[...]

PR #1564, opened 2021-08-31
Fix doctests

matbesancon:
Fixes the doctests to 1.6 behavior and builds the doc on 1.6. I started the branch from #1559 to avoid stepping on each other's commit. I'm targeting master but could be onto @odow's branch


odow:
Can we do this after #1559, or as a separate PR without the SingleVariable changes?

matbesancon:
yes the goal was to wait for #1559 to be merged, the diff appearing will then be smaller

odow:
Closing in favor of #1570

PR #1565, opened 2021-09-01
[docs] move NEWS to the documentation

odow:
The Python ecosystem do this. I think it works quite well: http://xarray.pydata.org/en/latest/whats-new.html

It's doubly important for this release of MOI because there are such a large number of changes.

### DIFF
--- a/docs/src/release_notes.md
+++ b/docs/src/release_notes.md
@@ -3,7 +3,7 @@
 v0.10.0 (unreleased)
 ---------------------------
 
-**MOI is a significant breaking release. There are a large number of
+**MOI v0.10 is a significant breaking release. There are a large number of
 user-visible breaking changes and code refactors, as well as a substantial
 number of new features.**
 
--- a/docs/src/release_notes.md
+++ b/docs/src/release_notes.md
@@ -1,8 +1,7 @@
-MathOptInterface (MOI) release notes
-====================================
+# Release notes
 
-v0.10.0 (In development)
-------------------------
+v0.10.0 (unreleased)
+---------------------------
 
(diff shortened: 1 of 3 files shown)


odow:
I think this is useful: https://jump.dev/MathOptInterface.jl/previews/PR1565/release_notes/

blegat:
Why not keeping `NEWS.md` where it is ? So that it can be accessed directly in addition to being in the docs ? An advantage of having it in the docs is that we can use Documenter features such as `@ref`. In that case, we can move it in `docs/src` but then we would need a clear link to if from the README

odow:
> Why not keeping NEWS.md where it is ? So that it can be accessed directly in addition to being in the docs ?

You can still link directly: https://github.com/jump-dev/MathOptInterface.jl/blob/od/release_notess/docs/src/release_notes.md

How often to people actually check the `NEWS.md` file? Adding it to the docs makes it much more visible. We could add a placeholder `NEWS.md` that just says "check the docs"?

> having it in the docs is that we can use Documenter features such as @ref

I thought about this, but then when we remove things like `SingleVariable` you just break all the previous links. We could get better at writing detailed release notes with links to the PRs through.

docs/src/release_notes.md
```diff
@@ -1,8 +1,7 @@
-MathOptInterface (MOI) release notes
-====================================
+# Release notes
 
-v0.10.0 (In development)
-------------------------
+v0.10.0 (unreleased)
+---------------------------
 
 **MOI is a significant breaking release. There are a large number of
```
matbesancon, line 6:
```suggestion
**MOI v0.10 is a significant breaking release. There are a large number of
```

PR #1569, opened 2021-09-01
Update release notes with latest changes

odow:


odow:
@blegat / @mlubin we should discuss how to make the next release.

Options are:
 1. Tag 0.10
 2. Tag a 0.10-beta

If we tag 0.10 it's easier to update solvers, but then if we find anything that needs to be changed we'll need a 0.11. Tagging a `-beta` is going to require two passes of the solvers because we'll need to add CI logic to grab a version not in General. 

My preference is 1. I don't think  it's the end of the world if we need 0.11. And we can have 0.10.x releases.

mlubin:
I don't have a strong opinion, whichever you think is easier for updating the solvers. It's fine if we need to release 0.11 shortly after 0.10.

blegat:
I agree that 1. would make things easier.

blegat (approved):

Issue #1572, opened 2021-09-02
supports AbstractConstraintAttribute broken for bridges


odow:
HiGHS doesn't support `ConstraintDualStart`, so any bridged constraint shouldn't either. But yet:
```julia
julia> model = MOI.Bridges.full_bridge_optimizer(HiGHS.Optimizer(), Float64)
MOIB.LazyBridgeOptimizer{HiGHS.Optimizer}
with 0 variable bridges
with 0 constraint bridges
with 0 objective bridges
with inner model A HiGHS model with 0 columns and 0 rows.

julia> MOI.supports(model, MOI.ConstraintDualStart(), MOI.ConstraintIndex{MOI.VectorOfVariables,MOI.Nonnegatives})
true

julia> MOI.supports(model.model, MOI.ConstraintDualStart(), MOI.ConstraintIndex{MOI.VectorOfVariables,MOI.Nonnegatives})
false
```

Here's the offending code:
https://github.com/jump-dev/MathOptInterface.jl/blob/9a54ba742ed1f0dbe63f7548c3c960dcac1dc8c1/src/Bridges/bridge_optimizer.jl#L1218-L1234

It ends up calling 
https://github.com/jump-dev/MathOptInterface.jl/blob/9a54ba742ed1f0dbe63f7548c3c960dcac1dc8c1/src/Bridges/Constraint/scalarize.jl#L121-L127

which is only true if the inner optimizer also supports it.

blegat:
Here is the fix for a similar problem with the slack bridge https://github.com/jump-dev/MathOptInterface.jl/pull/1383/files

odow:
Yeah is it sufficient to just look at the inner optimizer? If so, why is #1383 still a draft?

blegat:
Because it needs tests ^^

odow:
I've opened a few PRs to fix these
 - [ ] #1580 
 - [ ] #1579
 - [ ] #1578 
 - [ ] #1577 
 - [ ] #1576 
 - [ ] #1575
 - [ ] #1574

I still have the following to go
 - [ ] geomean_to_relentr.jl
 - [ ] function_conversion.jl
 - [ ] det.jl

odow:
Closing because this is now fixed. The IndicatorSOS1Bridge PR is different.

PR #1573, opened 2021-09-02
Change version to 0.10.0

odow:
Any objections to tagging once this is merged? There are still some bugs like #1572, but they can be worked around in the solvers by excluding tests. It's not a breaking change.


blegat (approved):
Let's merge https://github.com/jump-dev/MathOptInterface.jl/pull/1576 and release

PR #1574, opened 2021-09-03
[Bridges] maintenance of zero_one.jl

odow:
What does it mean to `ConstraintPrimalStart` a `ZeroOne` constraint? We should only be supporting it if it makes sense.

### DIFF
--- a/src/Bridges/Constraint/zero_one.jl
+++ b/src/Bridges/Constraint/zero_one.jl
@@ -23,7 +23,7 @@ function bridge_constraint(
 end
 
 function MOI.Bridges.added_constraint_types(
-    ::Type{<:ZeroOneBridge{T}}
+    ::Type{<:ZeroOneBridge{T}},
 ) where {T}
     return Tuple{Type,Type}[
         (MOI.VariableIndex, MOI.Interval{T}),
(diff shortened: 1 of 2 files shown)


src/Bridges/Constraint/zero_one.jl
```diff
@@ -74,32 +69,12 @@ end
...
-end
-
-function MOI.set(
-    model::MOI.ModelLike,
-    attr::MOI.ConstraintPrimalStart,
-    bridge::ZeroOneBridge{T},
-    value,
-) where {T}
-    MOI.set(model, attr, bridge.integer_index, value)
-    MOI.set(model, attr, bridge.interval_index, value)
```
odow, line 98: @blegat does `ConstraintPrimalStart` make sense for a `ZeroOne` constraint?

blegat: Yes, a ConstraintPrimalStart is the starting value of the function. It's well-defined for all constraints

odow: Sure. But does any solver actually implement it...

blegat: Since most solvers only support VariableIndex-in-Integer, and setting the starting value of this is a bit redundant, I don't think so. However, if a solver want to implemented ScalarAffineFunction in Integer, they will need this.

odow: To rephrase: in what situation would a solver benefit from being able to set `ConstraintPrimalStart` of a `ZeroOne` constraint.

odow:
Merged for now. We can always revisit in the future. For now, `ConstraintPrimalStart` isn't needed.

PR #1576, opened 2021-09-03
[Bridges] maintenance of slack.jl

odow:


odow:
@blegat this is broken because of 
https://github.com/jump-dev/MathOptInterface.jl/blob/3897b7cfb4f93335b1b367b609e0b50587a47bec/src/Bridges/bridge_optimizer.jl#L639-L644

What's the `||` part doing here? Because `variable_function && supports_constraint_bridges(b)` only means that one of the inner variables might be bridged, not that they actually are. Ideas of how to fix?

src/Bridges/bridge_optimizer.jl
```diff
@@ -1218,19 +1218,20 @@ end
...
+        return MOI.supports(b, attr, Constraint.concrete_bridge_type(b, F, S))
+    end
+    # If the bridge is capable of bridging the set for `VFT-in-S`, and `b` will
+    # be used to bridge the set, check the variable bridge supports attr.
+    VFT = MOI.Utilities.variable_function_type(S)
+    if F == VFT && is_bridged(b, S) && is_variable_bridged(b, S)
+        return MOI.supports(b, attr, Variable.concrete_bridge_type(b, S))
+    end
+    # We're not bridging, so we check if the inner model supports it.
+    return MOI.supports(b.model, attr, MOI.ConstraintIndex{F,S})
```
odow, line 1234: @blegat can you take a look at this?

blegat:
> What's the `||` part doing here? Because `variable_function && supports_constraint_bridges(b)` only means that one of the inner variables might be bridged, not that they actually are. Ideas of how to fix?

The issue was that we have to use `recursive_model`. This would have failed as well e.g. for a chordal bridge that bridges VAF-in-PSD -> VAF-in-PSD

blegat (approved):

PR #1581, opened 2021-09-03
[Bridges] maintenance of det.jl

odow:


src/Bridges/Constraint/det.jl
```diff
@@ -446,15 +440,27 @@ function MOI.get(
...
-# References
 function MOI.delete(model::MOI.ModelLike, bridge::RootDetBridge)
     MOI.delete(model, bridge.gmindex)
     MOI.delete(model, bridge.sdindex)
     MOI.delete(model, bridge.Δ)
     return
 end
 
-# Attributes, Bridge acting as a constraint
+function MOI.supports(
```
blegat, line 450: The setter for `ConstraintPrimalStart` is not yet implemented for the bridge so it does not support it yet

odow: Oops. I overlooked this. I just saw the `get` was implemented. I removed the `get` method for the start values then.

PR #1582, opened 2021-09-05
[Bridges] maintenance of scalarize.jl

odow:


odow:
Confirmed that GLPK now passes the `ConstraintPrimal/DualStart tests with this PR.

PR #1585, opened 2021-09-05
[breaking] [Bridges] Refactor IndicatorSOS1Bridge

odow:
In particular:

 * Remove MaybeBC type parameter
 * Remove and rename fields in struct
 * Fix MOI.supports for ConstraintPrimalStart

The fields and MaybeBC parameter were superfluous because we can
implicitly figure out what they are from the existing set.

### DIFF
--- a/src/Bridges/Constraint/indicator_sos.jl
+++ b/src/Bridges/Constraint/indicator_sos.jl
@@ -1,68 +1,61 @@
 """
-    IndicatorSOS1Bridge{T, BC <: MOI.AbstractScalarSet}
+    IndicatorSOS1Bridge{T,S<:MOI.AbstractScalarSet}
 
-The `IndicatorSOS1Bridge` replaces an indicator constraint of the following form:
+The `IndicatorSOS1Bridge` replaces an indicator constraint of the following
+form:
 ``z \\in \\mathbb{B}, z == 1 \\implies f(x) \\leq b`` with a SOS1 constraint:
-``z \\in \\mathbb{B}, w \\leq 0, f(x) + w \\leq b, SOS1(w, z)``.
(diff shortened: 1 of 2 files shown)


src/Bridges/Constraint/indicator_sos.jl
```diff
@@ -1,68 +1,61 @@
 """
-    IndicatorSOS1Bridge{T, BC <: MOI.AbstractScalarSet}
+    IndicatorSOS1Bridge{T,S<:MOI.AbstractScalarSet}
 
-The `IndicatorSOS1Bridge` replaces an indicator constraint of the following form:
+The `IndicatorSOS1Bridge` replaces an indicator constraint of the following
+form:
 ``z \\in \\mathbb{B}, z == 1 \\implies f(x) \\leq b`` with a SOS1 constraint:
-``z \\in \\mathbb{B}, w \\leq 0, f(x) + w \\leq b, SOS1(w, z)``.
+``z \\in \\mathbb{B}, slack \\leq 0, f(x) + slack \\leq b, SOS1(slack, z)``.
```
odow, line 7: Can anyone remind me why we need `w <= 0` here?

odow: I'm not sure this is a good enough argument: https://github.com/jump-dev/MathOptInterface.jl/pull/877#discussion_r324455787

We could simplify things a lot by removing it. If people are bridging indicator constraints, they probably don't care about an extra slack variable in some solvers.

odow: cc @matbesancon since you originally added this

odow: Talked to @blegat offline and we agreed that this change is okay.

PR #1591, opened 2021-09-07
Fix DeprecatedTest in modellike that were using copy_names

blegat:
This is needed by NLopt: https://github.com/JuliaOpt/NLopt.jl/pull/175


odow:
I had avoided changing these since they're going to be removed anyway. I guess it can go in 0.10.1

blegat:
Yes it's no big deal but it's easier when upgrading if these tests still work

PR #1592, opened 2021-09-07
Replace copy_to_and_optimize! with optimize_model!

blegat:
There is an issue with `copy_to_and_optimize!`. What if the user changes the optimizer attributes and then want to optimize again ?
The solver supports setting optimizer attributes so the state in the `CachingOptimizer` will remain `ATTACHED_OPTIMIZER`.
So it will call `optimize!` but the solver can't optimize since it did not cache the model data.
For solvers relying on `copy_to_and_optimize!`, `CachingOptimizer` should keep the state `EMPTY_OPTIMIZER`.
This PR allows to do that with a new `MOI.optimize_model!`.
I have kept `copy_to_and_optimize!` so that it's not breaking but we can get rid of it in MOI v1.0.

Closes https://github.com/jump-dev/MathOptInterface.jl/issues/1604

### DIFF
--- a/src/MathOptInterface.jl
+++ b/src/MathOptInterface.jl
@@ -171,7 +171,7 @@ end
 
 A single call similar to calling [`copy_to`](@ref) followed by
 [`optimize!`](@ref). Return an [`IndexMap`](@ref) and a `Bool` `copied`.
-If `copied` is `true`, `src` was copied to `dest` so [`optimize!](@ref)
+If `copied` is `true`, `src` was copied to `dest` so [`optimize!`](@ref)
 can be called again on `dest`, e.g., after changing the starting values
 or optimizer attributes. On the other hand, if `copied` is `false`, the
 model was not copied to `dest` so calling [`optimize!`](@ref) is not allowed.
--- a/src/MathOptInterface.jl
+++ b/src/MathOptInterface.jl
@@ -30,8 +30,43 @@ abstract type AbstractOptimizer <: ModelLike end
     optimize!(optimizer::AbstractOptimizer)
 
 Start the solution procedure.
+
+    optimize!(
+        dest::AbstractOptimizer,
+        src::ModelLike,
+    )::IndexMap
+
+A single call similar to calling [`copy_to(dest, src)`](@ref `copy_to`) followed
+by `optimize!(dest)`. Return an [`IndexMap`](@ref) and a `Bool` `copied`.
+The [`IndexMap`](@ref) object translates variable and constraint
+indices from the `src` model to the corresponding indices in the `dest`
+optimizer, similarly to the [`IndexMap`](@ref) returned by [`copy_to`](@ref).
+If `copied` is `true`, `src` was copied to `dest` so `optimize!(dest)`
+can be called again on `dest`, e.g., after changing the starting values
+or optimizer attributes. On the other hand, if `copied` is `false`, the
+model was not copied to `dest` so calling `optimize!(dest)` is not allowed.
+However, attributes for which [`is_set_by_optimize`](@ref) is true can be
+queried from `dest`.
+
+Subtypes, say `Optimizer`, of [`AbstractOptimizer`](@ref) should
+* either implement [`MOI.copy_to(::Optimizer, ::MOI.ModelLike)`](@ref `copy_to`) and
+  `optimize!(::Optimizer)` or
+* implement `optimize!(::Optimizer, ::MOI.ModelLike)`.
+
+!!! note
+    The main purpose of `optimize!` method with two arguments is for use in [`Utilities.CachingOptimizer`](@ref).
+
+!!! warning
+    The new `optimize!` method with two arguments is an experimental new feature of MOI v0.10.2 that may break in MOI v1.0.
 """
 function optimize! end
+
+function optimize!(dest, src)
+    # The arguments above are untyped to avoid ambiguities.
+    index_map = copy_to(dest, src)
+    optimize!(dest)
+    return index_map, true
+end
 
 """
     compute_conflict!(optimizer::AbstractOptimizer)
@@ -161,37 +196,6 @@ function copy_to(dest, src; kwargs...)
         )
     end
     return copy_to(dest, src)
-end
-
-"""
-    optimize_model!(
-        dest::AbstractOptimizer,
-        src::ModelLike,
-    )::IndexMap
-
-A single call similar to calling [`copy_to`](@ref) followed by
-[`optimize!`](@ref). Return an [`IndexMap`](@ref) and a `Bool` `copied`.
-If `copied` is `true`, `src` was copied to `dest` so [`optimize!`](@ref)
-can be called again on `dest`, e.g., after changing the starting values
-or optimizer attributes. On the other hand, if `copied` is `false`, the
-model was not copied to `dest` so calling [`optimize!`](@ref) is not allowed.
-However, attributes for which [`is_set_by_optimize`](@ref) is true can be
-queried from `dest`.
-
-An optimizer can decide to implement this function instead of implementing
-[`copy_to`](@ref) and [`optimize!`](@ref) individually.
-
-!!! note
-    The main purpose of this function is for use in [`Utilities.CachingOptimizer`](@ref).
-
-!!! warning
-    This is an experimental new feature of MOI v0.10.1 that may break in MOI v1.0.
-"""
-function optimize_model!(dest, src)
-    # The arguments above are untyped to avoid ambiguities.
-    index_map = copy_to(dest, src)
-    optimize!(dest)
-    return index_map, true
 end
 
 include("error.jl")
--- a/src/MathOptInterface.jl
+++ b/src/MathOptInterface.jl
@@ -202,7 +202,7 @@ An optimizer can decide to implement this function instead of implementing
 [`copy_to`](@ref) and [`optimize!`](@ref) individually.
 
 !!! warning
-    This is an experimental new feature of MOI v0.10 that may break in MOI v1.0.
+    This function will be removed in MOI v1.0, use `optimize_model!` instead.
 """
 function copy_to_and_optimize!(dest, src) # TODO remove in favor of `optimize_model!`
     # The arguments above are untyped to avoid ambiguities.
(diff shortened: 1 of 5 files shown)


odow:
I think I need to think about this a bit more. Couldn't SCS just error if the optimizer attributes are set after `optimize!`?

blegat:
A common use case is to run it for hours with a time limit. And then start again from where it left off. cc @kalmarek

odow:
Ah. I see we don't have a try-catch 
https://github.com/jump-dev/MathOptInterface.jl/blob/c2e237bdd3686db51753fd964fe6bd5087d0f202/src/Utilities/cachingoptimizer.jl#L917-L928

I'd prefer a different work-around than this PR. It's another special case to deal with. What happens if I just call `optimize!` twice with `SCS` without changing anything? And wouldn't restarting from the previous point count as setting a variable attribute which would reset the optimizer anyway?

blegat:
> Ah. I see we don't have a try-catch

What do you mean ? Solvers should always support setting optimizer attributes, these are kind of orthogonal to the `copy_to`/`empty!`/`is_empty`.

> What happens if I just call `optimize!` twice with `SCS` without changing anything?

Currently, the solution is set as starting value for the next call: https://github.com/jump-dev/SCS.jl/blob/500e75d8f3a3467b8ae2814a49e41aa05c31439e/src/geometric.jl#L139-L149

> And wouldn't restarting from the previous point count as setting a variable attribute which would reset the optimizer anyway?

At the moment, I was thinking of implementing `MOI.set` for starting values but indeed, maybe I should not so that the optimizer is reset.

> I'd prefer a different work-around than this PR.

If we decide not to support setting starting values outside `copy_to`, I guess another option is to say that in something like
```julia
MOI.set(optimizer, MOI.TimeLimitSec(), 1.0)
MOI.optimize!(optimizer)
MOI.set(optimizer, MOI.TimeLimitSec(), 2.0)
MOI.optimize!(optimizer)
```
the second `optimize!` doesn't do anything. However, it seems counter intuitive for the user.

odow:
The second call to `TimeLimitSec` should throw a `NotAllowed` error and force the caching optimizer to reset.

blegat:
> The second call to `TimeLimitSec` should throw a `NotAllowed` error and force the caching optimizer to reset.

Then when will the time limit be set to 2 seconds ? The optimizer attributes are set when the state goes from `NO_OPTIMIZER` to `EMPTY_OPTIMIZER`.

odow:
The time limit would get set in the `model_cache`, and copied on the next run. I haven't looked into the details, but doing something like this seems better than introducing a new code-pathway that is only hit in very limited circumstances. We want to try and minimize the API surface of MOI, not introduce new functions.

blegat:
Some solvers just work as function that take the data as input and return the solution, without keeping the problem data at all.
These solvers need a `CachingOptimizer` if they want to be used by the incremental MOI interface. The specificity here is that this cache is more than just caching the problem so that it can be copied at once (and not incrementally), it is really the only copy of the model that exists.
Currently, the `CachingOptimizer` assumes that the optimizer also has a cache of the problem since it assumes that it can do `copy_to` and then `optimize!` separately. We just need to be able to use `CachingOptimizer` as the *last cache*. It already serves quite well the role of the last cache. The only things that are asked of the optimizer is to implement getters of `is_set_by_optimize` and setters for objective attributes. The last assumption that remains is that assumption that after `copy_to_and_optimize!`, the optimizer keeps a cache.

odow:
Is there any reason that the wrapper couldn't just store a cache anyway? You could have a solver option for `keep_cache_of_data::Bool = true`, and users can set it to `false` if they have memory issues, with the caveat that they can't re-optimize after modifying some parameters. 

This is such a niche use-case, and I think there are multiple alternatives without needing an additional complexity in MOI.

blegat:
> Is there any reason that the wrapper couldn't just store a cache anyway?

Adding a few lines in MOI is much less works that a lot of complexity in many solvers.

> This is such a niche use-case, and I think there are multiple alternatives without needing an additional complexity in MOI.

It is really just to be used with CachingOptimizer. It can be added in MOIU if needed. Otherwise, we aleady have `copy_to_and_optimize!`. This is its replacement so it does not really add complexity.

odow:
Do we need both `copy_to_and_optimize` and `optimize_model` then?

Can't we just change `copy_to_and_optimize` to also return the bool? We're fine breaking things in 0.10.2. It's not like anyone has implemented it.

odow:
Is there anyone going to implement `copy_to_and_optimize` _and_ keep a cache?

blegat:
ECOS implements `copy_to_and_optimize!` and is already released, this is why I plan to remove it in MOI v1.0

odow:
Ah. That's why I wasn't releasing things just yet. Can we consider yanking the release? https://github.com/jump-dev/ECOS.jl/issues/123. No one will be using it yet, so it should be safe to do so.

blegat:
You're right. I removed `copy_to_and_optimize!`. We can either
* Remove `copy_to_and_optimize!` and add `optimize_model!`
* Return the `Bool` in `copy_to_and_optimize!`

I have chosen the first solution in this PR because `copy_to_and_optimize!` might be misleading as it may not copy. However, I'm fine with both name.

src/MathOptInterface.jl
```diff
@@ -164,27 +164,31 @@ function copy_to(dest, src; kwargs...)
...
-
-Keyword arguments are passed to [`copy_to`](@ref).
+A single call similar to calling [`copy_to`](@ref) followed by
+[`optimize!`](@ref). Return an [`IndexMap`](@ref) and a `Bool` `copied`.
+If `copied` is `true`, `src` was copied to `dest` so [`optimize!`](@ref)
+can be called again on `dest`, e.g., after changing the starting values
+or optimizer attributes. On the other hand, if `copied` is `false`, the
+model was not copied to `dest` so calling [`optimize!`](@ref) is not allowed.
+However, attributes for which [`is_set_by_optimize`](@ref) is true can be
+queried from `dest`.
```
odow, line 179: I think we need a 
```Julia
!!! note
    The main purpose of this function is for use in [`Utilities.CachingOptimizer`](@ref).
```
Or something similar? We should say _why_ this function exists, because it's a little unusual.

odow:
I hate to keep bikesheding this, but `one_shot_optimize!`? Or even just `optimize!(dest, src)`? Overloading `optimize!` seems better than introducing a new verb.

blegat:
> I hate to keep bikesheding this, but `one_shot_optimize!`? Or even just `optimize!(dest, src)`? Overloading `optimize!` seems better than introducing a new verb.

I like `one_shot_optimize!`, as its for solvers that are implemented with a one-shot function. `optimize!(dest, src)` is probably better though. Since these solvers are not going to implement `optimize!(dest)`, if we name it with the same function, we can at least keep the rule: every `AbstractOptimizer` should implement `optimize!`.

src/MathOptInterface.jl
```diff
@@ -30,9 +30,44 @@ abstract type AbstractOptimizer <: ModelLike end
     optimize!(optimizer::AbstractOptimizer)
 
 Start the solution procedure.
+
+    optimize!(
+        dest::AbstractOptimizer,
+        src::ModelLike,
+    )::IndexMap
```
odow, line 37:
```suggestion
    )::Tuple{IndexMap,Bool}
```

src/Utilities/cachingoptimizer.jl
```diff
@@ -306,7 +280,14 @@ MOI.is_empty(m::CachingOptimizer) = MOI.is_empty(m.model_cache)
 
 function MOI.optimize!(m::CachingOptimizer)
     if m.mode == AUTOMATIC && m.state == EMPTY_OPTIMIZER
-        _attach_optimizer(m, MOI.optimize_model!)
+        indexmap, copied = MOI.optimize!(m.optimizer, m.model_cache)
+        if copied
+            m.state = ATTACHED_OPTIMIZER
+        end
+        # MOI does not define the type of index_map, so we have to convert it
+        # into an actual IndexMap. Also load the reverse IndexMap.
```
odow, line 288: It does now: https://github.com/jump-dev/MathOptInterface.jl/blob/ddbb343858245cf562da4edfc921594a734c8900/src/MathOptInterface.jl#L113

src/MathOptInterface.jl
```diff
@@ -30,9 +30,44 @@ abstract type AbstractOptimizer <: ModelLike end
...
+can be called again on `dest`, e.g., after changing the starting values
+or optimizer attributes. On the other hand, if `copied` is `false`, the
+model was not copied to `dest` so calling `optimize!(dest)` is not allowed.
+However, attributes for which [`is_set_by_optimize`](@ref) is true can be
+queried from `dest`.
+
+Subtypes, say `Optimizer`, of [`AbstractOptimizer`](@ref) should
+* either implement [`MOI.copy_to(::Optimizer, ::MOI.ModelLike)`](@ref `copy_to`) and
+  `optimize!(::Optimizer)` or
+* implement `optimize!(::Optimizer, ::MOI.ModelLike)`.
```
odow, line 54: This is a bit nasty. We now have
 * The incremental interface
 * `copy_to` then single-argument `optimize!`
 * Two-argument `optimize!`

Do we need `supports_copy_to_interface` like `supports_incremental_interface`?

blegat: What would this be useful for? The supports_incremental_interface is there for instantiate

src/Utilities/cachingoptimizer.jl
blegat: This is copy-paste from attach_optimizer. Maybe do that in another PR?

src/MathOptInterface.jl
odow: How does a user know to call `optimize!(::AbstractOptimizer)` or `optimize!(::AbstractOptimizer, ::ModelLike)`? Or should they only ever call `optimize!(::AbstractOptimizer)` and use a caching optimizer? 

If this is only ever used by the `CachingOptimizer`, it should almost be a `Utilities.optimize!` method.

If you had `supports_`, you could do this:
```julia
if MOI.supports_copy_to_interface(m.optimizer)
    indexmap = MOI.copy_to(m.optimizer, m.model_cache)
    MOI.optimize!(m.optimizer)
    m.state = ATTACHED_OPTIMIZER
else
    indexmap = MOI.optimize!(m.optimizer, m.model_cache)
    m.state = EMPTY_OPTIMIZER
end
m.model_to_optimizer_map = indexmap
m.optimizer_to_model_map = _reverse_index_map(index_map)
```

odow (approved):

PR #1601, opened 2021-09-09
VariablePrimal: document undefined behaviour 

### LINKED ISSUE #1600: [docs/convention] VariablePrimal: what should solvers return when there is no solution?
When working on Chuffed.jl, I wondered what I should return for `VariablePrimal` when the model is infeasible. Currently, this is not documented (https://github.com/jump-dev/MathOptInterface.jl/blob/master/src/attributes.jl#L1188-L1195). 

Intuitively, two solutions make sense to me: either throw an appropriate exception (similarly to `InvalidCallbackUsage` or `OptimizeInProgress`) or return `nothing`.

dourouc05:
Fixes #1600. 

If you are OK with the wording, the same patch could be applied to `ConstraintDual`, `ConstraintBasisStatus`, `ObjectiveValue`, `DualObjectiveValue`, `VariableBasisStatus`, `ConstraintPrimal`.

### DIFF
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -1023,10 +1023,13 @@ attribute_value_type(::ObjectiveFunctionType) = Type{<:AbstractFunction}
 
 A model attribute for the objective value of the primal solution `result_index`.
 
-If the solver does not have a primal value for the objective (for instance, 
-because of an infeasibility or because there is no solution with the 
-requested `result_index`), the result is undefined. Users should first check
-[`PrimalStatus`](@ref) before accessing the `ObjectiveValue` attribute.
+If the solver does not have a primal value for the objective because the 
+`result_index` is beyond the available solutions (whose number is indicated by
+the [`ResultCount`](@ref) attribute), getting this attribute must throw a 
+[`ResultIndexBoundsError`](@ref). Otherwise, if the result is unavailable for 
+another reason (for instance, only a dual solution is available), the result is
+undefined. Users should first check [`PrimalStatus`](@ref) before accessing the
+`ObjectiveValue` attribute.
 
 See [`ResultCount`](@ref) for information on how the results are ordered.
 """
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -158,7 +158,7 @@ end
     check_result_index_bounds(model::ModelLike, attr)
 
 This function checks whether enough results are available in the `model` for 
-the requested `attr`, using its `result_index` member variable. If the model 
+the requested `attr`, using its `result_index` field. If the model 
 does not have sufficient results to answer the query, it throws a 
 [`ResultIndexBoundsError`](@ref).
 """
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -1190,6 +1190,10 @@ struct VariablePrimalStart <: AbstractVariableAttribute end
 
 A variable attribute for the assignment to some primal variable's value in
 result `result_index`. If `result_index` is omitted, it is 1 by default.
+        
+If the solver does not have a primal value for the variable (for instance, 
+because of an infeasibility or because there is no solution with the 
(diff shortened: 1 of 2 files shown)


src/attributes.jl
```diff
@@ -1190,6 +1190,10 @@ struct VariablePrimalStart <: AbstractVariableAttribute end
 
 A variable attribute for the assignment to some primal variable's value in
 result `result_index`. If `result_index` is omitted, it is 1 by default.
+        
+If the solver does not have a primal value for the variable (for instance, 
+because of an infeasibility or because there is no solution with the 
+requested `result_index`), the result is undefined.
```
odow, line 1196: Mention that people should use `PrimalStatus` to check if there is a solution?

odow (commented):
Yeah adding this to other attributes is a good idea.

dourouc05:
I added the comment (with the required adaptations, as far as I can see) for the other mentioned attributes.

src/attributes.jl
```diff
@@ -1383,6 +1415,11 @@ A constraint attribute for the `BasisStatusCode` of some constraint in result
 
 See [`ResultCount`](@ref) for information on how the results are ordered.
 
+If the solver does not have a basis status for the constraint (for instance, 
+because of an infeasibility or because there is no solution with the 
+requested `result_index`), the result is undefined. Users should first check
+[`PrimalStatus`](@ref) before accessing the `ConstraintBasisStatus` attribute.
```
odow, line 1421: There's also `VariableBasisStatus`

odow (approved):
LGTM

blegat (changes requested):
There is two cases here:
1) `result_index` > `MOI.get(model, MOI.ResultCount())`. In that case,`ResultIndexBoundsError` should be thrown.
2) `result_index` <= `MOI.get(model, MOI.ResultCount())` but only one of the dual and primal are available. In that case, it is undefined indeed.

With the current wording, it is confusing and not clear you are only talking about case 2)

dourouc05:
@blegat I reformulated for `ObjectiveValue`, is this better now? (I also added some docs for ResultIndexBoundsError, so that I can reference to it.)

src/attributes.jl
```diff
@@ -138,11 +138,30 @@ SubmitNotAllowed(sub::AbstractSubmittable) = SubmitNotAllowed(sub, "")
...
 struct ResultIndexBoundsError{AttrType} <: Exception
     attr::AttrType
     result_count::Int
 end
 
+"""
+    check_result_index_bounds(model::ModelLike, attr)
+
+This function checks whether enough results are available in the `model` for 
+the requested `attr`, using its `result_index` member variable. If the model
```
blegat, line 161: "member variables" -> "fields" ?

blegat (approved):

dourouc05:
I have adapted the new paragraph for the other attributes.

src/attributes.jl
```diff
@@ -138,11 +138,30 @@ SubmitNotAllowed(sub::AbstractSubmittable) = SubmitNotAllowed(sub, "")
...
+    struct ResultIndexBoundsError{AttrType} <: Exception
+        attr::AttrType
+        result_count::Int
+    end
+
+An error indicating that the requested attribute `attr` could not be retrieved,
+because the solver returned too few results compared to what was requested. 
+For instance, the user tries to retrieve `VariablePrimal(2)` when only one 
+solution is available, or when the model is infeasible and has no solution. 
+"""
```
odow, line 151:
```suggestion

See also: [`check_result_index_bounds`](@ref).
"""
```

odow (approved):

src/attributes.jl
dourouc05: I forgot about that one, indeed! (That's why I initially added docs for that function.)

PR #1610, opened 2021-09-14
Add set_with_dimension

blegat:
Closes https://github.com/jump-dev/MathOptInterface.jl/issues/1606
Closes https://github.com/jump-dev/MathOptInterface.jl/pull/1607


dourouc05:
I just found out this PR. I'm wondering if this new function could be used to replace JuMP's `moi_set` (as in https://github.com/jump-dev/JuMP.jl/blob/78ca5da3756b01b812a515578bb2c63cd58aad0b/docs/src/developers/extensions.md)? To me, these functions are really doing the same thing.

PR #1613, opened 2021-09-15
[Utilities] throw UnsupportedAttribute in pass_attributes

odow:
We should put a proper `supports` check here and throw the right error so down-stream users don't have to duplicate work.

Another idea from https://github.com/jump-dev/SCS.jl/pull/219


blegat:
This should be needed, the fallback of `MOI.set` is already throwing the error.

PR #1615, opened 2021-09-15
[Utilities] fix emptying of CachingOptimizer

odow:
Found by https://github.com/jump-dev/SCS.jl/pull/219#issuecomment-920460637


odow:
I have a version of SCS passing with this PR.

blegat (approved):
We maybe need another CachingOptimizer state but that is out of the scope of this PR

PR #1618, opened 2021-09-17
[Utilities] allow callbacks through CachingOptimizer

odow:
I hit another snag: callbacks through a CachingOptimizer require `optimizer_to_model_map` to be set before `optimize!` so that they are accessible for things like `CallbackVariablePrimal`. But, the two-argument version means that this isn't set until afterwards. 

As a work-around, we call `attach` if a callback is set. I don't think there will be any optimizers that require the two-argument version and support callbacks. There aren't any at the moment, at least.

### DIFF
--- a/src/Utilities/cachingoptimizer.jl
+++ b/src/Utilities/cachingoptimizer.jl
@@ -277,7 +277,7 @@ function MOI.optimize!(m::CachingOptimizer)
         # to check if we have an `AbstractCallback` set and attach the optimizer
         # before recalling `MOI.optimize!`.
         for attr in MOI.get(m, MOI.ListOfModelAttributesSet())
-            if typeof(attr) <: MOI.AbstractCallback
+            if attr isa MOI.AbstractCallback
                 attach_optimizer(m)
                 return MOI.optimize!(m)
             end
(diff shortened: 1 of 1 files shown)


src/Utilities/cachingoptimizer.jl
```diff
@@ -271,6 +271,17 @@ MOI.is_empty(m::CachingOptimizer) = MOI.is_empty(m.model_cache)
 
 function MOI.optimize!(m::CachingOptimizer)
     if m.mode == AUTOMATIC && m.state == EMPTY_OPTIMIZER
+        # Here is a special case for callbacks: we can't use the two-argument
+        # call to optimize! because we need the `optimizer_to_model_map` to be
+        # set _prior_ to starting the optimization process. Therefore, we need
+        # to check if we have an `AbstractCallback` set and attach the optimizer
+        # before recalling `MOI.optimize!`.
+        for attr in MOI.get(m, MOI.ListOfModelAttributesSet())
+            if typeof(attr) <: MOI.AbstractCallback
```
blegat, line 280: This seems to be a `isa` use case. If we don't use `isa` here, we might as well never use it

blegat (approved):

PR #1619, opened 2021-09-17
[Test] fix ListOfConstraintAttributesSet for BadModel

odow:
Found while updating Ipopt. No test because I want to see what happens (I assume it's not covered). I'll add a test once CI finishes.


blegat:
Indeed, it seems wrong

PR #1625, opened 2021-09-22
Allow skipping NLPBlockDual

blegat:
Should it be this or `if MOI.get(model, MOI.DualStatus()) != MOI.NO_SOLUTION` ?

### DIFF
--- a/src/Test/test_nonlinear.jl
+++ b/src/Test/test_nonlinear.jl
@@ -450,6 +450,7 @@ function test_nonlinear_hs071_NLPBlockDual(model::MOI.ModelLike, config::Config)
     @requires MOI.supports(model, MOI.NLPBlock())
     @requires _supports(config, MOI.optimize!)
     @requires MOI.supports(model, MOI.VariablePrimalStart(), MOI.VariableIndex)
+    @requires _supports(config, MOI.NLPBlockDual)
     n_variables = 4
     v = MOI.add_variables(model, n_variables)
     l = [1.1, 1.2, 1.3, 1.4]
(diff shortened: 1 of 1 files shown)


src/Test/test_nonlinear.jl
```diff
@@ -469,26 +469,29 @@ function test_nonlinear_hs071_NLPBlockDual(model::MOI.ModelLike, config::Config)
...
-        MOI.eval_objective_gradient(evaluator, g, x)
-        # Evaluate ∑ μᵢ * ∇cᵢ(x)
-        jtv = zeros(n_variables)
-        MOI.eval_constraint_jacobian_transpose_product(
-            evaluator,
-            jtv,
-            x,
-            con_dual,
-        )
+        if _supports(config, MOI.BLPBlockDual) &&
```
odow, line 472: No. `@requires _supports(config, MOI.NLPBlockDual)` at the top of this file.

blegat: Indeed, there isn't anything left in the test otherwise

odow: Yeah all these test functions should now aim to test one thing only. We should skip entire tests, rather than having switches with test parts of tests.

blegat: But if we do `if MOI.get(model, MOI.DualStatus()) != MOI.NO_SOLUTION`, at least the solver don't have to put anything in `config` since it already implement `DualStatus`

odow: I considered this. My thinking that it was better for solvers to explicitly opt-out of attributes that are set by optimize, rather than silently skip ones they should support. 

Solvers not supporting duals have to pass `MOI.ConstraintDual` to exclude, for example.

blegat: Indeed, that makes sense.

PR #1630, opened 2021-10-02
Zero one conflict test actually testing integrality and not UB = 1

joaquimg:


joaquimg:
codecov fail due to the `return` in case the model does not support the test.
This is good for merging.

PR #1638, opened 2021-10-19
Build docs as a PDF

odow:
Part of https://github.com/jump-dev/JuMP.jl/pull/2760

Success: https://jump.dev/MathOptInterface.jl/previews/PR1638/MathOptInterface.pdf


odow:
Merging for now since there are some good improvements. The next version of Documenter will improve the layout of the PDF.

PR #1641, opened 2021-10-20
[docs] Update documenter version

odow:
This should include my fix for the header issue.


odow:
Muuuuch better: https://jump.dev/MathOptInterface.jl/previews/PR1641/MathOptInterface.pdf

## Old

image

## New

image

PR #1644, opened 2021-10-28
Export num_rows

blegat:
This would be useful for
* ECOS: https://github.com/jump-dev/ECOS.jl/blob/b4074b777132b947d7f85749fb6a49bf68d72284/src/MOI_wrapper.jl#L203
* SCS: https://github.com/jump-dev/SCS.jl/blob/20d8521f5a2f4dd366beaea64550014630f5e6d1/src/MOI_wrapper/MOI_wrapper.jl#L265-L270
* SeDuMI: https://github.com/jump-dev/SeDuMi.jl/blob/1b93571a5d9e7a870688c7a5479fdbbc83196be5/src/MOI_wrapper.jl#L189-L190

### DIFF
--- a/docs/src/submodules/Utilities/reference.md
+++ b/docs/src/submodules/Utilities/reference.md
@@ -113,6 +113,7 @@ Utilities.set_index
 Utilities.set_types
 Utilities.add_set
 Utilities.rows
+Utilities.num_rows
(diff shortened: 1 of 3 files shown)


odow:
What's the status of this?

blegat:
Just fixed the tests

PR #1645, opened 2021-10-29
[Utilities] Fix getter in UniversalFallback

odow:
While thinking about https://github.com/jump-dev/JuMP.jl/issues/2770. The current behavior is:
```Julia
julia> using JuMP, Gurobi

julia> model = Model(Gurobi.Optimizer)
Academic license - for non-commercial use only - expires 2021-11-06
A JuMP Model
Feasibility problem with:
Variables: 0
Model mode: AUTOMATIC
Solver name: Gurobi

julia> MOI.get(model, Gurobi.ModelAttribute("IsMIP"))
ERROR: KeyError: key Gurobi.ModelAttribute("IsMIP") not found
Stacktrace:
 [1] getindex
   @ ./dict.jl:482 [inlined]
 [2] _get
   @ ~/.julia/packages/MathOptInterface/D0LUL/src/Utilities/universalfallback.jl:264 [inlined]
 [3] get
   @ ~/.julia/packages/MathOptInterface/D0LUL/src/Utilities/universalfallback.jl:309 [inlined]
 [4] get(model::MathOptInterface.Utilities.CachingOptimizer{MathOptInterface.Bridges.LazyBridgeOptimizer{Gurobi.Optimizer}, MathOptInterface.Utilities.UniversalFallback{MathOptInterface.Utilities.Model{Float64}}}, attr::Gurobi.ModelAttribute)
   @ MathOptInterface.Utilities ~/.julia/packages/MathOptInterface/D0LUL/src/Utilities/cachingoptimizer.jl:808
 [5] get(model::Model, attr::Gurobi.ModelAttribute)
   @ JuMP ~/.julia/dev/JuMP/src/JuMP.jl:1217
 [6] top-level scope
   @ REPL[3]:1
```

This. is clearly unhelpful.

We return `nothing` variables and constraints, so it seems sensible to do the same with other attributes:
https://github.com/jump-dev/MathOptInterface.jl/blob/f5cc48179a7b41908000259566fe1d5a1f59b45a/src/Utilities/universalfallback.jl#L260-L292

But that'd mean that `MOI.get(model, Gurobi.ModelAttribute("IsMIP"))` returned `nothing`, which is also unhelpful. We should instead return a meaningful error message.

Thoughts?

### DIFF
--- a/src/FileFormats/MOF/nonlinear.jl
+++ b/src/FileFormats/MOF/nonlinear.jl
@@ -62,12 +62,9 @@ function write_nlpblock(
     model::Model,
     name_map::Dict{MOI.VariableIndex,String},
 ) where {T<:Object}
-    # TODO(odow): is there a better way of checking if the NLPBlock is set?
-    nlp_block = try
-        MOI.get(model, MOI.NLPBlock())
-    catch ex
-        @assert isa(ex, KeyError)
-        return  # No NLPBlock set.
+    nlp_block = MOI.get(model, MOI.NLPBlock())
+    if nlp_block === nothing
+        return
     end
     MOI.initialize(nlp_block.evaluator, [:ExprGraph])
     variables = MOI.get(model, MOI.ListOfVariableIndices())
--- a/src/FileFormats/NL/NL.jl
+++ b/src/FileFormats/NL/NL.jl
@@ -243,10 +243,10 @@ MOI.initialize(::_LinearNLPEvaluator, ::Vector{Symbol}) = nothing
 function MOI.copy_to(dest::Model, model::MOI.ModelLike)
     mapping = MOI.Utilities.IndexMap()
     # Initialize the NLP block.
-    nlp_block = try
-        MOI.get(model, MOI.NLPBlock())
-    catch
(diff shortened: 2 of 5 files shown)


blegat:
> But that'd mean that `MOI.get(model, Gurobi.ModelAttribute("IsMIP"))` returned `nothing`, which is also unhelpful. We should instead return a meaningful error message.

Gurobi should probably implement `is_set_by_optimize` and return true. It's not really set by the optimization but `is_set_by_optimize` is only ever used by the CachingOptimizer to decide whether to dispatch to the cache or the optimizer and by JuMP to know whether it should check the termination status. Is the use case of being able to attach but not optimize and get the attribute important? 

If it is, we should probably have `is_set_by_attach`

odow:
So for the Gurobi case, making the attributes `is_set_by_optimize` is probably okay. 

Regardless, there's still a bug here: getting `KeyError` is unhelpful for users.

odow:
For example, we used to have to `try-catch` our NLPBlock checks, but now if the block is missing it returns `nothing`. That seems better.

blegat:
Yes, the new behavior with this PR seems better.

PR #1648, opened 2021-11-07
Fix ZeroOne 2 infeasibility test

joaquimg:
This test was wrong by doing integer division instead of real-valued division


joaquimg:
This actually calls for a new version. Because the test was not possible to pass. It only affect IIS, so few solvers did suffer.

joaquimg:
I don't know why docs are breaking the change was very minor.

blegat:
> I don't know why docs are breaking the change was very minor.

Looks like a network issue, I restarted it

blegat (approved):

PR #1649, opened 2021-11-07
Fix tests names

joaquimg:
When tests require a number because of a duplicate name they were using `ii` and `2`. I kept the latter because it is clearer.
We should highlight this in docs?

### DIFF
--- a/src/Test/test_conic.jl
+++ b/src/Test/test_conic.jl
@@ -6768,7 +6768,7 @@ function setup_test(
 end
 
 """
-    test_conic_SecondOrderCone_negative_post_bound_ii(
+    test_conic_SecondOrderCone_negative_post_bound_2(
         model::MOI.ModelLike,
         config::Config{T},
     ) where {T}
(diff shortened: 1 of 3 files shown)


odow (commented):
Okay for now, but we need to come up with some documented rules for when we will add new tests or modify existing tests. This will cause solver tests to fail if the previously excluded the full name.

Is it okay to break tests of solvers in a patch/minor/major release of MOI?

joaquimg:
> Is it okay to break tests of solvers in a patch/minor/major release of MOI?

If the test is wrong, we should fix it even in the patch.
If a test tests existing functionality it could go in a patch too.
These two cases would break solver in a patch just as much a bug fix could break them.

Renaming is a bit more complicated indeed.
We could check if some entry of the excluded list is never used, so the solver has a chance to clean up.

In this special case, we can leave the renaming for 0.11.

We certainly need clear rules for capitalization (use if and only if it refers to a struct) and for numbering.

odow (approved):
I think this is technically breaking, but I'm okay putting it in a patch release. This will only break tests of solvers, not their user-facing behavior.

blegat (approved):
Agreed to have this in a patch release

PR #1650, opened 2021-11-07
Prep for v0.10.5

joaquimg:


src/Test/test_solve.jl
```diff
@@ -1423,7 +1423,7 @@ function test_solve_conflict_zeroone_ii(
     c2 = MOI.add_constraint(
         model,
         MOI.ScalarAffineFunction(MOI.ScalarAffineTerm.(one(T), [x]), zero(T)),
-        MOI.EqualTo(div(one(T), T(2))),
+        MOI.EqualTo(one(T) / T(2)),
```
odow, line 1426: This test only makes sense if `(one(T) / T(2)) isa T` right?

joaquimg: correct!

odow (approved):

PR #1655, opened 2021-11-09
[Bridges] throw better error for nonconvex quadratic constraints

### LINKED ISSUE #1621: Exclude nonconvex quadratic test
Add an easy way to do 
https://github.com/jump-https://github.com/jump-dev/SCS.jl/blob/0f3806d55d83493574d1222f334b9c5a9ed19f95/test/MOI_wrapper.jl#L59-L65
in the Config

odow:
Closes #1621

### DIFF
--- a/src/Bridges/Constraint/quad_to_soc.jl
+++ b/src/Bridges/Constraint/quad_to_soc.jl
@@ -78,15 +78,17 @@ function bridge_constraint(
     F = try
         LinearAlgebra.cholesky(LinearAlgebra.Symmetric(Q))
     catch
-        error("""
-            Unable to transform a quadratic constraint into a second-order cone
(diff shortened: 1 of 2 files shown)


blegat:
What would be the `_error_handler` to fix https://github.com/jump-dev/MathOptInterface.jl/blob/2d76d9c8e1b2c6e77490dcc097acfca514b40315/src/Test/Test.jl#L248 ?

odow:
We throw an `UnsupportedConstraint` so that it uses this handler
 https://github.com/jump-dev/MathOptInterface.jl/blob/2d76d9c8e1b2c6e77490dcc097acfca514b40315/src/Test/Test.jl#L237-L246

blegat (approved):

PR #1656, opened 2021-11-09
[Utilties] don't throw UnsupportedAttribute in pass_attributes

### LINKED ISSUE #1614: Unsupported attributes
I think https://github.com/jump-dev/MathOptInterface.jl/pull/1613 should be reverted. The fallback of `set` is already throwing the expected error. Why was that PR needed?

odow:
Closes #1614

@blegat's argument being that there should already be some fallbacks for this, and catching it here might obscure other bugs.

### DIFF
--- a/src/Utilities/copy.jl
+++ b/src/Utilities/copy.jl
@@ -25,7 +25,6 @@ function pass_attributes(
             if attr == MOI.Name()
                 continue  # Skipping names is okay.
             end
-            throw(MOI.UnsupportedAttribute(attr))
         end
         _pass_attribute(dest, src, index_map, attr)
     end
@@ -66,7 +65,6 @@ function pass_attributes(
             if attr == MOI.VariableName() || attr == MOI.VariablePrimalStart()
                 continue  # Skipping names and start values is okay.
             end
-            throw(MOI.UnsupportedAttribute(attr))
         end
         _pass_attribute(dest, src, index_map, vis_src, attr)
     end
@@ -115,7 +113,6 @@ function pass_attributes(
             )
                 continue  # Skipping names and start values is okay.
             end
-            throw(MOI.UnsupportedAttribute(attr))
         end
         _pass_attribute(dest, src, index_map, cis_src, attr)
     end


odow:
I tested GLPK.jl and it passed, but I didn't run the MOI tests. I don't really see why we shouldn't throw here. We've already figured out we're going to throw an error, so why not now? There should be other tests that catch incorrect usage in other parts of the API.

blegat:
Can't the `if !supports(...)` be removed as well, i.e., so that it completely reverts https://github.com/jump-dev/MathOptInterface.jl/issues/1614 ?

> We've already figured out we're going to throw an error, so why not now?

If you completely revert it, "We've already figured out" won't apply anymore

odow:
We still need to be able to skip names and start values, so we need the `!MOI.supports`.

odow:
Bump @blegat

blegat:
Indeed, the PR moved this `if` but it was already there.

> I don't really see why we shouldn't throw here. We've already figured out we're going to throw an error, so why not now?

That's one way to read this but when I read this code, I don't understand why we throw since `set` is going to throw. Every line in a code should be there for a reason and here it's confusing because it doesn't have to be there.

> There should be other tests that catch incorrect usage in other parts of the API.

The top PR comment of https://github.com/jump-dev/MathOptInterface.jl/pull/1613 says

> throw the right error so down-stream users don't have to duplicate work.

So the motivation was that `set` might not work. If we assume `set` is working then there is no reason to throw here.

odow:
So is that a yay/nay on merging this then?

blegat (approved):

odow:
Ah. I found the problem:
```julia
using SCS
const MOI = SCS.MOI
model = MOI.instantiate(SCS.Optimizer, with_bridge_type = Float64)
MOI.copy_to(model, MOI.Test.BadModelAttributeModel())  # No error. Should throw unsupported attribute
```
Why? Because the model has a cache:
```Julia
julia> model
MOIB.LazyBridgeOptimizer{MOIU.CachingOptimizer{SCS.Optimizer, MOIU.UniversalFallback{MOIU.Model{Float64}}}}
with 2 variable bridges
with 0 constraint bridges
with 0 objective bridges
with inner model MOIU.CachingOptimizer{SCS.Optimizer, MOIU.UniversalFallback{MOIU.Model{Float64}}}
  in state EMPTY_OPTIMIZER
  in mode AUTOMATIC
    with 1 model attribute
  with optimizer SCS.Optimizer
```

We'll throw the error on `optimize!`, but not on `copy_to`.

I think we should revert this PR; it broke tests for SCS.

odow:
Yeah, this caused pretty widespread breakage in the SolverTests: https://github.com/jump-dev/SolverTests/runs/4316646996?check_suite_focus=true

PR #1660, opened 2021-11-16
Add docs and tests for infeasibility certificates

### LINKED ISSUE #433: What to do with constants for results of InfeasibilityCertificate
When the status is `InfeasibilityCertificate`, the results should be the result of the homogenized version.
For instance, for the problem
```
min x
x <= -1
x >= 0
```
with dual
```
max -y
1 - y >= 0
y <= 0
```
or equivalently
```
max -y
1 - y - z = 0
y <= 0
z >= 0
```
The dual has an infeasibility ray `y = -1` (with `z = 1`). When asking for the constraint dual of `x >= 0`, this the value of the constraint `1 - y >= 0` in the first case or to the value of `z` in the second case.
In the first case, one could think it should be `1 - (-1) = 2` but it should be `z = 1` in the second case.
The reason is that since it is a ray, it is a feasible solution to the problem with the constant terms removed (which is equivalent to setting the objective of the primal to zero).

By consistency, the when asking for `ConstraintPrimal` when the primal status is `InfeasibilityRay`, it should also not take the constant of the function into account.

Moreover, if the objective function has a constant, then it should not be included in the value of the objective value / objective bound for `InfeasibilityRay`.
Indeed, the objective can be moved to a constrant by adding a new variable and then by consistency, it should not be included since it is in a constraint.

Currently, in `MOI.Test`, we have tests with infeasibility rays in the dual and the values expected by the tests are the correct ones. The reason is that the MOI definition of the dual is the second case of the introduction example above.
However, we have not test with the objective of the dual nor with infeasibility ray in the primal and for these it is might be much more confusing.

What do you think ? Do you agree that the constants should be removed from the results of rays in `ConstraintPrimal` and objective value/bound ?
If yes, we should:

- [ ] Add a section about this in the doc
- [ ] Thoroughly test it
- [ ] Test `ConstraintPrimal` of `VectorizeBridge` with a `DUAL_INFEASIBLE` example with nonzero constant for a scalar constraint

### LINKED ISSUE #1201: Add tests for DualObjectiveValue when infeasible

odow:
This PR adds documentation and tests for primal and dual infeasibility certificates.

Preview link
https://jump.dev/MathOptInterface.jl/previews/PR1660/background/infeasibility_certificates/

Closes #433. 
Closes #1201

### DIFF
--- a/docs/src/background/infeasibility_certificates.md
+++ b/docs/src/background/infeasibility_certificates.md
@@ -65,7 +65,7 @@ and the dual is a minimization problem in standard conic form:
 
 ## Unbounded problems
 
-If a problem is unbounded, if and only if:
+A problem is unbounded if and only if:
  1. there exists a feasible primal solution
  2. the dual is infeasible.
 
@@ -139,7 +139,7 @@ If the solver has found a certificate of primal infeasibility:
 
 ### Infeasibility certificates of variable bounds
 
-Many linear solvers (e.g., Gurobi) do not provide explicit access to the dual
+Many linear solvers (e.g., Gurobi) do not provide explicit access to the
 infeasibility certificate of a variable bound. However, given a set of linear
 constraints:
 ```math
@@ -148,7 +148,7 @@ l_A \le A x \le u_A \\
 l_x \le x \le u_x,
 \end{align}
 ```
-the dual certificate of the variable bounds can be computed using  the
+the certificate of the variable bounds can be computed using  the
 certificate associated with the affine constraints, ``d``. (Note that ``d`` will
 have one element for each row of the ``A`` matrix, and that some or all of the
 elements in the vectors ``l_A`` and ``u_A`` may be ``\pm \infty``. If both
--- a/docs/src/background/infeasibility_certificates.md
+++ b/docs/src/background/infeasibility_certificates.md
@@ -1,0 +1,153 @@
+```@meta
+CurrentModule = MathOptInterface
+DocTestSetup = quote
+    using MathOptInterface
+    const MOI = MathOptInterface
+end
+DocTestFilters = [r"MathOptInterface|MOI"]
+```
+
+# Infeasibility certificates
+
+When given an infeasible or unbounded model, some (conic) solvers can produce a
+certificate or infeasibility. This page explains what a certificate of
+infeasibility is for unbounded and infeasible models.
+
+## Conic duality
+
+MathOptInterface uses conic duality, which we use to define infeasibility
+certificates. A full explanation is given in the section [Duality](@ref), but
+here is a brief overview.
+
+For a minimization problem in geometric conic form, the primal is:
+```math
+\begin{align}
+& \min_{x \in \mathbb{R}^n} & a_0^\top x + b_0
+\\
+& \;\;\text{s.t.} & A_i x + b_i & \in \mathcal{C}_i & i = 1 \ldots m,
+\end{align}
+```
+and the dual is a maximization problem in standard conic form:
+```math
+\begin{align}
+& \max_{y_1, \ldots, y_m} & -\sum_{i=1}^m b_i^\top y_i + b_0
+\\
+& \;\;\text{s.t.} & a_0 - \sum_{i=1}^m A_i^\top y_i & = 0
+\\
+& & y_i & \in \mathcal{C}_i^* & i = 1 \ldots m,
+\end{align}
+```
+where each ``\mathcal{C}_i`` is a closed convex cone and ``\mathcal{C}_i^*`` is
+its dual cone.
+
+For a maximization problem in geometric conic form, the primal is:
+```math
+\begin{align}
+& \max_{x \in \mathbb{R}^n} & a_0^\top x + b_0
+\\
+& \;\;\text{s.t.} & A_i x + b_i & \in \mathcal{C}_i & i = 1 \ldots m,
+\end{align}
+```
+and the dual is a minimization problem in standard conic form:
+```math
+\begin{align}
+& \min_{y_1, \ldots, y_m} & \sum_{i=1}^m b_i^\top y_i + b_0
+\\
+& \;\;\text{s.t.} & a_0 + \sum_{i=1}^m A_i^\top y_i & = 0
+\\
+& & y_i & \in \mathcal{C}_i^* & i = 1 \ldots m.
+\end{align}
+```
+
+## Unbounded models
+
+If a model is unbounded, two things are true:
+ 1. there exists a feasible primal solution
+ 2. the dual is infeasible.
+
+A feasible primal solution---if one exists---can be obtained by setting
+[`ObjectiveSense`](@ref) to `FEASIBILITY_SENSE` and then optimizing. Therefore,
+most solvers terminate after they prove the dual is infeasible via a certificate
+of dual infeasibility, but _before_ they have found a feasible primal solution.
+This is also the reason that MathOptInterface defines the `DUAL_INFEASIBLE`
+status instead of `UNBOUNDED`.
+
+A certificate of dual infeasibility is an improving ray of the primal problem.
+That is, there exists some vector ``d`` such that for all ``\eta > 0``:
+```math
+A_i (x + \eta d) + b_i \in \mathcal{C}_i,\ \ i = 1 \ldots m,
+```
+and (for minimization problems):
+```math
+a_0^\top (x + \eta d) + b_0 < a_0^\top x + b_0,
+```
+for any feasible point ``x``. The latter simplifies to ``a_0^\top d < 0``. For
+maximization problems, the inequality is reversed, so that ``a_0^\top d > 0``.
+
+If the solver has found a certificate of dual infeasibility:
+
+ * [`TerminationStatus`](@ref) must be `DUAL_INFEASIBLE`
+ * [`PrimalStatus`](@ref) must be `INFEASIBILITY_CERTIFICATE`
+ * [`VariablePrimal`](@ref) must be the corresponding value of ``d``
+ * [`ObjectiveValue`](@ref) must be the value ``a_0^\top d``. Note that this is
+   the value of the objective function at `d`, ignoring the constant `b_0`.
+
+!!! note
+    The choice of whether to scale the ray ``d`` to have magnitude `1` is left
+    to the solver.
+
+## Infeasible models
+
+A certificate of primal infeasibility is an improving ray of the dual problem.
+However, because infeasibility is independent of the objective function, we
+first homogenize the primal problem by removing its objective.
+
+Therefore, a dual improving ray is some vector ``d`` such that for all
+``\eta > 0``:
+```math
+\begin{align}
+\sum_{i=1}^m A_i^\top (y_i + \eta d_i) & = 0 \\
+(y_i + \eta d_i) & \in \mathcal{C}_i^* & i = 1 \ldots m,
+\end{align}
+```
+and:
+```math
+\sum_{i=1}^m b_i^\top (y_i + \eta d_i) < \sum_{i=1}^m b_i^\top y_i,
+```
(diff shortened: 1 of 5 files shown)


blegat:
Is this tested ?

docs/src/background/infeasibility_certificates.md
```diff
@@ -0,0 +1,113 @@
...
+```
+for any feasible point ``x``. The latter simplifies to ``a_0^\top d \le 0``. For
+maximization problems, the inequality is reversed, so that ``a_0^\top d \ge 0``.
+
+If the solver has found a certificate of infeasibility:
+
+ * [`TerminationStatus`](@ref) must be `DUAL_INFEASIBLE`
+ * [`PrimalStatus`](@ref) must be `INFEASIBILITY_CERTIFICATE`
+ * [`VariablePrimal`](@ref) must be the corresponding value of ``d``
+ * [`ObjectiveValue`](@ref) must be the value ``a_0^\top d``.
```
blegat, line 82:
```suggestion
 * [`ObjectiveValue`](@ref) must be the value ``a_0^\top d``. Note that this is the value of the objective function at `d` ignoring the constant `b_0`.
```

docs/src/background/infeasibility_certificates.md
```diff
@@ -0,0 +1,113 @@
...
+the primal problem. That is, there exists some vector ``d`` such that for all
+``\eta > 0``:
+```math
+A_i (x + \eta d) + b_i \in \mathcal{C}_i,\ \ i = 1 \ldots m,
+```
+and (for minimization problems):
+```math
+a_0^\top (x + \eta d) + b_0 \le a_0^\top x + b_0
+```
+for any feasible point ``x``. The latter simplifies to ``a_0^\top d \le 0``. For
```
blegat, line 74:
```suggestion
for any feasible point ``x``. The latter simplifies to ``a_0^\top d < 0``. For
```

docs/src/background/infeasibility_certificates.md
```diff
@@ -0,0 +1,113 @@
...
+maximization problems, the inequality is reversed, so that ``a_0^\top d \ge 0``.
```
blegat, line 75:
```suggestion
maximization problems, the inequality is reversed, so that ``a_0^\top d > 0``.
```

docs/src/background/infeasibility_certificates.md
```diff
@@ -0,0 +1,113 @@
...
+\\
+& \;\;\text{s.t.} & a_0 + \sum_{i=1}^m A_i^\top y_i & = 0
+\\
+& & y_i & \in \mathcal{C}_i^* & i = 1 \ldots m
+\end{align}
+```
+
+## Unbounded models
+
+For an unbounded model, a certificate of infeasibility is an improving ray of
```
blegat, line 64: "or an unbounded model, a certificate of infeasibility" can be confusing as it's not clear whether we're certifying unboundness or infeasiblity. I would add something like that:

For an unbounded model, a certificate of unboundedness is made of two separate certificates:
1) A feasible solution `x0` certifying that the primal is feasible
2) A direction `d` that is both feasible for the homogenized feasible set and is a direction of decrease for the objective (in case of minimization).

Combining these two certificates, by convexity of the feasible set and linearity of the objective function, we can see that `x0 + \eta d` is feasible for any \eta >= 0 and can have an objective value arbitrarily small for increasing values of \eta.

The direction `d` is also a certificate of infeasibility of the dual problem.
In fact, most solvers stop when they have found `d` and do not search for a feasible solution `x0` so they terminate with a complete certificate of dual infeasibility but an incomplete certificate of primal unboundedness.
In that situation, the dual is known to be infeasible but the primal can still be either infeasible or unbounded.
The two cases can be distinguished by setting the `ObjectiveSense` to `FEASIBILITY_SENSE` and calling `optimize!` on the modified model.

docs/src/background/infeasibility_certificates.md
```diff
@@ -0,0 +1,113 @@
...
+ * [`TerminationStatus`](@ref) must be `DUAL_INFEASIBLE`
+ * [`PrimalStatus`](@ref) must be `INFEASIBILITY_CERTIFICATE`
+ * [`VariablePrimal`](@ref) must be the corresponding value of ``d``
+ * [`ObjectiveValue`](@ref) must be the value ``a_0^\top d``.
+
+## Infeasible models
+
+For an infeasible model, a certificate of infeasibility is an improving ray of
+the dual problem. However, because infeasibility is independent of the objective
+function, we first homogenize the primal problem by removing the objective.
```
blegat, line 88:
```suggestion
function, we first homogenize the primal problem by removing the primal objective.
```

blegat: It's unclear whether it's the objective of the primal or dual

docs/src/background/infeasibility_certificates.md
```diff
@@ -0,0 +1,113 @@
...
+``\eta > 0``:
+```math
+\begin{align}
+\sum_{i=1}^m A_i^\top (y_i + \eta d_i) & = 0 \\
+(y_i + \eta d_i) & \in \mathcal{C}_i^* & i = 1 \ldots m,
+\end{align}
+```
+and:
+```math
+\sum_{i=1}^m b_i^\top (y_i + \eta d_i) \le \sum_{i=1}^m b_i^\top y_i
```
blegat, line 100:
```suggestion
\sum_{i=1}^m b_i^\top (y_i + \eta d_i) < \sum_{i=1}^m b_i^\top y_i
```

docs/src/background/infeasibility_certificates.md
```diff
@@ -0,0 +1,113 @@
...
+```
+for any feasible dual solution ``y``. The latter simplifies to
+``\sum_{i=1}^m b_i^\top d_i \le 0``. Note that conic duality is defined such
```
blegat, line 103:
```suggestion
``\sum_{i=1}^m b_i^\top d_i < 0``. Note that conic duality is defined such
```

docs/src/background/infeasibility_certificates.md
```diff
@@ -0,0 +1,113 @@
...
+that this inequality is independent the problem being minmization or
```
blegat, line 104:
```suggestion
that this inequality is independent of whether the objective sense being minimization or
```

odow:
No tests yet. I started, but then I opened #1661. I'll create a new set of tests, rather than modifying the existing ones. I think some of the ObjectiveValue and DualObjectiveValue tests might fail.

docs/src/background/infeasibility_certificates.md
odow: Yeah I should just call it primal and dual infeasibility certificates and add a note about solving with no objective.

src/Test/test_infeasibility_certificates.jl
```diff
@@ -0,0 +1,53 @@
...
+                MOI.INFEASIBILITY_CERTIFICATE,
+            )
+            d = MOI.get(model, MOI.VariablePrimal(), x)
+            if $sense == MOI.MIN_SENSE
+                @test d < -config.atol
+            else
+                @test d > config.atol
+            end
+            @test(
+                isapprox(2.2 * d, MOI.get(model, MOI.ObjectiveValue()), config),
```
odow, line 26: @blegat, we've obviously never tested or tried this. GLPK and Gurobi both fail the ObjectiveValue test.

docs/src/background/infeasibility_certificates.md
```diff
@@ -0,0 +1,132 @@
...
+a_0^\top (x + \eta d) + b_0 < a_0^\top x + b_0,
+```
+for any feasible point ``x``. The latter simplifies to ``a_0^\top d < 0``. For
+maximization problems, the inequality is reversed, so that ``a_0^\top d > 0``.
+
+If the solver has found a certificate of dual infeasibility:
+
+ * [`TerminationStatus`](@ref) must be `DUAL_INFEASIBLE`
+ * [`PrimalStatus`](@ref) must be `INFEASIBILITY_CERTIFICATE`
+ * [`VariablePrimal`](@ref) must be the corresponding value of ``d``
```
odow, line 91: The open question from #433 is what to do with `ConstraintPrimal`. For consistency, it makes the most sense for it to be `A_i d` (i.e., without the constant term `b_i`).

However, I'm not sure if solvers actually support querying this, so it's effort for solver wrappers to implement.

Alternatively, we could just say that querying `ConstraintPrimal` for an infeasibility certificate is undefined behavior.

docs/src/background/infeasibility_certificates.md
```diff
@@ -0,0 +1,154 @@
...
+modulo a `-` sign.)
+
+If the solver has found a certificate of infeasibility:
+
+ * [`TerminationStatus`](@ref) must be `INFEASIBLE`
+ * [`DualStatus`](@ref) must be `INFEASIBILITY_CERTIFICATE`
+ * [`ConstraintDual`](@ref) must be the corresponding value of ``d``
+ * [`DualObjectiveValue`](@ref) must be the value
+   ``-\sum_{i=1}^m b_i^\top d_i`` for minimization problems and
+   ``\sum_{i=1}^m b_i^\top d_i`` for maximization problems.
```
odow, line 129: There's also a question about what to do here. Currently, the dual fallback returns different signs, so that's what I've left it as.

odow: I've convinced myself that this is correct.

odow:
Looked at updating GLPK and Gurobi. They're actually pretty easy to update.

Although I got bit bad (again) by this bug in Gurobi 9.1: https://github.com/jump-dev/Gurobi.jl/pull/363#discussion_r517025607. Sign errors are the worst to track down.

docs/src/background/infeasibility_certificates.md
odow: Note for myself tomorrow: it's actually easy to implement if we update the `get_fallback`, so there's no reason not to explicitly define `ConstraintPrimal` here as `A_i d`.

blegat: Yes, `get_fallback` is the way and the constant should not be there, see also https://github.com/jump-dev/MathOptInterface.jl/issues/1310.

odow: Okay done. I still need to add tests for the scalarize and vectorize bridges.

odow:
SCS and GLPK are passing these tests on their respective PRs, so I think we're good to go on my end.

src/Utilities/results.jl
```diff
@@ -178,11 +178,13 @@ function get_fallback(
...
-    # TODO do not include constant if primal solution is a ray
-    return eval_variables(
-        vi -> MOI.get(model, MOI.VariablePrimal(attr.result_index), vi),
-        f,
-    )
+    c = eval_variables(f) do vi
+        return MOI.get(model, MOI.VariablePrimal(attr.result_index), vi)
+    end
+    if is_ray(MOI.get(model, MOI.PrimalStatus()))
+        c -= MOI.constant(f, Int)
```
blegat, line 185:
```suggestion
        c -= MOI.constant(f, typeof(c))
```

src/Utilities/results.jl
```diff
@@ -19,11 +19,16 @@ the `ObjectiveFunction` value.
...
-    # TODO do not include constant if primal solution is a ray
-    return eval_variables(
+    obj = eval_variables(
         vi -> MOI.get(model, MOI.VariablePrimal(attr.result_index), vi),
         f,
     )
+    if is_ray(MOI.get(model, MOI.PrimalStatus()))
+        # Dual infeasibiltiy certificates do not include the primal objective
+        # constant.
+        obj -= MOI.constant(f, Int)
```
odow, line 29:
```suggestion
        obj -= MOI.constant(f, typeof(obj))
```

blegat (approved):

PR #1663, opened 2021-11-18
[doc] clarify documentation of solvers

### LINKED ISSUE #1633: Add solver READMEs to documentation
For JuMP 1.0, I think we should add the solver READMEs to the documentation. Something like this (I mocked it up for MOI, but it should be for JuMP): 
image
This requires a 1.0 tag for each solver first so the API is stable. We don't want to release JuMP with the documentation baked in, and then releases a breaking version of the solver so that the documentation is out-of-date.

In doing so, we should standardize the README for each solver. At the moment solver documentation is lacking, and things like setting options aren't uniformly well documented. If the solver has proper documentation that's okay, the README can just link to that.

odow:
Closes #1633

I don't think we want to include out-dated READMEs in the documentation, and I don't think we want to mandate what the README should look like. We can aim for internal consistency without explicitly documenting this.

### DIFF
--- a/docs/src/tutorials/implementing.md
+++ b/docs/src/tutorials/implementing.md
@@ -197,10 +197,10 @@ involved option is to use [Documenter.jl](https://github.com/JuliaDocs/Documente
 
 Examples of packages with README-based documentation include:
  * [Cbc.jl](https://github.com/jump-dev/Cbc.jl)
- * [Gurobi.jl](https://github.com/jump-dev/Gurobi.jl)
+ * [HiGHS.jl](https://github.com/jump-dev/HiGHS.jl)
  * [SCS.jl](https://github.com/jump-dev/SCS.jl)
 
-Examples of packages with Documenter.jl-based documentation include:
+Examples of packages with Documenter-based documentation include:
  * [Alpine.jl](https://github.com/lanl-ansi/Alpine.jl)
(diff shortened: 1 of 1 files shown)


docs/src/tutorials/implementing.md
```diff
@@ -189,6 +189,22 @@ Project.toml
...
 
+### Documentation
+
+Your package must include documentation explaining how to use the package. The
+easiest approach is to include documentation in your `README.md`. A more
+involved option is to use [Documenter.jl](https://github.com/JuliaDocs/Documenter.jl).
+
+Examples of packages with README-based documentation include:
+ * [Cbc.jl](https://github.com/jump-dev/Cbc.jl)
+ * [Gurobi.jl](https://github.com/jump-dev/Gurobi.jl)
```
joaquimg, line 200: why not highs instead of gurobi?

joaquimg (commented):
Agree with not mandating.

In the first version of a solver, it is very common to start with a  readme.
Since the suggestion is looking at those, I bet people will be copy-and-paste'ing. In this case a suggested template would be very nice. I could even be Cbc.jl or other existing solver with a very good readme.

blegat (approved):
I agree on having templates and enforce consistency. Mentioning this explicitly in the doc is even better

PR #1664, opened 2021-11-18
Remove source of common test failure in LowerBoundAlreadySet

### LINKED ISSUE #1431: Bridges fail test_model_LowerBoundAlreadySet test
```Julia
julia> using GLPK

julia> GLPK.MOI.Test.test_model_LowerBoundAlreadySet(
           GLPK.Optimizer(), MOI.Test.Config()
       )

julia> GLPK.MOI.Test.test_model_LowerBoundAlreadySet(
           MOI.Bridges.full_bridge_optimizer(GLPK.Optimizer(), Float64), 
           MOI.Test.Config(),
       )
Test Failed at /Users/oscar/.julia/dev/MathOptInterface/src/Test/test_model.jl:815
  Expression: MOI.add_constraint(model, f, set2)
    Expected: MathOptInterface.LowerBoundAlreadySet{MathOptInterface.Semicontinuous{Float64}, MathOptInterface.GreaterThan{Float64}}(MathOptInterface.VariableIndex(1))
  No exception thrown
ERROR: There was an error during testing
```

Same happens with `test_model_UpperBoundAlreadySet`, but I assume it's the same problem.

odow:
I considered a few ways of fixing #1431 (e.g., adding an Interval constraint so the variable would have a bound), but couldn't find anything satisfactory. The bridges need something like #1097, but that's a much greater fix.

This stops people running into the common source of failure, and it includes a broken test for us to fix in the future.

Closes #1431. I've opened #1665 to track the broader issue.

### DIFF
--- a/src/Test/test_model.jl
+++ b/src/Test/test_model.jl
@@ -801,10 +801,7 @@ function test_model_LowerBoundAlreadySet(
     @requires _supports(config, MOI.delete)
     x = MOI.add_variable(model)
     lb = zero(T)
-    sets = [
-        MOI.EqualTo(lb),
-        MOI.Interval(lb, lb),
-    ]
+    sets = [MOI.EqualTo(lb), MOI.Interval(lb, lb)]
     set2 = MOI.GreaterThan(lb)
     for set1 in sets
         if !MOI.supports_constraint(model, MOI.VariableIndex, typeof(set1))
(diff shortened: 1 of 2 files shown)


blegat (approved):
Sounds good, excluding these tests for all solvers is annoying.

PR #1666, opened 2021-11-18
Add checker for version being tested

joaquimg:
See: https://github.com/jump-dev/MathOptInterface.jl/pull/1662#issuecomment-972923430

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -13,6 +13,7 @@ OrderedCollections = "bac558e1-5e72-5ebc-8fee-abe8a469f55d"
 Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+TOML = "fa267f1f-6049-4f14-aa54-33bafae1ed76"
 Unicode = "4ec0a83e-493e-50e2-b9ac-8f72acf5a8f5"
 
 [compat]
--- a/src/Test/Test.jl
+++ b/src/Test/Test.jl
@@ -1,5 +1,6 @@
 module Test
 
+import TOML
 using MathOptInterface
 const MOI = MathOptInterface
 const MOIU = MOI.Utilities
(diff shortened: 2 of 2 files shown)


odow (commented):
I understand the motivation, but is it really that hard for solvers to just write:
```Julia
module TestPackage

# This constant limits tests in MOI to those added in MOI v0.10.5 and earlier.
#
# We hard-code a limit so that MOI can introduce tests for new features without
# breaking our current tests. To ensure that you run the latest tests, you
# should update this version periodically, or whenever you update MOI to support
# a new feature.
const _MOI_VERSION_TESTED = v"0.10.5"

using MyPackage
using Test

@testset "MOI tests" begin
    MOI.Test.runtests(
        optimizer,
        config,
        exclude_tests_after = _MOI_VERSION_TESTED
    )
end

end
```

They can make this constant nice and obvious. It's also going to be a common request to have a lower bound of `MOI 1` in the `Project.toml`, but want to run the tests up to some point. So just limiting to what the bound is in `Project.toml` adds complexity without being significantly better than the alternatives.

joaquimg:
I don't get why would someone set the lower bound of MOI to 1 in the Project.toml, but want to run the tests up to some earlier version < 1.

I believe the version_added functionality might lead to people not testing new MOI functionality by accident and have their code not completely tested.

If you really think it makes things worse checking inside `runtests`.

We can also consider:

```julia
module TestPackage

using MyPackage
using Test
using MOI

# This constant limits tests in MOI to those added in MOI v0.10.5 and earlier.
#
# We hard-code a limit so that MOI can introduce tests for new features without
# breaking our current tests. To ensure that you run the latest tests, you
# should update this version periodically, or whenever you update MOI to support
# a new feature.
# Pay special attention to keep the version tested in sync with Project.toml
const _MOI_VERSION_TESTED = v"0.10.5"
MOI.Test.check_version(MyPackage, _MOI_VERSION_TESTED)

@testset "MOI tests" begin
    MOI.Test.runtests(
        optimizer,
        config,
        exclude_tests_after = _MOI_VERSION_TESTED
    )
end

end
```

odow:
I meant 
```
MathOptInterface = "1"
```

And then
```
exclude_tests_after = v"1.4.2",
```

I.e., the package supports any 1.x release of MOI and does not use new features. In addition, it passes all tests as at v1.4.2.

odow:
> I believe the version_added functionality might lead to people not testing new MOI functionality by accident and have their code not completely tested.

Sure. But that's on them for opting into the bound. It should be pretty explicit for package maintainers. And it's not the end of the world if they don't run the latest tests. Because of backwards compatibility, everything will continue to work.

PR #1668, opened 2021-11-19
[Test] add generic number types in test_linear

odow:
Part of #841

### DIFF
--- a/src/Test/test_linear.jl
+++ b/src/Test/test_linear.jl
@@ -96,24 +96,33 @@ function test_linear_integration(
         MOI.optimize!(model)
         @test MOI.get(model, MOI.TerminationStatus()) == config.optimal_status
         @test MOI.get(model, MOI.PrimalStatus()) == MOI.FEASIBLE_POINT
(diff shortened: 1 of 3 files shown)


src/Test/test_linear.jl
```diff
@@ -190,7 +174,7 @@ function test_linear_integration(
     else
         MOI.delete(model, c)
         cf = MOI.ScalarAffineFunction{T}(
-            MOI.ScalarAffineTerm{T}.([one(T), one(T), one(T)], v),
+            MOI.ScalarAffineTerm{T}.(one(T), v),
```
joaquimg, line 177: `T(1)`
are we standardizing?

joaquimg (changes requested):
We should decide between `T(1)` and `one(T)`, same do zero.

src/Test/test_linear.jl
odow: I switched to `T(1)` and `T(0)`. We assume you can go `T(3)`, so we might as well go all-in.

PR #1672, opened 2021-11-21
[Utilities] fix fallback of map_indices

### LINKED ISSUE #1670: CachingOptimizer shouldn't map Model and Optimizer attributes
From @chriscoey on Gitter:

image

I don't think we should map `AbstractOptimizerAttribute`:
https://github.com/jump-dev/MathOptInterface.jl/blob/c34a59493ab84e1af9aed027ecf01b8a1630d371/src/Utilities/cachingoptimizer.jl#L919

odow:
Closes #1670.

I think the original argument was that we shouldn't have a fallback to force everyone to implement it. However, most people implementing this method just implement the fallback:
 - https://github.com/atoptima/BlockDecomposition.jl/blob/v1.4.0/src/axis.jl#L16
 - https://github.com/atoptima/BlockDecomposition.jl/blob/v1.4.0/src/tree.jl#L20

If people are writing new attributes and not accounting for this, that's on them for not testing correctly.

### DIFF
--- a/src/Utilities/cachingoptimizer.jl
+++ b/src/Utilities/cachingoptimizer.jl
@@ -143,7 +143,7 @@ function reset_optimizer(m::CachingOptimizer, optimizer::MOI.AbstractOptimizer)
             continue
         end
         value = MOI.get(m.model_cache, attr)
-        optimizer_value = map_indices(m.model_to_optimizer_map, value)
+        optimizer_value = map_indices(m.model_to_optimizer_map, attr, value)
         MOI.set(m.optimizer, attr, optimizer_value)
     end
     return
@@ -727,7 +727,7 @@ end
 
 function MOI.set(m::CachingOptimizer, attr::MOI.AbstractModelAttribute, value)
     if m.state == ATTACHED_OPTIMIZER
-        optimizer_value = map_indices(m.model_to_optimizer_map, value)
+        optimizer_value = map_indices(m.model_to_optimizer_map, attr, value)
         if m.mode == AUTOMATIC
             try
                 MOI.set(m.optimizer, attr, optimizer_value)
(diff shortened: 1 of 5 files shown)


blegat:
I'm not in favor of this change. It's quite tricky to test as the index map is the identity most of the time.
This will lead to bug that are quite hard to debug.
I would rather have a fallback that throws a helpful error message.

odow:
The flip side is that people have to extend a `MOI.Utilities` method, and there are type piracy concerns. 

@chriscoey wanted to pass an `OptimizerWithAttributes`. Okay, we can add that to MOI. But what if someone wants to pass an arbitrary `MOI.ModelLike`? Or some type defined in a third-party package?

We've implemented and tested the ones that matter in MOI.

blegat:
We can implement it for `OptimizerWithAttributes` in MOI.
Passing a `MOI.ModelLike` seems odd, it contains indices so some may need to be mapped.
Passing it unmapped may only make sense on the context of some attribute.
We could define
```julia
map_indices(f, attr::MOI.AnyAttribute, value) = map_indices(f, value)
```
The `CachingOptimizer` would call `map_indices` with 3 arguments.
If you define a custom attribute you can also safely add a method for `map_indices` with 3 arguments with no risk of type piracy.

odow:
I guess my point was that we can't control what types are passed as attributes. So we have two options:
 1. Add this fallback
 2. Force people to implement `Utilities.map_indices`

However, for (2): Package A might want to pass a type from a Package B, e.g., something from `Base` like a dictionary. The method can't live in MOI because we don't know about B (and even if we do, we don't want to add a method for every possible type). And it probably shouldn't (or can't) live in Package B because that's a heavy dependency.

That leaves us adding this fallback, and relying on tests to catch people mis-using the interface.

blegat:
> And it probably shouldn't (or can't) live in Package B because that's a heavy dependency.

If package A defines the attribute, it can define the 3 argument version mentioned in my previous comment

odow:
This doesn't work for `RawOptimizerAttribute`, which was @chriscoey's use-case in Pavito.

odow:
What is RawOptimizerAttribute passes everything though unchanged, and otherwise we hit the fallbacks. Then new solvers can use RawOptimizerAttribute, or define a new attribute and implement `map_indices` for it if necessary.

blegat:
> What is RawOptimizerAttribute passes everything though unchanged, and otherwise we hit the fallbacks. Then new solvers can use RawOptimizerAttribute, or define a new attribute and implement `map_indices` for it if necessary.

This looks like a local optimum. Also consistent with the `Raw` in the name ^^

blegat (approved):

PR #1674, opened 2021-11-21
[Utilities] use ConstraintPrimal fallback in CachingOptimizer

### LINKED ISSUE #1310: Automatic fallback in CachingOptimizer
We have fallbacks for a few attributes such as `ConstraintPrimal` but currently, they are only used by `MockOptimizer`.
When you have a `CachingOptimizer{Solver.Optimizer}`, a common use case is to do
```julia
try
    MOI.get(model, MOI.ConstraintPrimal(), ...)
catch
    MOI.get_fallback(model, MOI.ConstraintPrimal(), ...)
end
```
However, as mentioned in https://github.com/jump-dev/MathOptInterface.jl/issues/1249, relying on try-catch for such normal behavior is not ideal. Moreover, we might catch other errors that shouldn't be ignored.
We decided against having a `supports` for getters and `Solver.Optimizer` cannot call `get_fallback` because it does not implement `MOI.ConstraintFunction`. Maybe we could have a function `use_fallback` and in the caching optimizer, we call `use_fallback` first and if it returns `true` then we call `get_fallback` instead of `get`.

Then if a solver does not natively support giving the `ConstraintPrimal` nor the `ConstraintFunction`, it can implement `use_fallback(::Solver.Optimizer, ::MOI.ConstraintPrimal) = true`. It would then be able to pass the MOI tests needed `ConstraintPrimal`.

odow:
Closes #1310.

I think this is fine for the short-term. It's a continual pain-point for solvers (e.g., @chriscoey bumped into it when updating Pavito).

### DIFF
--- a/src/Utilities/cachingoptimizer.jl
+++ b/src/Utilities/cachingoptimizer.jl
@@ -887,13 +887,10 @@ end
 # doesn't support the attribute directly, we can use the fallback to query the
 # function from the cache and the variable value from the optimizer.
 
-_mapped_indices(map, index::MOI.ConstraintIndex) = map[index]
-_mapped_indices(map, indices::Vector) = _mapped_indices(Ref(map), indices)
-
 function MOI.get(
     model::CachingOptimizer,
     attr::MOI.ConstraintPrimal,
-    index::Union{Vector{<:MOI.ConstraintIndex},MOI.ConstraintIndex},
+    index::MOI.ConstraintIndex,
 )
     if state(model) == NO_OPTIMIZER
         error(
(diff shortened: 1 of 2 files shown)


src/Utilities/cachingoptimizer.jl
```diff
@@ -878,6 +879,58 @@ function MOI.get(
...
+        )
+    catch
+        return get_fallback(model, attr, index)
+    end
+end
+
+function MOI.get(
+    model::CachingOptimizer,
+    attr::MOI.ConstraintPrimal,
+    indices::Vector{<:MOI.ConstraintIndex}
```
odow, line 915:
```suggestion
    indices::Vector{<:MOI.ConstraintIndex},
```

blegat (approved):

PR #1684, opened 2021-11-24
Check result index in get_fallback of ConstraintPrimal

blegat:
Follow up of https://github.com/jump-dev/MathOptInterface.jl/pull/1682/

### DIFF
--- a/src/Utilities/cachingoptimizer.jl
+++ b/src/Utilities/cachingoptimizer.jl
@@ -901,7 +901,6 @@ function MOI.get(
             "optimizer is attached.",
         )
     end
-    MOI.check_result_index_bounds(model, attr)
     try
         return MOI.get(
             model.optimizer,
@@ -924,7 +923,6 @@ function MOI.get(
             "optimizer is attached.",
         )
     end
-    MOI.check_result_index_bounds(model, attr)
     try
         return MOI.get(
             model.optimizer,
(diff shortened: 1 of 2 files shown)


odow (commented):
I considered this, but I wasn't convinced it was the right approach.

Solvers using the `get_fallback` can call `check_result_index_bounds` in their method. It doesn't have to be in the `get_fallback`.

This is also inconsistent with the other `get_fallback` method which _don't_ check that the result is in-bounds. 

It also means for the caching optimizer method that we enter the try, potentially throw a result-index error, enter the catch, and then re-throw the error. 

I decided it was simpler just to check before the try-catch.

blegat:
> Solvers using the `get_fallback` can call `check_result_index_bounds` in their method. It doesn't have to be in the `get_fallback`.

Yes but we might as well do it once in `get_fallback` rather than everywhere it's called from.

> This is also inconsistent with the other `get_fallback` method which _don't_ check that the result is in-bounds. 

The other ones should be changed. They were written in a time we were not checking these result bounds yet.

> It also means for the caching optimizer method that we enter the try, potentially throw a result-index error, enter the catch, and then re-throw the error.

That's worrying. It could also catch an `InvalidIndex` error. Shouldn't the `catch` check that the error is what it expects and rethrow otherwise ?

PR #1685, opened 2021-11-24
Revert "[Utilties] don't throw UnsupportedAttribute in pass_attribute…

odow:
This reverts commit 1a3c637dafd17e9ea59e66a774a2778e1f6ba772.

This caused widespread failures in SolverTests for solvers using a caching optimizer.
https://github.com/jump-dev/MathOptInterface.jl/pull/1656#issuecomment-978211649

### DIFF
--- a/src/Utilities/copy.jl
+++ b/src/Utilities/copy.jl
@@ -25,6 +25,7 @@ function pass_attributes(
             if attr == MOI.Name()
                 continue  # Skipping names is okay.
             end
+            throw(MOI.UnsupportedAttribute(attr))
         end
         _pass_attribute(dest, src, index_map, attr)
     end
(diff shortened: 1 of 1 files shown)


blegat:
When you have a CachingOptimizer in EMPTY_OPTIMIZER state, I think the behavior of `set`ting an attribute unsupported by the optimize or `copy_to`ing a model with an attribute unsupported by the optimizer should be the same. With this PR, it will throw on the copy but not on `set`. Adding the check in https://github.com/jump-dev/MathOptInterface.jl/blob/b55203669c197eb2ac21d4d8dd9713087add8813/src/Utilities/cachingoptimizer.jl#L729 would throw it in both cases.
Note that if we do it for the attributes, we should do it for `add_constraint` too.

Another (simpler IMO) solution would be to modify the test into something like
```julia
@test_throws ... begin
    copy_to(...)
    optimize!(...)
end
```

odow:
Yeah okay. Maybe it's a bad test. We don't have a lot of consistency around when errors are thrown.

odow:
Closing in favor of #1686

Issue #1628, opened 2021-09-29
Define supports for is_set_by_optimize attributes


odow:
`supports` is currently restricted to copyable attributes. But it would also be useful to extend the definition to optimizer attributes so you can know whether to query the attribute without risking an error.

We've discussed this before, but I can't find the issue. I think it may have been a comment on a pull request.

Related: https://github.com/jump-dev/JuMP.jl/pull/2704#discussion_r718030897

odow:
I'm not convinced by this. I see it as too breaking because every solver will need to be updated.

odow:
@blegat this is our remaining blocker.

blegat:
Currently, we could require `supports` to be implemented and start testing it but not use it in `CachingOptimizer` yet. When we release v0.11 or v1.0, we start using it in `CachingOptimizer` in something like:
```julia
if has_fallback(attr) && !MOI.supports(model.optimizer, attr)
    return get_fallback(model, attr, ...)
else
    return MOI.get(model.optimizer, attr, ...)
end
```
For attributes such that `has_fallback` is false, it won't make any changes. So unless we test `MOI.supports` in the tests, the solvers won't notice the difference.

odow:
I think the options are:

1. Do nothing. Release MOI v1.0 which is non-breaking compared with 0.10
2. Release a breaking v0.11 change with tests and updated code
3. Release a breaking v1.0 change with tests and updated code

I guess it's a pretty trivial update for authors upgrading to MOI 1.0 anyway; they just have to add `MOI.supports(::Optimizer, ::MOI.VariablePrimal, ::MOI.VariableIndex)` etc.

It's a bit annoying because there are other attributes like `ConstraintFunction` and `ConstraintSet` which are neither copyable or set by optimize.

blegat:
For `ConstraintFunction`, it could be confusing since some solvers could support setting it but not getting it.

odow:
Monthly call says: Don't add it, but document the reasons why, and the work-arounds.

PR #1690, opened 2021-11-29
Define supports for is_set_by_optimize

odow:
Part of #1628 

We need to discuss this on the monthly call.

### DIFF
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -236,12 +236,12 @@ function supports_fallback(
     ::ModelLike,
     attr::Union{AbstractModelAttribute,AbstractOptimizerAttribute},
 )
-    if !is_copyable(attr)
+    if !is_copyable(attr) || !is_set_by_optimize(attr)
         throw(
             ArgumentError(
-                "`supports` is not defined for $attr, it is only" *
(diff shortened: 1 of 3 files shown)


src/attributes.jl
```diff
@@ -178,46 +178,54 @@ function Base.showerror(io::IO, err::ResultIndexBoundsError)
...
+   [`get`](@ref) with the attribute `attr`.
+
+No other attribute should define `supports`.
+
+If the attribute is only not supported in specific circumstances, `supports`
+must still return `true`.
+
+## Methods
+
+The type  and number of `args` depends on the type of `attr`. Here are the
```
blegat, line 198:
```suggestion
The type and number of `args` depends on the type of `attr`. Here are the
```

odow:
I took another look at this, and I'm less keen on it every time I delve into the code. 

I don't think we've seen a case where `try-catch` was a bottleneck. Adding this, and then relying on it, is going to be quite breaking. It's also going to need a lot of work through the bridges and the utilities to catch all of the edge cases.

odow:
Closing for now based on the latest monthly call, detailed in #1628

PR #1692, opened 2021-11-30
[Utilities] Fix method invalidation

odow:
The existing method was pretty terrible for inference.

### DIFF
--- a/src/Utilities/functions.jl
+++ b/src/Utilities/functions.jl
@@ -3190,7 +3190,22 @@ function Base.one(F::Type{<:TypedScalarLike{T}}) where {T}
     return convert(F, one(T))
 end
 
-Base.promote_rule(::Type{F}, ::Type{T}) where {T,F<:TypedScalarLike{T}} = F
+# !!! note
+#     These two `promote_rules` are written explicitly instead of
+#     `TypedScalarLike` in order to avoid a method invalidation.
+function Base.promote_rule(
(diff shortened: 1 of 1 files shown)


blegat:
So if I understand correctly:
```julia
Base.promote_rule(::Type{F}, ::Type{T}) where {T,F<:TypedScalarLike{T}} = F
```
is bad,
```julia
Base.promote_rule(::Type{F}, ::Type{T}) where {T,F<:MOI.ScalarAffineFunction{T}} = F
```
is bad but
```julia
Base.promote_rule(::Type{MOI.ScalarAffineFunction{T}}, ::Type{T}) where {T} = MOI.ScalarAffineFunction{T}
```
is good ?

odow:
The underlying issue is
```Julia
julia> Union{} <: MOI.Utilities.TypedScalarLike{Float64}
true
```
which conflicts with
```julia
julia> @which promote_type(Union{}, Float64)
promote_type(::Type{Union{}}, ::Type{T}) where T in Base at promotion.jl:224
```

odow:
Why do we even need these methods? Coverage didn't hit them.

blegat:
I see, I didn't think about `Union{}` ^^

> Why do we even need these methods? Coverage didn't hit them.

Maybe because they are inlined ? I don't remember why they were added though.

odow:
Let's see if anything breaks.

odow:
I guess that's a false negative on the codecov then!

Issue #1698, opened 2021-12-04
Bug in bridges with variables-constrained-on-creation


chriscoey:
I'm on latest JuMP and MOI. This occurred running MOIPajarito's soc2 test here https://github.com/chriscoey/MOIPajarito.jl/blob/336edcf24cb3ee18dddbfaa955e9071a562848c9/test/JuMP_tests.jl#L87.
```
_soc2: Error During Test at /home/coey/.julia/dev/MOIPajarito/test/JuMP_tests.jl:29
  Got exception outside of a @test
  In `MathOptInterface.ScalarAffineFunction{Float64}`-in-`MathOptInterface.EqualTo{Float64}` constraint: Constant 2.5 of the function is not zero. The function constant should be moved to the set. You can use `MOI.Utilities.normalize_and_add_constraint` which does this automatically.
  Stacktrace:
    [1] throw_if_scalar_and_constant_not_zero
      @ ~/.julia/packages/MathOptInterface/hcTsY/src/constraints.jl:95 [inlined]
    [2] bridged_constraint_function
      @ ~/.julia/packages/MathOptInterface/hcTsY/src/Bridges/bridge_optimizer.jl:1925 [inlined]
    [3] add_constraint(b::MathOptInterface.Bridges.LazyBridgeOptimizer{MathOptInterface.Utilities.CachingOptimizer{MOIPajarito.Optimizer, MathOptInterface.Utilities.UniversalFallback{MathOptInterface.Utilities.Model{Float64}}}}, f::MathOptInterface.ScalarAffineFunction{Float64}, s::MathOptInterface.EqualTo{Float64})
      @ MathOptInterface.Bridges ~/.julia/packages/MathOptInterface/hcTsY/src/Bridges/bridge_optimizer.jl:1533
    [4] bridge_constraint(#unused#::Type{MathOptInterface.Bridges.Constraint.ScalarSlackBridge{Float64, MathOptInterface.ScalarAffineFunction{Float64}, MathOptInterface.Integer}}, model::MathOptInterface.Bridges.LazyBridgeOptimizer{MathOptInterface.Utilities.CachingOptimizer{MOIPajarito.Optimizer, MathOptInterface.Utilities.UniversalFallback{MathOptInterface.Utilities.Model{Float64}}}}, f::MathOptInterface.ScalarAffineFunction{Float64}, s::MathOptInterface.Integer)
[...]

PR #1699, opened 2021-12-05
Fix logic in cost_of_bridging

odow:
@blegat should this be a + or a -? Part of #1698 

With the `-`:
```Julia
julia> solver = MOI.instantiate(
           MOI.OptimizerWithAttributes(
               MOIPajarito.Optimizer,
               "oa_solver" => MOI.OptimizerWithAttributes(GLPK.Optimizer),
               "conic_solver" => MOI.OptimizerWithAttributes(ECOS.Optimizer),
           );
           with_bridge_type = Float64
       );

julia> MOI.Utilities._cost_of_bridging(solver, MOI.GreaterThan{Float64})
(0.0, false)

julia> MOI.Utilities._cost_of_bridging(solver, MOI.Integer)
(0.0, false)
```
The cost for GreaterThan is `1 - 1` but the cost for `Integer` is `0 - 0`. Or perhaps we need to return a triplet of `(x - y, x + y, false)`.

### DIFF
--- a/src/Utilities/copy.jl
+++ b/src/Utilities/copy.jl
@@ -397,8 +397,8 @@ of bridges that will need to be used.
 
 In fact, if the order does matter (in the sense that changing the order of the
 vector returned by this function leads to a different formulation), it means the
-variable _is_ in at least in one other variable constraint. Thus, in a sense we
-could do `x - y + sign(x)`` but `!iszero(x), x - y` is fine
+variable is in at least one other variable constraint. Thus, in a sense we
+could do `x - y + sign(x)`` but `!iszero(x), x - y` is fine.
 
 ## Example
 
--- a/src/Utilities/copy.jl
+++ b/src/Utilities/copy.jl
@@ -374,11 +374,10 @@ The sorting happens in the `_cost_of_bridging` function and has three main
 considerations:
 
 1. First add sets for which the `VariableBridgingCost` is `0`. This ensures that
-   we minimize the number of variable bridges that get added. (Variable bridges
-   are bad because they produce an expression that needs substituting into the
-   remainder of the model.)
+   we minimize the number of variable bridges that get added.
 2. Then add sets for which the VariableBridgingCost is smaller than the
-   `ConstraintBridgingCost` (again avoiding variable bridges).
+   `ConstraintBridgingCost` so they can get added with
+   `add_constrained_variable(s)`.
 3. Finally, break any remaining ties in favor of `AbstractVectorSet`s. This
    ensures we attempt to add large blocks of variables (e.g., such as PSD
    matrices) before we add things like variable bounds.
--- a/src/Utilities/copy.jl
+++ b/src/Utilities/copy.jl
@@ -349,8 +349,7 @@ function _cost_of_bridging(
 ) where {S<:MOI.AbstractScalarSet}
     x = MOI.get(dest, MOI.VariableBridgingCost{S}())
     y = MOI.get(dest, MOI.ConstraintBridgingCost{MOI.VariableIndex,S}())
-    # In case of ties, we give priority to vector sets. See issue #987.
-    return (x - y, false, x + y)
+    return !iszero(x), x - y, true
 end
 
 function _cost_of_bridging(
@@ -359,15 +358,30 @@ function _cost_of_bridging(
 ) where {S<:MOI.AbstractVectorSet}
     x = MOI.get(dest, MOI.VariableBridgingCost{S}())
     y = MOI.get(dest, MOI.ConstraintBridgingCost{MOI.VectorOfVariables,S}())
-    # In case of ties, we give priority to vector sets. See issue #987
-    return (x - y, true, x + y)
+    return !iszero(x), x - y, false
 end
 
 """
     sorted_variable_sets_by_cost(dest::MOI.ModelLike, src::MOI.ModelLike)
 
-Returns a `Vector{Type}` of the set types corresponding to `VariableIndex` and
+Return a `Vector{Type}` of the set types corresponding to `VariableIndex` and
 `VectorOfVariables` constraints in the order in which they should be added.
+
+The order is important because some solvers require variables to be added in
+particular order, and the order can also impact the bridging decisions.
+
+The sorting happens in the `_cost_of_bridging` function and has three main
+considerations:
+
+1. First add sets for which the `VariableBridgingCost` is `0`. This ensures that
+   we minimize the number of variable bridges that get added. (Variable bridges
+   are bad because they produce an expression that needs substituting into the
+   remainder of the model.)
+2. Then add sets for which the VariableBridgingCost is smaller than the
+   `ConstraintBridgingCost` (again avoiding variable bridges).
+3. Finally, break any remaining ties in favor of `AbstractVectorSet`s. This
+   ensures we attempt to add large blocks of variables (e.g., such as PSD
+   matrices) before we add things like variable bounds.
 """
 function sorted_variable_sets_by_cost(dest::MOI.ModelLike, src::MOI.ModelLike)
     constraint_types = MOI.get(src, MOI.ListOfConstraintTypesPresent())
--- a/src/Utilities/copy.jl
+++ b/src/Utilities/copy.jl
@@ -367,8 +367,7 @@ end
 Return a `Vector{Type}` of the set types corresponding to `VariableIndex` and
 `VectorOfVariables` constraints in the order in which they should be added.
 
-The order is important because some solvers require variables to be added in
-particular order, and the order can also impact the bridging decisions.
+## How the order is computed
 
 The sorting happens in the `_cost_of_bridging` function and has three main
 considerations:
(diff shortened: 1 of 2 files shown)


blegat:
It should be `-`, it's checking what will be the difference of cost if they are added as constrained-on-creation vs added free and then constrained so the `-` takes this difference.

odow:
Okay, but we should probably prefer sets that don't require bridging over those that do? So we should distinguish `1 - 1` from `0 - 0`?

odow:
Because for Pajarito the cost of bridging `GreaterThan` is greater than `Integer`.

odow:
@blegat I'm not convinced of the current ordering of the sets. For example, the comment says "give priority to vector sets" but the ordering actually does the opposite (they come last).

One issue with the costs is that they don't take into account modified costs if we choose one of the variable bridges. i.e., the act of bridging one set can invalidate the other bridges that we would have chosen. In this case, if we bridge on `GreaterThan` first, then the cost of `Integer` is no longer `0`.

I think we need to construct some test cases first, and then modify the behavior to match the test-cases.
```Julia
struct Optimizer1698 <: MOI.AbstractOptimizer end

function MOI.supports_constraint(
    ::Optimizer1698,
    ::Type{MOI.VariableIndex},
    ::Type{MOI.Integer},
)
    return true
end

function MOI.supports_constraint(
    ::Optimizer1698,
    ::Type{<:Union{MOI.VectorOfVariables,MOI.VectorAffineFunction{Float64}}},
    ::Type{<:Union{MOI.Nonnegatives,MOI.SecondOrderCone}},
)
    return true
end

function test_sorted_variable_sets_by_cost()
    src = MOI.Utilities.Model{Float64}()
    x = MOI.add_variable(src)
    y = MOI.add_variables(src, 2)
    MOI.add_constraint(src, x, MOI.GreaterThan(1.0))
    MOI.add_constraint(src, x, MOI.Integer())
    MOI.add_constraint(src, y, MOI.SecondOrderCone(2))
    dest = MOI.Bridges.full_bridge_optimizer(Optimizer1698(), Float64)
    @test MOI.Utilities.sorted_variable_sets_by_cost(dest, src) == [
        MOI.Integer,
        MOI.GreaterThan{Float64},
        MOI.SecondOrderCone,
    ]
    return
end
```

What about this one?
```julia
struct Optimizer1698_2 <: MOI.AbstractOptimizer end

function MOI.supports_constraint(
    ::Optimizer1698_2,
    ::Type{<:Union{MOI.VectorOfVariables,MOI.VectorAffineFunction{Float64}}},
    ::Type{<:Union{MOI.Nonnegatives,MOI.Nonpositives}},
)
    return true
end

function test_sorted_variable_sets_by_cost()
    src = MOI.Utilities.Model{Float64}()
    MOI.add_constrained_variables(src, MOI.Nonnegatives(2))
    MOI.add_constrained_variables(src, MOI.Zeros(2))
    dest = MOI.Bridges.full_bridge_optimizer(Optimizer1698_2(), Float64)
    @test MOI.Utilities.sorted_variable_sets_by_cost(dest, src) == [
        MOI.Nonnegatives,
        MOI.Zeros,
    ]
    return
end
```
I think it makes sense for `Nonnegatives` to come first. But at the moment `Zeros` does.

odow:
@chriscoey you can try MOIPajarito with this branch.

```Julia
julia> using JuMP, GLPK, ECOS, MOIPajarito

julia> solver = MOI.OptimizerWithAttributes(
           MOIPajarito.Optimizer,
           "oa_solver" => MOI.OptimizerWithAttributes(GLPK.Optimizer),
           "conic_solver" => MOI.OptimizerWithAttributes(ECOS.Optimizer),
       )
MathOptInterface.OptimizerWithAttributes(MOIPajarito.Optimizer, Pair{MathOptInterface.AbstractOptimizerAttribute, Any}[MathOptInterface.RawOptimizerAttribute("oa_solver") => MathOptInterface.OptimizerWithAttributes(GLPK.Optimizer, Pair{MathOptInterface.AbstractOptimizerAttribute, Any}[]), MathOptInterface.RawOptimizerAttribute("conic_solver") => MathOptInterface.OptimizerWithAttributes(ECOS.Optimizer, Pair{MathOptInterface.AbstractOptimizerAttribute, Any}[])])

julia> model = Model(solver)
A JuMP Model
Feasibility problem with:
Variables: 0
Model mode: AUTOMATIC
Solver name: Pajarito

julia> @variable(model, z >= 1, Int)
z

julia> optimize!(model)
no conic constraints need outer approximation
OA solver finished with status OPTIMAL, after 0.13000798225402832 seconds and 0 cuts
one tree method used 0 lazy callbacks and 0 heuristic callbacks
```

blegat:
Indeed, there is something I missed when writing this cost. For `Integer`, it is natively supported both if it's constrained on creation or not so it seems it will never need any bridging but it can also be forced-bridged into ScalarAffineFunction if the variable is constrained on creation.
It seems to me that we should first give priority to the variables such that constraining on creation is natively supported to try to have no variable bridge.
So the order should be
1. `Bool` equal to `supports_add_constrained_variable(s)` giving priority to `true`
2. Difference of bridge cost
3. `Bool` indicating whether it is vector, giving priority to `true`

> For example, the comment says "give priority to vector sets" but the ordering actually does the opposite (they come last).

Indeed, I'm surprised, I thought it was tested. This should be fixed.

odow:
That order makes sense.

odow:
How's this?

src/Utilities/copy.jl
```diff
@@ -359,15 +358,30 @@ function _cost_of_bridging(
...
+particular order, and the order can also impact the bridging decisions.
+
+The sorting happens in the `_cost_of_bridging` function and has three main
+considerations:
+
+1. First add sets for which the `VariableBridgingCost` is `0`. This ensures that
+   we minimize the number of variable bridges that get added. (Variable bridges
+   are bad because they produce an expression that needs substituting into the
+   remainder of the model.)
+2. Then add sets for which the VariableBridgingCost is smaller than the
```
blegat, line 380: > again avoiding variable bridges

This is not the reason. You just count the difference of bridges in both situation, this difference is taking each type of bridges with equal weight and importance

src/Utilities/copy.jl
```diff
@@ -359,15 +358,30 @@ function _cost_of_bridging(
...
+
+The order is important because some solvers require variables to be added in
+particular order, and the order can also impact the bridging decisions.
+
+The sorting happens in the `_cost_of_bridging` function and has three main
+considerations:
+
+1. First add sets for which the `VariableBridgingCost` is `0`. This ensures that
+   we minimize the number of variable bridges that get added. (Variable bridges
+   are bad because they produce an expression that needs substituting into the
```
blegat, line 378: > Variable bridges are bad because they produce an expression that needs substituting into the remainder of the model

What does "bad" mean ? It's a bit subjective and it's unclear why substitution is bad (it's not adding any additional variable or constraint so the reader why thing instead that it's quite good).
I would say that we give priority to passing the model as-is, without transformation. Even if the `ConstraintBridgingCost` is zero as well, the constraint will need to be forced-bridged into affine-in-set

odow: I removed the sentence.

blegat:
I'd like something along these lines to be either in the docstring or a comment in the code as I feel I will forget about that subtlety but I can add in a separate PR:

> We favor adding first variables that won't use variables bridges because then the variable constraints on the same variable can still be added as `VariableIndex` or `VectorOfVariables`.
> For the variable that do need variable bridges. If they are part of other variable constraint then they will be force-bridged into affine constraints so there is a hidden cost in terms of number of additional number of bridges that will need to be used. In fact, if the order matter, it means the variable is at least in one other variable constraint so in a sense we could do `x - y + sign(x)` but I think `!iszero(x), x - y` is fine.

blegat (approved):

odow:
I added a big chunk to the docstring, including the Pajarito example. An open research question for MOI 2.0 could be how to update the bridging costs in the presence of variable bridges. Getting that right would fix all of this.

src/Utilities/copy.jl
```diff
@@ -381,6 +380,54 @@ considerations:
...
+`VariableIndex` or `VectorOfVariables` constraints.
+
+If a variable does need variable bridges and is part of another variable
+constraint, then the other variable constraint will be force-bridged into affine
+constraints, so there is a hidden cost in terms of number of additional number
+of bridges that will need to be used.
+
+In fact, if the order does matter (in the sense that changing the order of the
+vector returned by this function leads to a different formulation), it means the
+variable _is_ in at least in one other variable constraint. Thus, in a sense we
```
blegat, line 400:
```suggestion
variable _is_ at least in one other variable constraint. Thus, in a sense we
```

src/Utilities/copy.jl
```diff
@@ -381,6 +380,54 @@ considerations:
...
+could do `x - y + sign(x)`` but `!iszero(x), x - y` is fine
```
blegat, line 401:
```suggestion
could do `x - y + sign(x)`` but `!iszero(x), x - y` is fine.
```

blegat (approved):

PR #1703, opened 2021-12-13
Add fallback error for compute_conflict

### LINKED ISSUE #1697: Add fallback for compute_conflict!
If you call `compute_conflict!` on an unsupported solver you currently get a method error:

https://github.com/jump-dev/MathOptInterface.jl/blob/8075edf61497ad8c74f1eb9fda5e927e5788b9c6/src/MathOptInterface.jl#L85-L103

odow:
Closes #1697

### DIFF
--- a/test/errors.jl
+++ b/test/errors.jl
@@ -320,7 +320,7 @@ function test_compute_conflict_fallback()
     model = Optimizer1697()
     @test_throws(
         ArgumentError(
-            "The optimizer $(typeof(optimizer)) does not support " *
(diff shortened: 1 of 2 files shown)


test/errors.jl
```diff
@@ -314,6 +314,20 @@ function test_errors_copy_to_fallback()
     return
 end
 
+struct Optimizer1697 <: MOI.AbstractOptimizer end
```
joaquimg, line 317: 1697?

odow: It's the issue number.

PR #1712, opened 2022-01-05
[Utilities] simplify MutableSparseMatrix to remove final_touch

### LINKED ISSUE #1711: Sparse conversion broken on julia 1.7.1 for empty MatrixOfConstraints
On julia 1.7.1,

The following code fails

```julia
using MathOptInterface
MOI = MathOptInterface

MOI.Utilities.@product_of_sets(Zeros, MOI.Zeros)

MOI.Utilities.@product_of_sets(Nonpositives, MOI.Nonpositives)

MOI.Utilities.@struct_of_constraints_by_set_types(
    StructCache,
    MOI.Zeros,
    MOI.Nonpositives,
)

OptimizerCache{T} = MOI.Utilities.GenericModel{
    T,
    MOI.Utilities.ObjectiveContainer{T},
    MOI.Utilities.VariablesContainer{T},
    StructCache{T}{
        MOI.Utilities.MatrixOfConstraints{
            T,
            MOI.Utilities.MutableSparseMatrixCSC{
                T,
                Int64,
                MOI.Utilities.OneBasedIndexing,
            },
            Vector{T},
            Zeros{T},
        },
        MOI.Utilities.MatrixOfConstraints{
            T,
            MOI.Utilities.MutableSparseMatrixCSC{
                T,
                Int64,
                MOI.Utilities.OneBasedIndexing,
            },
            Vector{T},
            Nonpositives{T},
        },
    },
}

model = MOI.Utilities.CachingOptimizer(
    MOI.Utilities.UniversalFallback(MOI.Utilities.Model{Float64}()),
    MOI.Utilities.AUTOMATIC,
)

x = MOI.add_variable(model)

cache = OptimizerCache{Float64}()
index_map = MOI.copy_to(cache, model)

Ab = MOI.Utilities.constraints(
    cache.constraints,
    MOI.VectorAffineFunction{Float64},
    MOI.Zeros,
)
A = Ab.coefficients
Gh = MOI.Utilities.constraints(
    cache.constraints,
    MOI.VectorAffineFunction{Float64},
    MOI.Nonpositives,
)
G = Gh.coefficients

@show G

using SparseArrays
convert(SparseMatrixCSC{Float64, Int}, G)
```

with

```julia
ERROR: ArgumentError: Illegal buffers for SparseMatrixCSC construction 1 [0, 0] Int64[] Float64[]
Stacktrace:
 [1] SparseMatrixCSC
   @ C:\Users\joaquimgarcia\AppData\Local\Programs\Julia-1.7.1\share\julia\stdlib\v1.7\SparseArrays\src\sparsematrix.jl:29 [inlined]
 [2] convert(#unused#::Type{SparseMatrixCSC{Float64, Int64}}, A::MathOptInterface.Utilities.MutableSparseMatrixCSC{Float64, Int64, MathOptInterface.Utilities.OneBasedIndexing})
   @ MathOptInterface.Utilities C:\Users\joaquimgarcia\.julia\dev\MathOptInterface\src\Utilities\sparse_matrix.jl:177
 [3] top-level scope
   @ REPL[18]:1
```

(works fine in Julia 1.6.4)

A `final_touch` is missing somewhere.
Just adding it here: https://github.com/jump-dev/MathOptInterface.jl/blob/dfc7c613ce5193f3ae872921668bd0f267f37c60/src/Utilities/struct_of_constraints.jl#L314
breaks other stuff.

odow:
Closes #1711 

The need for `final_touch` is just a recipe for bugs like this. We should endeavor to make it that we don't need to call `final_touch` for most things (now I think it's only the sets).

### DIFF
--- a/src/Utilities/sparse_matrix.jl
+++ b/src/Utilities/sparse_matrix.jl
@@ -27,8 +27,8 @@ allocation-free conversion of [`MutableSparseMatrixCSC`](@ref) to
 """
 struct OneBasedIndexing <: AbstractIndexing end
 
-_first_index(::ZeroBasedIndexing) = 0
-_first_index(::OneBasedIndexing) = 1
+_first_index(::Type{T}, ::ZeroBasedIndexing) where {T} = T(0)
+_first_index(::Type{T}, ::OneBasedIndexing) where {T} = T(1)
 
 _shift(x, ::T, ::T) where {T<:AbstractIndexing} = x
 _shift(x::Integer, ::ZeroBasedIndexing, ::OneBasedIndexing) = x + 1
@@ -65,8 +65,11 @@ mutable struct MutableSparseMatrixCSC{Tv,Ti<:Integer,I<:AbstractIndexing}
     colptr::Vector{Ti}
     rowval::Vector{Ti}
     nzval::Vector{Tv}
+    nz_added::Vector{Ti}
     function MutableSparseMatrixCSC{Tv,Ti,I}() where {Tv,Ti<:Integer,I}
-        return new{Tv,Ti,I}(I(), 0, 0, [zero(Ti)], Ti[], Tv[])
+        indexing = I()
+        colptr = [_first_index(Ti, indexing)]
+        return new{Tv,Ti,I}(indexing, 0, 0, colptr, Ti[], Tv[], Ti[])
     end
 end
 
@@ -80,43 +83,39 @@ function MOI.empty!(A::MutableSparseMatrixCSC{Tv,Ti}) where {Tv,Ti}
     A.m = 0
     A.n = 0
     resize!(A.colptr, 1)
-    A.colptr[1] = zero(Ti)
+    A.colptr[1] = _first_index(Ti, A.indexing)
     empty!(A.rowval)
     empty!(A.nzval)
+    empty!(A.nz_added)
     return
 end
 
 function add_column(A::MutableSparseMatrixCSC{Tv,Ti}) where {Tv,Ti}
     A.n += 1
-    push!(A.colptr, zero(Ti))
+    push!(A.colptr, _first_index(Ti, A.indexing))
+    push!(A.nz_added, Ti(0))
     return
 end
 
 function add_columns(A::MutableSparseMatrixCSC{Tv,Ti}, n) where {Tv,Ti}
-    A.n += n
     for _ in 1:n
-        push!(A.colptr, zero(Ti))
+        add_column(A)
     end
     return
 end
 
 function set_number_of_rows(A::MutableSparseMatrixCSC, num_rows)
     A.m = num_rows
+    offset = _first_index(Int, A.indexing)
     for i in 3:length(A.colptr)
-        A.colptr[i] += A.colptr[i-1]
+        A.colptr[i] += A.colptr[i-1] - offset
     end
-    resize!(A.rowval, A.colptr[end])
-    resize!(A.nzval, A.colptr[end])
+    resize!(A.rowval, A.colptr[end] - offset)
+    resize!(A.nzval, A.colptr[end] - offset)
     return
 end
 
-function final_touch(A::MutableSparseMatrixCSC)
-    for i in length(A.colptr):-1:2
-        A.colptr[i] = _shift(A.colptr[i-1], ZeroBasedIndexing(), A.indexing)
-    end
-    A.colptr[1] = _first_index(A.indexing)
-    return
-end
+final_touch(::MutableSparseMatrixCSC) = nothing
 
 _variable(term::MOI.ScalarAffineTerm) = term.variable
 _variable(term::MOI.VectorAffineTerm) = _variable(term.scalar_term)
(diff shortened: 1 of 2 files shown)


src/Utilities/sparse_matrix.jl
```diff
@@ -65,8 +65,11 @@ mutable struct MutableSparseMatrixCSC{Tv,Ti<:Integer,I<:AbstractIndexing}
     colptr::Vector{Ti}
     rowval::Vector{Ti}
     nzval::Vector{Tv}
+    nz_added::Vector{Ti}
```
joaquimg, line 68: add this field to docstring

joaquimg (approved):

odow:
Let's get @blegat to take a look

blegat:
I would prefer the same approach as with `MatrixOfConstraints`. A `final_touch` boolean and `convert` checks that it has been called.
It could even be more than a boolean. There is three states: allocating, loading and done.
And you have the transitions: allocating -`set_number_of_rows`-> loading -`final_touch`-> done -`MOI.empty!`-> allocating.
We could have an enum with the three states and always check that we are on the correct state.

joaquimg:
If the operations must follow a precise order I agree with an enum and `@assert`s to guarantee the order.
I like the new implementation with a no-op `final_touch`, am I missing any performance bottleneck?

I imagine that we can add the enum as extra protection and keep the current no-op `final_touch`. To get the best of both worlds.

About checking, I agree its fine to check `final_touch` in `convert` as extra protection, but actually relying on that would be worse. Or we should at least document that the user should always call `convert` or an `is_done` thing.

test/Utilities/sparse_matrix.jl
```diff
@@ -203,6 +218,11 @@ function test_ScalarAffine_OneBased()
     @test A.rowval == [1, 1, 2, 1, 2]
     @test A.nzval == [1.0, 1.0, 2.0, 1.0, 3.0]
     @test A.colptr == [1, 2, 4, 6]
+    MOI.empty!(A)
```
blegat, line 221: Remove the `final_touch` in the test above as well

blegat:
The extra vector could also be used to do sanity checks at `final_touch` if we want to check that the same amont that was allocated was also loaded but that can be left for another PR.

odow:
I dug into this more. The problem is our lazy creation of the constraint objects:
```julia
julia> model = MOI.Utilities.CachingOptimizer(
           MOI.Utilities.UniversalFallback(MOI.Utilities.Model{Float64}()),
           MOI.Utilities.AUTOMATIC,
       )
MOIU.CachingOptimizer{MOI.AbstractOptimizer, MOIU.UniversalFallback{MOIU.Model{Float64}}}
in state NO_OPTIMIZER
in mode AUTOMATIC
with optimizer nothing

julia> cache = OptimizerCache{Float64}()
MOIU.GenericModel{Float64, MOIU.ObjectiveContainer{Float64}, MOIU.VariablesContainer{Float64}, StructCache{Float64, MOIU.MatrixOfConstraints{Float64, MOIU.MutableSparseMatrixCSC{Float64, Int64, MOIU.OneBasedIndexing}, Vector{Float64}, Zeros{Float64}}, MOIU.MatrixOfConstraints{Float64, MOIU.MutableSparseMatrixCSC{Float64, Int64, MOIU.OneBasedIndexing}, Vector{Float64}, Nonpositives{Float64}}}}

julia> index_map = MOI.copy_to(cache, model)
MathOptInterface.Utilities.IndexMap with 1 entry:
  VariableIndex(1) => VariableIndex(1)

julia> cache.constraints.moi_nonpositives

```
At this point, the field is still `nothing`. It's only when we access the field that it gets created:
```Julia
julia> Gh = MOI.Utilities.constraints(
           cache.constraints,
           MOI.VectorAffineFunction{Float64},
           MOI.Nonpositives,
       )
MOIU.MatrixOfConstraints{Float64, MOIU.MutableSparseMatrixCSC{Float64, Int64, MOIU.OneBasedIndexing}, Vector{Float64}, Nonpositives{Float64}}

julia> G = Gh.coefficients
MathOptInterface.Utilities.MutableSparseMatrixCSC{Float64, Int64, MathOptInterface.Utilities.OneBasedIndexing}(MathOptInterface.Utilities.OneBasedIndexing(), 0, 1, [1, 1], Int64[], Float64[], [0])

julia> cache.constraints.moi_nonpositives
MOIU.MatrixOfConstraints{Float64, MOIU.MutableSparseMatrixCSC{Float64, Int64, MOIU.OneBasedIndexing}, Vector{Float64}, Nonpositives{Float64}}

```
This is now out-of-sync because we never got to call `final_touch` on it.

test/Utilities/sparse_matrix.jl
odow: It's still useful to leave it in case we add something back (and to check that it really does nothing).

odow:
I thought the point was that it didn't make sense to modify these after `final_touch`? We also have a working fix in this PR. No need for other fixes.

blegat:
Currently, we don't have any model type for which `final_touch` is not a no-op and modification is supported, indeed it's a bit counter-intuitive.

blegat (approved):

PR #1713, opened 2022-01-06
Add LP Base.read! and tests

### LINKED ISSUE #1675: Implement read! for LP file format
https://github.com/jump-dev/MathOptInterface.jl/blob/c34a59493ab84e1af9aed027ecf01b8a1630d371/src/FileFormats/LP/LP.jl#L289-L291

guilhermebodin:
This implementation is some copy from https://github.com/odow/LPWriter.jl/blob/master/src/reader.jl

There are still a few missing important features. 
- [x] Obj constant
- [x] SOS variables

I will probably implement them next week. 

I have some questions regarding the files we use as test sets. Do you have any suggestions?

Closes #1675

### DIFF
--- a/src/FileFormats/LP/LP.jl
+++ b/src/FileFormats/LP/LP.jl
@@ -351,7 +351,7 @@ function _verify_name(variable::String, maximum_length::Int)
     #     return false
     # end
     m = match(NAME_REG, variable)
-    if !isnothing(m)
+    if m !== nothing
         return false
     end
     return true
--- a/src/FileFormats/LP/LP.jl
+++ b/src/FileFormats/LP/LP.jl
@@ -321,14 +321,14 @@ const _KEYWORDS = Dict(
 )
 
 const _SENSE_ALIAS = Dict(
-    "max" => :Max,
-    "maximize" => :Max,
-    "maximise" => :Max,
-    "maximum" => :Max,
-    "min" => :Min,
-    "minimize" => :Min,
-    "minimise" => :Min,
-    "minimum" => :Min,
+    "max" => MOI.MAX_SENSE,
+    "maximize" => MOI.MAX_SENSE,
+    "maximise" => MOI.MAX_SENSE,
+    "maximum" => MOI.MAX_SENSE,
+    "min" => MOI.MIN_SENSE,
+    "minimize" => MOI.MIN_SENSE,
+    "minimise" => MOI.MIN_SENSE,
+    "minimum" => MOI.MIN_SENSE,
 )
 
 const _SUBJECT_TO_ALIAS = ["subject to", "such that", "st", "s.t."]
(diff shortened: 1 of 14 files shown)


odow:
Just use my tests here? It includes some annoying files like https://github.com/odow/LPWriter.jl/blob/master/test/model1_tricky.lp

src/FileFormats/LP/LP.jl
```diff
@@ -285,9 +285,429 @@ end
 #   Base.read!
 #
 # ==============================================================================
+const COMMENT_REG = r"(.*?)\\(.*)"
```
odow, line 288:
```suggestion
const _COMMENT_REG = r"(.*?)\\(.*)"
```

I'm going to be a bit pedantic on some naming (but I'll only comment on a couple). I know that the old code didn't have it. 

* https://jump.dev/JuMP.jl/dev/developers/style/#Naming
* https://jump.dev/JuMP.jl/dev/developers/style/#Exported-and-non-exported-names

src/FileFormats/LP/LP.jl
```diff
@@ -285,9 +285,429 @@ end
...
+function stripcomment(line::String)
```
odow, line 289:
```suggestion
function _strip_comment(line::String)
```

src/FileFormats/LP/LP.jl
```diff
@@ -285,9 +285,429 @@ end
...
+    if occursin("\\", line)
+        m = match(COMMENT_REG, line)
+        return strip(String(m[1]))
+    else
+        return strip(line)
+    end
+end
+
+# a list of section keywords in lower-case
+const KEYWORDS = Dict(
```
odow, line 299:
```suggestion
const _KEYWORDS = Dict(
```

src/FileFormats/LP/LP.jl
```diff
@@ -285,9 +285,429 @@ end
...
+    # end
+    m = match(NAME_REG, variable)
+    if !isnothing(m)
+        return false
+    end
+    return true
+end
+
+# TODO Obj constant and SOS Variables
+mutable struct TempLPModel
```
odow, line 366: It's useful to ask whether we actually need this. Can't we just write into the `::Model` directly? I get that it's a much bigger rewrite, so perhaps we do this first to get the function and tests all sorted. Then we can fix it in another PR.

src/FileFormats/LP/LP.jl
```diff
@@ -285,9 +285,429 @@ end
...
+        end
+        if haskey(KEYWORDS, lowercase(line)) # section has changed
+            section = KEYWORDS[lowercase(line)]
+            setsense!(section, data, line)
+            continue
+        end
+        parsesection!(section, data, line)
+    end
+    copy_to(model, data)
+    return data
```
odow, line 710:
```suggestion
    return
```

odow (commented):
Nice!

Most of these tests should be applicable:
https://github.com/odow/LPWriter.jl/blob/master/test/reader.jl

src/FileFormats/LP/LP.jl
```diff
@@ -285,9 +285,545 @@ end
...
+function _verify_name(variable::String, maximum_length::Int)
+    if length(variable) > maximum_length
+        return false
+    end
+    # m = match(START_REG, variable)
+    # if !isnothing(m)
+    #     return false
+    # end
+    m = match(NAME_REG, variable)
+    if !isnothing(m)
```
odow, line 354:
```suggestion
    if m !== nothing
```

src/FileFormats/LP/LP.jl
```diff
@@ -285,9 +285,545 @@ end
...
+    push!(data.col_upper, Inf)
+    push!(data.c, 0)
+    push!(data.colcat, :Cont)
+    push!(data.col_to_name, name)
+    return
+end
+
+function _get_variable_index!(data::TempLPModel, variable::String)
+    i = findfirst(isequal(variable), data.col_to_name)
+    if isnothing(i)
```
odow, line 420:
```suggestion
    if i === nothing
```

src/FileFormats/LP/LP.jl
```diff
@@ -285,9 +285,545 @@ end
...
+    return i
+end
+
+function _tokenize(line::AbstractString)
+    items = String.(split(line, " "))
+    return items[items.!=""]
+end
+
+function _parse_float(val::AbstractString)
+    if lowercase(val) == "-inf" || lowercase(val) == "-infinity"
```
odow, line 436: Not necessary just yet, but I saw in the MPS reader that calling `lowercase` was actually a significant bottleneck. We might want to call `lowercase(val)` once, rather than calling it 4 times (I don't think the compiler is smart enough to figure out it's the same computation with no side-effects).

test/FileFormats/LP/LP.jl
```diff
@@ -236,8 +236,74 @@ end
 
 function test_read()
     model = LP.Model()
-    exception = ErrorException("read! is not implemented for LP files.")
-    @test_throws exception MOI.read_from_file(model, LP_TEST_FILE)
+    MOI.read_from_file(model, joinpath(@__DIR__, "models", "example_lo1.lp"))
```
odow, line 239: Make each of these a separate test function, rather than one giant test.

src/FileFormats/LP/LP.jl
guilhermebodin: I think that is probably the best way. Maybe we should also test if it is slow enough to encourage the rewrite.

odow: Yeah a good first step is what you have at the moment. Get all the tests and plumbing working, then focus on performance.

odow:
@guilhermebodin what's the status of this? Do you want me to merge this as-is? Or are you going to benchmark the performance first?

guilhermebodin:
Hi @odow, since we already have a working version, I think it is worth benchmarking this solution and compare with a new one without the `TempLPModel`. After revisiting the code, I believe there are no significant design changes, so it should not be an extensive rewrite of the current PR.

odow:
Okay. So I'll hold off merging this then? Or do you want to work on that in a different PR?

guilhermebodin:
I was not clear in the previous message, let's hold this PR for a couple days.

src/FileFormats/LP/LP.jl
```diff
@@ -285,9 +285,472 @@ end
...
+
+function _add_new_variable!(model::Model, name::String)
+    var = MOI.add_variable(model)
+    MOI.set(model, MOI.VariableName(), var, name)
+    return var
+end
+
+function _get_variable_from_name(model::Model, variable_name::String)
+    variables_inside_model = MOI.get(model, MOI.ListOfVariableIndices())
+    for var_inside_model in variables_inside_model
```
odow, line 392: It might be worth constructing some benchmarks so you can profile this reader. This function is going to be very expensive. We should probably cache a dictionary mapping names to variable indices.

guilhermebodin:
I have changed a few definitions using the file generated with this code
```julia
model = Model()
@variable(model, x[1:1_000] >= 0)
@constraint(model, con[i=1:200], sum(x) >= i)
@objective(model, Min, sum(x))

io = open("test_big.lp", "w")
write(
    io,
    model;
    format= MOI.FileFormats.FORMAT_LP
)
```

The modifications that made the biggest effect were in the function `_tokenize` and `_get_variable_from_name `.

src/FileFormats/LP/LP.jl
```diff
@@ -372,6 +373,7 @@ mutable struct CacheLPModel
             false,
             "",
             0,
+            Dict{MOI.VariableIndex,String}(),
```
odow, line 376: This is not the same as the field you added.

src/FileFormats/LP/LP.jl
```diff
@@ -381,31 +383,31 @@ function _set_sense!(::Type{Val{:obj}}, model::Model, line)
...
-        if var_inside_model_name == variable_name
-            return var_inside_model
-        end
+function _get_variable_from_name(
+    model::Model,
+    data_cache::CacheLPModel,
+    variable_name,
+)
+    var_inside_model = get(data_cache.variables_in_model, variable_name, "")
+    if var_inside_model != ""
```
odow, line 399:
```suggestion
    if !isempty(var_inside_model)
```

or better yet make the default value `nothing` and then have `if var_inside_model !== nothing`

src/FileFormats/LP/LP.jl
```diff
@@ -285,9 +285,475 @@ end
...
+)
+
+function _verify_name(variable::String, maximum_length::Int)
+    if length(variable) > maximum_length
+        return false
+    end
+    # m = match(START_REG, variable)
+    # if !isnothing(m)
+    #     return false
+    # end
```
odow, line 352: Is this commented for a reason?

odow (commented):
Looking better.

There's lots of untested error scenarios. We should add a bunch of malformed LP files to test the error conditions.

src/FileFormats/LP/LP.jl
joaquimg: for readability or performance?
current version is type stable. the `nothing` version relies on the small unions compiler optimization.

src/FileFormats/LP/LP.jl
guilhermebodin: I forgot to fix this. I will look exactly what should be the STAR_REG because right now I think it blocks variables starting with `E` and I think it should allow them.

odow: I do love file formats with peculiar exceptions. Variables cannot start with an `E` or `e`:

image

joaquimg: I agree that they should be allowed in this reader. I'd add a comment on the code though.
I wouldn't mind blocking `{E/e}-E[0-9]*` and `{E/e}-[0-9]*` if you guys want to.

odow: I guess we can be liberal for what we read, and conservative for what we write. So then I'm okay reading names starting with `E`. The writer should add a prefix to them.

joaquimg: Agreed

odow (commented):
This is an improvement on what we currently have so it's good by me. There are still a few opportunities to improve things (e.g., `CacheLPModel` has abstract fields, I don't know about the cost of `try-catch`, we could use more tests to improve coverage), but we can do them in separate, smaller PRs.

guilhermebodin:
Just realized there exists a function `Base.tryparse` https://docs.julialang.org/en/v1/base/numbers/#Base.tryparse. I can rewrite the PR using this function.
Let's hold this PR until Monday and I will try to fix these issues today.

odow:
Let's do improvements in a separate PR. This one is already quite big, and the functionality is stand-alone so there's no problem with merging it.

PR #1714, opened 2022-01-12
Remove redefinition with same implementation

blegat:
The default implementation is the same.


blegat:
I forgot it was not a subtype of `MOI.ModelLike`, this always confuses me.

PR #1716, opened 2022-01-13
Add FreeVariables

blegat:
Needed by https://github.com/jump-dev/DiffOpt.jl/pull/187/ and by most conic solvers.

The problem when using `VariablesContainer` is that model that use this as inner representation will need to redefine `supports_constraint` and `supports_add_constrained_variable` to avoid variables bounds to go through. This is quite tedious and if you don't do it correctly, you will silently ignore constraints that went through hidden in this VariablesContainer while they should have been bridged. This is the problem with DiffOpt.
For conic solvers, when they use `VariablesContainer` in their `OptimizerCache`, they need to verify that no variable bounds have been set during `copy_to`. Also, you cannot just copy_to to the `OptimizerCache` as otherwise, that won't bridge the variable bounds, you are forced to use it as cache of a `CachingOptimizer` as the `CachingOptimizer` will check that this is supported as well for the solver.

### DIFF
--- a/src/Utilities/free_variables.jl
+++ b/src/Utilities/free_variables.jl
@@ -1,0 +1,39 @@
+"""
+    mutable struct FreeVariables <: MOI.ModelLike
+        n::Int64
+        FreeVariables() = new(0)
+    end
+
+A struct for storing free variables that can be used as the `variables` field
+of [`GenericModel`](@ref) or [`GenericModel`](@ref).
+"""
+mutable struct FreeVariables <: MOI.ModelLike
+    n::Int64
+    FreeVariables() = new(0)
+end
+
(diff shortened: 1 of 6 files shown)


src/Utilities/free_variables.jl
```diff
@@ -0,0 +1,39 @@
+"""
+    mutable struct FreeVariables <: MOI.ModelLike
+        n::Int64
+        FreeVariables() = new(0)
+    end
+
+A struct for storing free variables that can be used as the `variables` field
```
matbesancon, line 7: this does not explain what it does / represents

odow (approved):
LGTM.

odow:
Are you happy to merge this @blegat?

PR #1717, opened 2022-01-13
Improve quadratic duality in doc

blegat:
Preview at https://jump.dev/MathOptInterface.jl/previews/PR1717/

### DIFF
--- a/docs/src/background/duality.md
+++ b/docs/src/background/duality.md
@@ -160,24 +160,41 @@ If ``\mathcal{C}_i`` is a vector set, the discussion remains valid with
 between ``y_i`` and the vector of scalar-valued quadratic functions.
 
 !!! note
-    For quadratic programs with only affine constraints, the optimality condition
-    ``\nabla_x L(x, y^\star) = 0`` can be simplified as follows:
+    For quadratic programs with only affine conic constraints,
     ```math
-    0 = \nabla_x L(x, y^\star) = Q_0x + a_0 - \sum_{i = 1}^m y_i^\star a_i
+    \begin{align*}
+    & \min_{x \in \mathbb{R}^n} & \frac{1}{2}x^TQ_0x + a_0^T x + b_0
+    \\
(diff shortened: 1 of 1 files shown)


matbesancon:
is there a preview for it?

blegat:
> is there a preview for it?

Added in the PR description

matbesancon:
I would add a reference / link / explanation for the normal cone. This is typically not taught in basic optimization classes.

>   For instance, if ``\mathcal{C}_i`` is ``\{ x \in \mathbb{R} : x \le 0 \}``, this means that
  if ``\frac{1}{2}x^TQ_ix + a_i^T x + b_i`` is nonzero at ``x^\star`` then ``y_i^\star = 0``.
This is the classical complementary slackness condition.

For notation, I would not use `x` in the definition of the cone, since it's the affine image of x which is in the cone

matbesancon:
The affinely constrained part ends up being rather big, I would have it as a subsection rather than a note

docs/src/background/duality.md
```diff
@@ -149,35 +192,14 @@ A pair of primal-dual variables $(x^\star, y^\star)$ is optimal if
...
+  ``\{ z \in \mathbb{R} : z \le 0 \}``, this means that if
+  ``\frac{1}{2}x^TQ_ix + a_i^T x + b_i`` is nonzero at ``x^\star`` then
+  ``y_i^\star = 0``. This is the classical complementary slackness condition.
 
 If ``\mathcal{C}_i`` is a vector set, the discussion remains valid with
 ``y_i(\frac{1}{2}x^TQ_ix + a_i^T x + b_i)`` replaced with the scalar product
 between ``y_i`` and the vector of scalar-valued quadratic functions.
-
-!!! note
-    For quadratic programs with only affine constraints, the optimality condition
```
matbesancon, line 163: did you remove the affine constraint section altogether?

blegat: I moved it up

matbesancon (approved):

PR #1731, opened 2022-02-08
Fix performance of MutableArithmetics

odow:
To be used in conjunction with https://github.com/jump-dev/MutableArithmetics.jl/pull/147

This is my new favorite performance bug of all time. For some reason,
the extra N type parameter was preventing Julia from being clever, but
it's soooooo hard to notice.


matbesancon:
Oh wow I was not expecting this

matbesancon (approved):

PR #1732, opened 2022-02-08
Add Aqua.jl as a CI job and fix remaining errors

odow:
Damn. There are some pretty complicated ones here. I'll comment inline. GitHub action still to come because there are a few errors.

### DIFF
--- a/src/Utilities/functions.jl
+++ b/src/Utilities/functions.jl
@@ -2193,8 +2193,15 @@ function operate(
     return operate!(op, T, copy(f), g)
 end
 
-function Base.:+(args::VectorLike{T}...) where {T}
-    return operate(+, T, args...)
+function Base.:+(arg::VectorLike, args::VectorLike...)
+    T = _eltype(arg, args)
+    if T === nothing
+        error(
+            "Cannot add VectorOfVariables together without a coefficient " *
+            "type. Convert one argument to a VectorAffineFunction first.",
+        )
+    end
+    return operate(+, T, arg, args...)
 end
 
 # Base.:+(α::Vector{T}, f::VectorLike{T}...) is too general as it also covers
@@ -2207,8 +2214,15 @@ function Base.:+(f::VectorLike{T}, α::Vector{T}) where {T}
     return operate(+, T, f, α)
 end
 
-function Base.:-(args::VectorLike{T}...) where {T}
-    return operate(-, T, args...)
+function Base.:-(arg::VectorLike, args::VectorLike...)
+    T = _eltype(arg, args)
+    if T === nothing
+        error(
+            "Cannot subtract VectorOfVariables without a coefficient " *
+            "type. Convert one argument to a VectorAffineFunction first.",
+        )
+    end
+    return operate(-, T, arg, args...)
 end
 
 function Base.:-(f::VectorLike{T}, α::Vector{T}) where {T}
--- a/.github/workflows/aqua.yml
+++ b/.github/workflows/aqua.yml
@@ -1,0 +1,23 @@
+name: aqua-lint
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    types: [opened, synchronize, reopened]
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: julia-actions/setup-julia@latest
+        with:
+          version: '1'
+      - uses: actions/checkout@v1
+      - name: Aqua
+        shell: julia --color=yes {0}
+        run: |
+          using Pkg
+          Pkg.add(PackageSpec(name="Aqua"))
+          Pkg.develop(PackageSpec(path=pwd()))
+          using MathOptInterface, Aqua
+          Aqua.test_all(MathOptInterface; ambiguities = false)
--- a/src/Utilities/cachingoptimizer.jl
+++ b/src/Utilities/cachingoptimizer.jl
@@ -54,7 +54,7 @@ A `CachingOptimizer` may be in one of three possible states
 * `ATTACHED_OPTIMIZER`: The CachingOptimizer has an optimizer, and it is
   synchronized with the cached model.
 
-### modes
+### Modes
 
 A `CachingOptimizer` has two modes of operation (`CachingOptimizerMode`):
 


src/Utilities/mutable_arithmetics.jl
```diff
@@ -67,13 +67,31 @@ end
...
+    F::Type{MOI.ScalarAffineFunction{T}},
+    G::Type{MOI.VariableIndex},
+) where {T}
+    return promote_operation(op, T, F, G)
+end
+
 function MA.promote_operation(
     op::PROMOTE_IMPLEMENTED_OP,
-    F::Type{<:ScalarLike{T}},
-    G::Type{<:ScalarLike{T}},
```
odow, line 73: Here's another case of `ScalarLike`. This time there's only three possible cases, so we may as well just enumerate.

src/Utilities/struct_of_constraints.jl
```diff
@@ -365,13 +365,21 @@ function struct_of_constraint_code(struct_name, types, field_types = nothing)
...
     if type_parametrized
+        # A bit confusing: we need to add the extra type parameters to `{T}`.
+        # Here's the `where {T}`
         append!(constructor_code.args[1].args, field_types)
+        # And here's the `new{T}`
+        append!(constructor_code.args[2].args[end].args[1].args[1].args, field_types)
     end
-    push!(code.args, constructor_code)
+    # Here's we we put the inner constructor.
+    push!(code.args[2].args[3].args, constructor_code)
```
odow, line 382: Okay. Previously, this code made an external constructor. This could lead to errors like
```Julia
julia> module TestFoo

       struct Foo{T}
           x::Union{Nothing,Vector{T}}
       end

       function Foo{T}() where {T}
           return Foo{T}(nothing)
       end

       end
Main.TestFoo

julia> TestFoo.Foo(nothing)
ERROR: UndefVarError: T not defined
Stacktrace:
 [1] Main.TestFoo.Foo(x::Nothing)
   @ Main.TestFoo ./REPL[9]:4
 [2] top-level scope
   @ REPL[10]:1

julia> TestFoo.Foo()
ERROR: MethodError: no method matching Main.TestFoo.Foo()
Closest candidates are:
  Main.TestFoo.Foo(::Union{Nothing, Vector{T}}) where T at REPL[9]:4
Stacktrace:
 [1] top-level scope
   @ REPL[11]:1

julia> TestFoo.Foo{Float64}()
Main.TestFoo.Foo{Float64}(nothing)
```

This new code creates an internal constructor:
```Julia
julia> module TestFoo

       struct Foo{T}
           x::Union{Nothing,Vector{T}}
           function Foo{T}() where {T}
               return new{T}(nothing)
           end
       end

       end
Main.TestFoo

julia> TestFoo.Foo(nothing)
ERROR: MethodError: no method matching Main.TestFoo.Foo(::Nothing)
Stacktrace:
 [1] top-level scope
   @ REPL[6]:1

julia> TestFoo.Foo()
ERROR: MethodError: no method matching Main.TestFoo.Foo()
Stacktrace:
 [1] top-level scope
   @ REPL[7]:1

julia> TestFoo.Foo{Float64}()
Main.TestFoo.Foo{Float64}(nothing)
```

src/Utilities/functions.jl
```diff
@@ -1732,8 +1732,24 @@ function operate(
...
+_eltype(args::Tuple) = _eltype(first(args), Base.tail(args))
+_eltype(::Tuple{}) = nothing
+_eltype(::MOI.Utilities.TypedScalarLike{T}, tail) where {T} = T
+_eltype(::MOI.VariableIndex, tail) = _eltype(tail)
+_eltype(::MOI.Utilities.TypedVectorLike{T}, tail) where {T} = T
+_eltype(::MOI.VectorOfVariables, tail) = _eltype(tail)
+
 # To avoid type piracy, we add at least one `ScalarLike` outside of the `...`.
-function Base.:+(arg::ScalarLike{T}, args::ScalarLike{T}...) where {T}
+function Base.:+(arg::ScalarLike, args::ScalarLike...)
```
odow, line 1743: Okay. The problem here (and others below) is that `ScalarLike{T} = Union{MOI.VariableIndex,MOI.ScalarAffineFunction{T}`. 

The problem is that if you pass in one or more `MOI.VariableIndex`, then `T` is not defined. We previously worked around this by defining another fallback, but things get a bit confusing.

I'd say that writing methods like this works, but is probably bad practice. Unfortunately, I don't think there's a way to encode `T` using dispatch, so we have to use a look-up function `_eltype` internally. 

`_eltype` uses the same lispy recursion to extra type parameters.

The downside to this approach is that
```Julia

julia> y = 1 * x
MathOptInterface.ScalarAffineFunction{Int64}(MathOptInterface.ScalarAffineTerm{Int64}[MathOptInterface.ScalarAffineTerm{Int64}(1, MathOptInterface.VariableIndex(1))], 0)

julia> z = 1.0 * x
MathOptInterface.ScalarAffineFunction{Float64}(MathOptInterface.ScalarAffineTerm{Float64}[MathOptInterface.ScalarAffineTerm{Float64}(1.0, MathOptInterface.VariableIndex(1))], 0.0)

julia> x + y + z
ERROR: MethodError: no method matching operate!(::typeof(+), ::Type{Int64}, ::MathOptInterface.ScalarAffineFunction{Int64}, ::MathOptInterface.ScalarAffineFunction{Float64})
Closest candidates are:
  operate!(::typeof(+), ::Type{T}, ::Any, ::Any, ::Any, ::Any...) where T at /Users/oscar/.julia/dev/MathOptInterface/src/Utilities/functions.jl:1558
  operate!(::Function, ::Type{T}, ::Union{AbstractVector{T}, T}...) where T at /Users/oscar/.julia/dev/MathOptInterface/src/Utilities/functions.jl:1186
  operate!(::Union{typeof(+), typeof(-)}, ::Type{T}, ::MathOptInterface.VariableIndex, ::Union{MathOptInterface.VariableIndex, MathOptInterface.ScalarAffineFunction{T}, MathOptInterface.ScalarQuadraticFunction{T}, T} where T) where T at /Users/oscar/.julia/dev/MathOptInterface/src/Utilities/functions.jl:1581
  ...
Stacktrace:
 [1] operate(::typeof(+), ::Type{Int64}, ::MathOptInterface.VariableIndex, ::MathOptInterface.ScalarAffineFunction{Int64}, ::MathOptInterface.ScalarAffineFunction{Float64})
   @ MathOptInterface.Utilities ~/.julia/dev/MathOptInterface/src/Utilities/functions.jl:1618
 [2] +(::MathOptInterface.VariableIndex, ::MathOptInterface.ScalarAffineFunction{Int64}, ::MathOptInterface.ScalarAffineFunction{Float64})
   @ MathOptInterface.Utilities ~/.julia/dev/MathOptInterface/src/Utilities/functions.jl:1753
 [3] top-level scope
   @ REPL[7]:1
```

But we already had a similar method error:
```Julia

julia> y = 1 * x
MathOptInterface.ScalarAffineFunction{Int64}(MathOptInterface.ScalarAffineTerm{Int64}[MathOptInterface.ScalarAffineTerm{Int64}(1, MathOptInterface.VariableIndex(1))], 0)

julia> z = 1.0 * x
MathOptInterface.ScalarAffineFunction{Float64}(MathOptInterface.ScalarAffineTerm{Float64}[MathOptInterface.ScalarAffineTerm{Float64}(1.0, MathOptInterface.VariableIndex(1))], 0.0)

julia> x + y
MathOptInterface.ScalarAffineFunction{Int64}(MathOptInterface.ScalarAffineTerm{Int64}[MathOptInterface.ScalarAffineTerm{Int64}(1, MathOptInterface.VariableIndex(1)), MathOptInterface.ScalarAffineTerm{Int64}(1, MathOptInterface.VariableIndex(1))], 0)

julia> x + y + z
ERROR: MethodError: no method matching +(::MathOptInterface.ScalarAffineFunction{Int64}, ::MathOptInterface.ScalarAffineFunction{Float64})
Closest candidates are:
  +(::Any, ::Any, ::Any, ::Any...) at operators.jl:560
  +(::Union{MathOptInterface.ScalarAffineFunction{T}, MathOptInterface.ScalarQuadraticFunction{T}}, ::T) where T at /Users/oscar/.julia/packages/MathOptInterface/lRbIA/src/Utilities/functions.jl:1752
  +(::MutableArithmetics.Zero, ::Any) at /Users/oscar/.julia/packages/MutableArithmetics/UtY4H/src/rewrite.jl:54
  ...
Stacktrace:
 [1] +(::MathOptInterface.VariableIndex, ::MathOptInterface.ScalarAffineFunction{Int64}, ::MathOptInterface.ScalarAffineFunction{Float64})
   @ Base ./operators.jl:560
 [2] top-level scope
   @ REPL[19]:1
```

One thing I could do is check in `_eltype` that all types are equivalent instead of returning the first one.

odow:
This PR touches a few different concepts, so I'll split it up into different PRs.

src/Utilities/constraints.jl
```diff
@@ -15,7 +15,7 @@ function normalize_and_add_constraint(
     func::MOI.AbstractScalarFunction,
     set::MOI.AbstractScalarSet;
     allow_modify_function::Bool = false,
-) where {T}
+)
```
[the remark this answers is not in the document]
odow: #1734

src/Utilities/mutable_arithmetics.jl
odow: #1736

src/Utilities/struct_of_constraints.jl
odow: #1735

odow:
Okay. Once all the linked PRs are merged, I'll rebase this to add the GitHub action.

src/Utilities/functions.jl
blegat: Yes, it does not follow the methoderror principle but it should be ok

PR #1736, opened 2022-02-09
[Utilities] clean dispatch in MA.promote_operation

odow:
ScalarLike{T} includes VariableIndex, which does not have a type
parameter. So passing two variables to promote_operation would have
resulting in a 'T not defined error'.

_We_ know that this would have been caught by another method, but it
seems good coding practice to explicitly enumerate these methods.

Part of #1732

### DIFF
--- a/src/Utilities/mutable_arithmetics.jl
+++ b/src/Utilities/mutable_arithmetics.jl
@@ -65,15 +65,33 @@ function MA.promote_operation(
     return F
 end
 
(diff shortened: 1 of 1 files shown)


src/Utilities/mutable_arithmetics.jl
```diff
@@ -65,15 +65,33 @@ function MA.promote_operation(
     return F
 end
 
-# To avoid type piracy, we add at least one `ScalarLike` outside of the `...`.
+
```
odow, line 68:
```suggestion
```

PR #1738, opened 2022-02-09
[Utilities] implement Aqua.jl recommendation for functions.jl

odow:
Part of #1732


odow:
@blegat what are your thoughts on this one?

PR #1747, opened 2022-02-21
Add edge cases for RawStatusString

blegat:
This would have detected https://github.com/jump-dev/Ipopt.jl/pull/304


odow:
Given
```Julia
julia> MOI.is_set_by_optimize(MOI.RawStatusString())
true
```
I don't think it makes sense to test before `optimize!`?

blegat:
The issue is that it's called if the user calls `solution_summary` before `optimize!` is called.

odow:
Then we should fix that in JuMP.

PR #1754, opened 2022-02-23
Call final_touch in optimize in CachingOptimizer

blegat:
This allows to use a `CachingOptimizer` with a `MatrixOfConstraint` cache with the workflow: `add_variable`, `add_constraint`, `optimize`.

- [x] Test


odow (commented):
We haven't tested a lot of this, because I don't think we envisaged `MatrixOfConstraints` supporting an incremental build.

PR #1755, opened 2022-02-23
Canonicalize function in add_constraint for MatrixOfConstraints

blegat:
- [x] Add test

Detected by https://github.com/jump-dev/ECOS.jl/pull/137

### DIFF
--- a/src/Utilities/matrix_of_constraints.jl
+++ b/src/Utilities/matrix_of_constraints.jl
@@ -352,6 +352,7 @@ function _add_constraint(
     func::F,
     set::S,
 ) where {F<:MOI.AbstractFunction,S<:MOI.AbstractSet}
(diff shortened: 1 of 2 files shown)


src/Utilities/matrix_of_constraints.jl
```diff
@@ -352,6 +352,7 @@ function _add_constraint(
     func::F,
     set::S,
 ) where {F<:MOI.AbstractFunction,S<:MOI.AbstractSet}
+    func = MOI.Utilities.canonical(func)
```
odow, line 355: Why call canonical twice? (Here and `add_constraint`?)

blegat: Oops, this one shouldn't be

PR #1760, opened 2022-02-26
Add UniversalFallback utility to throw unsupported attributes and constraints

blegat:
This would be needed here:
https://github.com/jump-dev/Clp.jl/blob/6a6ebbf637f8933718645137e343bb969e1578e0/src/MOI_wrapper/MOI_wrapper.jl#L334-L339
Currently, the unsupported parts will just be ignored.
This is a blocker for Clp.jl and ECOS.jl releases. Using it allows to fix the last remaining failing tests in https://github.com/jump-dev/Clp.jl/pull/130

### DIFF
--- a/src/Utilities/universalfallback.jl
+++ b/src/Utilities/universalfallback.jl
@@ -129,6 +129,71 @@ function final_touch(uf::UniversalFallback, index_map)
     return final_touch(uf.model, index_map)
 end
 
+"""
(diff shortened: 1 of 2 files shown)


blegat:
@odow Good to merge ?

odow:
Yes. I just wanted to release v1.0.2 first. This will go out as v1.1.0 since we're adding new features.

blegat:
> This will go out as v1.1.0 since we're adding new features.

Right, we need to follow semantic versioning now

PR #1763, opened 2022-02-28
[Bridges] error in RSOCtoPSD when dimension is 2

### LINKED ISSUE #1761: Constraint.RSOCtoPSDBridge fails for dimension 2
The bridge transforms `(t, u, x)` from `2tu >= ||x||_2`, `t, u >= 0` to
```
t   x'
x   2uI
```
is PSD.
However, when the dimension is 2, i.e., the length of `x` is zero, this is transformed into the 1x1 PSD constraint `t >= 0`.
This is missing the constraint that `u` should be nonnegative.
Ideally the bridge should add two linear inequality constraints but in the meantime it should at least provide a nice error instead of being incorrect.

Detected while going from DeprecatedTest to Test in https://github.com/jump-dev/CSDP.jl/pull/72

odow:
Closes #1761 

The premise of using a RotatedSecondOrderCone in dimension 2 has always seemed a bit iffy to me.

### DIFF
--- a/src/Bridges/Constraint/soc_to_psd.jl
+++ b/src/Bridges/Constraint/soc_to_psd.jl
@@ -185,7 +185,7 @@ function MOIB.map_function(::Type{<:RSOCtoPSDBridge{T}}, func) where {T}
     if length(scalars) < 3
         error(
             "Unable to bridge RotatedSecondOrderCone to PSD because the ",
-            "dimension is too small.",
+            "dimension is too small: got $(length(scalars)), expected >= 3.",
         )
     end
     h = MOIU.operate!(*, T, scalars[2], convert(T, 2))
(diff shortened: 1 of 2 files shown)


src/Bridges/Constraint/soc_to_psd.jl
```diff
@@ -179,6 +182,12 @@ end
 
 function MOIB.map_function(::Type{<:RSOCtoPSDBridge{T}}, func) where {T}
     scalars = MOIU.eachscalar(func)
+    if length(scalars) < 3
+        error(
+            "Unable to bridge RotatedSecondOrderCone to PSD because the ",
+            "dimension is too small."
```
odow, line 188:
```suggestion
            "dimension is too small.",
```

PR #1768, opened 2022-03-01
Prep for v1.1.0 release

odow:


odow:
@blegat you're okay with this?

odow:
@blegat and I discussed this offline.

PR #1769, opened 2022-03-02
Use tolerance to compare value of blank objective in test

blegat:
With SDPA:
```julia
test_objective_ObjectiveFunction_blank: Test Failed at /home/runner/.julia/packages/MathOptInterface/FHFUH/src/Test/test_objective.jl:195
  Expression: MOI.get(model, MOI.ObjectiveValue()) == T(0)
   Evaluated: 1.3932339049530345e-8 == 0.0
```
See https://github.com/jump-dev/SDPA.jl/runs/5383637798?check_suite_focus=true#step:6:616


odow:
Why would it do this? 😢

odow:
Merging because this is correct. But I'd really like to know why SDPA would return a non-zero objective for `min 0x`!

blegat:
No idea, I was surprised as well

PR #1776, opened 2022-03-08
Instantiate with cache if needed

blegat:
Not so convinced by the name. What about:

* `with_incremental_interface_type`
* `with_cache_type`
* `with_incremental_interface_of_type`
* `incrementally_with_type`

See https://github.com/lanl-ansi/Alpine.jl/pull/188#issuecomment-1061114841

### DIFF
--- a/src/instantiate.jl
+++ b/src/instantiate.jl
@@ -96,6 +96,7 @@ end
     instantiate(
         optimizer_constructor,
         with_bridge_type::Union{Nothing, Type} = nothing,
+        with_incremental_interface = with_bridge_type,
     )
 
 Creates an instance of optimizer by either:
@@ -108,23 +109,27 @@ If `with_bridge_type` is not `nothing`, it enables all the bridges defined in
 the MathOptInterface.Bridges submodule with coefficient type `with_bridge_type`.
 
 If the optimizer created by `optimizer_constructor` does not support loading the
-problem incrementally (see [`supports_incremental_interface`](@ref)), then a
-[`Utilities.CachingOptimizer`](@ref) is added to store a cache of the bridged
-model.
+problem incrementally (see [`supports_incremental_interface`](@ref)) and either
+`with_incremental_interface` is not `nothing`, then a
+[`Utilities.CachingOptimizer`](@ref) is added to store a cache of coefficient
+type `with_incremental_interface` of the model. This cache stores the bridged
+model if `with_bridge_type` is not `nothing`.
 """
 function instantiate(
     optimizer_constructor;
     with_bridge_type::Union{Nothing,Type} = nothing,
+    with_incremental_interface::Union{Nothing,Type} = with_bridge_type,
 )
     optimizer = _instantiate_and_check(optimizer_constructor)
-    if with_bridge_type === nothing
-        return optimizer
-    end
-    if !supports_incremental_interface(optimizer)
-        cache = default_cache(optimizer, with_bridge_type)
+    if !isnothing(with_incremental_interface) &&
+       !supports_incremental_interface(optimizer)
+        cache = default_cache(optimizer, with_incremental_interface)
         optimizer = Utilities.CachingOptimizer(cache, optimizer)
     end
-    return Bridges.full_bridge_optimizer(optimizer, with_bridge_type)
+    if !isnothing(with_bridge_type)
+        optimizer = Bridges.full_bridge_optimizer(optimizer, with_bridge_type)
+    end
+    return optimizer
 end
 
 """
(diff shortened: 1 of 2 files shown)


blegat:
Is BARON's use case just the instantiation in `BARON/test/MOI_wrapper.jl` ? In this case, you a cache as outer layer of the instantiated optimizer. This create the `moi_backend` that JuMP would create.

odow:
I think if we're using MatrixOfConstraints, adding the cache should be the default. This is kinda what I mean a while ago; we didn't break JuMP usage by switching to MatrixOfConstraints, but we broke a lot of other usages, like Pavito, or any solver which uses Cbc.Optimizer internally. We should probably have a call about this to discuss.

blegat:
> I think if we're using MatrixOfConstraints, adding the cache should be the default.

If you do
```julia
optimizer = MOI.instantiate(solver)
MOI.copy_to(optimizer, model)
```
then you don't need the cache.
Solvers using `MatrixOfConstraints` and solver for which `implements_incremental_interface` if `false` are in the same category, I don't think they should be treated differently.
Maybe you mean that we should have a cache by default but then we should have a default value for the coefficient type.
Maybe Pavito should use `with_bridge_type=Float64` also.

odow:
```Julia
julia> Base.show(io::IO, ::Cbc.OptimizerCache) = print(io, "Cbc.OptimizerCache()")

julia> Base.show(io::IO, ::Type{Cbc.OptimizerCache}) = print(io, "Cbc.OptimizerCache")

julia> MOI.instantiate(Cbc.Optimizer)
Cbc.Optimizer

julia> MOI.instantiate(Cbc.Optimizer; with_bridge_type = Float64)
MOIB.LazyBridgeOptimizer{MOIU.CachingOptimizer{Cbc.Optimizer, MOIU.UniversalFallback{Cbc.OptimizerCache}}}
with 0 variable bridges
with 0 constraint bridges
with 0 objective bridges
with inner model MOIU.CachingOptimizer{Cbc.Optimizer, MOIU.UniversalFallback{Cbc.OptimizerCache}}
  in state EMPTY_OPTIMIZER
  in mode AUTOMATIC
  with model cache MOIU.UniversalFallback{Cbc.OptimizerCache}
    fallback for Cbc.OptimizerCache()
  with optimizer Cbc.Optimizer
```

The problem is that Pavito needs
```Julia
MOI.Utilities.CachingOptimizer(
    MOI.Utilities.UniversalFallback(MOI.Utilities.Model{Float64}()),
    MOI.instantiate(Cbc.Optimizer; with_bridge_type = Float64),
)
```

It's also a problem that
```Julia
julia> model = MOI.instantiate(Cbc.Optimizer; with_bridge_type = Float64)
MOIB.LazyBridgeOptimizer{MOIU.CachingOptimizer{Cbc.Optimizer, MOIU.UniversalFallback{Cbc.OptimizerCache}}}
with 0 variable bridges
with 0 constraint bridges
with 0 objective bridges
with inner model MOIU.CachingOptimizer{Cbc.Optimizer, MOIU.UniversalFallback{Cbc.OptimizerCache}}
  in state EMPTY_OPTIMIZER
  in mode AUTOMATIC
  with model cache MOIU.UniversalFallback{Cbc.OptimizerCache}
    fallback for Cbc.OptimizerCache()
  with optimizer Cbc.Optimizer

julia> MOI.supports_incremental_interface(model)
true
```

So MatrixOfConstraints works well if you're using JuMP, but it doesn't work if you have a sub-solver. And it's not obvious how to write a good detection for MatrixOfConstraints since it's nested so deeply in the solver.

odow:
Now I guess people could pass Pavito a caching optimizer of Cbc, but that also defeats the point and is pretty complicated.

blegat:
This isn't related to MatrixOfConstraints. If Pavito needs to be able to do modifications then it should do:
```julia
MOI.Utilities.CachingOptimizer(
    MOI.Utilities.UniversalFallback(MOI.Utilities.Model{Float64}()),
    MOI.instantiate(solver; with_bridge_type = Float64),
)
```
for any solver. Indeed, `MOI.instantiate(solver; with_bridge_type = Float64)` guarantees that it returns an optimizer that supports the incremental interface but it does not guarantee that it supports modification.

odow:
> guarantees that it returns an optimizer that supports the incremental interface but it does not guarantee that it supports modification

Oooff. This is really our problem. In the past anyone supporting the incremental interface _did_ support modification.

blegat:
> This is really our problem. In the past anyone supporting the incremental interface did support modification.

It depends which modifications though. That might be true for variable bounds for most but not all modifications.

odow:
Hmm.  I guess you're right. In general, Pavito would need a caching optimizer to ensure correctness.

odow:
So then perhaps we don't need this? Pavito should explicitly create a caching optimizer. No need for more complications here.

blegat:
Yes, this PR can be closed, we can reconsider once we have a use case.

odow:
I'm just working on updating Pavito now.

PR #1780, opened 2022-03-17
[FileFormats] fix copy_to for NL.Model when used with bridges

### LINKED ISSUE #1779: Writing to NL file fails
This used to work, so something must have changed:
```Julia
julia> model = Model()
A JuMP Model
Feasibility problem with:
Variables: 0
Model mode: AUTOMATIC
CachingOptimizer state: NO_OPTIMIZER
Solver name: No optimizer attached.

julia> @variable(model, x >= 0)
x

julia> @objective(model, Min, 2x + 1)
2 x + 1

julia> write_to_file(model, "test.nl")
ERROR: Model MathOptInterface.Bridges.LazyBridgeOptimizer{MathOptInterface.FileFormats.NL.Model} does not support copy_to.
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:33
 [2] default_copy_to(dest::MathOptInterface.Bridges.LazyBridgeOptimizer{MathOptInterface.FileFormats.NL.Model}, src::MathOptInterface.Utilities.CachingOptimizer{MathOptInterface.AbstractOptimizer, MathOptInterface.Utilities.UniversalFallback{MathOptInterface.Utilities.Model{Float64}}})
   @ MathOptInterface.Utilities ~/.julia/packages/MathOptInterface/FHFUH/src/Utilities/copy.jl:454
 [3] copy_to(dest::MathOptInterface.Bridges.LazyBridgeOptimizer{MathOptInterface.FileFormats.NL.Model}, src::MathOptInterface.Utilities.CachingOptimizer{MathOptInterface.AbstractOptimizer, MathOptInterface.Utilities.UniversalFallback{MathOptInterface.Utilities.Model{Float64}}}; kwargs::Base.Iterators.Pairs{Union{}, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
   @ MathOptInterface.Bridges ~/.julia/packages/MathOptInterface/FHFUH/src/Bridges/bridge_optimizer.jl:421
 [4] copy_to(dest::MathOptInterface.Bridges.LazyBridgeOptimizer{MathOptInterface.FileFormats.NL.Model}, src::MathOptInterface.Utilities.CachingOptimizer{MathOptInterface.AbstractOptimizer, MathOptInterface.Utilities.UniversalFallback{MathOptInterface.Utilities.Model{Float64}}})
   @ MathOptInterface.Bridges ~/.julia/packages/MathOptInterface/FHFUH/src/Bridges/bridge_optimizer.jl:421
 [5] copy_to(dest::MathOptInterface.Bridges.LazyBridgeOptimizer{MathOptInterface.FileFormats.NL.Model}, src::Model)
   @ JuMP ~/.julia/packages/JuMP/0STkJ/src/copy.jl:290
 [6] write_to_file(model::Model, filename::String; format::MathOptInterface.FileFormats.FileFormat)
   @ JuMP ~/.julia/packages/JuMP/0STkJ/src/file_formats.jl:27
 [7] write_to_file(model::Model, filename::String)
   @ JuMP ~/.julia/packages/JuMP/0STkJ/src/file_formats.jl:23
 [8] top-level scope
   @ REPL[13]:1
```

odow:
Closes #1779 

This was actually slightly more complicated that expected (there's a long comment in the code explaining why this is necessary). But essentially we were missing a test for the case where we tried to copy to a `Bridges.full_bridge_optimizer(NL.Model(), Float64)`. The other file formats work fine because they use `@model`, but NL is special because I cut out an extra cache and only implemented `copy_to`. (I can't have tried it via JuMP since I removed the cache.)

Other fixes, like JuMP using a CachingOptimizer, won't work because `NL.Model` is not an AbstractOptimizer. This is definitely a hole in the API, but since this is a one-off case we can keep it like this for now. If we encounter other places where this is useful, we could consider adding a `CachingModel` to `MOI.Utilities`.

```Julia
julia> using JuMP

julia> model = Model()
A JuMP Model
Feasibility problem with:
Variables: 0
Model mode: AUTOMATIC
CachingOptimizer state: NO_OPTIMIZER
Solver name: No optimizer attached.

julia> @variable(model, x >= 0)
x

julia> @objective(model, Min, 2x + 1)
2 x + 1

julia> write_to_file(model, "test.nl")

shell> cat test.nl
g3 1 1 0
 1 0 1 0 0 0
 0 1
 0 0
 0 0 0
 0 0 0 1
 0 0 0 0 0
 0 1
 0 0
 0 0 0 0 0
O0 0
n1
x1
0 0
b
2 0
G0 1
0 2

(copt) pkg> st
      Status `/private/tmp/copt/Project.toml`
  [4076af6c] JuMP v0.23.2
  [b8f27783] MathOptInterface v1.1.0 `~/.julia/dev/MathOptInterface`
```

### DIFF
--- a/src/FileFormats/NL/NL.jl
+++ b/src/FileFormats/NL/NL.jl
@@ -829,10 +829,7 @@ function MOI.get(model::_CachingModel, attr::MOI.AnyAttribute, args...)
     return MOI.get(model.cache, attr, args...)
 end
 
-function MOI.supports(
-    model::_CachingModel,
-    attr::MOI.AnyAttribute,
-    args...)
+function MOI.supports(model::_CachingModel, attr::MOI.AnyAttribute, args...)
     return MOI.supports(model.inner, attr, args...)
 end
 
@@ -846,7 +843,7 @@ end
 
 function MOI.copy_to(
     dest::MOI.Bridges.LazyBridgeOptimizer{MOI.FileFormats.NL.Model},
-    src::MOI.ModelLike
+    src::MOI.ModelLike,
 )
     model = _CachingModel(dest.model)
     index_map = MOI.Utilities.default_copy_to(model, src)
--- a/src/FileFormats/NL/NL.jl
+++ b/src/FileFormats/NL/NL.jl
@@ -846,7 +846,9 @@ function MOI.copy_to(
     src::MOI.ModelLike,
 )
     model = _CachingModel(dest.model)
-    index_map = MOI.copy_to(model, src)
+    # This needs to be `default_copy_to` so that it uses the incremental
+    # interface.
+    index_map = MOI.Utilities.default_copy_to(model, src)
     MOI.copy_to(model.inner, model.cache)
     return index_map
 end
--- a/src/FileFormats/NL/NL.jl
+++ b/src/FileFormats/NL/NL.jl
@@ -846,7 +846,7 @@ function MOI.copy_to(
     src::MOI.ModelLike,
 )
     model = _CachingModel(dest.model)
-    index_map = MOI.Utilities.default_copy_to(model, src)
+    index_map = MOI.copy_to(model, src)
     MOI.copy_to(model.inner, model.cache)
     return index_map
 end
--- a/src/FileFormats/NL/NL.jl
+++ b/src/FileFormats/NL/NL.jl
@@ -838,7 +838,7 @@ function MOI.supports_constraint(
     f::MOI.AbstractFunction,
     s::MOI.AbstractSet,
 )
-    return MOI.supports(model.inner, f, s)
+    return MOI.supports_constraint(model.inner, f, s)
 end
 
 function MOI.copy_to(
--- a/src/FileFormats/NL/NL.jl
+++ b/src/FileFormats/NL/NL.jl
@@ -773,4 +773,85 @@ function Base.write(io::IO, nlmodel::Model)
     return nlmodel
 end
 
+# This part is needed for JuMP's write_to_file method.
+#
+# JuMP calls:
+# ```julia
+# dest = MOI.FileFormats.Model(format = format, filename = filename)
+# bridged_dest = MOI.Bridges.full_bridge_optimizer(dest, Float64)
+# MOI.copy_to(bridged_dest, model)
+# MOI.write_to_file(dest, filename)
+# ```
+# So we need a way of calling `copy_to` from a model to a bridged NL.Model.
+# However because we don't support the incremental interface, this would require
+# a CachingOptimizer. Except we can't use a CachingOptimizer because NL.Model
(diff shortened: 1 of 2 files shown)


blegat:
IMO, the restriction for the optimizer of a `CachingOptimizer` to be a subtype of `AbstractOptimizer` should be removed.
It does not serve any purpose. I guess we were waiting for a use case like this one to remove it.

odow:
Yeah that could be a different pull request. It might lead to some weird errors, but should be okay for most things. 

However, can we merge this first? That lets JuMP 1.0 depend on MOI 1.0. We can drop the restriction in a future version of MOI, and then change JuMP to use that if/when we want to bump the minimum supported version of MOI.

blegat:
> That lets JuMP 1.0 depend on MOI 1.0. We can drop the restriction in a future version of MOI, and then change JuMP to use that if/when we want to bump the minimum supported version of MOI.

Why would it ? If JuMP depends on this PR, it also means that it will depend on MOI v1.1 no ?

blegat:
JuMP's code doesn't need to change if we allow `MOI.ModelLike` as optimizer field.

odow:
Because JuMP either needs to explicitly construct a caching optimizer, or call `MOI.instantiate`, and it can't do either of those with MOI 1.0:

```julia
julia> MOI.instantiate(MOI.FileFormats.NL.Model)
ERROR: The provided `optimizer_constructor` returned an object of type MathOptInterface.FileFormats.NL.Model. Expected a MathOptInterface.AbstractOptimizer.
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:33
 [2] _instantiate_and_check(optimizer_constructor::Type{MathOptInterface.FileFormats.NL.Model})
   @ MathOptInterface ~/.julia/packages/MathOptInterface/IQWV4/src/instantiate.jl:66
 [3] instantiate(optimizer_constructor::Type; with_bridge_type::Nothing)
   @ MathOptInterface ~/.julia/packages/MathOptInterface/IQWV4/src/instantiate.jl:119
 [4] instantiate(optimizer_constructor::Type)
   @ MathOptInterface ~/.julia/packages/MathOptInterface/IQWV4/src/instantiate.jl:119
 [5] top-level scope
   @ REPL[2]:1
```

blegat:
If we additionally fix `instantiate` to check for `MOI.ModelLike` and not for `MOI.AbstractOptimizer` then we don't need any change to JuMP, right ?
I feel there is nothing wrong with what `NL.Model` is doing and we might encounter other similar issues if we don't fix the root cause.

odow:
Okay. Let me put together a PR with the full fix. Maybe we do just lower-bound JuMP to MOI 1.1.2

odow:
Closing in favor of #1781

PR #1781, opened 2022-03-22
Relax AbstractOptimizer requirement in instantiate

odow:
Alternative to #1780

This will need JuMP to lower-bound it's dependence on MOI, but that's okay. This is really a bug-fix, so I plan to tag as 1.1.1.
```julia
julia> using MathOptInterface; const MOI = MathOptInterface
MathOptInterface

julia> MOI.instantiate(MOI.FileFormats.NL.Model)
An AMPL (.nl) model

julia> MOI.instantiate(MOI.FileFormats.NL.Model; with_bridge_type = Float64)
MOIB.LazyBridgeOptimizer{MOIU.CachingOptimizer{MOI.FileFormats.NL.Model, MOIU.UniversalFallback{MOIU.Model{Float64}}}}
with 0 variable bridges
with 0 constraint bridges
with 0 objective bridges
with inner model MOIU.CachingOptimizer{MOI.FileFormats.NL.Model, MOIU.UniversalFallback{MOIU.Model{Float64}}}
  in state EMPTY_OPTIMIZER
  in mode AUTOMATIC
  with optimizer An AMPL (.nl) model

julia> MOI.instantiate(MOI.FileFormats.MPS.Model; with_bridge_type = Float64)
MOIB.LazyBridgeOptimizer{MOIU.GenericModel{Float64, MOIU.ObjectiveContainer{Float64}, MOIU.VariablesContainer{Float64}, MOI.FileFormats.MPS.ModelFunctionConstraints{Float64}}}
with 0 variable bridges
with 0 constraint bridges
with 0 objective bridges
with inner model A Mathematical Programming System (MPS) model
```

### DIFF
--- a/src/Utilities/cachingoptimizer.jl
+++ b/src/Utilities/cachingoptimizer.jl
@@ -92,7 +92,7 @@ mutable struct CachingOptimizer{O,M<:MOI.ModelLike} <: MOI.AbstractOptimizer
 
     function CachingOptimizer(
         cache::MOI.ModelLike,
-        optimizer::MOI.AbstractOptimizer,
+        optimizer::MOI.ModelLike,
     )
         @assert MOI.is_empty(optimizer)
         model = new{typeof(optimizer),typeof(cache)}(
--- a/src/Utilities/cachingoptimizer.jl
+++ b/src/Utilities/cachingoptimizer.jl
@@ -90,10 +90,7 @@ mutable struct CachingOptimizer{O,M<:MOI.ModelLike} <: MOI.AbstractOptimizer
         )
     end
 
-    function CachingOptimizer(
-        cache::MOI.ModelLike,
-        optimizer::MOI.ModelLike,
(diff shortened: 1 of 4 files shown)


blegat (approved):

blegat:
> This will need JuMP to lower-bound it's dependence on MOI

I still don't understand why this would while the other fix wouldn't since none of them require a change in JuMP

odow:
JuMP doesn't use instantiate:
https://github.com/jump-dev/JuMP.jl/blob/7fdd4d41f5275f2f5a0775f5d89252c689128413/src/file_formats.jl#L51-L55
So when we switch to `MOI.instantiate(MOI.FileFormats.XXX.Model)` it won't work on an earlier version of MOI

odow:
This does need a change in JuMP. #1780 fixes it on the MOI side, so it doesn't need a change in JuMP.

PR #1782, opened 2022-03-22
[Test] add a test for empty conic problem

### LINKED ISSUE #1741: Add test for conic problem with constraints but no variables
See this issue in ECOS: https://github.com/jump-dev/ECOS.jl/issues/132#issuecomment-1039861056

We should see what happens for models of the form:
```Julia
f = MOI.VectorAffineFunction(MOI.VectorAffineTerm{Float64}[], [1.0, 0.5, 0.5])
MOI.add_constraint(model, f, MOI.SecondOrderCone(3))
```

<s>They might solve, or they might error, but they shouldn't segfault.</s>

Test for `MOI.OPTIMAL` or `MOI.INVALID_MODEL` as the termination status.

odow:
Closes #1741

### DIFF
--- a/src/Test/test_conic.jl
+++ b/src/Test/test_conic.jl
@@ -6799,7 +6799,7 @@ function test_conic_empty_matrix(
     @requires _supports(config, MOI.optimize!)
     F = MOI.VectorAffineFunction{T}
     @requires MOI.supports_constraint(model, F, MOI.SecondOrderCone)
-    terms = MOI.VectorAffineTerm{Float64}[]
+    terms = MOI.VectorAffineTerm{T}[]
     MOI.add_constraint(
         model,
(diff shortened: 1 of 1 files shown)


src/Test/test_conic.jl
```diff
@@ -6791,3 +6791,37 @@ function setup_test(
...
 end
+
+function test_conic_empty_matrix(
+    model::MOI.ModelLike,
+    config::Config{T},
+) where {T}
+    @requires _supports(config, MOI.optimize!)
+    F = MOI.VectorAffineFunction{T}
+    @requires MOI.supports_constraint(model, F, MOI.SecondOrderCone)
+    terms = MOI.VectorAffineTerm{Float64}[]
```
odow, line 6802:
```suggestion
    terms = MOI.VectorAffineTerm{T}[]
```

PR #1783, opened 2022-03-23
Prep for v1.1.1

odow:
One question: is a new test a patch or minor release?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "MathOptInterface"
 uuid = "b8f27783-ece8-5eb3-8dc8-9495eed66fee"
-version = "1.1.0"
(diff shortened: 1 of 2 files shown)


blegat:
IIUC, it depends on whether we consider it a new feature or a bug-fix. Since we had to change the docstring of `instantiate`, it might be considered a new feature

joaquimg:
Testing existing functionality could be framed as patch, as bug that was not seen due missing tests...

blegat:
I agree for the testing changes. The contentious one is https://github.com/jump-dev/MathOptInterface.jl/pull/1781

odow:
I think it can be a bug fix. It was an error, now it works, and we probably always intended it to work.

PR #1787, opened 2022-03-30
Allow throwing from ZeroOne tests with bounds

blegat:
This is needed for Mosek as it uses the `Bridges.Constraint.ZeroOneBridge`.

### DIFF
--- a/src/Test/test_constraint.jl
+++ b/src/Test/test_constraint.jl
@@ -537,14 +537,20 @@ function test_constraint_ZeroOne_bounds(
     try
         MOI.add_constraint(model, x, MOI.GreaterThan(T(0)))
     catch err
-        @test err isa MOI.LowerBoundAlreadySet
-        return
+        if err isa MOI.LowerBoundAlreadySet
+            return
+        else
+            rethrow(err)
+        end
     end
     try
         MOI.add_constraint(model, x, MOI.LessThan(T(1)))
     catch err
-        @test err isa MOI.UpperBoundAlreadySet
-        return
+        if err isa MOI.UpperBoundAlreadySet
+            return
+        else
+            rethrow(err)
+        end
     end
     MOI.set(model, MOI.ObjectiveSense(), MOI.MAX_SENSE)
     f = T(2) * x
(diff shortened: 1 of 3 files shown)


test/Bridges/Constraint/zero_one.jl
```diff
@@ -120,6 +120,19 @@ function test_ZeroOne()
...
+function test_double_bounds()
+    mock = MOI.Utilities.MockOptimizer(
+        MOI.Utilities.UniversalFallback(MOI.Utilities.Model{Float64}()),
+    )
+    config = MOI.Test.Config()
+    bridged_mock = MOI.Bridges.Constraint.ZeroOne{Float64}(mock)
+    MOI.Test.test_constraint_ZeroOne_bounds(bridged_mock, config)
+    MOI.Test.test_constraint_ZeroOne_bounds_2(bridged_mock, config)
+    MOI.Test.test_constraint_ZeroOne_bounds_3(bridged_mock, config)
+    MOI.Test.test_variable_solve_ZeroOne_with_0_upper_bound(bridged_mock, config)
```
odow, line 132:
```suggestion
    MOI.Test.test_variable_solve_ZeroOne_with_0_upper_bound(
        bridged_mock,
        config,
    )
```

test/Bridges/Constraint/zero_one.jl
```diff
@@ -120,6 +120,19 @@ function test_ZeroOne()
...
+    MOI.Test.test_variable_solve_ZeroOne_with_upper_bound(bridged_mock, config)
```
odow, line 133:
```suggestion
    MOI.Test.test_variable_solve_ZeroOne_with_upper_bound(
        bridged_mock,
        config,
    )
```

odow (commented):
Hmm. I wish we had a better solution for this, but it's okay, I guess.

blegat:
Actually, we need to rethrow so that it goes throw the `_error_handler` in case it's an `UnsupportedConstraint` for instance

src/Test/test_constraint.jl
```diff
@@ -528,8 +528,24 @@ function test_constraint_ZeroOne_bounds(
     @requires MOI.supports_constraint(model, MOI.VariableIndex, MOI.ZeroOne)
     x = MOI.add_variable(model)
     MOI.add_constraint(model, x, MOI.ZeroOne())
-    MOI.add_constraint(model, x, MOI.GreaterThan(T(0)))
-    MOI.add_constraint(model, x, MOI.LessThan(T(1)))
+    try
+        MOI.add_constraint(model, x, MOI.GreaterThan(T(0)))
+    catch err
+        if err isa MOI.LowerBoundAlreadySet
+            return
```
odow, line 535: I don't think this is right? It will skip the rest of the test. 

Shouldn't it be:
```julia
if !(err isa MOI.LowerBoundAlreadySet)
    rethrow(err)
end
```

odow:
I think this is the wrong fix. The ZeroOne bridge should add `1x in [0, 1]`, instead of `x in [0, 1]`. It's the bridge that is broken, not the tests.

PR #1797, opened 2022-04-02
Improve FileFormats.SDPA.Model docstring and include it in doc

blegat:
Preview doc at https://jump.dev/MathOptInterface.jl/previews/PR1797/

Closes https://github.com/jump-dev/MathOptInterface.jl/issues/1796


odow (approved):
I guess we should add the docstrings for the other file formats as well.

PR #1799, opened 2022-04-07
MOI.modify version for multiple changes at once

guilhermebodin:
Many solvers support changing many coefficients at once. 

Some examples are functions in Gurobi (https://www.gurobi.com/documentation/9.5/refman/c_xchgcoeffs.html) and Xpress (https://www.fico.com/fico-xpress-optimization/docs/latest/solver/optimizer/HTML/XPRSchgmcoef.html).

This is a draft about making an MOI version of such methods and a proper fallback. Honestly, I don't know what would be the best approach but I am happy to make a PR with the correct implementation.

### DIFF
--- a/src/modifications.jl
+++ b/src/modifications.jl
@@ -102,8 +102,24 @@ end
 
 function modify(
     model::ModelLike,
+    cis::Vector{ConstraintIndex},
+    changes::Vector{AbstractFunctionModification},
+)
+    return throw_modify_not_allowed.(cis, changes)
+end
+
+function modify(
+    model::ModelLike,
     attr::ObjectiveFunction,
     change::AbstractFunctionModification,
 )
     return throw_modify_not_allowed(attr, change)
+end
+
+function modify(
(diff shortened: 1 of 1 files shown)


src/modifications.jl
```diff
@@ -100,10 +100,26 @@ function modify(
...
+
 function modify(
     model::ModelLike,
     attr::ObjectiveFunction,
     change::AbstractFunctionModification,
 )
     return throw_modify_not_allowed(attr, change)
 end
+
+function modify(
```
joaquimg, line 119: It could be like this: https://github.com/jump-dev/MathOptInterface.jl/blob/6b50fa0f9e31ef97a06facaad60af7d0b6ae3d2a/src/variables.jl#L171
```julia
function modify(
    model::ModelLike,
    attr::ObjectiveFunction,
    changes::AbstractVector{<:AbstractFunctionModification},
)
    for change in changes
        modify(model, attr, change)
    end
    return
end
```

PR #1807, opened 2022-04-20
[FileFormats.MPS] Fix to stop writing OBJSENSE by default

### LINKED ISSUE #1806: FileFormats.MPS: add option to not write OBJSENSE
Introduced in https://github.com/jump-dev/MathOptInterface.jl/pull/1391, but broke https://github.com/joaquimg/BilevelJuMP.jl/issues/170.

We should add an option to skip
https://github.com/jump-dev/MathOptInterface.jl/blob/6b50fa0f9e31ef97a06facaad60af7d0b6ae3d2a/src/FileFormats/MPS/MPS.jl#L174-L178
and negate the objective coefficients.

cc @joaquimg

odow:
Closes #1806

### DIFF
--- a/src/FileFormats/MPS/MPS.jl
+++ b/src/FileFormats/MPS/MPS.jl
@@ -41,9 +41,10 @@ end
 
 struct Options
     warn::Bool
+    objsense::Bool
 end
 
-get_options(m::Model) = get(m.ext, :MPS_OPTIONS, Options(false))
+get_options(m::Model) = get(m.ext, :MPS_OPTIONS, Options(false, true))
 
 """
     Model(; kwargs...)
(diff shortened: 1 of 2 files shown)


src/FileFormats/MPS/MPS.jl
```diff
@@ -53,10 +54,11 @@ Create an empty instance of FileFormats.MPS.Model.
 Keyword arguments are:
 
  - `warn::Bool=false`: print a warning when variables or constraints are renamed.
+ - `print_objsense::Bool=true`: print the OBJSENSE section when writing
```
joaquimg, line 57: Default is false

odow, line 57:
```suggestion
 - `print_objsense::Bool=false`: print the OBJSENSE section when writing
```

PR #1808, opened 2022-04-21
[FileFormats] add option for generic names

### LINKED ISSUE #1617: .rew model file support in FileFormats
From [Discourse](https://discourse.julialang.org/t/rew-file-format-in-moi-fileformats-jump/68238).

Would it be possible to include support for .rew model file formats in JuMP?

It is sometimes useful to share models where a risk of leaking proprietary information is present (e.g. here) and .mps files contain constraint/variables names where .rew doesn’t.

Format explanation from [CPLEX documentation](https://www.ibm.com/docs/en/icos/20.1.0?topic=cplex-other-file-formats): 
>This format is simply an MPS file format with all variable (column) and constraint (row) names converted to generic names. Variables are relabeled x1 through xn, and rows are renamed c1 through cm. This format may be useful, for example, for problems that you consider highly proprietary.

odow:
Closes #1617

### DIFF
--- a/test/FileFormats/MPS/MPS.jl
+++ b/test/FileFormats/MPS/MPS.jl
@@ -581,7 +581,6 @@ function test_generic_names()
     MOI.add_constraint(model, y, MOI.GreaterThan(2.0))
     @test sprint(write, model) ==
           "NAME          \n" *
-          "OBJSENSE MIN\n" *
           "ROWS\n" *
           " N  OBJ\n" *
           " E  R1\n" *
--- a/test/FileFormats/MPS/MPS.jl
+++ b/test/FileFormats/MPS/MPS.jl
@@ -597,6 +597,62 @@ function test_generic_names()
           "ENDATA\n"
 end
 
+function test_rew_filename()
+    model = MOI.FileFormats.Model(; filename = "test.rew")
+    x = MOI.add_variable(model)
+    y = MOI.add_variable(model)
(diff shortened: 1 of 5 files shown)


test/FileFormats/MPS/MPS.jl
```diff
@@ -569,6 +569,35 @@ function test_sos_constraints()
...
+    y = MOI.add_variable(model)
+    c = MOI.add_constraint(
+        model,
+        MOI.ScalarAffineFunction(MOI.ScalarAffineTerm.(1.0, [x, y]), 0.0),
+        MOI.EqualTo(1.0),
+    )
+    MOI.add_constraint(model, y, MOI.GreaterThan(2.0))
+    @test sprint(write, model) ==
+          "NAME          \n" *
+          "OBJSENSE MIN\n" *
```
odow, line 584:
```suggestion
```

nickrobinson251:
Thank you for adding this! I was just about to post a question on Discourse asking how to write a JuMP Model to the `.rew` format (https://www.gurobi.com/documentation/9.5/refman/rew_format.html), and this looks to fulfil that need.

Is it perhaps worth mentioning `.rew` in the docstring here, or somewhere in the docs, in order to make this new `generic_names` functionality discoverable by people looking for that format?

odow:
Added. From JuMP, you'll be able to go `write_to_file(model, "model.rew")`.

PR #1819, opened 2022-04-26
[ReverseAD] Add support for user-defined hessians

odow:
Refactored from https://github.com/jump-dev/JuMP.jl/pull/2953

### DIFF
--- a/src/Nonlinear/ReverseAD/forward_over_reverse.jl
+++ b/src/Nonlinear/ReverseAD/forward_over_reverse.jl
@@ -336,7 +336,10 @@ function _forward_eval_ϵ(
                     for (i, c) in enumerate(children_idx)
                         f_input[i] = ex.forward_storage[children_arr[c]]
                     end
-                    H = _UnsafeHessianView(d.user_output_buffer, n_children)
+                    H = _UnsafeLowerTriangularMatrixView(
+                        d.user_output_buffer,
+                        n_children,
+                    )
                     has_hessian = Nonlinear.eval_multivariate_hessian(
                         user_operators,
                         user_operators.multivariate_operators[node.index],
--- a/src/Nonlinear/ReverseAD/forward_over_reverse.jl
+++ b/src/Nonlinear/ReverseAD/forward_over_reverse.jl
@@ -343,9 +343,9 @@ function _forward_eval_ϵ(
                         H,
                         f_input,
                     )
-                    if !has_hessian
-                        continue
-                    end
+                    # This might be `false` if we extend this code to all
+                    # multivariate functions.
+                    @assert has_hessian
                     for col in 1:n_children
                         dual = zero(ForwardDiff.Partials{N,T})
                         for row in 1:n_children
--- a/src/Nonlinear/ReverseAD/forward_over_reverse.jl
+++ b/src/Nonlinear/ReverseAD/forward_over_reverse.jl
@@ -126,6 +126,7 @@ function _hessian_slice_inner(d, ex, input_ϵ, output_ϵ, ::Type{T}) where {T}
     for i in ex.dependent_subexpressions
         subexpr = d.subexpressions[i]
         subexpr_forward_values_ϵ[i] = _forward_eval_ϵ(
+            d,
             subexpr,
             _reinterpret_unsafe(T, subexpr.forward_storage_ϵ),
             _reinterpret_unsafe(T, subexpr.partials_storage_ϵ),
@@ -135,6 +136,7 @@ function _hessian_slice_inner(d, ex, input_ϵ, output_ϵ, ::Type{T}) where {T}
         )
     end
     _forward_eval_ϵ(
+        d,
         ex,
         _reinterpret_unsafe(T, d.forward_storage_ϵ),
         _reinterpret_unsafe(T, d.partials_storage_ϵ),
@@ -178,6 +180,7 @@ end
 
 """
     _forward_eval_ϵ(
+        d::NLPEvaluator,
         ex::Union{_FunctionStorage,_SubexpressionStorage},
         storage_ϵ::AbstractVector{ForwardDiff.Partials{N,T}},
         partials_storage_ϵ::AbstractVector{ForwardDiff.Partials{N,T}},
@@ -195,6 +198,7 @@ components separate so that we don't need to recompute the real components.
 This assumes that `_reverse_model(d, x)` has already been called.
 """
 function _forward_eval_ϵ(
+    d::NLPEvaluator,
     ex::Union{_FunctionStorage,_SubexpressionStorage},
     storage_ϵ::AbstractVector{ForwardDiff.Partials{N,T}},
     partials_storage_ϵ::AbstractVector{ForwardDiff.Partials{N,T}},
@@ -328,10 +332,35 @@ function _forward_eval_ϵ(
                         recip_denominator,
                     )
                 elseif op > 6
-                    error(
-                        "User-defined operators not supported for hessian " *
-                        "computations",
+                    f_input = _UnsafeVectorView(d.jac_storage, n_children)
+                    for (i, c) in enumerate(children_idx)
+                        f_input[i] = ex.forward_storage[children_arr[c]]
+                    end
+                    H = _UnsafeHessianView(d.user_output_buffer, n_children)
+                    has_hessian = Nonlinear.eval_multivariate_hessian(
+                        user_operators,
+                        user_operators.multivariate_operators[node.index],
+                        H,
+                        f_input,
                     )
+                    if !has_hessian
+                        continue
+                    end
+                    for col in 1:n_children
+                        dual = zero(ForwardDiff.Partials{N,T})
+                        for row in 1:n_children
+                            # Make sure we get the lower-triangular component.
+                            h = row >= col ? H[row, col] : H[col, row]
+                            # Performance optimization: hessians can be quite
+                            # sparse
+                            if !iszero(h)
+                                i = children_arr[children_idx[row]]
+                                dual += h * storage_ϵ[i]
+                            end
+                        end
+                        i = children_arr[children_idx[col]]
+                        partials_storage_ϵ[i] = dual
+                    end
                 end
             elseif node.type == Nonlinear.NODE_CALL_UNIVARIATE
                 @inbounds child_idx = children_arr[ex.adj.colptr[k]]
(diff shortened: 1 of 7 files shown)


src/Nonlinear/ReverseAD/utils.jl
```diff
@@ -60,6 +60,68 @@ function _UnsafeVectorView(x::Vector, N::Int)
...
+`d.jac_storage` so that we don't have to worry about the GC removing
+`d.jac_storage` while `_UnsafeHessianView` exists. This lets us use a `Ptr{T}`
+and create a struct that is `isbitstype` and therefore does not allocate.
+
+## Unsafe behavior
+
+`_UnsafeHessianView` is unsafe because it assumes that the vector `x` remains
+valid during the usage of `_UnsafeHessianView`.
+"""
+struct _UnsafeHessianView <: AbstractMatrix{Float64}
```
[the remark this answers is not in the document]
odow: `MOI.Nonlinear` could conceivably be used for non-float64 nonlinear. But `Nonlinear.ReverseAD` is pretty tied to things being `Float64`.

test/Nonlinear/ReverseAD.jl
```diff
@@ -318,13 +346,43 @@ function test_hessian_sparsity_registered_rosenbrock()
...
-    # H = fill(NaN, 3)
-    # MOI.eval_hessian_lagrangian(evaluator, H, [1.0, 1.0], 1.5, Float64[])
-    # @test H == 1.5 .* [802, 200, -400]
+    @test :Hess in MOI.features_available(evaluator)
+    MOI.initialize(evaluator, [:Grad, :Jac, :Hess])
+    @test MOI.hessian_lagrangian_structure(evaluator) ==
+          [(1, 1), (2, 2), (2, 1)]
+    H = fill(NaN, 3)
+    MOI.eval_hessian_lagrangian(evaluator, H, [1.0, 1.0], 1.5, Float64[])
+    @test H == 1.5 .* [802, 200, -400]
```
frapac, line 355: Maybe specify directly an array of `Float64`:
```julia
@test H == 1.5 .* [802.0, 200.0, -400.0]
```

src/Nonlinear/ReverseAD/utils.jl
```diff
@@ -60,6 +60,68 @@ function _UnsafeVectorView(x::Vector, N::Int)
...
+end
+
+function _UnsafeHessianView(x::Vector, N::Int)
+    z = div(N * (N + 1), 2)
+    if length(x) < z
+        resize!(x, z)
+    end
+    for i in 1:z
+        x[i] = 0.0
+    end
```
frapac, line 121: How about using directly `fill!(x, 0.0)`?

frapac (commented):
Nice! As far as I understand, we assume by default that the Hessian of a custom operator is dense, right?

If so, would it be possible to emphasize this in the documentation (I can do it if that helps)?

src/Nonlinear/ReverseAD/utils.jl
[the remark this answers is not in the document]
odow: Done. That's a better name.

src/Nonlinear/ReverseAD/utils.jl
```diff
@@ -60,6 +60,68 @@ function _UnsafeVectorView(x::Vector, N::Int)
...
+
+function Base.setindex!(x::_UnsafeHessianView, value, i, j)
+    unsafe_store!(x.ptr, value, _linear_index(i, j))
+    return value
+end
+
+function _UnsafeHessianView(x::Vector, N::Int)
+    z = div(N * (N + 1), 2)
+    if length(x) < z
+        resize!(x, z)
```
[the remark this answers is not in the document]
odow: This is an internal method from inside ReverseAD, and we own the memory, so resizing is fine.

test/Nonlinear/ReverseAD.jl
odow: Done

src/Nonlinear/ReverseAD/utils.jl
odow: I don't think there is a difference? The loop is simpler because `fill!` actually calls a bunch of other methods like `size` etc.

src/Nonlinear/ReverseAD/mathoptinterface_api.jl
```diff
@@ -7,8 +7,10 @@
 function MOI.features_available(d::NLPEvaluator)
     # Check if we have any user-defined multivariate operators, in which case we
```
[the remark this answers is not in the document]
odow: Updated.

test/Nonlinear/ReverseAD.jl
```diff
@@ -318,13 +346,43 @@ function test_hessian_sparsity_registered_rosenbrock()
...
-    @test_broken :Hess in MOI.features_available(evaluator)
-    # TODO(odow): re-enable these tests when user-defined hessians are supported
-    # MOI.initialize(evaluator, [:Grad, :Jac, :Hess])
-    # @test MOI.hessian_lagrangian_structure(evaluator) == [(1, 1), (2, 2), (2, 1)]
-    # H = fill(NaN, 3)
-    # MOI.eval_hessian_lagrangian(evaluator, H, [1.0, 1.0], 1.5, Float64[])
-    # @test H == 1.5 .* [802, 200, -400]
+    @test :Hess in MOI.features_available(evaluator)
+    MOI.initialize(evaluator, [:Grad, :Jac, :Hess])
+    @test MOI.hessian_lagrangian_structure(evaluator) ==
```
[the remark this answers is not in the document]
odow: Yes, the Hessian of a user-defined function is dense. We can revisit this in future if it becomes a problem, but I don't expect people to be writing very large models where it makes a difference. They can always just not provide a Hessian and be in the same position we are today, so this is strictly an improvement over what we currently have.

src/Nonlinear/ReverseAD/utils.jl
frapac: oh ok, that makes sense then!

odow:
So I did some benchmarking of this:
```Julia
julia> using JuMP, Ipopt

julia> function main(N)
           set_silent(model)
           @variable(model, x[1:N])
           function foo(x...)
               return sum((x[i] - 1.1*i)^2 for i in 1:N)
           end
           register(model, :foo, N, foo; autodiff = true)
           @NLobjective(model, Min, foo(x...))
           optimize!(model)
           return value.(x)
       end

julia> main(10); # compilation

julia> for i in [10, 100, 1_000, 10_000]
           GC.gc()
           @time main(i)
       end
  0.003252 seconds (1.53 k allocations: 98.891 KiB)
  0.830141 seconds (3.08 M allocations: 175.975 MiB, 3.97% gc time)
  0.186039 seconds (591.61 k allocations: 153.440 MiB, 10.86% gc time)
  7.209030 seconds (33.99 M allocations: 13.457 GiB, 12.40% gc time)

julia> function main_1(N)
           set_silent(model)
           @variable(model, x[1:N])
           function foo(x...)
               return sum((x[i] - 1.1*i)^2 for i in 1:N)
           end
           function ∇foo(g, x...)
               for i in 1:N
                   g[i] = 2 * (x[i] - 1.1 * i)
               end
               return
           end
           register(model, :foo, N, foo, ∇foo)
           @NLobjective(model, Min, foo(x...))
           optimize!(model)
           return value.(x)
       end
main_1 (generic function with 1 method)

julia> main_1(10); # compilation

julia> for i in [10, 100, 1_000, 10_000]
           GC.gc()
           @time main_1(i)
       end
  0.002592 seconds (1.44 k allocations: 71.875 KiB)
  0.028930 seconds (80.08 k allocations: 4.595 MiB, 88.17% compilation time)
  0.032589 seconds (117.40 k allocations: 6.269 MiB, 75.08% compilation time)
  0.083687 seconds (495.45 k allocations: 23.409 MiB, 30.14% compilation time)

julia> function main_2(N)
           set_silent(model)
           @variable(model, x[1:N])
           function foo(x...)
               return sum((x[i] - 1.1*i)^2 for i in 1:N)
           end
           function ∇foo(g, x...)
               for i in 1:N
                   g[i] = 2 * (x[i] - 1.1 * i)
               end
               return
           end
           function ∇²foo(H, x...)
               for i in 1:N
                   H[i, i] = 2.0
               end
               return
           end
           register(model, :foo, N, foo, ∇foo, ∇²foo)
           @NLobjective(model, Min, foo(x...))
           optimize!(model)
           return value.(x)
       end
main_2 (generic function with 1 method)

julia> main_2(10); # compilation

julia> for i in [10, 100, 1_000] # , 10_000]
           GC.gc()
           @time main_2(i)
       end
  0.002018 seconds (2.67 k allocations: 162.812 KiB)
  0.047883 seconds (216.95 k allocations: 13.330 MiB, 71.50% compilation time)
  5.314729 seconds (13.42 M allocations: 822.473 MiB, 7.54% gc time, 0.63% compilation time)
```

As could be expected, it scales pretty poorly if you have large input vectors. But there's an easy escape hatch: just don't define the hessian and use first-order methods. This is really targeting problems which have lots of other structure, and a small user-defined function in one part that is disabling second-order information for the full problem.

src/Nonlinear/ReverseAD/forward_over_reverse.jl
```diff
@@ -328,10 +332,38 @@ function _forward_eval_ϵ(
...
                 elseif op > 6
-                    error(
-                        "User-defined operators not supported for hessian " *
-                        "computations",
+                    f_input = _UnsafeVectorView(d.jac_storage, n_children)
+                    for (i, c) in enumerate(children_idx)
+                        f_input[i] = ex.forward_storage[children_arr[c]]
+                    end
+                    H = _UnsafeLowerTriangularMatrixView(
+                        d.user_output_buffer,
```
mlubin, line 340: Where is it enforced that user_output_buffer is large enough?

odow: It calls this contractor: https://github.com/jump-dev/MathOptInterface.jl/pull/1819/files#diff-b6954ec7398499723bd82286762989c12564fda78890c07d49a3c1c06e0ea3a7R116

mlubin: So the constructor could break other unsafe views on the same array. That sounds extra unsafe... At least needs to be documented.

odow: Yes. For now though, the only usage of this is evaluating the user-defined hessians, and we use the `.user_output_buffer` as storage, so it's a safe usage of the unsafe behavior.

mlubin (approved):

PR #1821, opened 2022-04-27
Fix various performance issues

odow:
I'm investigating some performance problems with a facility location problem: https://github.com/jump-dev/JuMPPaperBenchmarks/blob/main/benchmark/facility.jl

```Julia
function _benchmark(n)
    model = Model(Gurobi.Optimizer)
    set_silent(model)
    solve_facility(model, n, n)
end
```

Before
```julia
julia> @time _benchmark(60);
Academic license - for non-commercial use only - expires 2022-06-19
  9.661598 seconds (57.29 M allocations: 4.051 GiB, 28.08% gc time)

julia> @time _benchmark(60);
Academic license - for non-commercial use only - expires 2022-06-19
  8.932473 seconds (57.29 M allocations: 4.051 GiB, 23.78% gc time)
```
After
```julia
julia> @time _benchmark(60);
Academic license - for non-commercial use only - expires 2022-06-19
  7.956889 seconds (47.67 M allocations: 3.210 GiB, 25.95% gc time)

julia> @time _benchmark(60);
Academic license - for non-commercial use only - expires 2022-06-19
  7.773976 seconds (47.67 M allocations: 3.210 GiB, 22.78% gc time)
```

### DIFF
--- a/src/Utilities/DoubleDicts.jl
+++ b/src/Utilities/DoubleDicts.jl
@@ -94,7 +94,10 @@ mutable struct IndexDoubleDictInner{F,S} <:
                AbstractDoubleDictInner{F,S,MOI.ConstraintIndex{F,S}}
     dict::Dict{Int64,Int64}
     function IndexDoubleDictInner{F,S}(d::IndexDoubleDict) where {F,S}
-        return new{F,S}(get!(d.dict, (F, S), Dict{Int64,Int64}()))
+        if !haskey(d.dict, (F, S))
+            d.dict[(F, S)] = Dict{Int64,Int64}()
+        end
+        return new{F,S}(d.dict[(F, S)])
     end
 end
 
--- a/src/Utilities/functions.jl
+++ b/src/Utilities/functions.jl
@@ -184,22 +184,25 @@ end
 # Functions
 
 function map_indices(index_map::F, f::MOI.VectorOfVariables) where {F<:Function}
-    return MOI.VectorOfVariables(index_map.(f.variables))
+    return MOI.VectorOfVariables([index_map(x) for x in f.variables])
 end
 
 function map_indices(
     index_map::F,
     f::Union{MOI.ScalarAffineFunction,MOI.VectorAffineFunction},
 ) where {F<:Function}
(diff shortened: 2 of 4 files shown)


src/Utilities/functions.jl
```diff
@@ -184,22 +184,25 @@ end
 # Functions
 
 function map_indices(index_map::F, f::MOI.VectorOfVariables) where {F<:Function}
-    return MOI.VectorOfVariables(index_map.(f.variables))
```
odow, line 187: Broadcasting is more of a problem for latency than runtime. But if we can get away with a generator, it's probably better.

blegat: Yes, I already noticed this. `MOI.VariableIndex[...]` is even faster but the different is not that big in my experience (at least in runtime, not in latency)

src/Utilities/DoubleDicts.jl
```diff
@@ -94,7 +94,10 @@ mutable struct IndexDoubleDictInner{F,S} <:
                AbstractDoubleDictInner{F,S,MOI.ConstraintIndex{F,S}}
     dict::Dict{Int64,Int64}
     function IndexDoubleDictInner{F,S}(d::IndexDoubleDict) where {F,S}
-        return new{F,S}(get!(d.dict, (F, S), Dict{Int64,Int64}()))
+        if !haskey(d.dict, (F, S))
+            d.dict[(F, S)] = Dict{Int64,Int64}()
+        end
+        return new{F,S}(d.dict[(F, S)])
```
joaquimg, line 100:
```julia
        return new{F,S}(get!(d.dict, (F, S)) do
            Dict{Int64,Int64}()
        )
```

joaquimg: this is possibly faster, because saves a haskey and does not create the dict if not needed.

odow: Yeah the problem previously was that it always created the dict. Your way needs an additional function compilation, which I guess is a trade-off against a key lookup? Anyway, it's a second-order consideration compared to creating a dict every time.

PR #1824, opened 2022-04-28
[docs] clarify usage of NLPBlockData

### LINKED ISSUE #1823: Nonlinear clarify role of Nonlinear.Model vs MOI.NLPBlockData
The two appear from the documentation as containers for the nonlinear parts of a model.

The difference seems to be that Model is iteratively constructed and then passed to the NLPBlockData which is not supposed to be modified anymore (since it doesn't have a add, delete API).

If the model can still be modified after adding the NLPBlockData, it means the Optimizer to which the NLPBlock is set should just hold onto the data but not process the nonlinear constraints yet and wait `MOI.optimize!` time for it (I don't think this is currently in the docs if that's the case)

odow:
cc @matbesancon does this close #1823?

### DIFF
--- a/docs/src/submodules/Nonlinear/overview.md
+++ b/docs/src/submodules/Nonlinear/overview.md
@@ -313,6 +313,9 @@ julia> model = MOI.Utilities.UniversalFallback(MOI.Utilities.Model{Float64}());
 
 julia> MOI.set(model, MOI.NLPBlock(), block);
 ```
+!!! warning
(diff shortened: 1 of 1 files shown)


matbesancon:
I would also add an explanation after the introduction of `Nonlinear.Model`

odow:
https://github.com/jump-dev/MathOptInterface.jl/pull/1803 adds more documentation. I'll clarify in that PR.

PR #1825, opened 2022-05-01
Add support for Constraint Programming

odow:
Part of https://github.com/jump-dev/MathOptInterface.jl/issues/1805

### DIFF
--- a/src/sets.jl
+++ b/src/sets.jl
@@ -1267,6 +1267,40 @@ function Base.:(==)(x::CountAtLeast, y::CountAtLeast)
     return x.n == y.n && x.partitions == y.partitions && x.set == y.set
 end
 
+"""
+    CountGreaterThan(dimension::Int)
+
+The set ``\\{(c, y, x) \\in \\mathbb{Z}^{1+1+d}\\}`` such that `c` is strictly
+greater than the number of occurances of `y` in `x`.
+
+## Also known as
+
+This constraint is called `count_gt` in MiniZinc.
+
+## Example
+
+```julia
+model = Utilities.Model{Float64}()
+c, _ = add_constrained_variable(model, Integer())
+y, _ = add_constrained_variable(model, Integer())
+x = [add_constrained_variable(model, Integer())[1] for _ in 1:3]
+add_constraint(model, VectorOfVariables([c; y; x]), CountGreaterThan(5))
+```
+"""
+struct CountGreaterThan <: AbstractVectorSet
+    dimension::Int
+    function CountGreaterThan(dimension::Base.Integer)
+        if dimension < 2
+            throw(
+                DimensionMismatch(
+                    "Dimension of CountGreaterThan must be >= 2.",
+                ),
+            )
+        end
+        return new(dimension)
+    end
+end
+
 # isbits types, nothing to copy
 function Base.copy(
     set::Union{
@@ -1305,6 +1339,7 @@ function Base.copy(
         CountDistinct,
         Among,
         CountAtLeast,
+        CountGreaterThan,
     },
 )
     return set
--- a/src/sets.jl
+++ b/src/sets.jl
@@ -1143,6 +1143,54 @@ struct AllDifferent <: AbstractVectorSet
     end
 end
 
+"""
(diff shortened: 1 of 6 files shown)


odow:
I've marked this as DNMY, because it is the base of a stack of PRs for constraint programming. We can merge the others into here, and then wait a while before merging this into MOI proper.

src/sets.jl
```diff
@@ -1116,6 +1116,33 @@ struct Complements <: AbstractVectorSet
     end
 end
 
+"""
+    AllDifferent(dimension::Int)
+
+Constrain the elements of a vector-valued function so that no two elements take
+the same value.
```
odow, line 1123:
```suggestion
the same value and each element is integer.
```

src/sets.jl
```diff
@@ -1143,6 +1143,361 @@ struct AllDifferent <: AbstractVectorSet
...
+        end
+        return new(dimension)
+    end
+end
+
+"""
+    Path(from::Vector{Int}, to::Vector{Int})
+
+Given a graph comprised of a set of nodes `1..N` and a set of arcs `1..E`
+represented by an edge from node `from[i]` to node `to[i]`, `Path` constraints
```
blegat, line 1455:
```suggestion
represented by an edge from node `from[i]` to node `to[i]`, `Path` constrains
```

src/sets.jl
```diff
@@ -1143,6 +1143,361 @@ struct AllDifferent <: AbstractVectorSet
...
+end
+
+function Base.:(==)(x::BinPacking{T}, y::BinPacking{T}) where {T}
+    return x.capacity == y.capacity && x.weights == y.weights
+end
+
+"""
+    Cumulative(dimension::Int)
+
+The set ``\\{(s, d, r, b) \\in \\mathbb{Z}^{length(s)+length(d)+length(r)+1}\\}``
```
blegat, line 1361:
```suggestion
The set ``\\{(s, d, r, b) \\in \\mathbb{R}^{length(s)+length(d)+length(r)+1}\\}``
```

blegat: I would imagine that MIP formulations would support these to be any real numbers

src/sets.jl
```diff
@@ -1143,6 +1143,361 @@ struct AllDifferent <: AbstractVectorSet
...
+
+"""
+    Path(from::Vector{Int}, to::Vector{Int})
+
+Given a graph comprised of a set of nodes `1..N` and a set of arcs `1..E`
+represented by an edge from node `from[i]` to node `to[i]`, `Path` constraints
+the set
+``(s, t, ns, es) \\in (1..N)\\times(1..N)\\times\\{0,1\\}^N\\times\\{0,1\\}^E``,
+to form subgraph that is a path from node `s` to node `t`, where node `n` is in
+the path if `ns[n]` is `1`, and edge `e` is in the path if `es[e]` is `1`.
```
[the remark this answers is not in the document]
odow: It must be acyclic and traverse each selected node and edge exactly once.

src/sets.jl
```diff
@@ -1143,6 +1143,361 @@ struct AllDifferent <: AbstractVectorSet
...
+
+Base.copy(set::Table) = Table(copy(set.table))
+
+Base.:(==)(x::Table{T}, y::Table{T}) where {T} = x.table == y.table
+
+"""
+    Circuit(dimension::Int)
+
+The set ``\\{x \\in \\{1..d\\}^d\\}`` that constraints ``x`` to be a circuit,
+such that ``x_i = j`` means that ``j`` is the successor of ``i``.
```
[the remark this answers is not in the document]
odow: No. Clarified this in the new docstring.

src/sets.jl
odow: I think most users and solvers of constraint programming in MOI are going to use `Model{Int}`, so I guess this becomes a moot point.

odow:
@blegat okay to merge? Then I'll start on the easy bridges.

odow:
Merging because I think we're pretty happy. We can always make changes in follow-up PRs.

PR #1826, opened 2022-05-01
Add CountDistinct set

odow:
Part of #1805 

Stacked on #1825 to avoid conflicts.

Issue #1805, opened 2022-04-20
Constraint programming support


odow:
This issue is to track support for a limited subset of constraint programming in MOI. At present, we don't envisage MOI having the same scope as MiniZinc, but we want users to be able to write down common constraint programs and solve them using a specialized constraint programming solver, or a MIP solver via bridges.

## Background

 * https://github.com/jump-dev/JuMP.jl/issues/2227
 * https://github.com/JuliaConstraints/ConstraintProgrammingExtensions.jl

## MathOptInterface

Add new sets to MOI. Definitions taken from: https://www.minizinc.org/doc-2.5.5/en/lib-globals.html

- [x] `all_different`:  #1825
   <img width="692" alt="image" src="https://user-images.githubusercontent.com/8177701/164132041-d442f21c-b9bc-465d-968e-985b89ca1c9e.png">
- [x] `nvalue`   #1826
   <img width="692" alt="image" src="https://user-images.githubusercontent.com/8177701/164132239-7a4f81f9-a741-4310-8d19-a82c0e0764df.png">
- [x] `among` #1827
   <img width="692" alt="image" src="https://user-images.githubusercontent.com/8177701/164132200-143f7df0-03b2-4b95-9d09-01ee99c73f3b.png">
- [x] `at_least`  #1828
   <img width="692" alt="image" src="https://user-images.githubusercontent.com/8177701/164132266-55818ed6-6616-4405-b5ae-2576cf150c7b.png">
- [x] `count_gt` #1829
   <img width="692" alt="image" src="https://user-images.githubusercontent.com/8177701/164132321-83b62a64-049b-47cc-9620-946dac5f7b99.png">
- [x] `bin_packing` https://github.com/jump-dev/MathOptInterface.jl/pull/1837
   <img width="711" alt="image" src="https://user-images.githubusercontent.com/8177701/164132364-a33ca518-ed1b-45ca-9fd0-dbb7c81dd3be.png">
- [x] `path` https://github.com/jump-dev/MathOptInterface.jl/pull/1837
   <img width="711" alt="image" src="https://user-images.githubusercontent.com/8177701/164132421-3de4905e-f835-4371-9be9-0e36df06bf16.png">
- [x] `cumulative` https://github.com/jump-dev/MathOptInterface.jl/pull/1837
   <img width="711" alt="image" src="https://user-images.githubusercontent.com/8177701/164132457-868ad658-8fc3-490b-b4c1-728ce994feab.png">
[...]

PR #1827, opened 2022-05-01
Add CountBelongs set

odow:
Part of #1805

Stacked on #1826 to avoid conflicts.

This set could use some bikeshedding. 

Minizinc calls it `among`, https://www.minizinc.org/doc-2.5.5/en/lib-globals.html, but what about:
 * `Count`
 * `CountAmong`
 * `CountIn`
 * `CountWithin`
 * `CountInSet`
 * `CountBelongs`

ConstraintProgrammingExtensions had a much more complicated setup, where there was `Count{<:AbstractSet}`, https://github.com/JuliaConstraints/ConstraintProgrammingExtensions.jl/blob/master/src/sets_count.jl. The generality was nice, but it would be quite hard to implement and might be too general.

### DIFF
--- a/src/sets.jl
+++ b/src/sets.jl
@@ -1186,6 +1186,36 @@ struct CountDistinct <: AbstractVectorSet
     end
 end
 
+"""
+    Among(dimension::Int, set::Set{Int})
+
+The set ``\\{(n, x) \\in \\mathbb{R}^{1+d}\\}`` such that `n` elements of the
+vector `x` take on of the values in `set`.
+
+This constraint is sometimes called `among`.
+
+## Example
+
+```julia
+model = Utilities.Model{Float64}()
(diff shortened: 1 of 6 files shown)


src/sets.jl
```diff
@@ -1186,6 +1186,36 @@ struct CountDistinct <: AbstractVectorSet
     end
 end
 
+"""
+    Among(dimension::Int, set::Set{Int})
+
+The set ``\\{(n, x) \\in \\mathbb{R}^{1+d}\\}`` such that `n` elements of the
```
blegat, line 1192:
```suggestion
The set ``\\{(n, x) \\in \\mathbb{R}^{1+dimension}\\}`` such that `n` elements of the
```

blegat:
I vote for `CountBelongs`

PR #1828, opened 2022-05-01
Add CountAtLeast set

odow:
Part of #1805

Stacked on #1827 to avoid conflicts.

PR #1829, opened 2022-05-01
Add CountGreaterThan set

odow:
Part of #1805

Stacked on #1828 to avoid conflicts.

### DIFF
--- a/src/sets.jl
+++ b/src/sets.jl
@@ -1265,6 +1265,38 @@ function Base.:(==)(x::CountAtLeast, y::CountAtLeast)
     return x.n == y.n && x.partitions == y.partitions && x.set == y.set
 end
 
+"""
+    CountGreaterThan(dimension::Int)
+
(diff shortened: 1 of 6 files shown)


src/sets.jl
```diff
@@ -1252,6 +1252,38 @@ function Base.:(==)(x::CountAtLeast, y::CountAtLeast)
     return x.dimension == y.dimension && x.n == y.n && x.set == y.set
 end
 
+"""
+    CountGreaterThan(dimension::Int)
+
+The set ``\\{(c, y, x) \\in \\mathbb{R}^{1+1+d}\\}`` such that `c` is strictly
```
odow, line 1258: So there's a decision whether we match MOI's epigraph-variable-first order, or MiniZinc's `x, y, c` order.

odow: Discussed with @blegat.

PR #1835, opened 2022-05-10
[FileFormats] add beginning of NL reader

odow:
* Expression tree parsing? Trivial.
 * Understanding the header and variable ordering? Insane. Not helped by the lack of documentation and the fact that AMPL doesn't generate NL files that make sense. I think some of the counters in the header are upper bounds on the storage needed.

```ampl
reset;
# continuous both
var x0 >= 0;
# discrete both
var x1 integer;
# continous constraint
var x2 <= 3.3, >= 2.2;
var x20 <= 3.3, >= 2.2;
var x21 <= 3.4, >= 2.3;
var x22 <= 3.5, >= 2.4;
var x23 <= 3.6, >= 2.5;
# discrete constraint
var x3 integer;
# continuous objective
var x4 == 4.4;
# discrete objective
var x5 integer;
# continous
var x6 <= 1.1;
# binary
var x7 binary;
# integer
var x8 integer;

minimize obj: 
	sin(x0 + x1) +
	sin(x4 + x5);
   
subject to nlcon: sin(x0 + x1) == sin(x2 + x3) + sin(x20 + x21 + x23 + x22);

option presolve 0;
write "gordering_test2";
```

produces 
```
g3 0 1 0	# problem ordering_test2
 10 1 1 0 1	# vars, constraints, objectives, ranges, eqns
 1 1	# nonlinear constraints, objectives
 0 0	# network constraints: nonlinear, linear
 8 10 2	# nonlinear vars in constraints, objectives, both
 0 0 0 1	# linear network variables; functions; arith, flags
 0 0 1 1 1	# discrete variables: binary, integer, nonlinear (b,c,o)
 8 4	# nonzeros in Jacobian, gradients
 0 0	# max name lengths: constraints, variables
 0 0 0 0 0	# common exprs: b,c,o,c1,o1
```
when there are clearly only 4 variables in the nonlinear objective, not 10...


odow:
Okay, this is in reasonable shape. 

I'm going to merge and then take a pass at unifying how we handle the reading and writing, rather than complicating this PR with changes to writing.

odow:
Okay, this is in reasonable shape. 

I'm going to merge and then take a pass at unifying how we handle the reading and writing, rather than complicating this PR with changes to writing.

Issue #1831, opened 2022-05-04
Missing attribute `ConstraintDual` for `ZerosBridge`


guilhermebodin:
This code gives the following error
```julia
using JuMP
using ECOS

model = Model(() -> ECOS.Optimizer())
@variable(model, x[1:2])
@variable(model, t[1:2])
@variable(model, p == 1)
@constraint(model, con1, [t[1], x[1] + p, x[1]] in SecondOrderCone())
@constraint(model, con2, [t[2], 2x[2] + p, x[2]] in SecondOrderCone())
@objective(model, Min, sum(t))
optimize!(model)
dual(FixRef(p))

julia> dual(FixRef(p))
ERROR: ArgumentError: Bridge of type `MathOptInterface.Bridges.Variable.ZerosBridge{Float64}` does not support 
accessing the attribute `MathOptInterface.ConstraintDual(1)`. If you encountered this error unexpectedly, it probably means your model has been reformulated using the bridge, and you are attempting to query an attribute that we haven't implemented yet for this bridge. Please open an issue at https://github.com/jump-dev/MathOptInterface.jl/issues/new and provide a reproducible example explaining what you were trying to do.
```

I was testing querying duals of parameters in some conic sets.

odow:
The problem is `p == 1`. See the discussion here: https://github.com/jump-dev/MathOptInterface.jl/issues/1767

> We can revisit if we get anyone encountering this issue in the wild.

I guess this now applies.

odow:
There's another work-around that would require adding methods to JuMP:
```Julia
julia> using JuMP

julia> using ECOS

julia> remove_bridge(model::Model, B) = remove_bridge(backend(model), B)
remove_bridge (generic function with 3 methods)

julia> remove_bridge(model::MOI.Utilities.CachingOptimizer, B) = remove_bridge(model.optimizer, B)
remove_bridge (generic function with 3 methods)

julia> remove_bridge(model::MOI.Bridges.AbstractBridgeOptimizer, B) = MOI.Bridges.remove_bridge(model, B)
remove_bridge (generic function with 3 methods)

julia> model = Model(ECOS.Optimizer)
A JuMP Model
Feasibility problem with:
Variables: 0
Model mode: AUTOMATIC
CachingOptimizer state: EMPTY_OPTIMIZER
Solver name: ECOS

julia> remove_bridge(model, MOI.Bridges.Variable.ZerosBridge{Float64})

[...]

PR #1838, opened 2022-05-12
[Bridges] improve error message for ZerosBridge and ConstraintDual

odow:
Part of #1831 

cc @guilhermebodin

### DIFF
--- a/src/Bridges/Variable/zeros.jl
+++ b/src/Bridges/Variable/zeros.jl
@@ -76,7 +76,7 @@ function MOI.get(
 ) where {T}
     return error(
         "Unable to query the dual of a variable bound that was reformulated " *
-        "using `ZerosBridge`. This usually arises in conic models when a " *
+        "using `ZerosBridge`. This usually arises in conic solvers when a " *
         "variable is fixed to a value. As a work-around, instead of creating " *
         "a fixed variable using variable bounds like `p == 1`, add an affine " *
         "equality constraint like `1 * p == 1` (or `[1 * p - 1,] in Zeros(1)`).",
--- a/src/Bridges/Variable/zeros.jl
+++ b/src/Bridges/Variable/zeros.jl
@@ -71,6 +71,20 @@ end
 
 function MOI.get(
     ::MOI.ModelLike,
+    ::MOI.ConstraintDual,
+    ::ZerosBridge{T},
+) where {T}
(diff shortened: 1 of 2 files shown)


test/Bridges/Variable/zeros.jl
```diff
@@ -127,17 +127,13 @@ function test_zeros()
...
-        "the attribute `$attr`. If you encountered this error " *
-        "unexpectedly, it probably means your model has been " *
-        "reformulated using the bridge, and you are attempting to query " *
-        "an attribute that we haven't implemented yet for this bridge. " *
-        "Please open an issue at https://github.com/jump-dev/MathOptInterface.jl/issues/new " *
-        "and provide a reproducible example explaining what you were " *
-        "trying to do.",
+    err = ErrorException(
+        "Unable to query the dual of a variable bound that was reformulated " *
+        "using `ZerosBridge`. This usually arises in conic models when a " *
```
blegat, line 133: Why conic models ? Because conic solvers do not implement variable bounds ?

odow: I put "usually" because most people are going to hit this in JuMP using a conic solver like SCS and `p == 1`. Even if it can happen in other cases, making the error more explanatory for the common case seems useful.

blegat: Then I would say `conic solvers`, not conic model. If you solve an LP or QP with SCS, you are going to get this error.

src/Bridges/Variable/zeros.jl
```diff
@@ -69,6 +69,20 @@ function MOI.get(
...
 end
 
+function MOI.get(
+    ::MOI.ModelLike,
+    ::MOI.ConstraintDual,
+    ::ZerosBridge{T},
+) where {T}
+    return error(
+        "Unable to query the dual of a variable bound that was reformulated " *
+        "using `ZerosBridge`. This usually arises in conic models when a " *
```
odow, line 79:
```suggestion
        "using `ZerosBridge`. This usually arises in conic solvers when a " *
```

test/Bridges/Variable/zeros.jl
odow, line 133:
```suggestion
        "using `ZerosBridge`. This usually arises in conic solvers when a " *
```

PR #1840, opened 2022-05-12
Doc feasibility sense behavior

### LINKED ISSUE #1832: Document the expected behavior with feasibility sense
Min and Max are less problematic, but it is less clear whether feasibility should erase the current objective function

matbesancon:
Closes #1832 
The second part is still unclear so not yet added to the docstring


src/attributes.jl
```diff
@@ -959,6 +959,7 @@ attribute_value_type(::Name) = String
 A model attribute for the objective sense of the objective function, which
 must be an `OptimizationSense`: `MIN_SENSE`, `MAX_SENSE`, or
 `FEASIBILITY_SENSE`. The default is `FEASIBILITY_SENSE`.
+When setting the sense as `FEASIBILITY_SENSE`, the objective function must be erased.
```
odow, line 962:
```suggestion

When setting the sense to `FEASIBILITY_SENSE`, the [`ObjectiveFunction`](@ref)
attribute must be erased.
```

matbesancon:
Then I think this is good like this, the second part is maybe not that crucial to add

odow (approved):

src/attributes.jl
```diff
@@ -959,6 +959,9 @@ attribute_value_type(::Name) = String
 A model attribute for the objective sense of the objective function, which
 must be an `OptimizationSense`: `MIN_SENSE`, `MAX_SENSE`, or
 `FEASIBILITY_SENSE`. The default is `FEASIBILITY_SENSE`.
+
+When setting the sense to `FEASIBILITY_SENSE`, the [`ObjectiveFunction`](@ref)
+attribute must be erased.
```
blegat, line 964: This isn't clear. Should the MOI implementation erase it or the user ?

PR #1841, opened 2022-05-16
Improve docstring of ObjectiveSense

matbesancon:
ping @blegat @odow 
you're right passive voice was not helping there


src/attributes.jl
```diff
@@ -960,8 +960,7 @@ A model attribute for the objective sense of the objective function, which
 must be an `OptimizationSense`: `MIN_SENSE`, `MAX_SENSE`, or
 `FEASIBILITY_SENSE`. The default is `FEASIBILITY_SENSE`.
 
-When setting the sense to `FEASIBILITY_SENSE`, the [`ObjectiveFunction`](@ref)
-attribute must be erased.
+When setting the sense to `FEASIBILITY_SENSE`, the optimizer must erase the [`ObjectiveFunction`](@ref) attribute.
```
blegat, line 963: I would add what it means, e.g., if you set `FEASIBILITY_SENSE` and the `MIN_SENSE` then the objective function should be zero.

odow:
How's this now?

matbesancon:
looks good

PR #1847, opened 2022-05-20
Docs for quadratic representation

joaquimg:
I get this question often...
I even considered adding the very same text of the functions docs in the manual.


odow:
Any and all improvements would be good. It can be quite confusing...

Issue #1504, opened 2021-08-02
Explain bridges in the documentation (better)


odow:
Since they're so complicated, I wonder if we need documentation in the style of literate programming where we explain in much more detail what's going on.

@matbesancon this blog post is a great start at explaining the bridges: 
https://matbesancon.github.io/post/2019-09-12-bridging-indicator/

Do you have any interest (or opposition) into adding something like it to the MOI documentation?

This is another one that could be included:
https://matbesancon.github.io/post/2019-11-04-binary-julia/

(Suggestion from @glebbelov.)

matbesancon:
@Wikunia also had some explanations on building bridges:
https://opensourc.es/blog/constraint-solver-bridges/
https://opensourc.es/blog/generic-bridges/

Wikunia:
I assume @dourouc05 would also be able to contribute here as he has 50+ bridges in https://github.com/dourouc05/ConstraintProgrammingExtensions.jl

dourouc05:
I can probably do a PR for this tomorrow, if you're interested. 

@Wikunia Closer to 70, right now :p! Even though that's a lot, most of these bridges are not really varied. 

I would have a question about the best practices: when a bridge declares some constrained variable types (`added_constrained_variable_types`), should these types be repeated in `added_constraint_types` (`SingleVariable`-in-whatever)?

dourouc05:
Here is the beginning of the tutorial: #1506. Unlike the blog posts that have been posted here, I tried to follow the principle of separating explanation of the concepts (manual) and the actual way of writing a bridge (tutorial).

dourouc05:
Now that #1506 is merged, is anything else required for this issue?

odow:
I'd like to keep this issue open. We still need to document the hyper graphs and bellman-ford algorithm etc.

blegat:
> I would have a question about the best practices: when a bridge declares some constrained variable types (`added_constrained_variable_types`), should these types be repeated in `added_constraint_types` (`SingleVariable`-in-whatever)?

No, `add_constraint_types` should only contain the type of constraints added with `add_constraint`, not those added with `add_constrained_variable`. Otherwise the bridge graph may not see possible bridging path and report unsupported constraints while they are feasible. Maybe this should be clarified in the doc.

odow:
I've been making a bunch of improvements here:

- #1876
- #1875 
- #1874 
- #1872 
- #1870 
- #1868 
- #1869
- #1867
[...]

PR #1850, opened 2022-05-24
[Bridges] improve formatting and content of docstrings

odow:
In the spirit of #1504, this PR tidies up various docstrings in `Bridges`.

### DIFF
--- a/docs/src/submodules/Bridges/implementation.md
+++ b/docs/src/submodules/Bridges/implementation.md
@@ -7,67 +7,80 @@ end
 DocTestFilters = [r"MathOptInterface|MOI"]
 ```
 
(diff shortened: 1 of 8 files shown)


odow:
There's a few quirks I found. 

 * We have the previously undocumented `supports_objective_function(model, F)` instead of the more obvious `MOI.supports(model, MOI.ObjectiveFunction{F}())`. This is [used by StochasticPrograms.jl](https://github.com/martinbiel/StochasticPrograms.jl/blob/cb38c4ca70366b770017c7225324bc19a4080c2a/src/types/decisions/bridges/functionize.jl#L34-L37), so changing would be breaking.

PR #1852, opened 2022-05-24
[Bridges] re-organize the Variable bridges folder

odow:
Same reason as #1851. This found some bridges hiding at the bottom of `set_map.jl`, which was confusing.


odow:
@blegat, maybe obvious in retrospect, but it still surprised me: the tests failed because I altered the order in which we added the bridges, so a different bridging path was chosen...

PR #1855, opened 2022-05-24
[Bridges] remove old hack for Julia 1.2

odow:
Testing to see if this is still needed.


blegat (approved):

odow:
Seems like whatever it was is now fixed.

PR #1861, opened 2022-05-25
[Bridges] simplify flip_sign bridges

odow:
This PR duplicates some of the (very simple) code `Bridges/flip_sign.jl` to `Variable/flip_sign.jl` and `Constraint/flip_sign.jl` in order to make things easier to read. 

Currently, the definition of the bridges is spread across a couple of files, and 
```
_NonnegToNonposMap{T} =
    Union{Variable.NonposToNonnegBridge{T},Constraint.NonnegToNonposBridge{T}}
```
is a bit confusing because the order of the flip is different in the variable and constraint case.

The bridge files should be as self-contained as possible; reading them is already complicated enough.


odow:
Coverage miss is because `adjoint_map_function` isn't needed or tested for Variable bridges, but that can be a separate PR.

PR #1883, opened 2022-06-02
[Bridges] add unified testing infrastructure

odow:
This is the start of an exploration to see if we can simplify how we test bridges. 

Alternative to #820
Alternative to #750 

Things that aren't addressed:
 - [x] Objective bridges
 - [ ] ConstraintPrimal
 - [ ] ConstraintDual
 - [x] ConstraintPrimalStart
 - [x] ConstraintDualStart
 - [x] Deletion

Out of scope
 - Variable bridges
 - VariablePrimalStart
 - Modifications
 - Updates

### DIFF
--- a/src/Bridges/Constraint/bridges/geomean.jl
+++ b/src/Bridges/Constraint/bridges/geomean.jl
@@ -218,9 +218,9 @@ function MOI.get(
 end
 
 function MOI.get(
-    b::GeoMeanBridge{T,F,G},
-    ::MOI.NumberOfConstraints{G,MOI.Nonnegatives},
-)::Int64 where {T,F,G}
+    b::GeoMeanBridge{T,F,G,H},
+    ::MOI.NumberOfConstraints{H,MOI.Nonnegatives},
+)::Int64 where {T,F,G,H}
     return b.d > 2 ? 0 : 1
 end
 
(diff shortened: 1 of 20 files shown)


odow:
This is turning out to be pretty useful. It found a bunch of problems with constraint primal/dual starts in the bridges.

src/Bridges/Constraint/bridges/geomean.jl
```diff
@@ -218,9 +218,9 @@ function MOI.get(
 end
 
 function MOI.get(
-    b::GeoMeanBridge{T,F,G},
-    ::MOI.NumberOfConstraints{G,MOI.Nonnegatives},
-)::Int64 where {T,F,G}
+    b::GeoMeanBridge{T,F,G,H},
+    ::MOI.NumberOfConstraints{H,MOI.Nonnegatives},
+)::Int64 where {T,F,G,H}
```
odow, line 223: These tests found a number of nasty bugs like this one!

odow:
I'm a bit stuck on how to test constraint primal/dual. 

Any ideas? We test the start values by round-tripping them. It doesn't really check correctness, only that get/set are inverses of each other. The problem with primal/dual values is that they're one-way only.

odow:
Merging since this is clearly a win. We can sort the primal/dual and `Variable` bridge tests in other PRs.

src/Bridges/Constraint/bridges/geomean.jl
blegat: Nice catch, this is a good validation for the new tests

PR #1884, opened 2022-06-06
[Bridges] add constraint programming bridges

odow:
This is quite a naïve reformulation. We could make it stronger by computing a smaller set of bins (e.g., greedily assignment), and we could add ordering constraints on the bins, so that you can only use bin 2 if bin 1 is full.

### DIFF
--- a/src/Bridges/Constraint/bridges/bin_packing.jl
+++ b/src/Bridges/Constraint/bridges/bin_packing.jl
@@ -17,7 +17,7 @@
     \\sum\\limits_{i=1}^n w_i z_{ij} \\le c & \\forall j \\\\
     \\sum\\limits_{j=1}^n j z_{ij} == x_i   & \\forall i
     \\end{aligned}
-  ```
+    ```
 ## Source node
 
 `BinPackingToMILPBridge` supports:
--- a/src/Bridges/Constraint/bridges/bin_packing.jl
+++ b/src/Bridges/Constraint/bridges/bin_packing.jl
@@ -13,9 +13,9 @@
     ```math
     \\begin{aligned}
     z_{ij} \\in \\{0, 1\\}                  & \\forall i, j \\\\
-    \\sum\\limits_{j=1}^n z_{ij} = 1        & \\forall i \\\\
-    \\sum\\limits_{i=1}^n w_i z_{ij} \\le c & \\forall j \\\\
-    \\sum\\limits_{j=1}^n j z_{ij} == x_i   & \\forall i
+    \\sum\\limits_{j=1}^d z_{ij} = 1        & \\forall i \\\\
+    \\sum\\limits_{i=1}^d w_i z_{ij} \\le c & \\forall j \\\\
+    \\sum\\limits_{j=1}^d j z_{ij} == x_i   & \\forall i
     \\end{aligned}
     ```
 ## Source node
--- a/src/Bridges/Constraint/bridges/bin_packing.jl
+++ b/src/Bridges/Constraint/bridges/bin_packing.jl
@@ -10,7 +10,14 @@
 `BinPackingToMILPBridge` implements the following reformulation:
 
   * ``x \\in BinPacking(c, w)`` into
-
+    ```math
+    \\begin{aligned}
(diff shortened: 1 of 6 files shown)


odow:
New tests make this waaaay easier.

src/Bridges/Constraint/bridges/table.jl
```diff
@@ -0,0 +1,173 @@
...
+    MOI.delete(model, bridge.z)
+    return
+end
+
+function MOI.get(bridge::TableToMILPBridge, ::MOI.NumberOfVariables)::Int64
+    return length(bridge.z)
+end
+
+function MOI.get(bridge::TableToMILPBridge, ::MOI.ListOfVariableIndices)
+    return copy(bridge.z)
```
blegat, line 141: Don't think we need to copy, we don't copy in other bridges nor in MOI models

blegat:
Not sure ordering would work. You could have other constraints that break the symmetry so you cannot choose the order.
Is the number of bins equal to the number of items ? This seems to be assumed in the bridge but in the docs, it seems to be given by the bounds on the items.

src/Bridges/Constraint/bridges/bin_packing.jl
```diff
@@ -0,0 +1,207 @@
...
+"""
+    BinPackingToMILPBridge{T,F} <: Bridges.Constraint.AbstractBridge
+
+`BinPackingToMILPBridge` implements the following reformulation:
+
+  * ``x \\in BinPacking(c, w)`` into
+    ```math
+    \\begin{aligned}
+    z_{ij} \\in \\{0, 1\\}                  & \\forall i, j \\\\
+    \\sum\\limits_{j=1}^n z_{ij} = 1        & \\forall i \\\\
```
blegat, line 16:
```suggestion
    \\sum\\limits_{j=1}^d z_{ij} = 1        & \\forall i \\\\
```

blegat: It's called `d` in the docstring of `BinPacking`

src/Bridges/Constraint/bridges/table.jl
odow: All the other bridges `copy` or create new vectors.

odow:
These are working with JuMP and HiGHS:
```Julia
julia> using JuMP, HiGHS

julia> model = Model(HiGHS.Optimizer)
A JuMP Model
Feasibility problem with:
Variables: 0
Model mode: AUTOMATIC
Solver name: HiGHS

julia> set_silent(model)

julia> @variable(model, x[1:3])
3-element Vector{VariableRef}:
 x[1]
 x[2]
 x[3]

julia> @constraint(model, x in MOI.Table(Float64[0 1 1; 1 1 0; 1 0 1]))
[x[1], x[2], x[3]] ∈ MathOptInterface.Table{Float64}([0.0 1.0 1.0; 1.0 1.0 0.0; 1.0 0.0 1.0])

julia> optimize!(model)

julia> value.(x)
3-element Vector{Float64}:
 0.0
 1.0
 1.0
```

```Julia
julia> using JuMP, HiGHS

julia> model = Model(HiGHS.Optimizer)
A JuMP Model
Feasibility problem with:
Variables: 0
Model mode: AUTOMATIC
Solver name: HiGHS

julia> set_silent(model)

julia> @variable(model, x[1:3])
3-element Vector{VariableRef}:
 x[1]
 x[2]
 x[3]

julia> @constraint(model, x in MOI.BinPacking(3.0, [1.0, 2.0, 3.0]))
[x[1], x[2], x[3]] ∈ MathOptInterface.BinPacking{Float64}(3.0, [1.0, 2.0, 3.0])

julia> @objective(model, Min, sum(x))
x[1] + x[2] + x[3]

julia> optimize!(model)

julia> value.(x)
3-element Vector{Float64}:
 1.0
 1.0
 2.0
```

odow:
Ah. I mis-understood the definition of `BinPacking`:
image

I'll revert adding it to the `add_all_bridges` and clarify the docstring.

PR #1890, opened 2022-06-08
WIP: split tests in CI

odow:
Trying out what https://github.com/SciML/OrdinaryDiffEq.jl does.


odow:
Not sure this is worth it. The current tests take ~50 minutes to run on CI. This cuts it down to ~40, because we hit a limit on the number of concurrent jobs. 

We should really try and switch to a threaded system, where we run the MOI.Test, MOI.Bridges, and MOI.Utilities tests in parallel on the same machine. That would help a lot.

PR #1839, opened 2022-05-12
[Bridges] Fix bug in dual of PSDConeSquare bridge

odow:
Closes #1830

If there are no symmetry constraints added, then we can copy-across the duals from the upper triangle to the lower triangle.

PR #1891, opened 2022-06-09
[Bridges] tidy SquareBridge

odow:
There's still the giant (and infamous https://github.com/jump-dev/JuMP.jl/issues/1062) "Let S₊ be the cone of symmetric semidefinite matrices" comment, which is related to #1839. If we come to a conclusion I can tidy it up and add it to the docs properly.

### DIFF
--- a/src/Bridges/Constraint/bridges/square.jl
+++ b/src/Bridges/Constraint/bridges/square.jl
@@ -41,17 +41,13 @@
 # In the last primal program, we have the variables Z = X + Xᵀ and a upper triangular matrix S such that X = Z + S - Sᵀ
 
 """
-    SquareBridge{T, F<:MOI.AbstractVectorFunction,
-                 G<:MOI.AbstractScalarFunction,
-                 TT<:MOI.AbstractSymmetricMatrixSetTriangle,
-                 ST<:MOI.AbstractSymmetricMatrixSetSquare} <: AbstractBridge
+    SquareBridge{T,F,G,TT,ST} <: Bridges.Constraint.AbstractBridge
 
-The `SquareBridge` reformulates the constraint of a square matrix to be in `ST`
-to a list of equality constraints for pair or off-diagonal entries with
-different expressions and a `TT` constraint the upper triangular part of the
-matrix.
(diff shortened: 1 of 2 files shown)


test/Bridges/Constraint/square.jl
```diff
@@ -90,6 +90,34 @@ function test_Square()
...
+        [x11, x12, x12, x22] in PositiveSemidefiniteConeSquare(2)
+        """,
+        """
+        variables: x11, x12, x22
+        [x11, x12, x22] in PositiveSemidefiniteConeTriangle(2)
+        """,
+    )
+    return
+end
+
```
odow, line 119:
```suggestion
```

src/Bridges/Constraint/bridges/square.jl
```diff
@@ -104,31 +112,20 @@ function bridge_constraint(
...
-            MOIU.canonicalize!(diff)
+            f_ij, f_ji = f_scalars[i+(j-1)*dim], f_scalars[j+(i-1)*dim]
+            diff = MOI.Utilities.operate!(-, T, f_ij, f_ji)
+            MOI.Utilities.canonicalize!(diff)
             # The value 1e-10 was decided in https://github.com/jump-dev/JuMP.jl/pull/976
             # This avoid generating symmetrization constraints when the
             # functions at entries (i, j) and (j, i) are almost identical
-            if !MOIU.isapprox_zero(diff, 1e-10)
-                if MOIU.isapprox_zero(diff, 1e-8)
-                    @warn "The entries ($i, $j) and ($j, $i) of the" *
```
blegat, line 116: Why have you removed this ?

odow: I removed this for a number of reasons:
 * The warning wasn't tested
 * Why even have the warning? The user wrote something. They expect it to work. If it's a problem, they can explicitly form a symmetric matrix.
 * Is there any code in the wild that actually triggers this warning?
 * Why 1-e8 and 1e-10? That's a pretty narrow band. It's unlikely to trigger from any linear algebra calculations, since that's going to result in errors much smaller than 1e-10. But if it comes from the solution of another optimizer, the error is likely 1e-6, which won't trigger this warning.

PR #1904, opened 2022-06-15
[docs] Document square PSD duality convention

odow:
Mainly a formatted version of @blegat's comment. Did I get it right?

### DIFF
--- a/docs/src/background/duality.md
+++ b/docs/src/background/duality.md
@@ -204,7 +204,7 @@ If ``\mathcal{C}_i`` is a vector set, the discussion remains valid with
 ``y_i(\frac{1}{2}x^TQ_ix + a_i^T x + b_i)`` replaced with the scalar product
 between ``y_i`` and the vector of scalar-valued quadratic functions.
 
-## Dual for square semi-definite matrices
+## Dual for square semidefinite matrices
 
 The set [`PositiveSemidefiniteConeTriangle`](@ref) is a self-dual. That is,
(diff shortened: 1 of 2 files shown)


docs/src/background/duality.md
```diff
@@ -203,3 +203,115 @@ A pair of primal-dual variables $(x^\star, y^\star)$ is optimal if
...
+    If you have a [`PositiveSemidefiniteConeSquare`](@ref) constraint, the
+    result matrix ``A`` from [`ConstraintDual`](@ref) is not positive
+    semidefinite. However, ``A + A^\top`` is positive semidefinite.
+
+Let ``\mathcal{S}_+`` be the cone of symmetric semidefinite matrices in
+the ``\frac{n(n+1)}{2}`` dimensional space of symmetric ``\mathbb{R}^{n \times n}``
+matrices. That is, ``\mathcal{S}_+`` is the set [`PositiveSemidefiniteConeTriangle`](@ref).
+It is well known that ``\mathcal{S}_+`` is a self-dual proper cone.
+
+Let ``\mathcal{P}_+`` be the cone of symmetric semidefinite matrices in
```
blegat, line 228:
```suggestion
Let ``\mathcal{P}_+`` be the cone of matrices in
```

odow:
I think its still quite a confusing explanation, but we can leave as-is for now and come back to it the next time someone complains. It's at least better than hiding it in a code comment, and the tl;dr summarizes things.

PR #1909, opened 2022-06-16
[docs] improve documentation of nonlinear API

### LINKED ISSUE #1908: Document constraint_expr and objective_expr
It has become apparent that these aren't well-documented enough:

 * https://github.com/SciML/Optimization.jl/pull/278
 * https://github.com/SciML/Optimization.jl/pull/290

In particular:

 * The limitations of syntax
 * The need for `x[VariableIndex(i)]`
 * The need for the bound (in)equalities to match those passed to NLPBlockData.

odow:
Closes #1908

### DIFF
--- a/src/nlp.jl
+++ b/src/nlp.jl
@@ -51,7 +51,7 @@ An initial assignment of the Lagrange multipliers on the constraints from the
 struct NLPBlockDualStart <: AbstractModelAttribute end
 
 """
-    NLPBoundsPair(lower,upper)
+    NLPBoundsPair(lower, upper)
 
 A struct holding a pair of lower and upper bounds. `-Inf` and `Inf`
 can be used to indicate no lower or upper bound, respectively.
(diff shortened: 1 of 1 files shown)


src/nlp.jl
```diff
@@ -90,147 +98,307 @@ struct NLPBlockData
...
 
-Returns the subset of features available for this problem instance, as a
-vector of symbols in the same format as in `initialize`.
+See [`initialize`](@ref) for the list of defined features.
 """
 function features_available end
 
 """
-    eval_objective(d::AbstractNLPEvaluator, x)
+    eval_objective(d::AbstractNLPEvaluator, x::AbstractVector{Float64})::Float64
```
frapac, line 148: Should we assume that we are always working with `Float64`? I would prefer being more generic there, if possible. For instance:

```julia
eval_objective(d::AbstractNLPEvaluator, x::AbstractVector{T})::T where T 

```

src/nlp.jl
```diff
@@ -90,147 +98,307 @@ struct NLPBlockData
...
+`Vector{Float64}`, but you may assume that it supports `setindex!` and `length`.
+For example, it may be the `view` of a vector.
 """
 function eval_constraint end
 
 """
-    eval_objective_gradient(d::AbstractNLPEvaluator, df, x)
+    eval_objective_gradient(
+        d::AbstractNLPEvaluator,
+        df::AbstractVector{Float64},
```
frapac, line 174: I would suggest renaming `df` as `grad` (`g` being taken by the constraints).

src/nlp.jl
```diff
@@ -90,147 +98,307 @@ struct NLPBlockData
...
+    )::Nothing
+
+Evaluate the gradient of the objective function ``\\nabla f(x)`` as a dense
+vector, storing the result in the vector `df`.
 
-Evaluate ``\\nabla f(x)`` as a dense vector, storing the result in the vector
-`df` which must be of the appropriate size.
+## Implementation notes
+
+When implementing this method, you must not assume that `g` is
```
frapac, line 183: I assume `g` means `df` here?

src/nlp.jl
```diff
@@ -90,147 +98,307 @@ struct NLPBlockData
...
-`(row, column)`, where each indicates the position of a structurally nonzero element.
-These indices are not required to be sorted and can contain duplicates, in which
-case the solver should combine the corresponding elements by adding them together.
+Returns a vector of tuples, `(row, column)`, where each indicates the position
+of a structurally nonzero element in the Jacobian matrix:
+``J_g(x) = \\left[ \\begin{array}{c} \\nabla g_1(x) \\\\ \\nabla g_2(x) \\\\ \\vdots \\\\ \\nabla g_m(x) \\end{array}\\right],``
+where ``g_i`` is the ``i\\text{th}`` component of the nonlinear constraints
+``g(x)``.
+
+Thee indices are not required to be sorted and can contain duplicates, in which
```
frapac, line 198: Thee -> These?

src/nlp.jl
```diff
@@ -90,147 +98,307 @@ struct NLPBlockData
...
-    eval_constraint_jacobian_product(d::AbstractNLPEvaluator, y, x, w)
+    eval_constraint_jacobian_product(
+        d::AbstractNLPEvaluator,
+        y::AbstractVector{Float64},
+        x::AbstractVector{Float64},
+        w::AbstractVector{Float64},
+    )::Nothing
+
+Computes the Jacobian-vector product ``J_g(x)w``, storing the result in the
+vector `y`.
```
frapac, line 255: I would maybe precise the dimension of `w` (n_var) and of `y` (n_cons)

frapac (commented):
It overall looks good to me, apart of a few comments.

src/nlp.jl
frapac: JuliaSmoothOptimizer has solvers supporting `Float32`, and they developed a bridge between NLPModels.jl and MOI. But you are right: if everything is hard-coded to use `Float64` in `Nonlinear`, it would be misleading to change the documentation to be type-independent

odow: Okay, I've removed `Float64` from the docstrings. But there's a lot of other `Float64`s hard-coded into `nlp.jl` and `MOI.Nonlinear`, so I don't think we have a viable way of making them generic. There's nothing stopping the creation of a different `AbstractNLPEvaluator` that is generic though.

frapac (approved):

PR #1915, opened 2022-06-19
[Test] fix ZeroOne tests with lower and upper bounds

odow:
Xpress had a bug https://github.com/jump-dev/Xpress.jl/pull/172, which was because we tested `min x <= 0` instead of `max x <= 0`, so it didn't appropriately check the upper bound.

I've also added a test for the lower bound.

cc @rafabench


odow:
Ah, @rafabench points out that we should also check bounds after deleting `ZeroOne` constraints: https://github.com/jump-dev/Xpress.jl/pull/172#issuecomment-1159804058

PR #1920, opened 2022-06-27
[Bridges] add CountAtLeastToCountBelongsBridge

odow:
Continues #1919 and #1901
Part of #1905


odow:
Needs a rebase and change from Utilities.final_touch to Bridges.final_touch

PR #1922, opened 2022-06-27
[Utilities] remove copy in modify_function

### LINKED ISSUE #1921: Unnecessary copy in MOI.modify?
We have noticed that changing objective and constraint coefficients seem to be slow due to the line 'terms = copy(terms)' in the MOIU._modifycoefficient function defined in MOI.Utilities.functions.jl. 

The copy seems unnecessary when used in the functions below
```
MOI.modify(
    o::MOIU.ObjectiveContainer,
    ::MOI.ObjectiveFunction,
    change::MOI.AbstractFunctionModification,
)

function MOI.modify(
    v::MOIU.VectorOfConstraints{F,S},
    ci::MOI.ConstraintIndex{F,S},
    change::MOI.AbstractFunctionModification,
) where {F,S}


For example, in the body of MOI.modify(
    o::MOIU.ObjectiveContainer,
    ::MOI.ObjectiveFunction,
    change::MOI.AbstractFunctionModification,
)
```
the MOIU._modifycoefficient function is used under the hood in the call to this line

o.scalar_affine = modify_function(o.scalar_affine, change)

which overwrites the first input anyway, so the copy of terms does not seem necessary.

By removing the copy in MOIU._modifycoefficient, we find that update time goes down from 3 sec to 0.3 sec in our test case, due to less allocation and GC.

odow:
Closes #1921

Just seeing if how this fares in CI. We'd have to be careful here. 

From what I can tell, no one is using `modify_function` other than StochasticPrograms, which should be okay with modifying the function: https://juliahub.com/ui/Search?q=modify_function&type=code

I'll see if I can come up with an example where this would matter. My guess is something like a function being re-used in an objective and a constraint.

### DIFF
--- a/src/Utilities/functions.jl
+++ b/src/Utilities/functions.jl
@@ -1044,7 +1044,6 @@ function _modifycoefficient(
     variable::MOI.VariableIndex,
     new_coefficient::T,
 ) where {T}
-    terms = copy(terms)
     i = something(findfirst(t -> t.variable == variable, terms), 0)
     if iszero(i)  # The variable was not already in the function
         if !iszero(new_coefficient)
@@ -1092,7 +1091,6 @@ function _modifycoefficients(
     variable::MOI.VariableIndex,
     new_coefficients::Vector{Tuple{Int64,T}},
 ) where {T}
-    terms = copy(terms)
     coef_dict = Dict(k => v for (k, v) in new_coefficients)
     elements_to_delete = Int[]
     for (i, term) in enumerate(terms)
--- a/src/Utilities/functions.jl
+++ b/src/Utilities/functions.jl
@@ -1002,44 +1002,59 @@ end
 """
     modify_function(f::AbstractFunction, change::AbstractFunctionModification)
 
-Return a new function `f` modified according to `change`.
+Return a copy of the function `f`, modified according to `change`.
 """
 function modify_function(
+    f::MOI.AbstractFunction,
+    change::MOI.AbstractFunctionModification,
+)
+    new_f = copy(f)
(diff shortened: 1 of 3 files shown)


odow:
Actually, I think this _is_ safe, because we always copy the user's input when adding to `Model`.
https://github.com/jump-dev/MathOptInterface.jl/blob/fc89aa4c3b35cfc4fdadd6769617060f8340cce9/src/Utilities/objective_container.jl#L112
https://github.com/jump-dev/MathOptInterface.jl/blob/fc89aa4c3b35cfc4fdadd6769617060f8340cce9/src/Utilities/objective_container.jl#L125
https://github.com/jump-dev/MathOptInterface.jl/blob/fc89aa4c3b35cfc4fdadd6769617060f8340cce9/src/Utilities/objective_container.jl#L138
https://github.com/jump-dev/MathOptInterface.jl/blob/fc89aa4c3b35cfc4fdadd6769617060f8340cce9/src/Utilities/universalfallback.jl#L565
https://github.com/jump-dev/MathOptInterface.jl/blob/fc89aa4c3b35cfc4fdadd6769617060f8340cce9/src/Utilities/vector_of_constraints.jl#L80

odow:
@blegat are you okay with this?

blegat (approved):
I think before MOI.Utilities.Model didn't own the function which is why I was doing that. This is not the case anymore so it makes sense to me.

odow:
> In fact, maybe we should have a modify_function! and modify_function 

Ah. Yeah. Maybe this makes more sense.

blegat (approved):

PR #1923, opened 2022-06-27
[Bridges] add AllDifferentToCountDistinctBridge

odow:
Refactored from #1901
Part of #1805


odow:
Merging because it doesn't need final_touch.

PR #1924, opened 2022-06-28
[Bridges] update SemiToBinaryBridge using final_touch

odow:
Refactored from #1901. I'll link to @blegat's previous comments inline.

Closes #1095

### DIFF
--- a/src/Bridges/Constraint/bridges/semi_to_binary.jl
+++ b/src/Bridges/Constraint/bridges/semi_to_binary.jl
@@ -75,19 +75,6 @@ function bridge_constraint(
     f::MOI.VariableIndex,
     s::S,
 ) where {T<:Real,S<:Union{MOI.Semicontinuous{T},MOI.Semiinteger{T}}}
-    F = MOI.VariableIndex
-    if MOI.is_valid(model, MOI.ConstraintIndex{F,MOI.GreaterThan{T}}(f.value))
-        throw(MOI.LowerBoundAlreadySet{MOI.GreaterThan{T},S}(f))
-    end
-    if MOI.is_valid(model, MOI.ConstraintIndex{F,MOI.LessThan{T}}(f.value))
-        throw(MOI.UpperBoundAlreadySet{MOI.LessThan{T},S}(f))
-    end
-    if MOI.is_valid(model, MOI.ConstraintIndex{F,MOI.EqualTo{T}}(f.value))
-        throw(MOI.LowerBoundAlreadySet{MOI.EqualTo{T},S}(f))
-    end
-    if MOI.is_valid(model, MOI.ConstraintIndex{F,MOI.Interval{T}}(f.value))
-        throw(MOI.LowerBoundAlreadySet{MOI.Interval{T},S}(f))
-    end
     binary, binary_con = MOI.add_constrained_variable(model, MOI.ZeroOne())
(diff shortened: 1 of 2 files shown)


src/Bridges/Constraint/bridges/semi_to_binary.jl
```diff
@@ -327,3 +340,25 @@ function MOI.get(
...
+    if MOI.is_valid(model, MOI.ConstraintIndex{F,MOI.LessThan{T}}(f.value))
+        throw(MOI.UpperBoundAlreadySet{S,MOI.LessThan{T}}(f))
+    end
+    if MOI.is_valid(model, MOI.ConstraintIndex{F,MOI.EqualTo{T}}(f.value))
+        throw(MOI.LowerBoundAlreadySet{S,MOI.EqualTo{T}}(f))
+    end
+    if MOI.is_valid(model, MOI.ConstraintIndex{F,MOI.Interval{T}}(f.value))
+        throw(MOI.LowerBoundAlreadySet{S,MOI.Interval{T}}(f))
+    end
+    return
```
odow, line 363: I think this part of `final_touch` is uncontroversial (if we keep these errors). We need to throw the correct error, and currently solvers which used this bridge had failing tests: https://github.com/jump-dev/MathOptInterface.jl/pull/1095.

src/Bridges/Constraint/bridges/semi_to_binary.jl
```diff
@@ -75,6 +75,19 @@ function bridge_constraint(
...
+    end
+    if MOI.is_valid(model, MOI.ConstraintIndex{F,MOI.LessThan{T}}(f.value))
+        throw(MOI.UpperBoundAlreadySet{MOI.LessThan{T},S}(f))
+    end
+    if MOI.is_valid(model, MOI.ConstraintIndex{F,MOI.EqualTo{T}}(f.value))
+        throw(MOI.LowerBoundAlreadySet{MOI.EqualTo{T},S}(f))
+    end
+    if MOI.is_valid(model, MOI.ConstraintIndex{F,MOI.Interval{T}}(f.value))
+        throw(MOI.LowerBoundAlreadySet{MOI.Interval{T},S}(f))
+    end
```
odow, line 90: This part is more controversial: https://github.com/jump-dev/MathOptInterface.jl/pull/1901#discussion_r908153753

I added these because if a bound already exists, then we need to error that adding Semi-set isn't allowed. If we want until later, we don't know in which order they were added.

On reflection though, I think this is actually a bug in the `AbstractBridgeOptimizer`, which should catch this at the modeling level instead of the bridge (the bridge can do what it likes, it's just illegal to add at the model level).

odow:
The argument for `LowerBoundAlreadySet` and `UpperBoundAlreadySet` was:

 * It doesn't make sense to add a `GreaterThan` and a `Interval` constraint, since only one can exist. It makes even less sense to add `Semiinteger` and an `EqualTo`, since if you add the Semiinteger and then the EqualTo, if we updated the bounds in-place without deleting the Semiinnteger, we'd end up with `{0} \cup {value}`, which isn't EqualTo!

There's a separate question about whether calling `add_constraint` twice for the same set-type should be allowed, but that's a different question from adding different sets.

blegat (approved):
We can merge this for now and remove it along with the rest in case we remove these errors

PR #1934, opened 2022-07-01
Change Vector{T} to AbstractVector{T} in _extract_reverse_pass and add test

### LINKED ISSUE #1933: Consider changing Vector{T} to AbstractVector{T} in _extract_reverse_pass
Here: https://github.com/jump-dev/MathOptInterface.jl/blob/24d2a0d44a077df1b9a4f86f1c12ed397933932d/src/Nonlinear/ReverseAD/reverse_mode.jl#L324

So that `MOI.eval_objective_gradient(evaluator, g, x)` works for `g = view(v, idx)`.
We test it on https://github.com/JuliaSmoothOptimizers/NLPModelsJuMP.jl/issues/116.

abelsiqueira:
Closes #1933 

I added a small test related to the issue as well.

### DIFF
--- a/src/Nonlinear/ReverseAD/reverse_mode.jl
+++ b/src/Nonlinear/ReverseAD/reverse_mode.jl
@@ -312,7 +312,7 @@ end
 
 """
     _extract_reverse_pass(
-        g::Vector{T},
+        g::AbstractVector{T},
         d::NLPEvaluator,
         f::Union{_FunctionStorage,_SubexpressionStorage},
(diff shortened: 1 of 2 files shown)


odow (approved):
Hmm yeah we obviously didn't have tests for this. I didn't consider wanting to use view for the gradient since it's meant to be dense.

odow:
I checked and this seems like the only instance of `::Vector` in `MOI.Nonlinear`.

PR #1935, opened 2022-07-05
[Nonlinear] improve performance of Jacobian products

### LINKED ISSUE #1903: Optimize eval_constraint_jacobian_product
Originally started here: https://github.com/jump-dev/JuMP.jl/pull/2239

odow:
Closes #1903

cc @amontoison finally got around to doing this!

### DIFF
--- a/src/Nonlinear/ReverseAD/mathoptinterface_api.jl
+++ b/src/Nonlinear/ReverseAD/mathoptinterface_api.jl
@@ -216,13 +216,16 @@ function MOI.eval_constraint_jacobian(d::NLPEvaluator, J, x)
 end
 
 function MOI.eval_constraint_jacobian_product(d::NLPEvaluator, y, x, w)
+    _reverse_mode(d, x)
     fill!(y, 0.0)
-    J_struct = MOI.jacobian_structure(d)
-    nnz = length(J_struct)
-    J = zeros(nnz)
-    MOI.eval_constraint_jacobian(d, J, x)
-    for (k, (i, j)) in enumerate(J_struct)
-        y[i] += J[k] * w[j]
+    for (row, expr) in enumerate(d.constraints)
+        for col in expr.grad_sparsity
+            d.jac_storage[col] = 0.0
+        end
+        _extract_reverse_pass(d.jac_storage, d, expr)
+        for (k, col) in enumerate(expr.grad_sparsity)
+            y[row] += d.jac_storage[k] * w[col]
+        end
     end
     return
 end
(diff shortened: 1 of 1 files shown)


src/Nonlinear/ReverseAD/mathoptinterface_api.jl
```diff
@@ -216,13 +216,16 @@ function MOI.eval_constraint_jacobian(d::NLPEvaluator, J, x)
 end
 
 function MOI.eval_constraint_jacobian_product(d::NLPEvaluator, y, x, w)
+    _reverse_mode(d, x)
     fill!(y, 0.0)
-    J_struct = MOI.jacobian_structure(d)
-    nnz = length(J_struct)
-    J = zeros(nnz)
-    MOI.eval_constraint_jacobian(d, J, x)
```
odow, line 223: @blegat the main benefit is that we remove the allocations of `J_struct` and `J` from every call.

blegat: Yes but in https://github.com/jump-dev/JuMP.jl/pull/2239, @amontoison was mentioning a bit slowdown

odow: It's not the same implementation. This is purely a refactor that avoids some allocations. The run-time (excluding allocations) is the same.

blegat: Ok, I don't know the internals well enough to review but I don't have any objection then :)

src/Nonlinear/ReverseAD/mathoptinterface_api.jl
```diff
@@ -216,13 +216,16 @@ function MOI.eval_constraint_jacobian(d::NLPEvaluator, J, x)
 end
 
 function MOI.eval_constraint_jacobian_product(d::NLPEvaluator, y, x, w)
+    _reverse_mode(d, x)
```
frapac, line 219: I am not sure to understand why we need `_reverse_mode` here. I would say using forward mode would be more appropriate to evaluate Jacobian-product.

frapac (commented):
It overall looks good to me, and this PR is a welcome addition to `ReverseAD`. I just a question about forward/reverse mode: I would say forward is faster than reverse to evaluate the Jacobian-product, whereas reverse is faster to evaluate the Jacobian-transpose-product. How difficult it would be to use forward mode AD to evaluate the Jacobian-product?

src/Nonlinear/ReverseAD/mathoptinterface_api.jl
odow: `_reverse_mode` is a just a caller for the process which updates the gradients of each expression.

> I would say using forward mode would be more appropriate to evaluate Jacobian-product.

That's potentially true, but that's a much bigger change. For now, this is a clear win because it removes some allocations.

frapac: I see. I agree with you: if this is not direct to implement, we should merge this PR first. How about opening a separate issue to track this, or add a `TODO` comment in the code?

odow: I'll mark it as improves #1903, but not closes.

blegat:
@odow It seems this broke JuMP tests, see https://github.com/jump-dev/SolverTests/runs/7483984144?check_suite_focus=true#step:4:326

PR #1937, opened 2022-07-07
[Bridges] remove reduce_bridged in favor of explicit methods

odow:
Chipping away at #1504, and the complications are largely limited to `bridge_optimizer.jl`. This `reduce_bridged` method confuses me everything I read it, because it uses too many magic anonymous function calls.

### DIFF
--- a/src/Bridges/bridge_optimizer.jl
+++ b/src/Bridges/bridge_optimizer.jl
@@ -635,62 +635,7 @@ end
 
 # Attributes
 
-"""
-    function reduce_bridged(
-        b::AbstractBridgeOptimizer,
-        args,
-        F::Type{<:MOI.AbstractFunction},
-        S::Type{<:MOI.AbstractSet},
-        value::T,
-        operate_variable_bridges!,
-        operate_constraint_bridges!,
-    )::T where {T}
-
-If `F`-in-`S` constraints may be added to `b.model`, starts with
-`value = MOI.get(b.model, args...)`, otherwise, starts with `value`.
-
-Then:
- * if `F`-in-`S` constraints may correspond to bridged variables, modify it with
-   `operate_variable_bridges!`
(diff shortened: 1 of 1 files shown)


src/Bridges/bridge_optimizer.jl
```diff
@@ -770,42 +724,40 @@ function MOI.get(b::AbstractBridgeOptimizer, attr::MOI.NumberOfVariables)::Int64
...
+        num += MOI.get(b.model, attr)
+    end
+    # If variable bridged, get the indices from the variable bridges.
+    if is_variable_function && is_bridged(b, S) && is_variable_bridged(b, S)
+        num += Variable.number_with_set(Variable.bridges(b), S)
+    end
+    # If constraint bridged, get the indices from the constraint bridges.
+    if is_bridged(b, F, S) ||
+       (is_variable_function && supports_constraint_bridges(b))
+       num += Constraint.number_of_type(
```
odow, line 746:
```suggestion
        num += Constraint.number_of_type(
```

src/Bridges/bridge_optimizer.jl
```diff
@@ -1398,25 +1350,30 @@ function MOI.supports(
...
     IndexType::Type{MOI.ConstraintIndex{F,S}},
 ) where {F,S}
-    return reduce_bridged(
-        b,
-        F,
-        S,
-        true,
-        () -> MOI.supports(b.model, attr, IndexType),
-        ok ->
-            ok && MOI.supports(
```
blegat, line 1408: @odow Maybe that needed to be `||` and we can keep your translation below and just revert the `if` and `return` to reflect that

odow: Hmm. Potentially. I refactored these methods because they were impenetrable, and I really didn't understand what they were trying to do.

Issue #1936, opened 2022-07-05
Create attributes for absolute and relative gap tolerances


mtanneau:
[EDIT] following the discussion and the distinction between model and optimizer attributes, I have renamed the PR and updated the proposals.

___

MOI currently has a [`RelativeGap`](https://jump.dev/MathOptInterface.jl/stable/reference/models/#MathOptInterface.RelativeGap) attribute for relative optimality gap.

Several solvers also support an absolute gap tolerance, and it is not uncommon to use both relative and absolute gap tolerances.

If the feature gets a green light, I can PR the corresponding changes and docs.

odow:
I think that's reasonable. It's a bit easier to define a fallback for as well.

mtanneau:
Precision: I want to use this not just to query the absolute gap after the optimization is done, but also to set it as tolerance for the solver.

That latter use case may fall under `AbstractOptimizerAttribute` rather than `AbstractModelAttribute`, which is the current super type of `MOI.RelativeGap`

odow:
See my comment here: https://github.com/jump-dev/Gurobi.jl/pull/484#issuecomment-1175579115

These are conflating two separate things. Not sure what the best solution is.

Querying is straight forward, and I'd accept that as a PR under `AbstractModelAttribute`. Setting might need a larger discussion.

odow:
Linking to previous discussions:

 * https://github.com/jump-dev/MathOptInterface.jl/issues/1012
 * https://github.com/jump-dev/MathOptInterface.jl/issues/61

mtanneau:
I see. 
I can propose to introduce `RelativeGapTolerance` and `AbsoluteGapTolerance` as optimizer attributes.
Absolute gap tolerance should be relatively solver-independent :) Relative gap tolerance is solver-dependent, same as `RelativeGap` already is.

I personally use these two only in the context of MIP, although such tolerances may be used in global optimization, or iterative algorithms like Benders decomposition, Frank-Wolfe, column generation, etc.
___

For context, this is more-or-less how we use them internally (though technically we miss-used `RelativeGap`):
```julia
set_optimizer_attribute(model, MOI.RelativeGap(), 1e-3)
set_optimizer_attribute(model, AbsoluteGap(), 10_000)
```

Having this exposed at the MOI level would make our life easier, but we can keep our current workflow.

blegat:
[...]

PR #1944, opened 2022-07-11
Create RelativeGapTolerance and AbsoluteGapTolerance optimizer attributes

mtanneau:
Addresses #1936 

I have created a `RelativeGapTolerance` attribute. If that's satisfactory code- and doc-wise, I'll do the same with `AbsoluteGapTolerance`.

TODO
* [x] `RelativeGapTolerance` attribute
* [x] `AbsoluteGapTolerance` attribute
* [x] Docs update?

### DIFF
--- a/src/Test/test_attribute.jl
+++ b/src/Test/test_attribute.jl
@@ -205,8 +205,8 @@ function test_attribute_AbsoluteGapTolerance(
     value = MOI.get(model, MOI.AbsoluteGapTolerance())
     MOI.set(model, MOI.AbsoluteGapTolerance(), 1e-2)
     @test MOI.get(model, MOI.AbsoluteGapTolerance()) == 1e-2
-    MOI.set(model, MOI.AbsoluteGapTolerance(), 100)
-    @test MOI.get(model, MOI.AbsoluteGapTolerance()) == 100
+    MOI.set(model, MOI.AbsoluteGapTolerance(), 100.0)
+    @test MOI.get(model, MOI.AbsoluteGapTolerance()) == 100.0
     MOI.set(model, MOI.AbsoluteGapTolerance(), value)
     @test value == MOI.get(model, MOI.AbsoluteGapTolerance())
     _test_attribute_value_type(model, MOI.AbsoluteGapTolerance())
--- a/src/Test/test_attribute.jl
+++ b/src/Test/test_attribute.jl
@@ -196,7 +196,10 @@ end
 Test that the [`MOI.AbsoluteGapTolerance`](@ref) attribute is implemented for
 `model`.
 """
-function test_attribute_AbsoluteGapTolerance(model::MOI.AbstractOptimizer, ::Config)
+function test_attribute_AbsoluteGapTolerance(
+    model::MOI.AbstractOptimizer,
+    ::Config,
+)
     @requires MOI.supports(model, MOI.AbsoluteGapTolerance())
     # Get the current value to restore it at the end of the test
     value = MOI.get(model, MOI.AbsoluteGapTolerance())
@@ -226,7 +229,10 @@ end
 Test that the [`MOI.RelativeGapTolerance`](@ref) attribute is implemented for
 `model`.
 """
-function test_attribute_RelativeGapTolerance(model::MOI.AbstractOptimizer, ::Config)
+function test_attribute_RelativeGapTolerance(
+    model::MOI.AbstractOptimizer,
+    ::Config,
+)
     @requires MOI.supports(model, MOI.RelativeGapTolerance())
     # Get the current value to restore it at the end of the test
     value = MOI.get(model, MOI.RelativeGapTolerance())
--- a/src/Test/test_attribute.jl
+++ b/src/Test/test_attribute.jl
@@ -191,6 +191,66 @@ function setup_test(
 end
 
 """
+    test_attribute_AbsoluteGapTolerance(model::MOI.AbstractOptimizer, config::Config)
+
+Test that the [`MOI.AbsoluteGapTolerance`](@ref) attribute is implemented for
+`model`.
+"""
+function test_attribute_AbsoluteGapTolerance(model::MOI.AbstractOptimizer, ::Config)
+    @requires MOI.supports(model, MOI.AbsoluteGapTolerance())
+    # Get the current value to restore it at the end of the test
+    value = MOI.get(model, MOI.AbsoluteGapTolerance())
+    MOI.set(model, MOI.AbsoluteGapTolerance(), 1e-2)
+    @test MOI.get(model, MOI.AbsoluteGapTolerance()) == 1e-2
+    MOI.set(model, MOI.AbsoluteGapTolerance(), 100)
+    @test MOI.get(model, MOI.AbsoluteGapTolerance()) == 100
+    MOI.set(model, MOI.AbsoluteGapTolerance(), value)
+    @test value == MOI.get(model, MOI.AbsoluteGapTolerance())
+    _test_attribute_value_type(model, MOI.AbsoluteGapTolerance())
+    return
+end
+test_attribute_AbsoluteGapTolerance(::MOI.ModelLike, ::Config) = nothing
+
+function setup_test(
+    ::typeof(test_attribute_AbsoluteGapTolerance),
+    model::MOIU.MockOptimizer,
+    ::Config,
+)
+    MOI.set(model, MOI.AbsoluteGapTolerance(), nothing)
+    return
+end
+
+"""
+    test_attribute_RelativeGapTolerance(model::MOI.AbstractOptimizer, config::Config)
+
+Test that the [`MOI.RelativeGapTolerance`](@ref) attribute is implemented for
+`model`.
+"""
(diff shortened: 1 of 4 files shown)


src/attributes.jl
```diff
@@ -819,6 +819,37 @@ struct NumberOfThreads <: AbstractOptimizerAttribute end
...
 attribute_value_type(::NumberOfThreads) = Union{Nothing,Int}
 
+"""
+    RelativeGapTolerance()
+
+An optimizer attribute for setting the relative gap tolerance for an optimization.
+This is an _optimizer_ attribute, and should be set before calling `optimize!`.
+
+If you are looking for the relative gap of the current best solution, see
+    [RelativeGap](@ref). If no limit nor issue is encountered during the optimization,
```
blegat, line 829:
```suggestion
[RelativeGap](@ref). If no limit nor issue is encountered during the optimization,
```

src/attributes.jl
```diff
@@ -819,6 +819,37 @@ struct NumberOfThreads <: AbstractOptimizerAttribute end
...
+    the value of `RelativeGap` should be at most as large as `RelativeGapTolerance`.
```
blegat, line 830:
```suggestion
the value of `RelativeGap` should be at most as large as `RelativeGapTolerance`.
```

src/attributes.jl
```diff
@@ -819,6 +819,37 @@ struct NumberOfThreads <: AbstractOptimizerAttribute end
...
+# The relative gap tolerance has not changed...
+MOI.get(model, MOI.RelativeGapTolerance())  # returns 1e-3
+# ... and the relative gap of the obtained solution is smaller or equal to the tolerance
+MOI.get(model, MOI.RelativeGap())  # should return something ≤ 1e-3
+```
+
+!!! warning
+    The mathematical definition of "relative gap", and its allowed range, are solver-dependent.
+    Typically, solvers expect a value between `0` and `1`.
+
```
blegat, line 847: Remove empty line

mtanneau:
Is there anything I should change in the docs? Maybe around here?

https://github.com/jump-dev/MathOptInterface.jl/blob/868aa29bffaa6fd727a1858897ef961cc7b34e12/docs/src/reference/models.md?plain=1#L66-L77

src/attributes.jl
```diff
@@ -819,6 +819,55 @@ struct NumberOfThreads <: AbstractOptimizerAttribute end
 
 attribute_value_type(::NumberOfThreads) = Union{Nothing,Int}
 
+"""
+    RelativeGapTolerance()
+
+An optimizer attribute for setting the relative gap tolerance for an optimization.
+This is an _optimizer_ attribute, and should be set before calling `optimize!`.
```
odow, line 826:
```suggestion
This is an _optimizer_ attribute, and should be set before calling [`optimize!`](@ref).
```

src/attributes.jl
```diff
@@ -819,6 +819,55 @@ struct NumberOfThreads <: AbstractOptimizerAttribute end
...
+"""
+struct RelativeGapTolerance <: AbstractOptimizerAttribute end
+
+attribute_value_type(::RelativeGapTolerance) = Union{Nothing,Float64}
+
+"""
+    AbsoluteGapTolerance()
+
+An optimizer attribute for setting the absolute gap tolerance for an optimization.
+This is an _optimizer_ attribute, and should be set before calling `optimize!`.
```
odow, line 856:
```suggestion
This is an _optimizer_ attribute, and should be set before calling [`optimize!`](@ref).
```

odow (commented):
Yip. That's the place to add the docstrings.

You could also add it to:
https://github.com/jump-dev/MathOptInterface.jl/blob/master/docs/src/tutorials/implementing.md#implement-attributes

The attribute value type should be tested:
https://github.com/jump-dev/MathOptInterface.jl/blob/868aa29bffaa6fd727a1858897ef961cc7b34e12/test/attributes.jl#L225

It also needs a test for solvers implementing this. Something like:
https://github.com/jump-dev/MathOptInterface.jl/blob/868aa29bffaa6fd727a1858897ef961cc7b34e12/src/Test/test_attribute.jl#L164-L191

src/Test/test_attribute.jl
```diff
@@ -190,6 +190,72 @@ function setup_test(
...
+function test_attribute_AbsoluteGapTolerance(
+    model::MOI.AbstractOptimizer,
+    ::Config,
+)
+    @requires MOI.supports(model, MOI.AbsoluteGapTolerance())
+    # Get the current value to restore it at the end of the test
+    value = MOI.get(model, MOI.AbsoluteGapTolerance())
+    MOI.set(model, MOI.AbsoluteGapTolerance(), 1e-2)
+    @test MOI.get(model, MOI.AbsoluteGapTolerance()) == 1e-2
+    MOI.set(model, MOI.AbsoluteGapTolerance(), 100)
```
odow, line 208: This should probably be `100.0`? The `attribute_value_type` is Union{Nothing,Float64}`.

You should also test setting `nothing`.

mtanneau: I put an `Int` on purpose, because I thought some users (including myself) would probably pass an integer at some point, and I thought the code should be able to support the conversion.

What would setting the attribute to `nothing` mean? I put `Union{Nothing,Float64}` as attribute type mainly because that's what the other attributes were doing.

odow: > What would setting the attribute to nothing mean?

Setting `nothing` would re-set to the default. This should probably be documented in the docstring. If you don't want that, then remove `Nothing` from the attribute value type.

> I put an Int on purpose, because I thought some users (including myself) would probably pass an integer at some point, and I thought the code should be able to support the conversion.

Hmm. This seems problematic. I'd prefer we forced the user to pass `Float64`. If you want to let them pass an `Int`, then the attribute value type should be `Union{Float64,Int}`

mtanneau: I can follow the convention of the other attributes, and update the docs to mention that `nothing` uses solver defaults.

However, when I look at the Gurobi wrapper, setting an attribute like, e.g., `TimeLimitSec` to `nothing` is not supported
```julia
using Gurobi
using MathOptInterface
const MOI = MathOptInterface

grb = Gurobi.Optimizer()
MOI.set(grb, MOI.TimeLimitSec(), 1.0)  # no issue
MOI.set(grb, MOI.TimeLimitSec(), nothing)  # not supported
```

mtanneau: Re testing `nothing`: this is not done when testing the other attributes, e.g., when testing `TimeLimitSec` or `NumberOfThreads`.
The `setup_test` function does set the attribute to `nothing`, but it's only defined for a mock optimizer (I did the same as for other attributes)

odow: Ah. If it's not tested, then that's probably why Gurobi doesn't support it.

mtanneau: If I add a test for `nothing`, then downstream solvers will have to support that. It came as a surprise to me, but there's no way that I know of, in Gurobi, to reset a specific parameter to its default value. As far as I can tell, the only two options are to reset all parameters to default, or to set the parameter to a value that happens to be the default. The issue being that default parameter values may change from one release to the next (it's unlikely, but it might).

I don't think it's a good idea to explicitly test for `nothing` values here.

odow: Okay. I'm okay not setting `nothing`. We should remove from the attribute value type then.

mtanneau: So.... removing `Nothing` as attribute type fails the `_test_attribute_value_type` when testing against `MOIU.MockOptimizer`.
I guess it's because  `MOI.get(mock::MOI.MockOptimizer, attr::MOI.AbstractOptimizerAttribute)` will return `nothing` if said attribute has not been set yet.

This leaves the following options:
* we keep `Union{Nothing,Float64}` as attribute type and...
  * ... we don't test `set/get` with `nothing` input
  * ... we test `set/get` with `nothing` input, and deal with the consequences in downstream solvers
* we change the default behavior of `MockOptimizer{MT}` so that it returns a `Float` when queried for those attributes. I don't know about the consequences of that.

I think the first option (keeping `Nothing` but not testing the `set/get` with `nothing` as input) is the most viable. It's what is currently done for other attributes like `TimeLimitSec` and `NumberOfThreads`.

odow: Okay. I guess setting an option to `nothing` isn't a common operation that is requested.

mtanneau: The only place I can think of it happening is when passing model around, if the optimizer attributes are copied too. Not even sure that would be the case.

On the other hand, if someone tries to set those attributes to `nothing` in, e.g., CPLEX or Gurobi, they will get an error message because the `MOI.set` functions will require a `::Real` argument (that's the case for, e.g., `TimeLimitSec`). It won't silently fail nor cause segfaults.

The fact that nobody has reported such issue so far indeed suggests that it's not a common operation :)

odow: Fair enough. I think the actionable thing I'd like to see is `100` become `100.0`.

odow (approved):

PR #1948, opened 2022-07-13
Add missing promotion rules

matbesancon:
Just got bitten by this.
Before:
```
julia> T1 = eltype(variable_blocks)
MathOptInterface.VariableIndex

julia> T2 = eltype(1.0 * variable_blocks)
MathOptInterface.ScalarAffineFunction{Float64}

julia> promote_type(T1, T2)
MathOptInterface.AbstractScalarFunction
```

After:
```
julia> promote_type(T1, T2)
MathOptInterface.ScalarAffineFunction{Float64}
```

### DIFF
--- a/src/Utilities/functions.jl
+++ b/src/Utilities/functions.jl
@@ -3207,7 +3207,7 @@ function Base.promote_rule(
 end
 
 function Base.promote_rule(
-    F::Type{<:Union{MOI.ScalarAffineFunction,MOI.ScalarQuadraticFunction},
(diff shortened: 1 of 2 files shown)


matbesancon:
One open question, should we push it as far as promoting:
```julia
VariableIndex, T <: Real --> MOI.ScalarAffineFunction{T}
```

odow (approved):

odow:
> One open question, should we push it as far as promoting

Are there any unintended consequences, like `1 * x` promoting both to scalar affine and then creating a quadratic?

odow:
Merged since this is a win. We can discuss the other rule in a different PR.

PR #1951, opened 2022-07-13
[Bridges] fix tests for SplitComplexZerosBridge

odow:
The amount of little niggly things here shows how hard it is to write a bridge without these new tests.

### DIFF
--- a/src/Bridges/Constraint/bridges/split_complex_zeros.jl
+++ b/src/Bridges/Constraint/bridges/split_complex_zeros.jl
@@ -34,6 +34,7 @@ struct SplitComplexZerosBridge{
     G<:MOI.Utilities.TypedLike{Complex{T}},
 } <: AbstractBridge
     dimension::Int
+    func::G  # Simplify querying ConstraintFunction.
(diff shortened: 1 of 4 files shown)


src/Bridges/Constraint/bridges/split_complex_zeros.jl
```diff
@@ -34,6 +34,7 @@ struct SplitComplexZerosBridge{
     G<:MOI.Utilities.TypedLike{Complex{T}},
 } <: AbstractBridge
     dimension::Int
+    func::G  # Simplify querying ConstraintFunction.
```
blegat, line 37: We don't want to simplify it. It's ok for it to be slow. However, we want bridges to have low memory footprint

odow: I've been prioritizing correctness and completeness over performance. But yeah, we can fix this in a different PR if you like.

blegat: Yes, thanks for making the bridges tests working with complex! Just wanted to check we're on the same page about this which is why I also clarified it in the docs now

PR #1952, opened 2022-07-14
Clarify how to implement the ConstraintFunction getter

blegat:
See https://github.com/jump-dev/MathOptInterface.jl/pull/1951#discussion_r920778838

### DIFF
--- a/docs/src/tutorials/bridging_constraint.md
+++ b/docs/src/tutorials/bridging_constraint.md
@@ -327,7 +327,7 @@ end
     footprint of the bridges as the garbage collector won't be able to delete
     that object. The convention is to not store the function in the bridge
     and not care too much about the efficiency of these getters. If the user
-    needs efficient getters for [`ConstraintFunction`](@ref) then he should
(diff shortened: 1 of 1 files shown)


docs/src/tutorials/bridging_constraint.md
```diff
@@ -296,6 +296,40 @@ function get(
...
+```
+
+!!! warning
+    Alternatively, one could store the original function and set in `SignBridge`
+    during [`Bridges.Constraint.bridge_constraint`](@ref) to make these getters
+    simpler and more efficient. On the other hand, this will increase the memory
+    footprint of the bridges as the garbage collector won't be able to delete
+    that object. The convention is to not store the function in the bridge
+    and not care too much about the efficiency of these getters. If the user
+    needs efficient getters for [`ConstraintFunction`](@ref) then he should
```
odow, line 330:
```suggestion
    needs efficient getters for [`ConstraintFunction`](@ref) then they should
```

odow (approved):

PR #1955, opened 2022-07-20
Add Reified set and MILP bridges

### LINKED ISSUE #1805: Constraint programming support

odow:
Closes #1805

I've implemented these in MiniZinc: https://github.com/jump-dev/MiniZinc.jl/pull/13, but I don't know if we should add to MOI.

They're a little harder to write MILP bridges for, and unlike Indicator, there isn't a range of solvers supporting a small number of constraint types (like `Indicator{GreaterThan}` and `Indicator{LessThan}`). Instead, it's pretty arbitrary what sets you can wrap in a `Reified`.


src/sets.jl
```diff
@@ -1534,6 +1534,24 @@ Base.copy(set::Table) = Table(copy(set.table))
 
 Base.:(==)(x::Table{T}, y::Table{T}) where {T} = x.table == y.table
 
+"""
+    Reified(set::AbstractSet)
+
+The constraint ``[z; f(x)] \\in Reified(S)`` ensures that ``f(x) \\in S`` if and
```
dourouc05, line 1540: Why `f(x)`? I would simply use `x`, with `x \\in S`.

test/sets.jl
```diff
@@ -322,6 +322,16 @@ function test_sets_Interval()
...
 end
 
+function test_sets_reified()
+    set = MOI.Reified(MOI.AllDifferent(2))
+    @test MOI.dimension(set) == 3
+    @test set == copy(set)
+    set = MOI.Reified(MOI.GreaterThan(1))
+    @test MOI.dimension(set) == 2
+    @test set == copy(set)
+    return
```
dourouc05, line 332: The `return` is not present for the three functions above, maybe that should be fixed?

src/sets.jl
```diff
@@ -1534,6 +1534,24 @@ Base.copy(set::Table) = Table(copy(set.table))
 
 Base.:(==)(x::Table{T}, y::Table{T}) where {T} = x.table == y.table
 
+"""
+    Reified(set::AbstractSet)
```
dourouc05, line 1538: With the exception of `GreaterThan`, `LessThan`, and `EqualTo`, all set names are nouns and not modifiers (like adjectives).

odow: There's also `Indicator`. Does that count?

src/sets.jl
odow: We've been using this notation to mean that it also supports `VectorAffineFunction` etc, not just `VectorOfVariables`.

test/sets.jl
odow: Yeah, some functions that end in a macro call are missing a `return` statement, because the formatter cannot automatically infer that `return @set == copy(set)` or a blank `return` is appropriate. 

I fix them when convenient, but it's not a big deal.

odow:
@blegat can you take a look at this?

chriscoey:
LGTM - pretty straightforward.

odow:
Okay. I've added a bridge from `Reified{AllDifferent}` to `Reified{CountDistinct}` and one from `Reified{CountDistinct}` to MILP. I've also added a test for `Reified{AlDifferent}` and verified that HiGHS passes the test with the new bridges.

I think that shows an avenue to making the MILP bridges work, even if the reformulations are a little more work.

odow:
I think I'm pretty happy with this, so I'll leave open for a few days if anyone wants to comment, otherwise I'll merge.

PR #1958, opened 2022-07-21
[Bridges] add IndicatorSetMapBridge

### LINKED ISSUE #1223: Bridge request: inner constraint of Indicator
It seems to make sense to use bridges for the right hand side or what I call inner constraint of an indicator constraint.

Such that `@constraint(b => { x + y >= z})` can be bridged with the `GreaterToLessBridge` to `@constraint(b => { -x - y <= -z})`

Currently this error is thrown:
```
ERROR: LoadError: Constraints of type MathOptInterface.VectorAffineFunction{Float64}-in-MathOptInterface.IndicatorSet{MathOptInterface.ACTIVATE_ON_ONE,MathOptInterface.GreaterThan{Float64}} are not supported by the solver and there are no bridges that can reformulate it into supported constraints.
```

I tried to implement it myself but I'm unsure how. Thought that I need to define sth like:

```
function MOI.supports_constraint(::Type{MOIBC.GreaterToLessBridge},
    ::Type{MOI.VectorAffineFunction}, ::Type{IS}) where {A,IS<:MOI.IndicatorSet{A,MOI.GreaterThan}}
    return true
end
```

Maybe someone can point me into the right direction on how the specialized version can be implemented as well as how one could implement the general version to use all available bridges for the inner constraint.

odow:
Closes #1223. 

I don't know if we want to add the more complicated bridged mentioned in #1223. It's implemented https://github.com/Wikunia/ConstraintSolver.jl/blob/master/src/MOI_wrapper/Bridges/indicator.jl, but it requires hard-coding the inner bridge type. 

It's probably simpler to just manually implement the reformulations as needed. Writing bridges is now a lot less of a hassle with tests and docs.


docs/src/submodules/Bridges/list_of_bridges.md
```diff
@@ -50,6 +50,7 @@ Bridges.Constraint.SquareBridge
 Bridges.Constraint.RootDetBridge
 Bridges.Constraint.LogDetBridge
 Bridges.Constraint.IndicatorActiveOnFalseBridge
+Bridges.Constraint.IndicatorFlipSign
```
odow, line 53:
```suggestion
Bridges.Constraint.IndicatorFlipSignBridge
```

blegat:
> it requires hard-coding the inner bridge type.

I don't understand what's the issue. Being able to create a new bridge with one line by parametrizing the bridge with another one is still easier than doing a new bridge from scratch

odow:
So I tried this, and it's not that simple. We'd end up with `IndicatorGreaterToLessBridge` and `IndicatorLessToGreaterBridge`, and things aren't actually that much simplified. We also have only one other use-case for this in ConstraintSolver.jl, which already has the bridge written. 

I'm really leaning to simpler, more verbose bridges. They're just really hard to reason about if we add more complexity. New bridges are now easy to write and easy to test.

blegat:
> So I tried this, and it's not that simple. We'd end up with `IndicatorGreaterToLessBridge` and `IndicatorLessToGreaterBridge`, and things aren't actually that much simplified.

Could you elaborate on that ? Why can't we just do the same as in ConstraintSolver ?

odow:
> We'd end up with IndicatorGreaterToLessBridge and IndicatorLessToGreaterBridge

You can get the src constraint type, but not the dest, so CS uses a non-concrete type for the dest constraint index:
https://github.com/Wikunia/ConstraintSolver.jl/blob/e99db3814e88818222174e086a3024b9076fcaa1/src/MOI_wrapper/Bridges/indicator.jl#L2
To specify a concrete type you need to manually encode the bridge that's used, and the two sets that are being mapped.

And because CS manually adds the bridges:
https://github.com/Wikunia/ConstraintSolver.jl/blob/e99db3814e88818222174e086a3024b9076fcaa1/src/MOI_wrapper/MOI_wrapper.jl#L47-L67

I think if someone wants a new bridge for indicator, let them write their own. This is the main use-case, and they can copy what we have if needed.

odow:
Okay, here it is. I'm not convinced it's simpler, but it's a demonstration of a bridge using another bridge.

src/Bridges/Constraint/bridges/indicator_flipsign.jl
```diff
@@ -0,0 +1,185 @@
...
+
+  * [`MOI.VectorAffineFunction{T}`](@ref) in [`MOI.Indicator{A,S1}`](@ref)
+
+## Target nodes
+
+`IndicatorSetMapBridgeBridge` creates:
+
+  * [`MOI.VectorAffineFunction{T}`](@ref) in [`MOI.Indicator{A,S2}`](@ref)
+"""
+struct IndicatorSetMapBridgeBridge{T,B,S1,S2,A} <: AbstractBridge
```
blegat, line 27:
```suggestion
struct IndicatorSetMapBridge{T,B,S1,S2,A} <: AbstractBridge
```

blegat:
It looks good, just the name is a bit weird with `BridgeBridge`

src/Bridges/Constraint/bridges/indicator_flipsign.jl
odow: Hmm. I thought I pushed a commit fixing this. Looks like I didn't.

PR #1960, opened 2022-07-22
[docs] tidy standard form page and add global constraints

odow:
Preview https://jump.dev/MathOptInterface.jl/previews/PR1960/manual/standard_form/

I think the tables are tidier and easier to read than the bulleted list.


blegat (approved):
I agree, it looks better

PR #1961, opened 2022-07-22
Add HyperRectangle set

### LINKED ISSUE #1233: Set request: Box or HyperRectangle
```julia
struct Box{T}
    lower::Vector{T}
    upper::Vector{T}
end
```
which is a cartesian product if `Interval{T}` so `lower` and `upper` are allowed to have infinite values.
This has been requested by @migarstka for a long time and once again in https://github.com/oxfordcontrol/COSMO.jl/pull/129#issuecomment-771536349
It is also often requested from people in control where it's common to work with hyper-rectangles.

We would also have the bridges:
* `Box` to `Interval`s (`Bridges.Constraint.Scalarize`)
* `Interval` to `Box` (`Bridges.Constraint.Vectorize`)
* `Zeros`, `Nonnegatives` and `Nonpositives` to `Box` (otherwise, it would be scalarized and then vectorized to many 1-dimension boxes ^^)
* `Box` to `Nonnegatives` and `Nonpositives` (otherwise, it would be scalarized to `Interval` then bridged to a `GreaterThan` and a `LessThan` and then vectorized and it's many small 1-dim constraints)

odow:
Closes #1233 

Both SCS and COSMO (and maybe others) support hyper-rectangle sets.


src/Bridges/Constraint/bridges/split_hyperrectangle.jl
```diff
@@ -0,0 +1,119 @@
...
+# Use of this source code is governed by an MIT-style license that can be found
+# in the LICENSE.md file or at https://opensource.org/licenses/MIT.
+
+"""
+    SplitHyperRectangleBridge{T,G,F} <: Bridges.Constraint.AbstractBridge
+
+`SplitHyperRectangleBridge` implements the following reformulation:
+
+  * ``f(x) \\in \\textsf{HyperRectangle}(l, u)`` to
+    ``[f(x) - l; u - f(x)] \\in \\mathbb{R}_+``.
```
odow, line 13: Actually I need to fix the case when `l` or `u` contains `inf`.

src/Bridges/Constraint/bridges/split_hyperrectangle.jl
```diff
@@ -0,0 +1,118 @@
...
+    G = MOI.Utilities.promote_operation(-, T, F, Vector{T})
+    return SplitHyperRectangleBridge{T,G,F}
+end
+
+function MOI.get(
+    ::MOI.ModelLike,
+    ::MOI.ConstraintFunction,
+    bridge::SplitHyperRectangleBridge,
+)
+    return bridge.func
```
odow, line 90: @blegat any ideas for how to recover the function in the general case? Like if we have a `-Inf, Inf` row, then we'll drop the row from the bridge, so we at least need to store those functions, and if someone passes `f(x) + c in HyperRectangle(l, u)`, then we need to store `l` and `u` to be able to recover non-zero constants `c`.

src/Bridges/Constraint/bridges/split_hyperrectangle.jl
```diff
@@ -0,0 +1,118 @@
...
+    scalars = MOI.Utilities.eachscalar(f)
+    lower = [
+        MOI.Utilities.operate(-, T, fi, s.lower[i]) for
+        (i, fi) in enumerate(scalars) if isfinite(s.lower[i])
+    ]
+    upper = [
+        MOI.Utilities.operate(-, T, s.upper[i], fi) for
+        (i, fi) in enumerate(scalars) if isfinite(s.upper[i])
+    ]
+    g = MOI.Utilities.operate(vcat, T, lower..., upper...)
```
blegat, line 51: This will need to compile a new method for new sizes of rectangles. If the rectangle has 1000 entries, it will compile a method with 2000 arguments?

src/Bridges/Constraint/bridges/split_hyperrectangle.jl
blegat: We could error on nonzero constant like for the scalar affine sets. With infinite bounds, we could store the function yes

src/Bridges/Constraint/bridges/split_hyperrectangle.jl
odow: True. I'll have another go.

odow: Done

src/Bridges/Constraint/bridges/split_hyperrectangle.jl
odow: This is probably okay for now. We can see if anyone complains? I don't think it's a common set.

blegat: We should add a TODO or FIXME with a link to the docs where we say why we don't want that then it's fine to merge

odow: Done.

src/Bridges/Constraint/bridges/split_hyperrectangle.jl
```diff
@@ -88,6 +88,9 @@ function MOI.get(
     ::MOI.ConstraintFunction,
     bridge::SplitHyperRectangleBridge,
 )
+    # We cache the function to simplify cases in which the box is (-Inf, Inf).
+    # This contributes some memory overhead, so if someone complains in future
```
blegat, line 92: I don't agree that we should wait for someone to complain, it's a TODO. See the warning in https://jump.dev/MathOptInterface.jl/dev/tutorials/bridging_constraint/#Attribute-retrieval that was added in https://github.com/jump-dev/MathOptInterface.jl/pull/1952/

odow: Fully agree that in the common case, we shouldn't cache the functions. But this is a pretty niche bridge that in all likelihood won't get used much. I don't want to spend much time optimizing if for an edge-case.

blegat: You don't have to do it but because it's a convention, we should at least mark it as a TODO or FIXME otherwise we're being inconsistent.
Memory overhead is quite hard to track down so users won't come complaining about that.
It's not that niche though, box constraints are very common in control.
Note that the bookkeeping needed to implement this will be useful for `ConstraintPrimal` and `ConstraintDual` anyway.
Anyway, no big deal, I can do that in a follow up PR to clean up the TODO

odow: I've been omitting `Constraint{Primal/Dual}{Start}` for simplicity. We can come back to them. It's a lower priority.

blegat: `ConstraintFunction` has a much lower priority than `ConstraintDual`. `ConstraintFunction` is mostly used in MOI tests but users are recommended to use a cache if they need to query the model. I feel the issue is that the bridge tests need `ConstraintFunction` to be implemented. Maybe we could add an option for the test to not test the `ConstraintFunction` attribute ? If you just cache the constraint function as in this PR, it's not like this is testing anything anyway. It seems that the fact that `MOI.Bridges.runtests` forces `ContraintFunction` to be implemented might be annoying. You might want to first make it work without having to worry about that attribute

odow: `Constraint{Function/Set}` are really useful because they allow printing of the bridged models. If you have a cache, It's harder to see what the bridge is really doing.

odow: Okay, I'll add the starts now.

blegat: What do you mean, it will print the same as the cache no?

odow: In the `Bridges.runtests`, I often print the bridge when I'm writing it as a way to check things. But yeah, I guess we could make the test optional, but at this point, every bridge supports it.

odow:
As discussed with @blegat, made `ConstraintFunction` optional.

odow:
Merging because I think this is reasonable. Once merged and tagged we can update SCS etc to support it.

PR #1963, opened 2022-07-23
Prep for v1.6.1

blegat:
This includes two bug fixes that would be nice to have to the ICCOPT workshop tomorrow so that v1.7 can wait for https://github.com/jump-dev/MathOptInterface.jl/pull/1962


odow:
I'm okay with either. The Hermitian stuff can be in 1.8. (I don't think we need to go slow in the minor version numbers. It's okay to get to v1.23.0.)

PR #1964, opened 2022-07-24
Fix complex tests for local solver

blegat:
See https://github.com/jump-dev/SolverTests/runs/7483984523?check_suite_focus=true#step:4:386

### DIFF
--- a/src/Test/test_linear.jl
+++ b/src/Test/test_linear.jl
@@ -3913,20 +3913,29 @@ function test_linear_complex_Zeros(optimizer, config::Config{T}) where {T}
         MOI.Zeros(1),
     )
     if _supports(config, MOI.optimize!)
+        @test MOI.get(model, MOI.TerminationStatus()) == MOI.OPTIMIZE_NOT_CALLED
         MOI.optimize!(optimizer)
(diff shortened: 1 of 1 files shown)


src/Test/test_linear.jl
```diff
@@ -3913,20 +3913,29 @@ function test_linear_complex_Zeros(optimizer, config::Config{T}) where {T}
         MOI.Zeros(1),
     )
     if _supports(config, MOI.optimize!)
+        @test MOI.get(model, MOI.TerminationStatus()) == MOI.OPTIMIZE_NOT_CALLED
```
odow, line 3916:
```suggestion
        @test MOI.get(optimizer, MOI.TerminationStatus()) ==
              MOI.OPTIMIZE_NOT_CALLED
```

src/Test/test_linear.jl
```diff
@@ -3971,20 +3980,21 @@ function test_linear_complex_Zeros_duplicate(
...
```
odow, line 3983:
```suggestion
        @test MOI.get(optimizer, MOI.TerminationStatus()) ==
              MOI.OPTIMIZE_NOT_CALLED
```

PR #1969, opened 2022-07-25
[Test] add test for open Interval sets

### LINKED ISSUE #881: Add tests cases with infinite bounds

odow:
Closes #881. 

If we merge this before tagging 1.7, we might want to run SolverTests.

### DIFF
--- a/src/Test/test_linear.jl
+++ b/src/Test/test_linear.jl
@@ -4020,3 +4020,49 @@ function setup_test(
     )
     return
 end
+
+"""
+    test_linear_open_intervals(
+        model::MOI.ModelLike,
+        config::Config{T},
+    ) where {T}
+
+Test that the solver supports open-intervals like `(-Inf, u]`, `[l, Inf)`, and
+`(-Inf, Inf)`.
+"""
+function test_linear_open_intervals(
+    model::MOI.ModelLike,
+    config::Config{T},
+) where {T}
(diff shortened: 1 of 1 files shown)


odow:
I think the bridge issue that prevented this earlier is fixed. I've tested in HiGHS, Gurobi, SCS, etc, and all pass the test.

PR #1974, opened 2022-08-01
[FileFormats.LP] add support for quadratic problems

### LINKED ISSUE #1973: FileFormats.LP: support quadratic functions
Goes with #1972. 

I took a read of the different docs:

CPLEX: https://www.ibm.com/docs/en/icos/12.9.0?topic=representation-quadratic-terms-in-lp-file-format
Gurobi: https://www.gurobi.com/documentation/9.5/refman/lp_format.html
Mosek: https://docs.mosek.com/latest/pythonfusion/lp-format.html

 * Quadratic terms are enclosed in `[ ] / 2`
 * CPLEX and Gurobi don't include the `/ 2` in constraints
 * We should support `x * x` and `x^2`, but when writing, always use `^`

odow:
Closes #1973


matbesancon (approved):

odow:
We can read `[ ... ]/2` and `[ ... ]` varieties, so it shouldn't matter what format they're in. When writing, we copy Gurobi and CPLEX.

PR #1976, opened 2022-08-05
Clarify assumption for SOC bridge

blegat:
I was a bit confused by `|x|`, I was wondering what was the absolute value of a vector. I agree it's a common notation for the cardinality of set but here I didn't think about it since it's not a set.

### DIFF
--- a/src/Bridges/Constraint/bridges/soc_rsoc.jl
+++ b/src/Bridges/Constraint/bridges/soc_rsoc.jl
@@ -13,7 +13,7 @@
 
 ## Assumptions
 
-  * `SOCtoRSOCBridge` assumes that the dimension of the cone is at least one.
+  * `SOCtoRSOCBridge` assumes that the length of `x` is at least one.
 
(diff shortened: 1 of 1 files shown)


src/Bridges/Constraint/bridges/soc_rsoc.jl
```diff
@@ -13,7 +13,7 @@
 
 ## Assumptions
 
-  * `SOCtoRSOCBridge` assumes that ``|x| \\ge 1``.
+  * `SOCtoRSOCBridge` assumes that the dimension of the cone is at least one.
```
odow, line 16:
```suggestion
  * `SOCtoRSOCBridge` assumes that the length of `x` is at least one.
```

I don't think it works if you have `[t] in SecondOrderCone(1)`.

blegat: Ah yes, I meant strictly greater

PR #1977, opened 2022-08-08
[Bridges] Add hermitian bridge for constraints

blegat:
This is the last piece needed from MOI for complex number support in JuMP.
This was not part of ComplexOptInterface and will allow to transform hermitian LMIs into LMIs. This missing feature of ComplexOptInterface was requested in https://discourse.julialang.org/t/advice-on-optimizing-large-sdp-model-generation-with-jump-complexoptinterface/82629

cc @cgois


odow (approved):
Happy if you're happy.

PR #1979, opened 2022-08-12
[Bridges] add Objective.Quadratize bridge

### LINKED ISSUE #1978: [Bridges] add objective bridge from affine to quadratic
x-ref https://github.com/darnstrom/DAQP.jl/issues/4

odow:
Closes #1978

This should simplify QP solvers and allow then to support only a quadratic objective.

There's a lot of double up with the `Functionize` bridge, but it was simpler to copy-paste than to extend the existing definition of `Functionize` to add new type parameters.


blegat (approved):
I agree it's best to create a new bridge. If we want to remove duplicate code, we can do something similar to `Bridges.Constraint.AbstractFunctionConversionBridge`, e.g., create a `Bridges.Objective.AbstractFunctionConversionBridge`.

PR #1981, opened 2022-08-15
Prep for v1.7.0

odow:
There's a lot of stuff building up, so let's get a release.


odow:
Needs https://github.com/jump-dev/MathOptInterface.jl/pull/1982 first.

odow:
SolverTests passing for most solvers (few unrelated issues that I can fix in the solvers):  https://github.com/jump-dev/SolverTests/runs/7847579947

So I'm going to hit the button.

PR #1991, opened 2022-09-05
[Bridges] fix supports for VariableIndex attributes

odow:
xref:
 * https://github.com/jump-dev/Ipopt.jl/pull/333
 * https://github.com/jump-dev/Ipopt.jl/pull/334


odow:
Hmm. I broke SDPA. Perhaps I need to take another look

PR #1995, opened 2022-09-12
[Utilities] add PenaltyRelaxation

odow:
Part of https://github.com/jump-dev/JuMP.jl/issues/3034. This PR explores the options we have to add a feasibility relaxation function to MOI.

The main decisions would be: 

* Should it modify in-place or create a new model?
* What should we do for constraints which don't support in-place modification like variable bounds?

The other option, not implemented here, is some sort of `Optimizer` like `Dualization.Optimizer`. That might be a better solution, but would add more code and overhead.

### DIFF
--- a/src/Utilities/feasibility_relaxation.jl
+++ b/src/Utilities/feasibility_relaxation.jl
@@ -111,7 +111,7 @@ end
 function MOI.supports(
     ::MOI.ModelLike,
     ::FeasibilityRelaxation,
-    ::Type{MOI.ConstraintIndex{F,<:MOI.AbstractScalarSet}},
+    ::Type{<:MOI.ConstraintIndex{F,<:MOI.AbstractScalarSet}},
 ) where {T,F<:Union{MOI.ScalarAffineFunction{T},MOI.ScalarQuadraticFunction{T}}}
     return true
 end
--- a/src/Utilities/feasibility_relaxation.jl
+++ b/src/Utilities/feasibility_relaxation.jl
@@ -28,7 +28,7 @@ constraint that is being relaxed. If no value exists, the default is `1.0`.
 The feasibility relaxation is limited to modifying constraint types for which
 `MOI.supports(model, ::FeasibilityRelaxation, MOI.ConstraintIndex{F,S})` is
 `true`. By default, this is only true for [`MOI.ScalarAffineFunction`](@ref) and
-[`MOI.MOI.ScalarQuadraticFunction`](@ref) constraints in the linear sets
+[`MOI.ScalarQuadraticFunction`](@ref) constraints in the linear sets
 [`MOI.LessThan`](@ref), [`MOI.GreaterThan`](@ref), [`MOI.EqualTo`](@ref) and
 [`MOI.Interval`](@ref). It does not include variable bound or integrality
 constraints, because these cannot be modified in-place.
--- a/src/Utilities/feasibility_relaxation.jl
+++ b/src/Utilities/feasibility_relaxation.jl
@@ -1,0 +1,176 @@
+# Copyright (c) 2017: Miles Lubin and contributors
+# Copyright (c) 2017: Google Inc.
+#
+# Use of this source code is governed by an MIT-style license that can be found
+# in the LICENSE.md file or at https://opensource.org/licenses/MIT.
+
+"""
+    FeasibilityRelaxation(
+        penalties = Dict{MOI.ConstraintIndex,Float64}(),
+    ) <: MOI.AbstractModelAttribute
+
+A model attribute that, when set, destructively modifies the model in-place to
+create a feasibility relxation.
+
+!!! warning
+    This is a destructive routine that modifies the model in-place. If you don't
+    want to modify the original model, use `copy_model` to create a copy before
+    setting this attribute.
+
+## Reformulation
+
+The feasibility relaxation modifies constraints of the form ``f(x) \\in S`` into
+``f(x) + y - z \\in S``, where ``y, z \\ge 0``, and then it introduces a penalty
+term into the objective of ``a \\times (y + z)`` (if minimizing, else ``-a``),
+where `a` is the value in the `penalties` dictionary associated with the
+constraint that is being relaxed. If no value exists, the default is `1.0`.
+
+The feasibility relaxation is limited to modifying constraint types for which
+`MOI.supports(model, ::FeasibilityRelaxation, MOI.ConstraintIndex{F,S})` is
+`true`. By default, this is only true for [`MOI.ScalarAffineFunction`](@ref) and
+[`MOI.MOI.ScalarQuadraticFunction`](@ref) constraints in the linear sets
+[`MOI.LessThan`](@ref), [`MOI.GreaterThan`](@ref), [`MOI.EqualTo`](@ref) and
+[`MOI.Interval`](@ref). It does not include variable bound or integrality
+constraints, because these cannot be modified in-place.
+
+## Example
+
+```jldoctest
+julia> model = MOI.Utilities.Model{Float64}();
+
+julia> x = MOI.add_variable(model);
+
+julia> c = MOI.add_constraint(model, 1.0 * x, MOI.LessThan(2.0));
+
+julia> MOI.set(model, MOI.Utilities.FeasibilityRelaxation(Dict(c => 2.0)))
+
+julia> print(model)
+Minimize ScalarAffineFunction{Float64}:
+ 0.0 + 2.0 v[2]
+
+Subject to:
+
+ScalarAffineFunction{Float64}-in-LessThan{Float64}
+ 0.0 + 1.0 v[1] - 1.0 v[2] <= 2.0
+
+VariableIndex-in-GreaterThan{Float64}
+ v[2] >= 0.0
+```
+"""
+mutable struct FeasibilityRelaxation{T} <: MOI.AbstractModelAttribute
+    penalties::Dict{MOI.ConstraintIndex,T}
+    scale::T
+    function FeasibilityRelaxation(p::Dict{MOI.ConstraintIndex,T}) where {T}
+        return new{T}(p, zero(T))
+    end
+end
+
+function FeasibilityRelaxation()
+    return FeasibilityRelaxation(Dict{MOI.ConstraintIndex,Float64}())
+end
+
+function FeasibilityRelaxation(d::Dict{<:MOI.ConstraintIndex,T}) where {T}
+    return FeasibilityRelaxation(convert(Dict{MOI.ConstraintIndex,T}, d))
+end
+
+function MOI.set(
+    model::MOI.ModelLike,
+    relax::FeasibilityRelaxation{T},
+) where {T}
+    sense = MOI.get(model, MOI.ObjectiveSense())
+    if sense == MOI.FEASIBILITY_SENSE
+        relax.scale = one(T)
+        MOI.set(model, MOI.ObjectiveSense(), MOI.MIN_SENSE)
+        f = zero(MOI.ScalarAffineFunction{T})
+        MOI.set(model, MOI.ObjectiveFunction{typeof(f)}(), f)
+    elseif sense == MOI.MIN_SENSE
+        relax.scale = one(T)
+    elseif sense == MOI.MAX_SENSE
+        relax.scale = -one(T)
+    end
+    for (F, S) in MOI.get(model, MOI.ListOfConstraintTypesPresent())
+        MOI.set(model, relax, F, S)
+    end
+    return
+end
+
+function MOI.set(
+    model::MOI.ModelLike,
+    relax::FeasibilityRelaxation,
+    ::Type{F},
+    ::Type{S},
+) where {F,S}
+    if MOI.supports(model, relax, MOI.ConstraintIndex{F,S})
+        for ci in MOI.get(model, MOI.ListOfConstraintIndices{F,S}())
+            MOI.set(model, relax, ci)
+        end
+    end
+    return
+end
+
+function MOI.supports(
+    ::MOI.ModelLike,
+    ::FeasibilityRelaxation,
+    ::Type{MOI.ConstraintIndex{F,<:MOI.AbstractScalarSet}},
+) where {T,F<:Union{MOI.ScalarAffineFunction{T},MOI.ScalarQuadraticFunction{T}}}
+    return true
+end
+
+function MOI.supports_fallback(
+    ::MOI.ModelLike,
+    ::FeasibilityRelaxation,
+    ::Type{MOI.ConstraintIndex{F,S}},
+) where {F,S}
+    return false
+end
+
+function MOI.set(
+    model::MOI.ModelLike,
+    relax::FeasibilityRelaxation,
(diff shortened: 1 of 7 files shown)


src/Utilities/feasibility_relaxation.jl
```diff
@@ -0,0 +1,176 @@
...
+# Use of this source code is governed by an MIT-style license that can be found
+# in the LICENSE.md file or at https://opensource.org/licenses/MIT.
+
+"""
+    FeasibilityRelaxation(
+        penalties = Dict{MOI.ConstraintIndex,Float64}(),
+    ) <: MOI.AbstractModelAttribute
+
+A model attribute that, when set, destructively modifies the model in-place to
+create a feasibility relxation.
```
blegat, line 13:
```suggestion
create a feasibility relaxation.
```

blegat:
> That might be a better solution, but would add more code and overhead.

Not if we use `MOI.Utilities.ModelFilter`

src/Utilities/feasibility_relaxation.jl
```diff
@@ -0,0 +1,176 @@
...
+`MOI.supports(model, ::FeasibilityRelaxation, MOI.ConstraintIndex{F,S})` is
+`true`. By default, this is only true for [`MOI.ScalarAffineFunction`](@ref) and
+[`MOI.MOI.ScalarQuadraticFunction`](@ref) constraints in the linear sets
+[`MOI.LessThan`](@ref), [`MOI.GreaterThan`](@ref), [`MOI.EqualTo`](@ref) and
+[`MOI.Interval`](@ref). It does not include variable bound or integrality
+constraints, because these cannot be modified in-place.
+
+## Example
+
+```jldoctest
```
odow, line 38:
```suggestion
```jldoctest; setup=:(import MathOptInterface; const MOI = MathOptInterface)
```

src/Utilities/feasibility_relaxation.jl
```diff
@@ -0,0 +1,176 @@
...
+The feasibility relaxation modifies constraints of the form ``f(x) \\in S`` into
+``f(x) + y - z \\in S``, where ``y, z \\ge 0``, and then it introduces a penalty
+term into the objective of ``a \\times (y + z)`` (if minimizing, else ``-a``),
+where `a` is the value in the `penalties` dictionary associated with the
+constraint that is being relaxed. If no value exists, the default is `1.0`.
+
+The feasibility relaxation is limited to modifying constraint types for which
+`MOI.supports(model, ::FeasibilityRelaxation, MOI.ConstraintIndex{F,S})` is
+`true`. By default, this is only true for [`MOI.ScalarAffineFunction`](@ref) and
+[`MOI.MOI.ScalarQuadraticFunction`](@ref) constraints in the linear sets
```
odow, line 31:
```suggestion
[`MOI.ScalarQuadraticFunction`](@ref) constraints in the linear sets
```

odow:
Bump. Does anyone want to take a look at this?

odow:
> Any reason why is it not handling vector function-in-set's?

Only that it is a little more complicated to implement. I think this should cover 99% of the requests.

src/Utilities/feasibility_relaxation.jl
```diff
@@ -0,0 +1,176 @@
...
+    FeasibilityRelaxation(
+        penalties = Dict{MOI.ConstraintIndex,Float64}(),
+    ) <: MOI.AbstractModelAttribute
+
+A model attribute that, when set, destructively modifies the model in-place to
+create a feasibility relaxation.
+
+!!! warning
+    This is a destructive routine that modifies the model in-place. If you don't
+    want to modify the original model, use `copy_model` to create a copy before
```
blegat, line 17:
```suggestion
    want to modify the original model, use `JuMP.copy_model` to create a copy before
```

src/Utilities/feasibility_relaxation.jl
```diff
@@ -0,0 +1,176 @@
...
+            MOI.set(model, relax, ci)
+        end
+    end
+    return
+end
+
+function MOI.supports(
+    ::MOI.ModelLike,
+    ::FeasibilityRelaxation,
+    ::Type{<:MOI.ConstraintIndex{F,<:MOI.AbstractScalarSet}},
```
blegat, line 114: You should implement `supports_fallback` when you don't have a concret type of model to avoid ambiguity

src/Utilities/feasibility_relaxation.jl
```diff
@@ -0,0 +1,176 @@
...
+
+!!! warning
+    This is a destructive routine that modifies the model in-place. If you don't
+    want to modify the original model, use `copy_model` to create a copy before
+    setting this attribute.
+
+## Reformulation
+
+The feasibility relaxation modifies constraints of the form ``f(x) \\in S`` into
+``f(x) + y - z \\in S``, where ``y, z \\ge 0``, and then it introduces a penalty
```
blegat, line 23: Mention that `y` or`z` are not added for `GreaterThan`/`LessThan`

mlubin:
It seems weird conceptually to use attributes to (destructively) transform a model.

blegat:
Indeed, maybe `MOI.set(model, MOI.Utilities.FeasibilityRelaxation(Dict(c => 2.0)))` could be `MOI.modify(model, c, MOI.Utilities.FeasibilityRelaxation(2.0))`

odow:
Yes, `MOI.modify` is a much better verb. I don't know why I didn't think of that.

odow:
Any comments now that I've swapped to `MOI.modify`?

odow:
- [ ] Throw warning on unsupported constraints
 - [ ] Document that bounds should be rewritten as linear constraints
 - [ ] Potentially renamed to PenaltyRelaxation
 - [ ] Return a map between the constraint and the penalty

odow:
This works quite nicely. Here's what it looks like from JuMP:
```julia
julia> using JuMP, HiGHS

julia> model = Model(HiGHS.Optimizer);

julia> set_silent(model)

julia> @variable(model, x >= 0);

julia> @objective(model, Max, 2x + 1);

julia> @constraint(model, c, 2x - 1 <= -2);

julia> optimize!(model)

julia> function penalty_relaxation!(
           model::Model,
           penalties;
           default::Union{Nothing,Real} = 1.0,
       )
           if default !== nothing
               default = Float64(default)
           end
           moi_penalties = Dict{MOI.ConstraintIndex,Float64}(
               index(k) => Float64(v) for (k, v) in penalties
           )
           map = MOI.modify(
               backend(model),
               MOI.Utilities.PenaltyRelaxation(moi_penalties; default = default),
           )
           return Dict(
               ConstraintRef(model, k, ScalarShape()) => jump_function(model, v) for
               (k, v) in map
           )
       end
penalty_relaxation! (generic function with 2 methods)

julia> function penalty_relaxation!(model::Model; kwargs...)
           return penalty_relaxation!(model, Dict(); kwargs...)
       end
penalty_relaxation! (generic function with 2 methods)

julia> penalties = penalty_relaxation!(model; default = 2)
Dict{ConstraintRef{Model, MathOptInterface.ConstraintIndex{MathOptInterface.ScalarAffineFunction{Float64}, MathOptInterface.LessThan{Float64}}, ScalarShape}, AffExpr} with 1 entry:
  c : 2 x - _[2] ≤ -1.0 => _[2]

julia> print(model)
Max 2 x - 2 _[2] + 1
Subject to
 c : 2 x - _[2] ≤ -1.0
 x ≥ 0.0
 _[2] ≥ 0.0

julia> optimize!(model)

julia> value(penalties[c])
1.0
```

odow:
Bump.

src/Utilities/penalty_relaxation.jl
```diff
@@ -0,0 +1,239 @@
...
+    want to modify the original model, use `JuMP.copy_model` to create a copy
+    before setting this attribute.
+
+## Reformulation
+
+The penalty relaxation modifies constraints of the form ``f(x) \\in S`` into
+``f(x) + y - z \\in S``, where ``y, z \\ge 0``, and then it introduces a penalty
+term into the objective of ``a \\times (y + z)`` (if minimizing, else ``-a``),
+where `a` is the value in the `penalties` dictionary associated with the
+constraint that is being relaxed. If no value exists, the default is `default`.
```
blegat, line 27: This does not say what happens if `default` is `nothing`

src/Utilities/penalty_relaxation.jl
```diff
@@ -0,0 +1,239 @@
...
+) where {T,F<:Union{MOI.ScalarAffineFunction{T},MOI.ScalarQuadraticFunction{T}}}
+    y = MOI.add_variable(model)
+    z = MOI.add_variable(model)
+    MOI.add_constraint(model, y, MOI.GreaterThan(zero(T)))
+    MOI.add_constraint(model, z, MOI.GreaterThan(zero(T)))
+    MOI.modify(model, ci, MOI.ScalarCoefficientChange(y, one(T)))
+    MOI.modify(model, ci, MOI.ScalarCoefficientChange(z, -one(T)))
+    sense = MOI.get(model, MOI.ObjectiveSense())
+    scale = sense == MOI.MIN_SENSE ? one(T) : -one(T)
+    a = scale * get(relax.penalties, ci, relax.default)
```
blegat, line 197: It's a bit weird that a constraint would receive a dictionary and not its value directly. Maybe we should have a different attribute for `modify` with a constraint index that just contains a value. `ScalarCoefficientChange` contains just the value, not a dictionary so it would be inconsistent to have a dictionary here.

src/Utilities/penalty_relaxation.jl
odow: Done

src/Utilities/penalty_relaxation.jl
odow: I added `ScalarPenaltyRelaxation`. Does that help?

src/Utilities/penalty_relaxation.jl
```diff
@@ -0,0 +1,310 @@
...
+modifies the model in-place to create a penalized relaxation of the constraints.
+
+!!! warning
+    This is a destructive routine that modifies the model in-place. If you don't
+    want to modify the original model, use `JuMP.copy_model` to create a copy
+    before calling [`MOI.modify`](@ref).
+
+## Reformulation
+
+The penalty relaxation modifies constraints of the form ``f(x) \\in S`` into
```
blegat, line 170: I would just redirect to the docstring of `ScalarPenaltyRelaxation`

src/Utilities/penalty_relaxation.jl
```diff
@@ -0,0 +1,310 @@
...
+ 0.0 + 1.0 v[1] - 1.0 v[2] <= 2.0
+
+VariableIndex-in-GreaterThan{Float64}
+ v[2] >= 0.0
+
+julia> f isa MOI.ScalarAffineFunction{Float64}
+true
+```
+"""
+struct ScalarPenaltyRelaxation{T}
```
blegat, line 63: Isn't it an `AbstractFunctionModification` ?

src/Utilities/penalty_relaxation.jl
```diff
@@ -0,0 +1,310 @@
...
+    penalty::T
+end
+
+function MOI.supports(
```
blegat, line 67: `supports` is only defined for attributes and submittable.

src/Utilities/penalty_relaxation.jl
```diff
@@ -0,0 +1,310 @@
...
+    ::Type{S},
+) where {T,F,S}
+    for ci in MOI.get(model, MOI.ListOfConstraintIndices{F,S}())
+        penalty = get(relax.penalties, ci, relax.default)
+        if penalty === nothing
+            continue
+        end
+        attr = ScalarPenaltyRelaxation(penalty)
+        if !MOI.supports(model, attr, MOI.ConstraintIndex{F,S})
+            @warn("Skipping PenaltyRelaxation of constraints of type $F-in-$S")
```
blegat, line 304: `supports` is only defined for attributes and submittable. Is that `if` important ?

src/Utilities/penalty_relaxation.jl
odow: Changed to a default fallback.

src/Utilities/penalty_relaxation.jl
odow: Added a comment explaining

src/Utilities/penalty_relaxation.jl
```diff
@@ -0,0 +1,302 @@
...
+For each constraint `ci`, the penalty passed to [`Utilities.ScalarPenaltyRelaxation`](@ref)
+is `get(penalties, ci, default)`. If the value is `nothing`, because `ci` does
+not exist in `penalties` and `default = nothing`, then the constraint is
+skipped.
+
+## Return value
+
+`MOI.modify(model, PenaltyRelaxation())` returns a
+`Dict{MOI.ConstraintIndex,MOI.ScalarAffineFunction}` that maps constraint
+indices to a [`MOI.ScalarAffineFunction`](@ref) comprised of `y + z`. In an
```
joaquimg, line 178: > comprised of `y + z`

is not clear...

src/Utilities/penalty_relaxation.jl
```diff
@@ -0,0 +1,302 @@
...
+    MOI.set(model, MOI.ObjectiveFunction{typeof(f)}(), f)
+    return MOI.MIN_SENSE
+end
+
+function MOI.modify(
+    ::MOI.ModelLike,
+    ::MOI.ConstraintIndex,
+    ::ScalarPenaltyRelaxation,
+)
+    # A generic fallback if modification is not supported.
```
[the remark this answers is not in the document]
odow: No, it's for any set that isn't supported.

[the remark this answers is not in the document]
odow: We either need `supports` or we need this fallback.

src/Utilities/penalty_relaxation.jl
```diff
@@ -0,0 +1,302 @@
...
+"""
+struct ScalarPenaltyRelaxation{T} # <: MOI.AbstractFunctionModification
+    # We don't make this a subtype of AbstractFunctionModification to avoid some
+    # ambiguities with generic methods in Utilities and Bridges. The underlying
+    # reason is that these reformulations can be written using the high-level
+    # MOI API, so we don't need special handling for bridges and utilities.
+    penalty::T
+end
+
+function _change_set_to_min_if_necessary(
```
blegat, line 70:
```suggestion
function _change_sense_to_min_if_necessary(
```

src/Utilities/penalty_relaxation.jl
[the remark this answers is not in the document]
blegat: > No, it's for any set that isn't supported.

Isn't the method below defined for any `AbstractScalarSet` ?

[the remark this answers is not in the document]
odow: I added the `try-catch`.

blegat (approved):
I would probably consider adding a default fallback throwing `ModifyConstraintNotAllowed` but then we should do it for all modifications so it's better done in a separate PR

PR #2008, opened 2022-09-26
[Utilities] fix unsupported get in model.jl

### LINKED ISSUE #2005: `MOIU._single_variable_flag`
When I was implementing a new variable set `Spin <: AbstractScalarSet` for $s \in \\\{-1, 1\\\}$ I kept running into calls to `MOIU._single_variable_flag`.

After correctly defining the JuMP extension interface things seemed to work but the problem persisted as I needed to perform `MOI` queries such as `MOI.get(model, MOI.ListOfConstraintIndices{VI, Spin}())`.

Here is a MWE:
```julia
# MWE imports:
using JuMP
const MOIU = MOI.Utilities

# actual lib code:
# ~*~ MOI ~*~ #
@doc raw"""
    Spin()

The set ``\left\lbrace{}{-1, 1}\right\rbrace{}``.
""" struct Spin <: MOI.AbstractScalarSet end

function MOIU._to_string(options::MOIU._PrintOptions, ::Spin)
    return string(MOIU._to_string(options, ∈), " {-1, 1}")
end

function MOIU._to_string(::MOIU._PrintOptions{MIME"text/latex"}, ::Spin)
    return raw"\in \left\lbrace{}{-1, 1}\right\rbrace{}"
end

# ~*~ JuMP ~*~ #
# Ref: https://jump.dev/JuMP.jl/stable/developers/extensions/#extend_variable_macro

struct SpinInfo
    info::JuMP.VariableInfo
end

function JuMP.build_variable(
    ::Function,
    info::JuMP.VariableInfo,
    ::Type{Spin};
    kwargs...
)
    return SpinInfo(info)
end

function JuMP.add_variable(
    model::JuMP.Model,
    info::SpinInfo,
    name::String,
)
    x = JuMP.add_variable(model, JuMP.ScalarVariable(info.info), name)

    JuMP.@constraint(model, x ∈ Spin())

    return x
end

JuMP.in_set_string(::MIME"text/plain", ::Spin) = "spin"
JuMP.in_set_string(::MIME"text/latex", ::Spin) = raw"\in \left\langle{}-1, 1\right\rangle{}}"

# MWE again:
m = MOIU.Model{Float64}()
MOI.get(m, MOI.ListOfConstraintIndices{MOI.VariableIndex, Spin}()) # Method Error: no method matching _single_variable_flag(::Type{Spin})
# `MOI.ConstraintIndex{MOI.VariableIndex, Spin}[]` is the expected return
```

A pirate fix for this was to define
```julia
MOIU._single_variable_flag(::Type{Spin}) = 0x0100 # last flag used was 0x0080
```

The polite fix was to ask if `MOI.supports(model, VI, Spin)` before asking for the possibly empty list.

I think that maybe other `MOI.get` methods involving constraints might fail as they reach `_single_variable_flag` too.

AFAIK, this whole flag system is only used to do basic comparison between variable types by calling `!iszero(mask & flag)` on them.
Is this still necessary to avoid type comparison `===`/inclusion `<:` or should we consider replacing this mechanism?

odow:
Closes #2005

### DIFF
--- a/src/Utilities/model.jl
+++ b/src/Utilities/model.jl
@@ -416,7 +416,10 @@ end
 function MOI.get(
     model::AbstractModel,
     attr::MOI.NumberOfConstraints{MOI.VariableIndex,S},
-) where {S}
+)::Int64 where {S}
+    if !MOI.supports_constraint(model, MOI.VariableIndex, S)
+        return 0
+    end
     return MOI.get(model.variables, attr)
 end
 
@@ -441,6 +444,9 @@ function MOI.get(
     model::AbstractModel,
     attr::MOI.ListOfConstraintIndices{MOI.VariableIndex,S},
 ) where {S}
+    if !MOI.supports_constraint(model, MOI.VariableIndex, S)
+        return MOI.ConstraintIndex{MOI.VariableIndex,S}[]
+    end
     return MOI.get(model.variables, attr)
 end
 
--- a/test/Utilities/model.jl
+++ b/test/Utilities/model.jl
@@ -493,6 +493,19 @@ function test_extension_dictionary()
     return
 end
 
+struct _UnsupportedSetIssue2005 <: MOI.AbstractScalarSet end
+
+function test_get_unsupported_constraint()
+    model = MOI.Utilities.Model{Float64}()
+    S = _UnsupportedSetIssue2005
+    for F in (MOI.VariableIndex, MOI.ScalarAffineFunction{Float64})
+        @test MOI.get(model, MOI.NumberOfConstraints{F,S}()) == 0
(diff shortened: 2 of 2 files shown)


blegat:
Wasn't it already fixed by https://github.com/jump-dev/MathOptInterface.jl/pull/2007 ?

blegat:
Getting a `MethodError` is the right behavior so I don't thing there is anything to fix. For non-variable constraints, this is also what you get. We could have a default fallback in `src/attributes` if we want to replace the `MethodError` by something else. Or we could say that all implementation should do it explicitly but then we need to update all of them in MOI.Utilities at least to have something consistent (and tests as well would be even better)

PR #2010, opened 2022-09-30
Revert "[Utilities] fix unsupported get in model.jl (#2008)"

blegat:
This reverts https://github.com/jump-dev/MathOptInterface.jl/pull/2008 for the reason detailed in https://github.com/jump-dev/MathOptInterface.jl/pull/2008#issuecomment-1260701844

### DIFF
--- a/src/Utilities/model.jl
+++ b/src/Utilities/model.jl
@@ -416,10 +416,7 @@ end
 function MOI.get(
     model::AbstractModel,
     attr::MOI.NumberOfConstraints{MOI.VariableIndex,S},
-)::Int64 where {S}
-    if !MOI.supports_constraint(model, MOI.VariableIndex, S)
-        return 0
-    end
+) where {S}
     return MOI.get(model.variables, attr)
 end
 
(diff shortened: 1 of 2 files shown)


odow:
For clarity, the comment was

> Getting a MethodError is the right behavior so I don't thing there is anything to fix. For non-variable constraints, this is also what you get. We could have a default fallback in src/attributes if we want to replace the MethodError by something else. Or we could say that all implementation should do it explicitly but then we need to update all of them in MOI.Utilities at least to have something consistent (and tests as well would be even better)

The MethodError isn't what you get for non-variable constraints:
```Julia
julia> using MathOptInterface

julia> const MOI = MathOptInterface
MathOptInterface

julia> struct MySet <: MOI.AbstractVectorSet end

julia> model = MOI.Utilities.Model{Float64}()
MOIU.Model{Float64}

julia> MOI.get(model, MOI.NumberOfConstraints{MOI.VectorAffineFunction{Float64},MySet}())
0

julia> MOI.get(model, MOI.ListOfConstraintIndices{MOI.VectorAffineFunction{Float64},MySet}())
MathOptInterface.ConstraintIndex{MathOptInterface.VectorAffineFunction{Float64}, MySet}[]
```

We should improve the tests.

blegat:
Good point, then https://github.com/jump-dev/MathOptInterface.jl/pull/2011 might be better

PR #2011, opened 2022-09-30
ListOfConstraintIndices and NumberOfConstraints

blegat:
With https://github.com/jump-dev/MathOptInterface.jl/pull/2008, it now returns a correct value for unsupported constraints. If we want this behavior to be consistent with other implemenation of `MOI.ModelLike`, we don't want to have to implement it for all of them so we could add the following fallbacks so that we don't have to implement it for all implementations.

Closes https://github.com/jump-dev/MathOptInterface.jl/pull/2010

### DIFF
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -1145,7 +1145,7 @@ function _may_have_been_added(
 ) where {F, S}
     return MOI.supports_constraint(model, F, S) ||
         (S === MOI.VariableIndex && MOI.supports_add_constrained_variable(model, S)) ||
-        (S == MOI.VectorOfVariables && MOI.supports_add_constrained_variables(model, S))
+        (S === MOI.VectorOfVariables && MOI.supports_add_constrained_variables(model, S))
 end
 
 function get(
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -1148,7 +1148,7 @@ function _may_have_been_added(
         (S == MOI.VectorOfVariables && MOI.supports_add_constrained_variables(model, S))
 end
 
-function get_fallback(
+function get(
     model::ModelLike,
     attr::ListOfConstraintIndices{F,S},
 ) where {F,S}
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -1138,6 +1138,27 @@ indices of type `F`-in-`S` in the model (i.e., of length equal to the value of
 """
 struct ListOfConstraintIndices{F,S} <: AbstractModelAttribute end
 
+function _may_have_been_added(
+    model::ModelLike,
+    ::Type{F},
+    ::Type{S}
+) where {F, S}
+    return MOI.supports_constraint(model, F, S) ||
+        (S === MOI.VariableIndex && MOI.supports_add_constrained_variable(model, S)) ||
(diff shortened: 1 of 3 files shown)


src/attributes.jl
```diff
@@ -1138,6 +1138,27 @@ indices of type `F`-in-`S` in the model (i.e., of length equal to the value of
...
+    model::ModelLike,
+    ::Type{F},
+    ::Type{S}
+) where {F, S}
+    return MOI.supports_constraint(model, F, S) ||
+        (S === MOI.VariableIndex && MOI.supports_add_constrained_variable(model, S)) ||
+        (S == MOI.VectorOfVariables && MOI.supports_add_constrained_variables(model, S))
+end
+
+function get_fallback(
```
blegat, line 1151:
```suggestion
function get(
```

src/attributes.jl
```diff
@@ -1148,6 +1169,17 @@ struct NumberOfConstraints{F,S} <: AbstractModelAttribute end
 
 attribute_value_type(::NumberOfConstraints) = Int64
 
+function get_fallback(
```
blegat, line 1172:
```suggestion
function get(
```

src/attributes.jl
```diff
@@ -1138,6 +1138,27 @@ indices of type `F`-in-`S` in the model (i.e., of length equal to the value of
...
 struct ListOfConstraintIndices{F,S} <: AbstractModelAttribute end
 
+function _may_have_been_added(
+    model::ModelLike,
+    ::Type{F},
+    ::Type{S}
+) where {F, S}
+    return MOI.supports_constraint(model, F, S) ||
+        (S === MOI.VariableIndex && MOI.supports_add_constrained_variable(model, S)) ||
+        (S == MOI.VectorOfVariables && MOI.supports_add_constrained_variables(model, S))
```
joaquimg, line 1148:
```suggestion
        (S === MOI.VectorOfVariables && MOI.supports_add_constrained_variables(model, S))
```
for consistency?

odow:
I need to think about this. This feels like the wrong fix

src/attributes.jl
```diff
@@ -1138,6 +1138,27 @@ indices of type `F`-in-`S` in the model (i.e., of length equal to the value of
...
+        (S === MOI.VariableIndex && MOI.supports_add_constrained_variable(model, S)) ||
+        (S === MOI.VectorOfVariables && MOI.supports_add_constrained_variables(model, S))
+end
+
+function get(
+    model::ModelLike,
+    attr::ListOfConstraintIndices{F,S},
+) where {F,S}
+    if _may_have_been_added(model, F, S)
+        error("Not implemented")
```
odow, line 1156: Why have this error instead of just returning the empty vector? Same with erroring instead of returning `0` below.

blegat: The error should be improved but basically, the MOI implementation is responsible for implementing this in case the constraint type is supported. Returning an empty vector would be incorrect in case the MOI implementation does not implement it. I don't see situation where this fallback would be correct. It would mean that an MOI implementation supports a constraint but knows that there is always going to be zero constraints of that type

PR #2016, opened 2022-10-14
[FileFormats.MPS] fix parsing of OBJSENSE

### LINKED ISSUE #2015: FileFormats.MPS OBJSENSE
I have a model like the following:
```
NAME          model
OBJSENSE
  MIN
ROWS
 N  Obj 
 L  c1 
```

it seems the MPS parser does not support the OBJSENSE and sense on two different lines (or the OBJSENSE altogether?)
For what it's worth, the Wikipedia entry has this specification

odow:
Closes #2015


matbesancon (approved):

matbesancon:
So apparently, the optional sense is a section so it could even make sense to write it on two separate lines and not one
https://www.ibm.com/docs/en/icos/22.1.0?topic=extensions-objective-sense-name-offset-in-mps-files

Issue #2017, opened 2022-10-17
Test problems on nightly


odow:
Recent jobs have been failing in different ways:

* Segfault https://github.com/jump-dev/MathOptInterface.jl/actions/runs/3261321958/jobs/5355955760
* Suprious allocations https://github.com/jump-dev/MathOptInterface.jl/actions/runs/3255175420/jobs/5344217939

I've been trying, without much luck, to reproduce the segfault locally. The allocations are a change in sorting: https://github.com/JuliaLang/julia/pull/45222

PR #2018, opened 2022-10-17
Omit allocation test due to changes in sort on nightly

odow:
Part of #2017 

Recent changes to sorting have introduced allocations: 
 * https://github.com/JuliaLang/julia/issues/45326
 * https://github.com/JuliaLang/julia/pull/45222


odow:
Merging because this is causing tests to fail, but I won't close`#2017` just yet.

PR #2021, opened 2022-10-18
Add fallback for ListOfConstraintIndices and NumberOfConstraints

odow:
This PR adds a fallback for `ListOfConstraintIndices` and `NumberOfConstraints` to return default values in the case that the solver does't support the particular function-set combination.

This has come up a few times:

Closes #2011 
Closes #2010
Follow-up to #2008

We already had something similar in two places:

https://github.com/jump-dev/MathOptInterface.jl/blob/71defad7679ca34987860bbfd660fc290b1e2290/src/Utilities/struct_of_constraints.jl#L142-L150

https://github.com/jump-dev/MathOptInterface.jl/blob/71defad7679ca34987860bbfd660fc290b1e2290/src/Utilities/struct_of_constraints.jl#L132-L140

#2008 added it to two more:

https://github.com/jump-dev/MathOptInterface.jl/blob/71defad7679ca34987860bbfd660fc290b1e2290/src/Utilities/model.jl#L416-L424

https://github.com/jump-dev/MathOptInterface.jl/blob/71defad7679ca34987860bbfd660fc290b1e2290/src/Utilities/model.jl#L443-L451

So this PR is essentially just making some implied behavior concrete.

I've tested with Gurobi:

```julia
julia> model = Gurobi.Optimizer();

julia> MOI.Test.test_attribute_unsupported_list_of_constraint_indices(model, config)

```

But once this PR is open, I'll run SolverTests to see the damage: https://github.com/jump-dev/SolverTests/pull/19

### DIFF
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -1142,8 +1142,8 @@ function get_fallback(
     model::ModelLike,
     ::ListOfConstraintIndices{F,S},
 ) where {F,S}
-    if MOI.supports_constraint(model, F, S)
-        throw(MOI.GetAttributeNotAllowed(MOI.ListOfConstraintIndices{F,S}()))
+    if supports_constraint(model, F, S)
+        throw(GetAttributeNotAllowed(ListOfConstraintIndices{F,S}()))
     end
     return ConstraintIndex{F,S}[]
 end
@@ -1159,8 +1159,8 @@ struct NumberOfConstraints{F,S} <: AbstractModelAttribute end
 attribute_value_type(::NumberOfConstraints) = Int64
 
 function get_fallback(model::ModelLike, ::NumberOfConstraints{F,S}) where {F,S}
-    if MOI.supports_constraint(model, F, S)
-        throw(MOI.GetAttributeNotAllowed(MOI.NumberOfConstraints{F,S}()))
+    if supports_constraint(model, F, S)
+        throw(GetAttributeNotAllowed(NumberOfConstraints{F,S}()))
     end
     return Int64(0)
 end
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -1138,7 +1138,13 @@ indices of type `F`-in-`S` in the model (i.e., of length equal to the value of
 """
 struct ListOfConstraintIndices{F,S} <: AbstractModelAttribute end
 
-function get_fallback(::ModelLike, ::ListOfConstraintIndices{F,S}) where {F,S}
+function get_fallback(
+    model::ModelLike,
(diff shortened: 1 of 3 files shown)


odow:
No issues identified by SolverTests, https://github.com/jump-dev/SolverTests/pull/19, so I'd say this is okay.

blegat:
Since all MOI wrappers tested by SolverTests are correctly implemented, this is not different from https://github.com/jump-dev/MathOptInterface.jl/pull/2011.
However, as detailed in https://github.com/jump-dev/MathOptInterface.jl/pull/2011#discussion_r998141214, I don't see in which situation, it is a good thing to add a fallback for supported constraint types.
If a solver supports a constraint type and does not implement the getter, returning zero is never a good implementation, a clear error saying that this is not implemented by the MOI implementation so a cache should be used if that feature is needed is what is most suited in that situation

odow:
Added a `GetAttributeNotAllowedError`

blegat:
`supports_constraint` is not enough, consider for instance CSDP with VectorOfVariables in PSD, we need some condition like in https://github.com/jump-dev/MathOptInterface.jl/pull/2011

odow:
Done.

blegat (approved):
Looks good, maybe add a comment in the code since it is not so obvious why we are doing what we are doing ?

src/attributes.jl
```diff
@@ -1138,6 +1138,18 @@ indices of type `F`-in-`S` in the model (i.e., of length equal to the value of
 """
 struct ListOfConstraintIndices{F,S} <: AbstractModelAttribute end
 
+function get_fallback(
+    model::ModelLike,
+    ::ListOfConstraintIndices{F,S},
+) where {F,S}
+    if supports_constraint(model, F, S) ||
```
odow, line 1145:
```suggestion
    # Throw error here so that we don't return an incorrect value if solvers
    # don't implement this method for some constraints they support.
    if supports_constraint(model, F, S) ||
```

src/attributes.jl
```diff
@@ -1148,6 +1160,15 @@ struct NumberOfConstraints{F,S} <: AbstractModelAttribute end
 
 attribute_value_type(::NumberOfConstraints) = Int64
 
+function get_fallback(model::ModelLike, ::NumberOfConstraints{F,S}) where {F,S}
+    if supports_constraint(model, F, S) ||
```
odow, line 1164:
```suggestion
    # Throw error here so that we don't return an incorrect value if solvers
    # don't implement this method for constraints they support.
    if supports_constraint(model, F, S) ||
```

PR #2024, opened 2022-10-21
[Test] add requires for test_constraint_get_ConstraintIndex

matbesancon:
I got surprised the test fails for an optimizer even though it does not specify that it supports constraint names, shouldn't there be a `@requires` there?
Possibly also checking the requires with `Type{CI{F,S}}`

### DIFF
--- a/src/Test/test_constraint.jl
+++ b/src/Test/test_constraint.jl
@@ -16,10 +16,10 @@ function test_constraint_get_ConstraintIndex(
     model::MOI.ModelLike,
     ::Config{T},
 ) where {T}
+    x = MOI.add_variable(model)
     f = T(1) * x
     CI = MOI.ConstraintIndex{typeof(f),MOI.GreaterThan{T}}
     @requires MOI.supports(model, MOI.ConstraintName(), CI)
-    x = MOI.add_variable(model)
     c1 = MOI.add_constraint(model, f, MOI.GreaterThan(T(1)))
     MOI.set(model, MOI.ConstraintName(), c1, "c1")
     c2 = MOI.add_constraint(model, f, MOI.LessThan(T(2)))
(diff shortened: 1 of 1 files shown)


src/Test/test_constraint.jl
```diff
@@ -16,6 +16,7 @@ function test_constraint_get_ConstraintIndex(
     model::MOI.ModelLike,
     ::Config{T},
 ) where {T}
+    @requires MOI.supports(model, MOI.ConstraintName(), MOI.ConstraintIndex)
```
odow, line 19: I don't think this is quite right, because every solver should fail the `MOI.ConstraintIndex` check. 

It's asking for a method like `MOI.supports(::MOI.ModelLike, ::MOI.ConstraintName, ::Type{MOI.ConstraintIndex})` which is different to `MOI.supports(::MOI.ModelLike, ::MOI.ConstraintName, ::Type{MOI.ConstraintIndex{F,S}) where {F,S}`.

But we could use explicit `F = MOI.ScalarAffineFunction{T}`  and `S = MOI.GreaterThan{T}` (and another for `S=MOI.LessThan{T}`.

matbesancon: just did

src/Test/test_constraint.jl
```diff
@@ -16,10 +16,12 @@ function test_constraint_get_ConstraintIndex(
     model::MOI.ModelLike,
     ::Config{T},
 ) where {T}
+    f = T(1) * x
+    @requires MOI.supports(model, MOI.ConstraintName(), MOI.ConstraintIndex{typeof(f), MOI.GreaterThan{T}})
```
odow, line 20:
```suggestion
    CI = MOI.ConstraintIndex{typeof(f),MOI.GreaterThan{T}}
    @requires MOI.supports(model, MOI.ConstraintName(), CI)
```

PR #2029, opened 2022-10-27
[FileFormats] add default fallback for read_from_file and write_to_file

### LINKED ISSUE #2028: write_to_file and read_from_file fallbacks
Everytime I need to read/write from files I end up copy-pasting from the MOI.FileFormats documentation, could we add:

```julia
function MOI.write_to_file(src::MOI.ModelLike, file)
    dest = MOI.FileFormats.Model(filename=file)
    MOI.copy_to(dest, src)
    MOI.write_to_file(dest, file)
end
```

and 

```julia
function MOI.read_from_file(dest::MOI.ModelLike, file)
    src = MOI.FileFormats.Model(filename=file)
    MOI.copy_to(dest, src)
    # assuming here the indices map would not be needed
    return nothing
end
```

odow:
Closes #2028

cc @matbesancon is this what you had in mind?

### DIFF
--- a/src/MathOptInterface.jl
+++ b/src/MathOptInterface.jl
@@ -118,28 +118,41 @@ end
 """
     write_to_file(model::ModelLike, filename::String)
 
-Writes the current model data to the given file.
+Write the current model to the file at `filename`.
+
 Supported file types depend on the model type.
 """
-function write_to_file end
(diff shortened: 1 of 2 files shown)


matbesancon (approved):

matbesancon:
yes great!

PR #2032, opened 2022-10-31
[Nonlinear] add Vector argument as possible cause to FORWARD_DIFF_METHOD_ERROR

### LINKED ISSUE #2031: Misleading error message with user defined functions?
Hi!

I recently starting using JuMP for nonlinear optimization. Here I stumbled over an error message when registering my own function to a model. The error was telling me that my function does not support differentiation by ForwardDiff and that I can debug this in an easier way using ForwardDiff.jl to try whether im able to calculate gradients directly with ForwardDiff.gradient(). This worked perfectly fine for my function. After a while I found out that the problem was not that my function could not be differentiated with ForwardDiff, but rather that user defined functions with vectors as input are not supported. 

Maybe the error message is a bit misleading?

Here is a basic example of the problem I encountered:
(JuMP v1.4.0, ForwardDiff v0.10.32) 

```julia
julia> using JuMP, ForwardDiff

julia> foo(x) = sum(x.^2)
foo (generic function with 1 method)

julia> x = [1.0, 2.0]
2-element Vector{Float64}:
 1.0
 2.0

julia> ForwardDiff.gradient(foo, x)
2-element Vector{Float64}:
 2.0
 4.0

julia> model = Model()
A JuMP Model
Feasibility problem with:
Variables: 0
Model mode: AUTOMATIC
CachingOptimizer state: NO_OPTIMIZER
Solver name: No optimizer attached.

julia> register(model, :foo, 2, foo; autodiff=true)
ERROR: Unable to register the function :foo because it does not support differentiation via ForwardDiff.

Common reasons for this include:
```

odow:
Closes #2031

@mpirke is this better?

### DIFF
--- a/src/Nonlinear/operators.jl
+++ b/src/Nonlinear/operators.jl
@@ -143,12 +143,23 @@ end
 const _FORWARD_DIFF_METHOD_ERROR_HELPER = raw"""
 Common reasons for this include:
 
- * the function assumes `Float64` will be passed as input, it must work for any
+ * The function takes `f(x::Vector)` as input, instead of the splatted
+   `f(x...)`.
+ * The function assumes `Float64` will be passed as input, it must work for any
    generic `Real` type.
- * the function allocates temporary storage using `zeros(3)` or similar. This
+ * The function allocates temporary storage using `zeros(3)` or similar. This
    defaults to `Float64`, so use `zeros(T, 3)` instead.
 
-As an example, instead of:
+As examples, instead of:
+```julia
+my_function(x::Vector) = sum(x.^2)
+```
+use:
+```julia
+my_function(x::T...) where {T<:Real} = sum(x[i]^2 for i in 1:length(x))
+```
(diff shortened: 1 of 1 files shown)


src/Nonlinear/operators.jl
```diff
@@ -143,12 +143,23 @@ end
...
+ * The function takes `f(x::Vector)` as input, instead of the splatted
+   `f(x...)`.
+ * The function assumes `Float64` will be passed as input, it must work for any
    generic `Real` type.
- * the function allocates temporary storage using `zeros(3)` or similar. This
+ * The function allocates temporary storage using `zeros(3)` or similar. This
    defaults to `Float64`, so use `zeros(T, 3)` instead.
 
-As an example, instead of:
+As examples, instead of:
```
mpirke, line 153: That makes things clear!

mpirke:
Yes. Great!

PR #2034, opened 2022-11-01
[Utilities] Re-enable allocation test from #2018

### LINKED ISSUE #2017: Test problems on nightly

odow:
Tracking to see if #2017 changed.

But I don't expect it to have. This is a regression introduced by https://github.com/JuliaLang/julia/pull/45222.

Closes #2017

### DIFF
--- a/src/Utilities/functions.jl
+++ b/src/Utilities/functions.jl
@@ -833,7 +833,7 @@ function _sort_and_compress!(x::Vector)
     if length(x) == 0
         return
     end
-    sort!(x, QuickSort, Base.Order.ord(isless, MOI.term_indices, false))
+    sort!(x, InsertionSort, Base.Order.ord(isless, MOI.term_indices, false))
     i = 1
     @inbounds for j in 2:length(x)
         if MOI.term_indices(x[i]) == MOI.term_indices(x[j])
--- a/src/Utilities/functions.jl
+++ b/src/Utilities/functions.jl
@@ -833,7 +833,7 @@ function _sort_and_compress!(x::Vector)
     if length(x) == 0
         return
     end
-    sort!(x; by = MOI.term_indices)
+    sort!(x, QuickSort, Base.Order.ord(isless, MOI.term_indices, false))
     i = 1
     @inbounds for j in 2:length(x)
         if MOI.term_indices(x[i]) == MOI.term_indices(x[j])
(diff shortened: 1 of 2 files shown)


test/Utilities/functions.jl
```diff
@@ -1058,7 +1058,9 @@ function _test_canonicalization(
     @test MOI.Utilities.canonical(g) !== g
     # There are some changes to sorting in Julia v1.9 that now mean sorting
     # allocates.
-    # @test @allocated(MOI.Utilities.canonicalize!(f)) == 0
+    if VERSION < v"1.9"
```
odow, line 1061:
```suggestion
    if VERSION < v"1.9.0-DEV"
```

src/Utilities/functions.jl
```diff
@@ -833,7 +833,7 @@ function _sort_and_compress!(x::Vector)
     if length(x) == 0
         return
     end
-    sort!(x; by = MOI.term_indices)
+    sort!(x, QuickSort, Base.Order.ord(isless, MOI.term_indices, false))
```
odow, line 836:
```suggestion
    sort!(x, InsertionSort, Base.Order.ord(isless, MOI.term_indices, false))
```

test/Utilities/functions.jl
```diff
@@ -1058,7 +1058,9 @@ function _test_canonicalization(
     @test MOI.Utilities.canonical(g) !== g
     # There are some changes to sorting in Julia v1.9 that now mean sorting
     # allocates.
-    # @test @allocated(MOI.Utilities.canonicalize!(f)) == 0
+    if VERSION < v"1.9.0-DEV"
```
odow, line 1061:
```suggestion
    if true # VERSION < v"1.9.0-DEV"
```

odow:
@blegat I've changed to use InsertionSort since QuickSort allocates on Julia 1.9. Thoughts?

odow:
Merging because insertion sort is probably a better choice here anyway. We mostly work with small arrays because most constraints are sparse and have few terms.

PR #2038, opened 2022-11-02
[Bridges] Use is_coefficient_type

blegat:
The bridge will fail with the error given in https://github.com/jump-dev/MathOptInterface.jl/issues/2035 whenever the function type is different from `T` so this should fix all these cases, not only the `Complex` case.

### DIFF
--- a/src/Bridges/Objective/bridges/slack.jl
+++ b/src/Bridges/Objective/bridges/slack.jl
@@ -81,24 +81,10 @@ function supports_objective_function(
 end
 
 function supports_objective_function(
-    ::Type{<:SlackBridge},
-    ::Type{<:MOI.AbstractScalarFunction},
-)
-    return true
-end
-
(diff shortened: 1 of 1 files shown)


src/Bridges/Objective/bridges/slack.jl
```diff
@@ -81,24 +81,10 @@ function supports_objective_function(
...
-end
-
-function supports_objective_function(
-    ::Type{<:SlackBridge},
-    ::Type{<:MOI.ScalarQuadraticFunction{<:Complex}},
-)
-    return false
+    ::Type{<:SlackBridge{T}},
+    ::Type{F},
+) where {T,F<:<:MOI.AbstractScalarFunction}
```
odow, line 86:
```suggestion
) where {T,F<:MOI.AbstractScalarFunction}
```

Issue #2033, opened 2022-10-21
Implement Utilities.distance_to_set for more sets


freemin7:
**Is your feature request related to a problem? Please describe.**
I have a model which contains a MathOptInterface.PowerCone and want to verify a the primal feasibility of a point.

**Describe the solution you'd like**
Support feasibility checking of all standard sets maintained by jump-dev. Short term i need a workaround.

**Describe alternatives you've considered**
I considered monkey-patching `_distance_to_set(var::Vector{Float64},cone::MathOptInterface.PowerCone) = distance_to_set(var::Vector{Float64},cone::MathOptInterface.PowerCone)` from MathOptSetDistances by @matbesancon however that did not work.

odow:
Assuming we merge #2048, what distance would we use for this?

MOSD uses:

https://github.com/matbesancon/MathOptSetDistances.jl/blob/a449c4b178855db0b461d1aa13d60a5ae81b0eac/src/distance_sets.jl#L171-L185

freemin7:
I think we established that what distance is used doesn't matter for the sake of feasibility checking as long as it implies the same [discrete metric](https://mathworld.wolfram.com/DiscreteMetric.html). 
Since i am not aware of any code base which depends on something else than the discrete metric pick and since this was "broken" before for those cones no code should change by the decision. I'd suggest we pick whatever is cheapest to compute and document what was used and that we guarantee that under <=0 it satisfies discrete metric. Should different features be required in the future by a code base this decision can be revisited in light of more information and multiple dispatch on what kind the user wants.

odow:
@matbesancon what are your thoughts on this? We could potentially add MathOptSetDistances as a pkg extension?

matbesancon:
yes that makes sense. The current dependencies are:
```
BlockDiagonals = "0a1fb500-61f7-11e9-3c65-f5ef3456f9f0"
FillArrays = "1a297f60-69ca-5386-bcde-b61e274b549b"
LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
MathOptInterface = "b8f27783-ece8-5eb3-8dc8-9495eed66fee"
StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
```

matbesancon:
one of the key things to use these functionalities in solvers would be an in-place version of the projection to set

blegat:
[...]

PR #2048, opened 2022-11-14
[Utilities] add distance_to_set

odow:
Docs: https://jump.dev/MathOptInterface.jl/previews/PR2048/submodules/Utilities/reference/#Set-utilities

Reviving the ghost of https://github.com/jump-dev/MathOptInterface.jl/pull/1023

These definitions have been working fine in JuMP, and providing a public API will let external packages add new definitions (even if it's piracy). x-ref #2033.

The up-for-debate topic is the fact, after some thinking, I called it an upper bound on the Euclidean distance. I think this resolves the debate of which distance to use, and how to define it for various sets.

It's _always_ the Euclidean distance between the point and the set, it just might not be the smallest distance. It will be `0` if the point is within the set.

That'd let us define:
```julia
function distance_to_set(
    point::AbstractVector{T},
    set::MOI.ExponentialCone,
) where {T<:Real}
    _check_dimension(point, set)
    x, y, z = point
    # y * e^(x / y) <= z, y > 0
    if y <= 0
        return sqrt((y + eps(T))^2 + min(0, z)^2)
    end
    return max(0.0, y * exp(x / y) - z)
end
```

with tests like
```Julia
function test_exponentialcone()
    @test_throws(
        DimensionMismatch,
        MOI.Utilities.distance_to_set([-1.0, 1.0], MOI.ExponentialCone())
    )
    for (x, d) in [
        [1.0, 1.0, 1.0] => exp(1) - 1.0,
        [1.0, 1.0, 2.0] => exp(1) - 2.0,
        [1.0, 1.0, exp(1)] => 0.0,
        [1.0, 1.0, exp(1) + 0.1] => 0.0,
        [2.0, 0.5, 0.5 * exp(2 / 0.5)] => 0.0,
        [2.0, 0.0, 1.0] => eps(Float64),
        [2.0, -1.0, 1.0] => 1 + eps(Float64),
    ]
        @test MOI.Utilities.distance_to_set(x, MOI.ExponentialCone()) ≈ d
    end
    return
end
```

### DIFF
--- a/docs/src/submodules/Utilities/reference.md
+++ b/docs/src/submodules/Utilities/reference.md
@@ -213,6 +213,14 @@ Utilities.is_diagonal_vectorized_index
 Utilities.side_dimension_for_vectorized_dimension
 ```
 
+## Function utilities
+
+The following utilities are available for sets:
+
+```@docs
+Utilities.distance_to_set
+```
+
 ## DoubleDicts
 
 ```@docs
--- a/docs/src/submodules/Utilities/reference.md
+++ b/docs/src/submodules/Utilities/reference.md
@@ -219,7 +219,7 @@ The following utilities are available for sets:
 
 ```@docs
 Utilities.AbstractDistance
-Utilities.DefaultDistance
+Utilities.ProjectionUpperBoundDistance
 Utilities.distance_to_set
 Utilities.set_dot
 ```
--- a/docs/src/submodules/Utilities/reference.md
+++ b/docs/src/submodules/Utilities/reference.md
@@ -213,12 +213,13 @@ Utilities.is_diagonal_vectorized_index
 Utilities.side_dimension_for_vectorized_dimension
 ```
 
-## Function utilities
+## Set utilities
 
 The following utilities are available for sets:
 
 ```@docs
 Utilities.distance_to_set
+Utilities.set_dot
 ```
 
 ## DoubleDicts
--- a/docs/src/submodules/Utilities/reference.md
+++ b/docs/src/submodules/Utilities/reference.md
@@ -218,6 +218,8 @@ Utilities.side_dimension_for_vectorized_dimension
 The following utilities are available for sets:
 
 ```@docs
+Utilities.AbstractDistance
+Utilities.DefaultDistance
 Utilities.distance_to_set
 Utilities.set_dot
 ```
--- a/src/Utilities/distance_to_set.jl
+++ b/src/Utilities/distance_to_set.jl
@@ -13,7 +13,7 @@ An abstract type used to enabble dispatch of
 abstract type AbstractDistance end
 
 """
-    DefaultDistance() <: AbstractDistance
+    ProjectionUpperBoundDistance() <: AbstractDistance
 
 An upper bound on the minimum distance between `point` and the closest
 feasible point in `set`.
@@ -47,17 +47,17 @@ If the distance is not the smallest upper bound, the docstring of the
 appropriate `distance_to_set` method _must_ describe the way that the distance
 is computed.
 """
-struct DefaultDistance <: AbstractDistance end
+struct ProjectionUpperBoundDistance <: AbstractDistance end
 
 """
     distance_to_set(
-        [d::AbstractDistance = DefaultDistance()],]
+        [d::AbstractDistance = ProjectionUpperBoundDistance()],]
         point::T,
         set::MOI.AbstractScalarSet,
     ) where {T}
 
     distance_to_set(
-        [d::AbstractDistance = DefaultDistance(),]
+        [d::AbstractDistance = ProjectionUpperBoundDistance(),]
         point::AbstractVector{T},
         set::MOI.AbstractVectorSet,
     ) where {T}
@@ -65,9 +65,11 @@ struct DefaultDistance <: AbstractDistance end
 Compute the distance between `point` and `set` using the distance metric `d`. If
 `point` is in the set `set`, this function _must_ return `zero(T)`.
 
-If `d` is omitted, the default distance is [`Utilities.DefaultDistance`](@ref).
+If `d` is omitted, the default distance is [`Utilities.ProjectionUpperBoundDistance`](@ref).
 """
-distance_to_set(point, set) = distance_to_set(DefaultDistance(), point, set)
+function distance_to_set(point, set)
+    return distance_to_set(ProjectionUpperBoundDistance(), point, set)
+end
 
 function distance_to_set(d::AbstractDistance, ::Any, set::MOI.AbstractSet)
     return error(
@@ -81,7 +83,7 @@ end
 ###
 
 function distance_to_set(
-    ::DefaultDistance,
+    ::ProjectionUpperBoundDistance,
     x::T,
     set::MOI.LessThan{T},
 ) where {T<:Real}
@@ -89,7 +91,7 @@ function distance_to_set(
 end
 
 function distance_to_set(
-    ::DefaultDistance,
+    ::ProjectionUpperBoundDistance,
     x::T,
     set::MOI.GreaterThan{T},
 ) where {T<:Real}
@@ -97,7 +99,7 @@ function distance_to_set(
 end
 
 function distance_to_set(
-    ::DefaultDistance,
+    ::ProjectionUpperBoundDistance,
     x::T,
     set::MOI.EqualTo{T},
 ) where {T<:Number}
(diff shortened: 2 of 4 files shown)


blegat:
The Euclidean distance is not be what we want for the PSD cone. Note that we already have this issue with `set_dot` and we just said that when you define a set, you can also choose the scalar product and this choice is communicated in the docstring and in the implementation of `set_dot`. This choice has impact on what is the adjoint so on the dual. `get_fallback` and Dualization are using `set_dot` for that reason and the bridges are also paying attention to this when defining the `ConstraintDual` and `ConstraintDualStart`.
I think it would be best to say that `distance_to_set` is with respect to the distance given by `d(x, y) = ||x - y||` where the norm is defined as `||x|| = \sqrt{<x, x>}` where `<., .>` is `set_dot`.
Can you remind us for which set we compute an upper bound instead of the exact value ?

src/Utilities/distance_to_set.jl
```diff
@@ -0,0 +1,123 @@
...
+
+Compute an upper bound on the minimum distance between `point` and the closest
+feasible point in `set`. If `point` is in the set `set`, this function _must_
+return `zero(T)`.
+
+## Definition of distance
+
+The minimum distance is computed as
+```math
+d(x, \\mathcal{K}) = \\min\\_{y \\in \\mathcal{K}} || x - y ||
```
odow, line 22:
```suggestion
d(x, \\mathcal{K}) = \\min_{y \\in \\mathcal{K}} || x - y ||
```

src/Utilities/distance_to_set.jl
```diff
@@ -0,0 +1,123 @@
...
+```
+where ``x`` is `point` and ``\\mathcal{K}`` is `set`. The norm is computed as
+```math
+||x|| = \\sqrt{\\texttt{set\\_dot}(x, x, \\mathcal{K})}
```
odow, line 26:
```suggestion
||x|| = \\sqrt{\\texttt{set_dot}(x, x, \\mathcal{K})}
```

src/Utilities/distance_to_set.jl
```diff
@@ -0,0 +1,123 @@
...
+```
+where ``\\texttt{set\\_dot}`` is [`Utilities.set_dot`](@ref).
```
odow, line 28:
```suggestion
where ``\\texttt{set_dot}`` is [`Utilities.set_dot`](@ref).
```

blegat:
I see, for the set `{ (t, x) | f(x) <= t }`, then if `t < f(x)`, we could choose the projection `(f(x), x)` which may not be the Euclidean projection but it is an upper bound. In general, we can choose an easy projection and that gives an upper bound.
I think it is ok but maybe we should say that sets for which `distance_to_set` is a non-tight upper bound should say in their docstring what is being done in `distance_to_set`.

src/Utilities/distance_to_set.jl
```diff
@@ -0,0 +1,123 @@
...
+where ``\\texttt{set_dot}`` is [`Utilities.set_dot`](@ref).
+
+In the default case, where the set does not have a specialized method for
+[`Utilities.set_dot`](@ref), the norm is equivalent to the Euclidean norm
+``||x|| = \\sqrt{\\sum x_i^2}``.
+
+## Why an upper bound?
+
+In most cases, `distance_to_set` should return the smallest upper bound, but it
+may return a larger value if the smallest upper bound is expensive to compute.
```
blegat, line 37: Add something like "in which case, this will be specified in the docstring of the set"

odow:
One useful thing about the docs mentioning an upper bound is that there's nothing stopping us using the epigraph distance first, and then later refining to the projection distance in a non-breaking PR.

joaquimg:
> It's always the Euclidean distance between the point and the set, it just might not be the smallest distance.

Can you explain this more precisely?

odow:
> It's always the Euclidean distance between the point and the set, it just might not be the smallest distance.

I guess I meant 

It's always the Euclidean distance between the point and **a point in** the set, it just might not be the smallest distance.

joaquimg:
Got it.

What do you consider the problem with the MathOptSetDistances style?

odow:
> What do you consider the problem with the MathOptSetDistances style?

It's a different trade-off. 

The problem is:

 * If you pick a distance, some sets are harder to compute than others

So there are two options:

 * Let the user pick the distance, and only implement some distances for each set (MathOptSetDistances)
 * Relax the definition of "distance" that we compute (this PR)

We don't have to re-litigate the first option, but it has a few problems, mostly that most users won't know/care what distance they are using, and that it adds complexity. The pros are that the distances are exact.

This PR is useful for getting a distance if all you want to use it for is checking feasibility, where `> 0` is "infeasible" and `=0` is "feasible." It's not useful if you want to use it in an algorithm where you need stable distances or the exact euclidean projection.

joaquimg:
I think having well-defined distances is a strictly superior approach:

- No type piracy: if the user needs to extend/redefine distances or use their own
- Seamless: there will be a default that beginners don't need to know much about.

The default will be your relaxed definition of distance.

odow:
How's this then? 

cc @matbesancon: the more things change, the more they stay the same.

docs/src/submodules/Utilities/reference.md
```diff
@@ -213,6 +213,17 @@ Utilities.is_diagonal_vectorized_index
 Utilities.side_dimension_for_vectorized_dimension
 ```
 
+## Set utilities
+
+The following utilities are available for sets:
+
+```@docs
+Utilities.AbstractDistance
+Utilities.DefaultDistance
```
odow, line 222:
```suggestion
Utilities.UpperBoundProjectionDistance
```

src/Utilities/distance_to_set.jl
```diff
@@ -0,0 +1,185 @@
...
+"""
+    abstract type AbstractDistance end
+
+An abstract type used to enabble dispatch of
+[`Utilities.distance_to_set`](@ref).
+"""
+abstract type AbstractDistance end
+
+"""
+    DefaultDistance() <: AbstractDistance
```
odow, line 16:
```suggestion
    ProjectionUpperBoundDistance() <: AbstractDistance
```

src/Utilities/distance_to_set.jl
```diff
@@ -0,0 +1,185 @@
...
+
+For example, given an epigraph from of a conic set, ``\\{(t, x) | f(x) \\le t\\}``,
+it may be simpler to return ``\\delta`` such that ``f(x) \\le t + \\delta``,
+rather than computing the nearest projection onto the set.
+
+If the distance is not the smallest upper bound, the docstring of the
+appropriate `distance_to_set` method _must_ describe the way that the distance
+is computed.
+"""
+struct DefaultDistance <: AbstractDistance end
```
odow, line 50:
```suggestion
struct ProjectionUpperBoundDistance <: AbstractDistance end
```

src/Utilities/distance_to_set.jl
```diff
@@ -0,0 +1,185 @@
...
+
+"""
+    distance_to_set(
+        [d::AbstractDistance = DefaultDistance()],]
```
odow, line 54:
```suggestion
        [d::AbstractDistance = ProjectionUpperBoundDistance()],]
```

src/Utilities/distance_to_set.jl
```diff
@@ -0,0 +1,185 @@
...
+        point::T,
+        set::MOI.AbstractScalarSet,
+    ) where {T}
+
+    distance_to_set(
+        [d::AbstractDistance = DefaultDistance(),]
```
odow, line 60:
```suggestion
        [d::AbstractDistance = ProjectionUpperBoundDistance(),]
```

src/Utilities/distance_to_set.jl
```diff
@@ -0,0 +1,185 @@
...
+        point::AbstractVector{T},
+        set::MOI.AbstractVectorSet,
+    ) where {T}
+
+Compute the distance between `point` and `set` using the distance metric `d`. If
+`point` is in the set `set`, this function _must_ return `zero(T)`.
+
+If `d` is omitted, the default distance is [`Utilities.DefaultDistance`](@ref).
```
odow, line 68:
```suggestion
If `d` is omitted, the default distance is [`Utilities.ProjectionUpperBoundDistance`](@ref).
```

blegat (approved):

odow:
@matbesancon any comments?

odow:
Bumping this one too: @joaquimg @matbesancon

src/Utilities/distance_to_set.jl
```diff
@@ -0,0 +1,195 @@
...
+
+An abstract type used to enabble dispatch of
+[`Utilities.distance_to_set`](@ref).
+"""
+abstract type AbstractDistance end
+
+"""
+    ProjectionUpperBoundDistance() <: AbstractDistance
+
+An upper bound on the minimum distance between `point` and the closest
```
matbesancon, line 18:
```suggestion
An upper bound on the minimum Euclidean distance between `point` and the closest
```

odow: This depends on the norm defined by `Utilities.set_dot`. In theory it is, but triangle/square matrices confuse things a little, so it isn't always the norm in the input space.

matbesancon:
only drawback is that there could be multiple upper bounds on Euclidean projection so this is not wrong but not necessarily helpful as a default distance.

But overall I'm fine with this concept, we can merge and adapt MOD to use it

src/Utilities/distance_to_set.jl
matbesancon: ah gotcha, there could be an adjective to mention the norm but given it is below that's fine

odow:
> only drawback is that there could be multiple upper bounds on Euclidean projection so this is not wrong but not necessarily helpful as a default distance.

Yeah. This is on purpose. Our main usage internally is for detecting feasibility. If you want to use it in anger in a solver algorithm, not very useful.

joaquimg (approved):

PR #2052, opened 2022-11-20
[Utilities] add DoubleDicts.outer_keys

joaquimg:


odow:
Where is this needed?

joaquimg:
We have already used the internal field `.dict` to get these outer keys a few times, for instance:

https://github.com/jump-dev/DiffOpt.jl/blob/47f06f08cf3af0ef830340990248b55eba6e1044/src/diff_opt.jl#L375

https://github.com/jump-dev/ParametricOptInterface.jl/blob/ae74f92e554aa181cb26595aceb48245b521bafb/src/duals.jl#L25

`outer_keys` makes this very useful step available to the general user that would not know about the internal fields.

It could be useful in wrappers to return `ListOfConstraintsTypesPreset`

src/Utilities/DoubleDicts.jl
```diff
@@ -299,6 +299,55 @@ function Base.keys(d::AbstractDoubleDictInner{F,S}) where {F,S}
...
+        # ...
+    end
+end
+For performance, it is recommended that the inner loop lies in a separate
+function to gurantee type-stability.
+If you only want an iterator rather than a vector use [`outer_keys`](@ref).
+```
+"""
+function outer_keys_vector(d::AbstractDoubleDict; only_nonempty::Bool = true)
+    out = Tuple{DataType,DataType}[]
```
odow, line 342:
```suggestion
    out = Tuple{Type,Type}[]
```

We should never use `DataType`, because it has a few quirks
```julia
julia> Union{Int,Float64} isa DataType
false

julia> Union{Int,Float64} isa Type
true
```

src/Utilities/DoubleDicts.jl
```diff
@@ -299,6 +299,55 @@ function Base.keys(d::AbstractDoubleDictInner{F,S}) where {F,S}
...
+"""
+outer_keys(d::AbstractDoubleDict) = keys(d.dict)
+
+"""
+    outer_keys_vector(d::AbstractDoubleDict; only_nonempty::Bool = true)
+
+Return a vector of outer keys of the `AbstractDoubleDict` `d`.
+If `only_nonempty` is set to true only outer keys that have a nonempty
+set of inner keys will be returned.
+Each outer key is a `Tuple{DataType,DataType}` so that a double loop can
```
odow, line 328:
```suggestion
Each outer key is a `Tuple{Type,Type}` so that a double loop can
```

src/Utilities/DoubleDicts.jl
```diff
@@ -299,6 +299,55 @@ function Base.keys(d::AbstractDoubleDictInner{F,S}) where {F,S}
     return MOI.ConstraintIndex{F,S}.(keys(d.dict))
 end
 
+"""
+    outer_keys(d::AbstractDoubleDict)
+
+Return an iterator over the outer keys of the `AbstractDoubleDict` `d`.
+Each outer key is a `Tuple{DataType,DataType}` so that a double loop can
```
odow, line 306:
```suggestion
Each outer key is a `Tuple{Type,Type}` so that a double loop can
```

src/Utilities/DoubleDicts.jl
```diff
@@ -299,6 +299,55 @@ function Base.keys(d::AbstractDoubleDictInner{F,S}) where {F,S}
...
+for (F, S) in DoubleDicts.outer_keys(dict)
+    for (k, v) in dict[F, S]
+        # ...
+    end
+end
+For performance, it is recommended that the inner loop lies in a separate
+function to gurantee type-stability.
+Some outer keys `(F, S)` might lead to an empty `dict[F, S]`.
+If you only want nonempty `dict[F, S]` use [`outer_keys_vector`](@ref).
+```
```
odow, line 318:
```suggestion
```

src/Utilities/DoubleDicts.jl
```diff
@@ -299,6 +299,55 @@ function Base.keys(d::AbstractDoubleDictInner{F,S}) where {F,S}
...
+
+Return an iterator over the outer keys of the `AbstractDoubleDict` `d`.
+Each outer key is a `Tuple{DataType,DataType}` so that a double loop can
+be easily used
+```julia
+for (F, S) in DoubleDicts.outer_keys(dict)
+    for (k, v) in dict[F, S]
+        # ...
+    end
+end
```
odow, line 313:
````suggestion
end
```

````

File: src/Utilities/DoubleDicts.jl [hunk 6]
```diff
@@ -299,6 +299,55 @@ function Base.keys(d::AbstractDoubleDictInner{F,S}) where {F,S}
...
+For performance, it is recommended that the inner loop lies in a separate
+function to gurantee type-stability.
+Some outer keys `(F, S)` might lead to an empty `dict[F, S]`.
+If you only want nonempty `dict[F, S]` use [`outer_keys_vector`](@ref).
```
odow, line 317:
```suggestion
If you want only nonempty `dict[F, S]`, use [`outer_keys_vector`](@ref).
```

src/Utilities/DoubleDicts.jl
```diff
@@ -299,6 +299,55 @@ function Base.keys(d::AbstractDoubleDictInner{F,S}) where {F,S}
     return MOI.ConstraintIndex{F,S}.(keys(d.dict))
 end
 
+"""
+    outer_keys(d::AbstractDoubleDict)
+
+Return an iterator over the outer keys of the `AbstractDoubleDict` `d`.
+Each outer key is a `Tuple{DataType,DataType}` so that a double loop can
+be easily used
```
odow, line 307:
```suggestion
be easily used:
```

src/Utilities/DoubleDicts.jl
```diff
@@ -299,6 +299,55 @@ function Base.keys(d::AbstractDoubleDictInner{F,S}) where {F,S}
...
+If you only want nonempty `dict[F, S]` use [`outer_keys_vector`](@ref).
+```
+"""
+outer_keys(d::AbstractDoubleDict) = keys(d.dict)
+
+"""
+    outer_keys_vector(d::AbstractDoubleDict; only_nonempty::Bool = true)
+
+Return a vector of outer keys of the `AbstractDoubleDict` `d`.
+If `only_nonempty` is set to true only outer keys that have a nonempty
```
odow, line 326:
```suggestion

If `only_nonempty` is set to true only outer keys that have a nonempty
```

src/Utilities/DoubleDicts.jl
odow, line 328:
```suggestion

Each outer key is a `Tuple{DataType,DataType}` so that a double loop can
```

src/Utilities/DoubleDicts.jl
```diff
@@ -299,6 +299,55 @@ function Base.keys(d::AbstractDoubleDictInner{F,S}) where {F,S}
...
+be easily used
+```julia
+for (F, S) in DoubleDicts.outer_keys_vector(dict)
+    for (k, v) in dict[F, S]
+        # ...
+    end
+end
+For performance, it is recommended that the inner loop lies in a separate
+function to gurantee type-stability.
+If you only want an iterator rather than a vector use [`outer_keys`](@ref).
```
odow, line 338:
```suggestion

If you want an iterator rather than a vector use [`outer_keys`](@ref).
```

src/Utilities/DoubleDicts.jl
joaquimg: interesting

odow: More specifically, `DataType <: Type`, and `DataType` is the "type" of every concrete type. But using `DataType` in dispatch doesn't work for `Union`s or `UnionAll`.

Basically, there's the type system, which at a high level you can just think of as `Type` or `Type{T}`. But because types are also objects in the language, they themselves need a type:
```julia
julia> isa.(Int, (DataType, Union, UnionAll))
(true, false, false)

julia> isa.(Union{Int,Float64}, (DataType, Union, UnionAll))
(false, true, false)

julia> isa.((Vector{T} where {T}), (DataType, Union, UnionAll))
(false, false, true)
```
So restricting the vector to contain `DataType` doesn't let it hold every possible type in Julia.

... and just for good measure, there's this:
```julia
julia> isa.(Union{}, (DataType, Union, UnionAll))
(false, false, false)

julia> typeof(Union{})
Core.TypeofBottom
```

docs/src/submodules/Utilities/reference.md
```diff
@@ -220,4 +220,6 @@ Utilities.DoubleDicts.DoubleDict
 Utilities.DoubleDicts.DoubleDictInner
 Utilities.DoubleDicts.IndexDoubleDict
 Utilities.DoubleDicts.IndexDoubleDictInner
+Utilities.DoubleDicts.outer_keys
+Utilities.DoubleDicts.outer_keys_vector
```
odow, line 224: I guess one question I had: we we need both of these? What about making the second a keyword argument to `outer_keys`?

odow: This works just as well, and if you want the vector you can call `collect`?
````Julia
"""
    outer_keys(d::AbstractDoubleDict; only_nonempty::Bool = false)

Return an iterator over the outer keys of the `AbstractDoubleDict` `d`.

Each outer key is a `Tuple{Type,Type}` so that a double loop can
be easily used (with a function-barrier to ensure the inner loop is
type-stable):
```julia
function function_barrier(dict, ::Type{F}, ::Type{S}) where {F,S}
    for (k, v) in dict[F, S]
        # ...
    end
end
```

If `only_nonempty = true`, only outer keys that have a non-empty set of inner
keys will be returned.
"""
function outer_keys(d::AbstractDoubleDict; only_nonempty::Bool = false)
    return ((F, S) for ((F, S), d) in d.dict if !only_nonempty || length(d) > 0)
end
````

Comment (joaquimg) [reply docs/src/submodules/Utilities/reference.md:224]: This is not Lazy.

I like both versions because one is Lazy and the other is very useful for things like ListOfConstraintTypePresent.

One way to simplify would be:

Keep the current `outer_keys` and add:
```julia
function nonempty_outer_keys(d::AbstractDoubleDict)
    return ((F, S) for ((F, S), d) in d.dict if length(d) > 0)
end
```

src/Utilities/DoubleDicts.jl
```diff
@@ -299,6 +299,59 @@ function Base.keys(d::AbstractDoubleDictInner{F,S}) where {F,S}
...
+    end
+end
+```
+
+For performance, it is recommended that the inner loop lies in a separate
+function to gurantee type-stability.
+Some outer keys `(F, S)` might lead to an empty `dict[F, S]`.
+If you want only nonempty `dict[F, S]`, use [`outer_keys_vector`](@ref).
+"""
+outer_keys(d::AbstractDoubleDict) = keys(d.dict)
```
blegat, line 321: For consistency, this one could also have the `only_nonempty` keyword and return a `Base.Iterators.Filter` in that case. The return type would then be a `Union` though

joaquimg: how about:

Keep the current `outer_keys` and add:

```julia
function nonempty_outer_keys(d::AbstractDoubleDict)
    # using Base.Iterators.Filter
end
```

?

blegat: Yes, I would prefer that. It's easy for the user to call `collect` on top of either functions if a vector is needed

odow (approved):

PR #2056, opened 2022-11-28
[FileFormats.MOF] add Constraint{Primal,Dual}Start attributes

LukasBarner:
x-ref https://github.com/jump-dev/Ipopt.jl/issues/341#issuecomment-1327987869
cc @odow 
This is a first shot at including (primal and dual) warm starts in MOF.
Did I miss anything?


src/FileFormats/MOF/mof.1.2.schema.json
```diff
@@ -40,9 +40,9 @@
                     "name": {
                         "type": "string"
                     },
-                    "primal_start": {
-                        "description": "An initial value that the optimizer may use to warm-start the solution process.",
-                        "type": "number"
+                    "VariablePrimalStart": {
```
odow, line 43: This a breaking change to the schema. Please leave as `primal_start`.

src/FileFormats/MOF/mof.1.2.schema.json
```diff
@@ -40,9 +40,9 @@
...
+                        "description": "An initial variable value that the optimizer may use to warm-start the solution process.",
+                        "type": ["number", "null"]
```
odow, line 45: Please leave as `number`. If it is `null`, we should just omit the field.

src/FileFormats/MOF/mof.1.2.schema.json
```diff
@@ -82,6 +82,10 @@
                 "properties": {
                     "name": {
                         "type": "string"
+                    },
+                    "ConstraintDualStart": {
```
odow, line 86: I guess we went with `primal_start`, so this should be `dual_start`.

src/FileFormats/MOF/mof.1.2.schema.json
```diff
@@ -82,6 +82,10 @@
...
+                        "description": "An initial value for constraint duals that the optimizer may use to warm-start the solution process.",
+                        "type": ["number","null"]
```
odow, line 88: Okay, so this section actually has to go below in the `oneOf` section, because it needs to be a `number` for scalar-valued constraints, and an `array` for vector-valued constraints.

src/FileFormats/MOF/read.jl
```diff
@@ -75,6 +75,7 @@ function read_variables(model::Model, object::Object)
     name_map = Dict{String,MOI.VariableIndex}()
     for variable::typeof(object) in object["variables"]
         name = get(variable, "name", "")::String
+        primal_start = get(variable, "VariablePrimalStart", nothing)
```
odow, line 78:
```suggestion
        primal_start = get(variable, "primal_start", nothing)
```

src/FileFormats/MOF/read.jl
```diff
@@ -83,6 +84,7 @@ function read_variables(model::Model, object::Object)
         end
         name_map[name] = index = MOI.add_variable(model)
         MOI.set(model, MOI.VariableName(), index, name)
+        MOI.set(model, MOI.VariablePrimalStart(), index, primal_start)
```
odow, line 87:
```suggestion
        if primal_start !== nothing
            MOI.set(model, MOI.VariablePrimalStart(), index, primal_start)
        end
```

odow (commented):
On the right track. But we need scalar-valued dual starts for scalar-valued constraints and vector-valued dual starts for vector-valued constraints.

LukasBarner:
saw https://github.com/jump-dev/MathOptFormat/pull/22 a bit too late, now constraint primal starts are also included. Also wrote tests for this.

src/FileFormats/MOF/mof.1.2.schema.json
```diff
@@ -92,6 +92,13 @@
                         },
                         "set": {
                             "$ref": "#/definitions/scalar_sets"
+                        },
+                        "primal_start": {
+                            "type": "number"
+                        },
+                        "dual_start": {
+                            "description": "An initial scalar value for constraint duals that the optimizer may use to warm-start the solution process.",
```
odow, line 100: If we're going to add descriptions, we should add them also to `primal_start`, and we should update them in https://github.com/jump-dev/MathOptFormat/pull/22 as well.

src/FileFormats/MOF/mof.1.2.schema.json
```diff
@@ -41,7 +41,7 @@
                         "type": "string"
                     },
                     "primal_start": {
-                        "description": "An initial value that the optimizer may use to warm-start the solution process.",
+                        "description": "An initial variable value that the optimizer may use to warm-start the solution process.",
```
odow, line 44:
```suggestion
                        "description": "An initial value for the primal variable solution that the optimizer may use to warm-start the solution process.",
```

src/FileFormats/MOF/mof.1.2.schema.json
```diff
@@ -92,6 +92,14 @@
                         },
                         "set": {
                             "$ref": "#/definitions/scalar_sets"
+                        },
+                        "primal_start": {
+                            "description": "An initial scalar value for the constraint primal start that the optimizer may use to warm-start the solution process.",
```
odow, line 97:
```suggestion
                            "description": "An initial scalar value for the primal constraint solution that the optimizer may use to warm-start the solution process.",
```

src/FileFormats/MOF/mof.1.2.schema.json
```diff
@@ -92,6 +92,14 @@
...
+                            "type": "number"
+                        },
+                        "dual_start": {
+                            "description": "An initial scalar value for the constraint dual that the optimizer may use to warm-start the solution process.",
```
odow, line 101:
```suggestion
                            "description": "An initial scalar value for the constraint's dual solution that the optimizer may use to warm-start the solution process.",
```

src/FileFormats/MOF/mof.1.2.schema.json
```diff
@@ -102,6 +110,20 @@
                         },
                         "set": {
                             "$ref": "#/definitions/vector_sets"
+                        },
+                        "primal_start": {
+                            "description": "An initial vector for a constraints primal starts that the optimizer may use to warm-start the solution process.",
```
odow, line 115:
```suggestion
                            "description": "An initial vector for the constraint's primal solution that the optimizer may use to warm-start the solution process.",
```

src/FileFormats/MOF/mof.1.2.schema.json
```diff
@@ -102,6 +110,20 @@
...
+                            "type": "array",
+                            "items": {
+                                "type": "number"
+                            }
+                        },
+                        "dual_start": {
+                            "description": "An initial vector for duals of a constraint that the optimizer may use to warm-start the solution process.",
```
odow, line 122:
```suggestion
                            "description": "An initial vector for the constraint's dual solution that the optimizer may use to warm-start the solution process.",
```

odow:
> Also out of curiosity, is there an application case for primal starts to vector-valued constraints?

Some conic solvers like SCS formulate `b - Ax in K` into `Ax + Is = b`, `s in K`. So the `ConstraintPrimalStart` is used as a primal start for the vector `s`. In theory, the solver could compute `b - Ax` as the start using the `VariablePrimalStart`. But there might be a small difference between `s` and `b - Ax`.

odow (approved):

src/FileFormats/MOF/mof.1.2.schema.json
```diff
@@ -11,7 +11,7 @@
             "required": ["minor", "major"],
             "properties": {
                 "minor": {
-                    "const": 1
+                    "const": 2
```
odow, line 14:
```suggestion
                    "enum": [0, 1, 2]
```

odow:
Thanks!

LukasBarner:
Happy to, thanks a lot also for your guidance 😄

PR #2059, opened 2022-12-08
Add ScalarNonlinearFunction support

odow:
## Overview

This PR is an implementation of the idea in https://github.com/jump-dev/MathOptInterface.jl/issues/846, and is based on our work at the JuMP-level: https://github.com/jump-dev/JuMP.jl/pull/3106.

## Preview links

There are two key bits of documentation to read. The actual implementation is very straight forward:

 * https://jump.dev/MathOptInterface.jl/previews/PR2059/reference/standard_form/#MathOptInterface.ScalarNonlinearFunction

One complicating issue is how to deal with user-defined function. I've added a new attribute:

 * https://jump.dev/MathOptInterface.jl/previews/PR2059/reference/models/#MathOptInterface.UserDefinedFunction

I've also added `ListOfSupportedNonlinearOperators <: AbstractModelAttribute` and `UnsupportedNonlinearOperator <: UnsupportedError`
 * https://jump.dev/MathOptInterface.jl/previews/PR2059/reference/models/#MathOptInterface.ListOfSupportedNonlinearOperators
 * https://jump.dev/MathOptInterface.jl/previews/PR2059/reference/errors/#MathOptInterface.UnsupportedNonlinearOperator
 
## Solver support

Ipopt is adds support for `ScalarNonlinearFunction` in https://github.com/jump-dev/Ipopt.jl/pull/346.

## What to do with NLPBlock

There's an open question about what we do with the existing NLPBlock stuff. The easiest answer is just to leave it, and say that solvers should support both, although they can error if both appear in the same model.

It'll be a harder decision at the JuMP-level, but we can wait until this is merged to figure that out.

## Blockers

There's nothing major blocking this from my end. It's actually pretty lightweight. The harder part is JuMP and documenting the two different nonlinear approaches. But this can be added without breaking anything because it is strictly a feature addition.

Needs:

 - [x] Tests to improve code coverage


odow:
Here's a current demo with the Ipopt PR:

```julia
julia> import Ipopt

julia> const MOI = Ipopt.MOI
MathOptInterface

julia> function main()
           model = Ipopt.Optimizer()
           x = MOI.add_variable(model)
           con_f = MOI.ScalarNonlinearFunction{Float64}(:^, Any[x, 2])
           MOI.add_constraint(model, con_f, MOI.LessThan(1.0))
           MOI.set(model, MOI.ObjectiveSense(), MOI.MIN_SENSE)
           my_f(x) = (x - 0.5)^2
           MOI.set(model, MOI.UserDefinedFunction(:my_f, 1), (my_f,))
           obj_f = MOI.ScalarNonlinearFunction{Float64}(:my_f, Any[x])
           MOI.set(model, MOI.ObjectiveFunction{typeof(obj_f)}(), obj_f)
           MOI.optimize!(model)
           return MOI.get(model, MOI.VariablePrimal(), x)
       end

julia> main()
This is Ipopt version 3.14.4, running with linear solver MUMPS 5.4.1.

Number of nonzeros in equality constraint Jacobian...:        0
Number of nonzeros in inequality constraint Jacobian.:        1
Number of nonzeros in Lagrangian Hessian.............:        2

Total number of variables............................:        1
                     variables with only lower bounds:        0
                variables with lower and upper bounds:        0
                     variables with only upper bounds:        0
Total number of equality constraints.................:        0
Total number of inequality constraints...............:        1
        inequality constraints with only lower bounds:        0
   inequality constraints with lower and upper bounds:        0
        inequality constraints with only upper bounds:        1

iter    objective    inf_pr   inf_du lg(mu)  ||d||  lg(rg) alpha_du alpha_pr  ls
   0  2.5000000e-01 0.00e+00 1.00e+00  -1.0 0.00e+00    -  0.00e+00 0.00e+00   0
   1  6.2500000e-02 0.00e+00 4.90e-01  -1.7 2.50e-01    -  1.00e+00 1.00e+00f  1
   2  1.1441994e-04 0.00e+00 1.74e-03  -1.7 2.39e-01    -  1.00e+00 1.00e+00h  1
   3  7.8304570e-06 0.00e+00 2.89e-04  -2.5 6.50e-02    -  1.00e+00 1.00e+00h  1
   4  1.5182438e-08 0.00e+00 2.74e-05  -3.8 2.72e-03    -  1.00e+00 1.00e+00h  1
   5  1.6255746e-12 0.00e+00 5.28e-08  -5.7 1.29e-04    -  1.00e+00 1.00e+00h  1
   6  2.8086636e-18 0.00e+00 6.35e-12  -8.6 1.29e-06    -  1.00e+00 1.00e+00h  1

Number of Iterations....: 6

                                   (scaled)                 (unscaled)
Objective...............:   2.8086636097379302e-18    2.8086636097379302e-18
Dual infeasibility......:   6.3507046478545526e-12    6.3507046478545526e-12
Constraint violation....:   0.0000000000000000e+00    0.0000000000000000e+00
Variable bound violation:   0.0000000000000000e+00    0.0000000000000000e+00
Complementarity.........:   2.5091160166026672e-09    2.5091160166026672e-09
Overall NLP error.......:   2.5091160166026672e-09    2.5091160166026672e-09


Number of objective function evaluations             = 7
Number of objective gradient evaluations             = 7
Number of equality constraint evaluations            = 0
Number of inequality constraint evaluations          = 7
Number of equality constraint Jacobian evaluations   = 0
Number of inequality constraint Jacobian evaluations = 7
Number of Lagrangian Hessian evaluations             = 6
Total seconds in IPOPT                               = 0.045

EXIT: Optimal Solution Found.
0.4999999983240932
```

Things are plumbed end-to-end, including user-defined functions.

I'm still undecided on how to implement nonlinear parameters and expressions.

src/attributes.jl
```diff
@@ -1856,6 +1856,36 @@ function attribute_value_type(::ConstraintConflictStatus)
...
+    return
+end
+MOI.set(model, MOI.UserDefinedFunction(:f, 2), (f,))
+MOI.set(model, MOI.UserDefinedFunction(:f, 2), (f, ∇f))
+MOI.set(model, MOI.UserDefinedFunction(:f, 2), (f, ∇f, ∇²f))
+```
+"""
+struct UserDefinedFunction <: AbstractModelAttribute
+    name::Symbol
+    N::Int
```
dourouc05, line 1886: The name is a bit odd, when all the other variable names are lower case. `arity` or `n`?

dourouc05 (commented):
Only minor comments in the code. I'm wondering what the API would look like to rewrite such functions using bridges, for instance for CP. (Or this issue has been discussed elsewhere and I'm not aware of it.)

src/Nonlinear/parse.jl
```diff
@@ -36,6 +36,44 @@ function parse_expression(::Model, ::Expression, x::Any, ::Int)
...
+    node_type = if length(x.args) == 1 && id !== nothing
+        NODE_CALL_UNIVARIATE
+    else
+        id = get(data.operators.multivariate_operator_to_id, x.head, nothing)
+        @assert id !== nothing
+        NODE_CALL_MULTIVARIATE
+    end
+    push!(expr.nodes, Node(node_type, id, parent))
+    parent = length(expr.nodes)
+    # Args need to be pushed onto the stack in reverse
```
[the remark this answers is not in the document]
odow: Updated the comment. The stack is first-in last-out.

src/attributes.jl
odow: Changed to `arity`

odow:
> I'm wondering what the API would look like to rewrite such functions using bridges, for instance for CP.

I think, instead of a bridge, it'd likely be a transformation layer, like `Dualization.Optimizer`.

We don't really have tools for bridging based on the content of a constraint, only on the type, so you'd likely need to be able to see the entire problem in order to construct a valid transformation.

blegat:
> We don't really have tools for bridging based on the content of a constraint, only on the type, so you'd likely need to be able to see the entire problem in order to construct a valid transformation.

I don't think it's about content vs type. If a transformation can be done independently for each constraint then it can be done by a bridge.

For the non-breaking path, I don't need we need to be non-breaking for the solvers. Once we have the NL API set up, we can drop support for NLPBlock in Ipopt and tag Ipopt v2.0. What we need is to continue supporting solvers requiring NLPBlock from MOI until MOI v2.0.
One way we could do that is to have an MOI layer that transforms ScalarNonlinearFunction into NLPBlock.
In `MOI.instantiate`, if `supports(::Optimizer, ::NLPBlock)` is `true` then we add this layer.

I still think transforming from NL expression to gradient/hessian oracle dealing with AD, `UserDefinedFunction`, etc... is best done by `LazyBridgeOptimizer` instead of each NL solver as detailed in https://github.com/jump-dev/MathOptInterface.jl/issues/846. Do you still agree with this long term goal ?

odow:
> Do you still agree with this long term goal ?

No. Computing derivatives really needs the entirety of the problem in one go, whereas the bridges work on a constraint-by-constraint basis. Dealing with the derivatives is not complicated (look how small the Ipopt PR is https://github.com/jump-dev/Ipopt.jl/pull/346) because we've abstracted all of the helpers into `MOI.Nonlinear.Model`.

blegat:
> Computing derivatives really needs the entirety of the problem in one go, whereas the bridges work on a constraint-by-constraint basis.

Why computing derivatives cannot be done on a constraint-by-constraint basis ?

odow:
> Why computing derivatives cannot be done on a constraint-by-constraint basis ?

Because that's a muuuuuch larger change to our AD engines. It also makes it a lot more complicated to deal with sparsity etc. We don't really want to have to deal with 10,000 constraints, each with 10,000 instances of an AD tape. JuMP's reverse diff amortizes a lot of allocations over the constraints. It also makes it hard to have shared common subexpressions.

mlubin:
Shared sub-expressions is a good reason. Sparsity of the hessian of the lagrangian is another one, although this could probably be refactored if we really wanted to.

blegat:
Could an MOI layer that does the transformation from `MOI.ScalarNonlinearFunction` to `MOI.NLPBlock` be a good first step ? I would be needed for retro-compatibility anyway and we can see if it can produce oracles as `MOI.AbstractScalarFunction` later and if it can be done as bridged even later

frapac:
> we can see if it can produce oracles as MOI.AbstractScalarFunction later and if it can be done as bridged even later

To me, the question is: what would be the use case for this? 

Usually, the nonlinear solver requires only access to the full Jacobian and to the full Hessian, discarding the individual constraints. Seeing all the constraints as a single multivalued function has also its advantages, such as: 
- we can evaluate the full Jacobian `J` efficiently using forward-mode AD to evaluate `p` vector products `J v`, with `p` a number of colors associated to the sparsity pattern of the Jacobian.  If we look at the individual constraints, evaluating the full Jacobian would require `m` evaluations of reverse-mode AD to evaluate the gradients of the `m` constraints one by one. 
- we can evaluate the Hessian of the Lagrangian efficiently using forward-over-reverse AD, using the same coloring trick to compress the  number of Hessian-vector product required.

odow:
We should check and confirm support for complex-valued nonlinear functions (polynomials of complex variables)

odow:
Here's a proof of concept of PATHSolver solving a nonlinear complementarity problem with `VectorNonlinearFunction-in-Complements`: https://github.com/chkwon/PATHSolver.jl/pull/82

Still an open question as to what the proper definition is. We can also revert the commit and start it in a separate PR if we decide to merge ScalarNonlinearFunction first.

src/functions.jl
```diff
@@ -249,6 +249,46 @@ struct ScalarNonlinearFunction{T} <: AbstractScalarFunction
...
 
+"""
+    VectorNonlinearFunction{T<:Number}(args::Vector{Any})
+
+The vector-valued nonlinear function composed of a vector of scalar functions.
+
+## `args`
+
+The vector `args` contains the scalar elements of the nonlinear function. Each
+element must be one of the folowing:
```
chriscoey, line 260:
```suggestion
element must be one of the following:
```

chriscoey:
> Still an open question as to what the proper definition is.

For VNF here you mean? This seems like the obvious approach to me - what alternatives are there?

odow:
> what alternatives are there

Some sort of vector-valued expression tree? Not sure. You could imagine `A * x` being `VectorNonlinearExpr(:*, Any[A, x])`.

odow:
#2094 is a resolution to the "how to we deal with parameters" question. Ipopt would declare support for `MOI.Parameter`, and then it could add parameters to the `MOI.Nonlinear.Model` object, and replace any parameter variables in the expression with the `Nonlinear.ParameterIndex`. It doesn't resolve the nonlinear subexpression question, but perhaps that's for the solver or `MOI.Nonlinear.Model` to deal with.

odow:
`Parameter` is added, so I'll have a go at updating this PR to use it.

odow:
Updated to use parameters here: https://github.com/jump-dev/Ipopt.jl/pull/346.

There are a bunch of issues with VectorNonlinearFunction and the bridges, so I think I'm going to remove VectorNonlinearFunction from this PR; we can re-add it at a later date.

odow:
At the very least, https://github.com/chkwon/PATHSolver.jl/pull/82 proves that VectorNonlinearFunction is feasible with the current stack.

odow:
The other reason for `VectorNonlinearFunction`: nonlinear multi-objective problems https://discourse.julialang.org/t/jump-non-linear-optimization/94020/14?u=odow

blegat:
I am in favor of VectorNonlinearFunction, I want to experiment with writing an MOI meta-solver for sequential nonlinear conic (as in https://arxiv.org/abs/2210.02142). I'll give it a shot with this PR next week

odow:
> I'll give it a shot with this PR next week

Note that I've removed `VectorNonlinearFunction` for now to simplify this PR.

odow:
@blegat this is ready for a thorough review. Can we hold off `VectorNonlinearFunction` until this is merged?

src/functions.jl
```diff
@@ -278,6 +278,67 @@ function Base.copy(f::ScalarQuadraticFunction)
...
+ * [`Nonlinear.DEFAULT_MULTIVARIATE_OPERATORS`](@ref)
+
+Additional operators can be registered by setting a [`UserDefinedFunction`](@ref)
+attribute.
+
+## `args`
+
+The vector `args` contains the arguments to the nonlinear function. If the
+operator is univariate, it must contain one element. Otherwise, it may contain
+multiple elements. Each element must be one of the folowing:
```
chriscoey, line 306: I see the allowed elements are only scalar types. I'm wondering about cases like `+` `norm` where the input could be a very long list of scalars - will that be performant? If not would allowing vector type (VV/VAF etc) inputs be OK down the track? I'm also wondering about a case like `f(vec1, vec2)` where you are dealing with a function with two vector arguments (e.g. relative entropy) - I'm not sure how to handle that if we only allow `head(args...)` where args are scalars.

src/functions.jl
```diff
@@ -794,6 +871,50 @@ function Base.convert(
     )
 end
 
+# ScalarNonlinearFunction
+
+function Base.convert(
```
chriscoey, line 876: Just for my own future reference, how do you think we can allow the user to convert in the opposite directions, e.g. SNF to SAF (which can potentially fail of course). I imagine this would be similar for e.g. VAF to VV, SQF to SAF. Could we implement these as `convert`s with errors, or do we need another interface like bridges?

src/attributes.jl
```diff
@@ -1856,6 +1856,86 @@ function attribute_value_type(::ConstraintConflictStatus)
...
+
+Set a user-defined function by the name of `name` with `arity` arguments.
+
+The value to be set is a tuple containg one to three functions, representing the
+zero, first, and second-order derivatives of the function.
+
+For uni-variate functions with `arity == 1`, the functions must have the form:
+
+ * `f(x::T)::T`: returns the evaluation of the function at `x`
+ * `∇f(x::T)::T`: return the first-order deriative of `f` with respect to `x`
```
chriscoey, line 1870:
```suggestion
 * `∇f(x::T)::T`: return the first-order derivative of `f` with respect to `x`
```

src/attributes.jl
```diff
@@ -1856,6 +1856,86 @@ function attribute_value_type(::ConstraintConflictStatus)
...
+ * `∇²f(x::T)::T`: returns the second-order derviative of `f` with respect to `x`.
```
chriscoey, line 1871:
```suggestion
 * `∇²f(x::T)::T`: returns the second-order derivative of `f` with respect to `x`.
```

src/attributes.jl
```diff
@@ -1856,6 +1856,86 @@ function attribute_value_type(::ConstraintConflictStatus)
     return ConflictParticipationStatusCode
 end
 
+"""
+    UserDefinedFunction(name::Symbol, arity::Int) <: AbstractModelAttribute
+
+Set a user-defined function by the name of `name` with `arity` arguments.
+
+The value to be set is a tuple containg one to three functions, representing the
```
chriscoey, line 1864:
```suggestion
The value to be set is a tuple containing one to three functions, representing the
```

src/functions.jl
odow: The current AD system doesn't support vector-valued arguments. We could potentially relax this restriction in future.

src/attributes.jl
```diff
@@ -1856,6 +1856,86 @@ function attribute_value_type(::ConstraintConflictStatus)
...
+zero, first, and second-order derivatives of the function.
```
chriscoey, line 1865: zeroth? or evaluation and first two derivatives? or just remove the fragment after "representing the ..." because it is redundant/repetitive with what's immediately below.

src/attributes.jl
```diff
@@ -1856,6 +1856,86 @@ function attribute_value_type(::ConstraintConflictStatus)
...
+
+For uni-variate functions with `arity == 1`, the functions must have the form:
+
+ * `f(x::T)::T`: returns the evaluation of the function at `x`
```
chriscoey, line 1869:
```suggestion
 * `f(x::T)::T`: returns the value of the function at `x`
```
not a strong opinion though

src/functions.jl
odow: You can add `convert` and throw an `InexactError`. MOI does this in places.

```julia
julia> convert(Int, 1.5)
ERROR: InexactError: Int64(1.5)
Stacktrace:
 [1] Int64
   @ ./float.jl:723 [inlined]
 [2] convert(#unused#::Type{Int64}, x::Float64)
   @ Base ./number.jl:7
 [3] top-level scope
   @ REPL[1]:1

help?> InexactError
search: InexactError

  InexactError(name::Symbol, T, val)

  Cannot exactly convert val to type T in a method of function name.

  Examples
  ≡≡≡≡≡≡≡≡≡≡

  julia> convert(Float64, 1+2im)
  ERROR: InexactError: Float64(1 + 2im)
  Stacktrace:
  [...]
```

src/attributes.jl
```diff
@@ -1856,6 +1856,86 @@ function attribute_value_type(::ConstraintConflictStatus)
...
 
+"""
+    UserDefinedFunction(name::Symbol, arity::Int) <: AbstractModelAttribute
+
+Set a user-defined function by the name of `name` with `arity` arguments.
+
+The value to be set is a tuple containg one to three functions, representing the
+zero, first, and second-order derivatives of the function.
+
+For uni-variate functions with `arity == 1`, the functions must have the form:
```
chriscoey, line 1867:
```suggestion
For uni-variate functions with `arity == 1`, the functions in the tuple must have the form:
```

src/attributes.jl
```diff
@@ -1856,6 +1856,86 @@ function attribute_value_type(::ConstraintConflictStatus)
...
+
+ * `f(x::T)::T`: returns the evaluation of the function at `x`
+ * `∇f(x::T)::T`: return the first-order deriative of `f` with respect to `x`
+ * `∇²f(x::T)::T`: returns the second-order derviative of `f` with respect to `x`.
+
+For multi-variate functions, the functions must have the form:
```
chriscoey, line 1873:
```suggestion
For multi-variate functions, the functions in the tuple must have the form:
```

chriscoey: Can we remove the hyphen in "multi-variate" everywhere? I see in e.g. https://jump.dev/MathOptInterface.jl/dev/submodules/Nonlinear/overview/#Multivariate-operators that the hyphen is not used

src/attributes.jl
```diff
@@ -1856,6 +1856,86 @@ function attribute_value_type(::ConstraintConflictStatus)
...
+
+ * `f(x::T...)::T`: returns the evaluation of the function at `x`
```
chriscoey, line 1875:
```suggestion
 * `f(x::T...)::T`: returns the value of the function at `x`
```

src/functions.jl
```diff
@@ -278,6 +278,67 @@ function Base.copy(f::ScalarQuadraticFunction)
...
 
+"""
+    ScalarNonlinearFunction{T<:Number}(head::Symbol, args::Vector{Any})
+
+The scalar-valued nonlinear function `head(args...)`, represented as a symbolic
+expression tree, with the call operator `head` and ordered arguments in `args`.
+
+## `head`
+
+The `head::Symbol` must be a scalar operator recognized by the model.
```
chriscoey, line 289: "recognized by the model" feels a little unclear to me, but maybe it's fine

src/functions.jl
```diff
@@ -278,6 +278,67 @@ function Base.copy(f::ScalarQuadraticFunction)
...
+
+By default, the list of supported uni-variate operators is given by the list:
```
chriscoey, line 291:
```suggestion
By default, the list of supported univariate operators is given by the list:
```

src/functions.jl
```diff
@@ -278,6 +278,67 @@ function Base.copy(f::ScalarQuadraticFunction)
...
+
+ * [`Nonlinear.DEFAULT_UNIVARIATE_OPERATORS`](@ref)
+
+and the list of supported multi-variate operators is given by the list:
```
chriscoey, line 295:
```suggestion
and the list of supported multivariate operators is given by the list:
```

src/functions.jl
chriscoey: "list" was said twice. I feel the new sentence is also clearer.

src/functions.jl
chriscoey, line 306:
```suggestion
multiple elements. Each element must be one of the following:
```

chriscoey:
Thanks @odow, very cool. Please review request me on any related nonlinear PRs when they're ready.

src/functions.jl
odow, line 295:
```suggestion
and the default list of supported multivariate operators is given by:
```

src/functions.jl
chriscoey: Ok, I'm thinking about CSP situations (where AD does not apply), so please make it as easy as possible to relax this restriction in the near future (I should be able to contribute).

src/functions.jl
chriscoey: Ah ok, I'd be interested in seeing these convert methods added in future. I may be able to contribute.

chriscoey:
Sorry if I missed part of the discussion/code/tests, but I'm wondering what we do (and what's the best we can do) in the case of an SNF involving SAF/SQF arguments. For example, suppose we have `saf::SAF` and we construct an SNF `log(saf)`. It seems the most efficient thing to do would be to preserve `saf` as-is rather than convert to an SNF.

odow:
ScalarNonlinear functions can have `ScalarAffineFunction` as arguments: `ScalarNonlinearFunction{Float64}(:log, Any[saf])`.

src/functions.jl
odow: There's nothing stopping this other than convention (the args are `Any[]`, so they can include vectors). So this should work:
```julia
X = reshape(MOI.VariableIndex.(1:4), 2, 2)
f = ScalarNonlinearFunction(:log, Any[ScalarNonlinearFunction(:det, Any[X])])
```
The hard part is adding it into the AD system.

And I don't know what to do if the function returns a vector argument.

chriscoey: Nice. Is it worth adding some simple tests for cases like this?

chriscoey:
Thanks @odow. I didn't immediately see any tests for cases like `ScalarNonlinearFunction{Float64}(:log, Any[saf])`. So this doesn't cause any trouble for AD?

odow:
I need to add some more tests. Internally, `MOI.Nonlinear` expands out the affine term.

src/attributes.jl
```diff
@@ -1856,6 +1856,88 @@ function attribute_value_type(::ConstraintConflictStatus)
     return ConflictParticipationStatusCode
 end
 
+"""
+    UserDefinedFunction(name::Symbol, arity::Int) <: AbstractModelAttribute
+
+Set a user-defined function by the name of `name` with `arity` arguments.
+
+The value to be set is a tuple containing one, two, or three functions.
```
chriscoey, line 1864: For the case of a symbolic function recognized by a particular solver and for which we don't want to bother specifying evaluation (and maybe can't give derivatives), can we make things convenient at the MOI level? I'm looking [here in the Minizinc logical operators PR](https://github.com/jump-dev/MiniZinc.jl/pull/22/files#diff-3b9314a6f9f2d7eec1d0ef69fa76cfabafdbe6d0df923768f9ec32f27a249c63R1143) and thinking it would be cleaner to be able to specify just `MOI.Nonlinear.register_operator(nlp, :!, 1)`.

chriscoey:
> I need to add some more tests. Internally, `MOI.Nonlinear` expands out the affine term.

My point was that that seems a little inefficient.

src/attributes.jl
chriscoey: Maybe I'm confused here by the name and docstring... is the purpose of UserDefinedFunction simply to allow passing in eval+derivs for solvers that require these? i.e. is it optional to specify UserDefinedFunction, and not needed for solvers that recognize function symbols directly (like MiniZinc)? If so, then I feel like UserDefinedFunction should be renamed to something like UserDefinedFunctionEvaluators to be a lot clearer (or the interface for specifying arity, f, ∇f, ∇²f should be redesigned), and the docstring `Set a user-defined function by the name of `name` with `arity` arguments.` should be rewritten to make it clear when/why to set these.

chriscoey:
I've updated MiniZinc.jl to use this PR, over at https://github.com/jump-dev/MiniZinc.jl/pull/23

odow:
Chris suggests we add `ListOfSupportedNonlinearOperators` and `UnsupportedNonlinearOperatorError`

chriscoey:
Also it's convenient to be able to write e.g.
```MOI.Nonlinear.add_constraint(nlp, :(($a => !($b ⊻ $c))), MOI.EqualTo(1.0))```
but I don't think we can do this outside MOI.Nonlinear, to create a SNF. Could the Julia Expr parser stuff in MOI.Nonlinear be moved to Utilities so that we can create SNFs more easily?

**Comment by odow** [comment]
> Could the Julia Expr parser stuff in MOI.Nonlinear be moved to Utilities so that we can create SNFs more easily?

We could potentially do something that makes it easy. The function is pretty trivial to write though:

```julia
to_scalar_nonlinear_function(f) = f

function to_scalar_nonlinear_function(f::Expr)
    if Meta.isexpr(f, :call)
        return ScalarNonlinearFunction{Float64}(
            f.args[1], 
            Any[to_scalar_nonlinear_function(arg) for arg in f.args[2:end]],
        )
    elseif Meta.isexpr(f, :&&) || Meta.isexpr(f, :||)
        return ScalarNonlinearFunction{Float64}(
            f.head, 
            Any[to_scalar_nonlinear_function(arg) for arg in f.args],
        )
    else
        error("Unsupported expression $f")
    end
end
```

odow:
Let me flesh it out a little. Give me two ticks

odow:
```julia

julia> scalar_nonlinear_function(f) = f
scalar_nonlinear_function (generic function with 1 method)

julia> function scalar_nonlinear_function(::Val{F}, args::Vector{Any}) where {F}
           return MOI.ScalarNonlinearFunction{Float64}(
               F, 
               scalar_nonlinear_function.(args),
           )
       end
scalar_nonlinear_function (generic function with 2 methods)

julia> function scalar_nonlinear_function(f::Expr)
           if Meta.isexpr(f, :call)
               return scalar_nonlinear_function(Val(f.args[1]), f.args[2:end])
           elseif Meta.isexpr(f, :&&) || Meta.isexpr(f, :||)
               return scalar_nonlinear_function(Val(f.head), f.args)
           else
               error("Unsupported expression $f")
           end
       end
scalar_nonlinear_function (generic function with 3 methods)

julia> X = [x, x]
2-element Vector{MathOptInterface.VariableIndex}:

julia> scalar_nonlinear_function(:(log($x)))
MathOptInterface.ScalarNonlinearFunction{Float64}(:log, Any[MathOptInterface.VariableIndex(1)])

julia> scalar_nonlinear_function(:((1 + log($x))^2))
MathOptInterface.ScalarNonlinearFunction{Float64}(:^, Any[MathOptInterface.ScalarNonlinearFunction{Float64}(:+, Any[1, MathOptInterface.ScalarNonlinearFunction{Float64}(:log, Any[MathOptInterface.VariableIndex(1)])]), 2])

julia> scalar_nonlinear_function(:(norm($X)))
MathOptInterface.ScalarNonlinearFunction{Float64}(:norm, Any[MathOptInterface.VariableIndex[MathOptInterface.VariableIndex(1), MathOptInterface.VariableIndex(1)]])

julia> function scalar_nonlinear_function(::Val{:norm}, args::Vector{Any})
           args2 = [MOI.ScalarNonlinearFunction{Float64}(:^, Any[a, 2]) for a in args]
           return MOI.ScalarNonlinearFunction{Float64}(
               :sqrt,
               Any[MOI.ScalarNonlinearFunction{Float64}(:+, args2)],
           )
       end
scalar_nonlinear_function (generic function with 4 methods)

julia> scalar_nonlinear_function(:(norm($X)))
MathOptInterface.ScalarNonlinearFunction{Float64}(:sqrt, Any[MathOptInterface.ScalarNonlinearFunction{Float64}(:+, Any[MathOptInterface.ScalarNonlinearFunction{Float64}(:^, Any[MathOptInterface.VariableIndex[MathOptInterface.VariableIndex(1), MathOptInterface.VariableIndex(1)], 2])])])
```

chriscoey:
Nice! Should be able to pass in the number type parameter for SNF I guess too. For CSP this would often be Int. 

I forgot to ask today... For SAT models we want to work with Bool rather than Int - would that work with this PR? I ask because I think there was a JuMP discussion about letting a model be parametrized by a Bool type and things seemed a bit unclear there.

chriscoey:
Having to specify the numeric type of each SNF is a bit annoying actually...

What should the numeric types be in the case of e.g. SNF(:<, [SNF(:log, [x]), 1]). The user probably wants to set the type of the inner SNF (log) to be Float64, and the type of the outer SNF to be Bool, like SNF{Bool}(:<, [SNF{Float64}(:log, [x]), 1]). IDK just seems like that type info is of limited utility.

blegat:
> Chris suggests we add `ListOfSupportedNonlinearOperators` and `UnsupportedNonlinearOperatorError`

Why not `supports_nonlinear_operator` similar to `supports` and `supports_constraint` ?

> For SAT models we want to work with Bool rather than Int - would that work with this PR? I ask because I think there was a JuMP discussion about letting a model be parametrized by a Bool type and things seemed a bit unclear there. 

With https://github.com/jump-dev/JuMP.jl/pull/3191, for SAT you would set `Bool` as the type of variable values in JuMP. The coefficients themself can then either be `Bool` or `Complex{Bool}`. Why would you need to use a `log` in a SAT ?

chriscoey:
> With [jump-dev/JuMP.jl#3191](https://github.com/jump-dev/JuMP.jl/pull/3191), for SAT you would set `Bool` as the type of variable values in JuMP. The coefficients themself can then either be `Bool` or `Complex{Bool}`. 

OK thanks, good to know that JuMP and MOI will allow parametrizing models by a "Bool" type (e.g. for SAT).

> Why would you need to use a `log` in a SAT ?

Not SAT, but CSP more generally. I guess my question is: what is the meaning of the numeric type parameter in SNF? Is it simply the same type as the model/variables? Is the argument/result types of the symbolic operator irrelevant? And second question (maybe more annoying to answer...) is: why does SNF need to be parametrized by a type?

odow:
> why does SNF need to be parametrized by a type?

This is a fair question. Perhaps it doesn't. It depends whether we expect solvers to distinguish `ScalarNonlinearFunction{Bool}` from `ScalarNonlinearFunction{Float64}`. But perhaps that can be distinguished from the set type.

Currently, our only non-typed functions are VariableIndex and VectorOfVariables, so I added a type because it felt right. But if we're going abstract `Vector{Any}` then perhaps we don't need to.

odow:
> Why not supports_nonlinear_operator similar to supports and supports_constraint ?

How would someone discover what operators are supported? `supports_` works if you say "Is X supported?" but that is different to "What X are supported?"

chriscoey:
The purpose of coefficient types for affine/quadratic functions is pretty clear, but yeah the purpose and semantics of types is not clear to me for SNF. Sounds like it would be worth trying to remove the types in this PR and seeing if that causes any issues.

odow:
Last commit drops the type parameter.

src/attributes.jl
odow: Thinking on this more, I think the function evaluation is always required. Otherwise, how is the solver meant to know what function `!` refers to? With the introduction of `ListOfSupportedNonlinearOperators`, solvers can pre-define a list of operators they support and recognize.

src/attributes.jl
```diff
@@ -1859,9 +1859,18 @@ end
...
 
-The value to be set is a tuple containing one, two, or three functions.
+Once registered, `name` will appear in
+[`ListOfSupportedNonlinearOperators`](@ref).
+
+## Value type
+
+The value to be set is a tuple containing one, two, or three functions to
+evaluate the function, the first-order derivative, and the second-order
+derivative respectively. Both dervatives are optional, and if you pass the
```
chriscoey, line 1872:
```suggestion
derivative respectively. Both derivatives are optional, and if you pass the
```

src/attributes.jl
```diff
@@ -1859,9 +1859,18 @@ end
...
+second-order derivative you must also pass the first-order derivative.
 
 For uni-variate functions with `arity == 1`, the functions in the tuple must
```
chriscoey, line 1875:
```suggestion
For univariate functions with `arity == 1`, the functions in the tuple must
```

blegat:
> @blegat this is ready for a thorough review. Can we hold off `VectorNonlinearFunction` until this is merged?

Sounds good, I'll have some time next week

odow:
@chriscoey asks why not a design like:
```Julia
abstract type NonlinearOperator end

function num_arguments end

struct Addition <: NonlinearOperator end
num_arguments(::Addition) = nothing
(::Addition)(args...) = +(args...)

struct Subtraction <: NonlinearOperator end
num_arguments(::Subtraction) = 2
(::Subtraction)(x, y) = x - y

struct Multiplication <: NonlinearOperator end
num_arguments(::Multiplication) = nothing
(::Multiplication)(args...) = *(args...)

struct Division <: NonlinearOperator end
num_arguments(::Division) = 2
(::Division)(x, y) = x / y

struct Power <: NonlinearOperator end
num_arguments(::Power) = 2
(::Power)(x, y) = x ^ y

struct IfElse <: NonlinearOperator end
num_arguments(::IfElse) = 3
(::IfElse)(flag, x, y) = ifelse(flag, x, y)

struct ScalarNonlinearFunction
    head::NonlinearOperator
    args::Vector{Any}
end

evaluate(f::ScalarNonlinearFunction) = f.head(f.args...)

f = ScalarNonlinearFunction(Addition(), Any[1, 2])

evaluate(f)  # 3
```

This is reasonable, but it's heading more in the direction of Symbolics.jl. There's definitely a tension as to how much JuMP is a computer algebraic system for optimization.

We've largely stuck with the `Symbol` design for simplicity, because that's what JuMP has done in the past, and because that's how Julia represents expressions (and it's what we get from the macros).

This question was motivated by our discussion on the nonlinear call today, where MOI is one source of truth for what each Symbol means, but solvers can support additional ones. And there could easily be a clash if PkgA.jl decides that `:logsum` means the `sum(log10.(x))` and PkgB.jl decides `:logsum` means `sum(log.(x))`.

But perhaps all we need to do is enforce that packages introduce unique symbols for ambiguous meanings. So you'd have to write `@constraint(model, a_logsum(x...) <= 1)` and `@constraint(model, pkgb_logsum(x...) <= 1)`.

Even something like this should work:
```julia
struct PkgA.LogSum end
(::PkgA.MyLogSum)(args...) = MOI.ScalarNonlinearFunction(:PkgA_logsum, Any[args...])
@constraint(model, PkgA.LogSum(x...) <= 1)
```

odow:
```Julia
struct Power <: NonlinearOperator
    p::Int
end
num_arguments(::Power) = 1
(f::Power)(x) = x ^ f.p
(f::Power)(x) = MOI.ScalarNonlinearFunction(:^, Any[x, f.p])

@constraint(model, Power(2)(x) <= 1)
square = Power(2)
@constraint(model, square(x) <= 1)
```

blegat:
From a first look, it looks quite nice, I guess the main point of argument is whether we should parametrize it with the type `T` and whether we should convert numbers to `T` when we add them to the expression.
For instance, what does `MOI.Utilities.operate(+, Float64, 3, nl_expr)` does ? Does it add `3` or `3.0` ?
We do the conversion for scalar and quadratic because we want a concretely-typed list of terms but for nonlinear expressions we don't so it's less obvious.
Note that since the type `T` is always passed as second argument to `operate`, we don't have to add `T` as type parameter if we want to do the conversion.

odow:
> I guess the main point of argument is whether we should parametrize it with the type T and whether we should convert numbers to T when we add them to the expression.

I had that originally. But I talked to @chriscoey a lot and we decided it was redundant and we could drop. It just made things more complicated.

https://github.com/jump-dev/MathOptInterface.jl/pull/2059/commits/93b289699b97f0fdee0d1a5cbde714219bd1205c

blegat:
Then I guess the question about `MOI.Utilities.operate(+, Float64, 3, nl_expr)` does not really apply then since `MOI.operate` with `T` only expects numbers of type `T` as input the it's not a valid call.

src/Nonlinear/parse.jl
```diff
@@ -36,6 +36,58 @@ function parse_expression(::Model, ::Expression, x::Any, ::Int)
     )
 end
 
+function parse_expression(
+    data::Model,
+    expr::Expression,
+    x::MOI.ScalarNonlinearFunction,
+    parent::Int,
```
mlubin, line 43: What's the parent argument? Please document.

src/Test/test_nonlinear.jl
```diff
@@ -1115,3 +1115,560 @@ function setup_test(
...
+    @requires T == Float64
+    @requires _supports(config, MOI.optimize!)
+    F = MOI.ScalarNonlinearFunction
+    @requires MOI.supports(model, MOI.ObjectiveFunction{F}())
+    @requires MOI.supports_constraint(model, F, MOI.GreaterThan{Float64})
+    MOI.Utilities.loadfromstring!(
+        model,
+        """
+variables: w, x, y, z
+minobjective: ScalarNonlinearFunction(w * z * (w + x + y) + y)
```
mlubin, line 1132: This is a big improvement for being able to test nlp solvers.

src/Utilities/functions.jl
```diff
@@ -946,10 +955,6 @@ end
     filter_variables(keep::Function, f::AbstractFunction)
 
 Return a new function `f` with the variable `vi` such that `!keep(vi)` removed.
-
-WARNING: Don't define `filter_variables(::Function, ...)` because Julia will
```
mlubin, line 950: When did this change?

src/attributes.jl
```diff
@@ -1856,6 +1856,108 @@ function attribute_value_type(::ConstraintConflictStatus)
     return ConflictParticipationStatusCode
 end
 
+"""
+    UserDefinedFunction(name::Symbol, arity::Int) <: AbstractModelAttribute
+
+Set this attribute to register a user-defined function by the name of `name`
```
mlubin, line 1862: Should this attribute be set or submitted?

src/attributes.jl
```diff
@@ -1856,6 +1856,108 @@ function attribute_value_type(::ConstraintConflictStatus)
...
+with `arity` arguments.
```
mlubin, line 1863: Are you allowed to have multiple `UserDefinedFunction`s with the same `name` but different `arity`?

test/Bridges/Constraint/interval.jl
```diff
@@ -456,6 +456,18 @@ function test_runtests()
         """,
         eltype = Rational{Int},
     )
+    MOI.Bridges.runtests(
```
mlubin, line 459: I don't think I've looked at `Bridges.runtests` recently. This is an amazingly clear test given the complexity of what's being tested.

src/functions.jl
```diff
@@ -278,6 +278,109 @@ function Base.copy(f::ScalarQuadraticFunction)
...
+See the full list of operators supported by a [`ModelLike`](@ref) by querying
+[`ListOfSupportedNonlinearOperators`](@ref).
+
+## `args`
+
+The vector `args` contains the arguments to the nonlinear function. If the
+operator is univariate, it must contain one element. Otherwise, it may contain
+multiple elements. Each element must be one of the following:
+
+ * A constant value of type `T<:Number`
```
mlubin, line 311: Should the element type of `args` be explicitly typed as a union of these types listed?

src/functions.jl
```diff
@@ -278,6 +278,109 @@ function Base.copy(f::ScalarQuadraticFunction)
...
+ * A [`ScalarAffineFunction`](@ref)
+ * A [`ScalarQuadraticFunction`](@ref)
+ * A [`ScalarNonlinearFunction`](@ref)
+
+## Unsupported operators
+
+If the optimizer does not support `head`, an [`UnsupportedNonlinearOperator`](@ref)
+error will be thrown.
+
+There is no guarantee about when this error will be thrown; it may be thrown
```
mlubin, line 322: This is kind of weird from an error handling perspective. Are there other similar cases? Should `optimize!` fail with an `INVALID_MODEL` termination status instead of throwing?

mlubin (commented):
I don't have any major concerns so far. Here are comments from a first pass.

src/Utilities/functions.jl
odow: It hasn't. I don't think this change was intentional. It's unrelated to Nonlinear and not present in `master`. I'll revert.

src/Test/test_nonlinear.jl
odow: Yeah. I was surprised how little modifications is required to the parsing code

test/Bridges/Constraint/interval.jl
odow: There was a period of weeks/months, where I felt like all I was doing was tidying up tests and refactoring things. Twas a long slog, but now it's trivial to write new bridges and I'm a lot more confident that the bridge code is correct.

src/functions.jl
odow: This can happen with `UnsupportedConstraint`, which depends whether there is a cache or whether `copy_to` or `add_constraint` is used.

I added `UnsupportedNonlinearOperator` because it's a pretty serious error that can't be recovered from. It's not like an invalid model that needs a starting point changed or at least one constraint.

src/functions.jl
odow: @chriscoey and I had a chat about this.

The main argument for `Any` is simplicity. The downside is that the user might pass a disallowed argument like a vector of variables. When olvers like Ipopt which use `MOI.Nonlinear` encounter such an object, they will throw
https://github.com/jump-dev/MathOptInterface.jl/blob/acd7a0e72e4b736354d9dbe3b6c641ca3a541b05/src/Nonlinear/parse.jl#L33-L37

The other benefit is that if we are permissible in what types we accept, then a solver like MiniZinc could support vector arguments and work fine. But that's be non-standard syntax.

We could use the Union, but because the union is quite wide, it adds no performance benefit, and it'd make things clunkier to construct. I guess we could add a validator to the constructor?

src/attributes.jl
odow: Not sure. I've just copied the existing data structures for now. But really, we need only know whether the function is unary so we know the gradient and hessian calls.

src/functions.jl
mlubin: The method error principle should push us to type the array so that errors happen sooner rather than later.

src/attributes.jl
odow: So I remembered why I made this a model attribute: we don't cache `AbstractSubmittable` in the `.model_cache`. That means, if you change the solver, we'd loose the user-defined function information and need to re-register it.

Currently, the only submittables are `LazyConstraint`, `HeuristicSolution` and `UserCut` which are set during a callback.

src/Nonlinear/parse.jl
odow: Documented to some extent in
https://github.com/jump-dev/MathOptInterface.jl/blob/978418089e5e24534f192e378bbd6c4b763707dd/src/Nonlinear/parse.jl#L18-L32

src/attributes.jl
odow: > Are you allowed to have multiple UserDefinedFunctions with the same name but different arity?

No. But you're potentially allowed a univariate and multivariate operator with the same name.

We use the arity to initialize some of the buffers in the AD system. But we could shift that to resize dynamically in each call site. It only gets used in a couple of places.

Then we could add `UserDefinedFunction(name::Symbol, is_univariate::Bool)`.

src/attributes.jl
```diff
@@ -1856,6 +1856,108 @@ function attribute_value_type(::ConstraintConflictStatus)
...
+For multivariate functions with `arity > 1`, the functions in the tuple must
+have the form:
+
+ * `f(x::T...)::T`: returns the value of the function at `x`
+ * `∇f(g::AbstractVector{T}, x::T...)::Nothing`: fills the components of `g`,
+   with `g[i]` being the first-order partial derivative of `f` with respect to
+   `x[i]`
+ * `∇²f(H::AbstractMatrix{T}, x::T...)::Nothing`: fills the components of `H`,
+    with `H[i, j]` being the second-order partial derivative of `f` with respect
+    to `x[i]` and then `x[j]`. It does not need to fill in the structural
```
blegat, line 1892: This is not clear, what is the difference between a structural and non-structural zero ? Can't we just say that `H` can be assumed to be a zero matrix at first so whatever is not set is just zero ?

test/Bridges/Constraint/slack.jl
```diff
@@ -324,6 +324,18 @@ function test_runtests()
...
     )
+    MOI.Bridges.runtests(
+        MOI.Bridges.Constraint.ScalarSlackBridge,
+        """
+        variables: x
+        ScalarNonlinearFunction(log(x)) <= 2.0
+        """,
+        """
+        variables: x, y
+        ScalarNonlinearFunction(log(x) - y) == 0.0
```
blegat, line 335: It's nice to see it working through the bridges with just a few `operate` methods implemented

odow: Yeah. The problem is that it's hard to know which `operate` methods are needed :smile:

src/Utilities/objective_container.jl
```diff
@@ -167,6 +174,17 @@ function MOI.set(
     return
 end
 
+function MOI.set(
+    o::ObjectiveContainer,
+    ::MOI.ObjectiveFunction{MOI.ScalarNonlinearFunction},
+    f::MOI.ScalarNonlinearFunction,
+)
+    _empty_keeping_sense(o)
```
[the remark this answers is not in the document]
odow: Yeah now there is no rule. It's just normal behavior.

src/attributes.jl
odow: A structural zero is an element that is always zero. A non-structural zero is an element that is zero right now, but may be non-zero at a different evaluation point.

But yeah, too complicated. Assumed to be zero works better :+1:

src/Utilities/copy.jl
```diff
@@ -23,7 +26,11 @@ function pass_attributes(
     src::MOI.ModelLike,
     index_map::IndexMap,
 )
-    for attr in MOI.get(src, MOI.ListOfModelAttributesSet())
+    attrs = MOI.get(src, MOI.ListOfModelAttributesSet())
+    # We need to deal with the UserDefinedFunctions first, so that they are in
+    # the model before we deal with the objective function or the constraints.
+    sort!(attrs; by = _is_user_defined_function, rev = true)
```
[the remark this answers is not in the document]
odow: Added a more general sort function. I don't know about tests forcing the order for `get(::Model, ::ListOfModelAttributesSet)`. It doesn't matter what order the get is, just when your setting. And for most people, they'll be writing a specific solver interface which will have a particular order things need to go in. It's only relevant for us writing MOI-level layers.

It's probably a separate discussion after/outside of this PR.

test/Bridges/Constraint/slack.jl
blegat: Follow the `MethodError`s :laughing:

blegat (approved):

src/attributes.jl
odow: We could treat that number as the maximum arity of a user-defined function. I think this should mostly just work. Then if you want to register `max(args...)`, you don't need to register `max2`, `max3`, `max4`, etc.

mlubin: I don't think the current behavior needs to change. I was asking more because the documentation wasn't clear.

mlubin (approved):
LGTM to merge.

In terms of breaking this in the future, I think we should do our best to stick to semver. So let's test this out and see what needs to change. Then MOI 2.0 can include any breaking changes to this API, plus deleting the old `NLPBlockData` interface?

odow:
> So let's test this out and see what needs to change. Then MOI 2.0 can include any breaking changes to this API, plus deleting the old NLPBlockData interface?

Sounds good. The actual interface is pretty minimal, so I think there are probably a bunch of missing methods and things that we'll need to add, rather than breaking changes.

PR #2060, opened 2022-12-09
[Utilities] add distance_to_set for SecondOrderCone

odow:
Part of https://github.com/jump-dev/MathOptInterface.jl/issues/2033. 

We'll need to decide what sets we use exact methods for, and what we make up.

Boyd's book is: https://web.stanford.edu/~boyd/papers/pdf/prox_algs.pdf

### DIFF
--- a/src/Utilities/distance_to_set.jl
+++ b/src/Utilities/distance_to_set.jl
@@ -204,35 +204,13 @@ function distance_to_set(
     # Parikh, N., & Boyd, S. (2014). Proximal algorithms. Foundations and
     # trends in Optimization, page 184, section 6.3.2.
     t, rhs = x[1], LinearAlgebra.norm(@views x[2:end])
+    if t >= rhs
+        return zero(T) # The point is feasible!
+    end
     if rhs <= -t  # Projection to the point (0, [0])
         return LinearAlgebra.norm(x)
-    elseif t >= rhs  # Projection to the point (t, x)
-        return zero(T)
     end
     # Projection to the point (t, x) + 0.5 * (|x|_2 - t, (t/|x|_2 - 1) * x)
-    c = t / rhs - one(T)
-    return sqrt((rhs - t)^2 + sum((c * x[i])^2 for i in 2:length(x))) / 2
-end
+    return sqrt(2) / 2 * abs(t - rhs)
 
-"""
-    distance_to_set(::ProjectionUpperBoundDistance, x, ::MOI.ExponentialCone)
-
-Given `x = (u, v, w)`, this is distance:
-
- * `d`, such that `v * exp(u / v) <= w + d`, if `v > 0`
- * `√(1^2 + d^2)`, such that `1 * exp(u / 1) <= w + d`, if `v <= 0`.
-"""
-function distance_to_set(
-    ::ProjectionUpperBoundDistance,
-    x::AbstractVector{T},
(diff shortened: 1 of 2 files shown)


tjdiamandis:
A great reference for this is https://docs.mosek.com/whitepapers/expcone-proj.pdf (Section 5 and Lemma 5.1, specifically).

odow:
> it is still ongoing research due to its numerical instability

We could just convert to `BigFloat`. Performance likely isn't a big concern.

odow:
I added a test for an `Inf` distance to the exponential cone. I don't know what else we should do.

src/Utilities/distance_to_set.jl
```diff
@@ -193,3 +193,46 @@ function distance_to_set(
...
+    # trends in Optimization, page 184, section 6.3.2.
+    t, rhs = x[1], LinearAlgebra.norm(@views x[2:end])
+    if rhs <= -t  # Projection to the point (0, [0])
+        return LinearAlgebra.norm(x)
+    elseif t >= rhs  # Projection to the point (t, x)
+        return zero(T)
+    end
+    # Projection to the point (t, x) + 0.5 * (|x|_2 - t, (t/|x|_2 - 1) * x)
+    c = t / rhs - one(T)
+    return sqrt((rhs - t)^2 + sum((c * x[i])^2 for i in 2:length(x))) / 2
```
blegat, line 214: This can be simplified to sqrt(2)/2 abs(t - rhs)

src/Utilities/distance_to_set.jl
```diff
@@ -193,3 +193,46 @@ function distance_to_set(
...
+    ::ProjectionUpperBoundDistance,
+    x::AbstractVector{T},
+    set::MOI.SecondOrderCone,
+) where {T<:Real}
+    _check_dimension(x, set)
+    # Projections come from:
+    # Parikh, N., & Boyd, S. (2014). Proximal algorithms. Foundations and
+    # trends in Optimization, page 184, section 6.3.2.
+    t, rhs = x[1], LinearAlgebra.norm(@views x[2:end])
+    if rhs <= -t  # Projection to the point (0, [0])
```
blegat, line 207: You mean (t, [0]) ?

src/Utilities/distance_to_set.jl
```diff
@@ -193,3 +193,46 @@ function distance_to_set(
...
+        return LinearAlgebra.norm(x)
```
blegat, line 208: Why don't we also do sqrt(2)/2 abs(t - rhs) ?

src/Utilities/distance_to_set.jl
```diff
@@ -193,3 +193,46 @@ function distance_to_set(
...
+    elseif t >= rhs  # Projection to the point (t, x)
```
blegat, line 209: I would simply do abs(t) >= rhs

odow: Why?

src/Utilities/distance_to_set.jl
odow: No, I think I mean `(0, [0])`? I just followed Section 6.3.2 here: https://web.stanford.edu/~boyd/papers/pdf/prox_algs.pdf

blegat (approved):

PR #2070, opened 2023-01-06
Add support for vector-valued objective functions

odow:
This is a re-hash of https://github.com/jump-dev/MathOptInterface.jl/pull/968

The last PR lost steam because of the discussion in https://github.com/jump-dev/JuMP.jl/issues/2099

My sense is that this is still the right API approach at the MOI-level, but that it's the wrong approach at the JuMP level. I'm working on a PR to JuMP: https://github.com/jump-dev/JuMP.jl/pull/3176, <s>so let's hold off comments until things are working fully and we can discuss the full API, rather than the pros and cons of treating multi-criteria as vector optimization.</s>

I have a proof-of-concept solver at https://github.com/odow/MOO.jl and a PR for Gurobi is https://github.com/jump-dev/Gurobi.jl/pull/497.

## Alternative approaches

Instead of having a vector-valued objective, we could instead add a vector of scalar objectives.

The start to PR implementing that is https://github.com/jump-dev/MathOptInterface.jl/pull/2086.

Closes #2086

### DIFF
--- a/docs/src/submodules/Bridges/list_of_bridges.md
+++ b/docs/src/submodules/Bridges/list_of_bridges.md
@@ -78,6 +78,7 @@ These bridges are subtypes of [`Bridges.Objective.AbstractBridge`](@ref).
 Bridges.Objective.FunctionizeBridge
 Bridges.Objective.QuadratizeBridge
 Bridges.Objective.SlackBridge
+Bridges.Objective.VectorSlackBridge
 ```
 
 ## [Variable bridges](@id variable_bridges_ref)
--- a/src/Bridges/Bridges.jl
+++ b/src/Bridges/Bridges.jl
@@ -341,7 +341,7 @@ function _general_bridge_tests(bridge::B) where {B<:AbstractBridge}
         MOI.get(bridge, MOI.NumberOfVariables())
     )
     if B <: Objective.AbstractBridge
-        Test.@test set_objective_function_type(B) <: MOI.AbstractScalarFunction
+        Test.@test set_objective_function_type(B) <: MOI.AbstractFunction
     end
     return
 end
--- a/src/Bridges/Constraint/single_bridge_optimizer.jl
+++ b/src/Bridges/Constraint/single_bridge_optimizer.jl
@@ -122,7 +122,7 @@ end
 
 function MOI.Bridges.is_bridged(
     ::SingleBridgeOptimizer,
-    ::Type{<:MOI.AbstractScalarFunction},
+    ::Type{<:MOI.AbstractFunction},
 )
     return false
 end
--- a/src/Bridges/Objective/Objective.jl
+++ b/src/Bridges/Objective/Objective.jl
@@ -17,6 +17,7 @@ include("single_bridge_optimizer.jl")
 include("bridges/functionize.jl")
 include("bridges/quadratize.jl")
 include("bridges/slack.jl")
+include("bridges/vector_slack.jl")
 
 """
     add_all_bridges(model, ::Type{T}) where {T}
@@ -29,6 +30,7 @@ function add_all_bridges(model, ::Type{T}) where {T}
     MOI.Bridges.add_bridge(model, FunctionizeBridge{T})
     MOI.Bridges.add_bridge(model, QuadratizeBridge{T})
     MOI.Bridges.add_bridge(model, SlackBridge{T})
+    MOI.Bridges.add_bridge(model, VectorSlackBridge{T})
     return
 end
 
--- a/src/Bridges/Objective/bridge.jl
+++ b/src/Bridges/Objective/bridge.jl
@@ -23,7 +23,7 @@ abstract type AbstractBridge <: MOI.Bridges.AbstractBridge end
 """
     supports_objective_function(
         BT::Type{<:MOI.Bridges.Objective.AbstractBridge},
-        F::Type{<:MOI.AbstractScalarFunction},
+        F::Type{<:MOI.AbstractFunction},
     )::Bool
 
 Return a `Bool` indicating whether the bridges of type `BT` support bridging
@@ -37,7 +37,7 @@ objective functions of type `F`.
 """
 function supports_objective_function(
     ::Type{<:AbstractBridge},
-    ::Type{<:MOI.AbstractScalarFunction},
+    ::Type{<:MOI.AbstractFunction},
 )
     return false
 end
@@ -45,7 +45,7 @@ end
 """
     concrete_bridge_type(
         BT::Type{<:MOI.Bridges.Objective.AbstractBridge},
-        F::Type{<:MOI.AbstractScalarFunction},
+        F::Type{<:MOI.AbstractFunction},
     )::Type
 
 Return the concrete type of the bridge supporting objective functions of type
@@ -56,14 +56,14 @@ This function can only be called if `MOI.supports_objective_function(BT, F)` is
 """
 function concrete_bridge_type(
     ::Type{BT},
-    ::Type{<:MOI.AbstractScalarFunction},
+    ::Type{<:MOI.AbstractFunction},
 ) where {BT}
     return BT
 end
 
 function concrete_bridge_type(
     b::MOI.Bridges.AbstractBridgeOptimizer,
-    F::Type{<:MOI.AbstractScalarFunction},
+    F::Type{<:MOI.AbstractFunction},
 )
     return concrete_bridge_type(MOI.Bridges.bridge_type(b, F), F)
 end
@@ -72,7 +72,7 @@ end
     bridge_objective(
         BT::Type{<:MOI.Bridges.Objective.AbstractBridge},
         model::MOI.ModelLike,
-        func::MOI.AbstractScalarFunction,
+        func::MOI.AbstractFunction,
     )::BT
 
 Bridge the objective function `func` using bridge `BT` to `model` and returns
@@ -86,7 +86,7 @@ a bridge object of type `BT`.
 function bridge_objective(
     ::Type{<:AbstractBridge},
     ::MOI.ModelLike,
-    func::MOI.AbstractScalarFunction,
+    func::MOI.AbstractFunction,
 )
     return throw(
         MOI.UnsupportedAttribute(MOI.ObjectiveFunction{typeof(func)}()),
--- a/src/Bridges/Objective/bridges/vector_slack.jl
+++ b/src/Bridges/Objective/bridges/vector_slack.jl
@@ -39,7 +39,8 @@ struct VectorSlackBridge{
     F<:MOI.AbstractVectorFunction,
     G<:MOI.AbstractVectorFunction,
 } <: AbstractBridge
-    slack::MOI.VectorOfVariables
+    variables::Vector{MOI.VariableIndex}
+    slacked_objectives::Vector{Int}
     constraint::MOI.ConstraintIndex{F,MOI.Nonnegatives}
 end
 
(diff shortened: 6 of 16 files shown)


odow:
JuMP PR is https://github.com/jump-dev/JuMP.jl/pull/3176. I'm pretty happy with this. The bridge stuff isn't really tested yet; all I did was rename things from `AbstractScalarFunction` to `AbstractFunction`. There might be some issues we need to fix, but let's leave that to a new PR, and once we get some solvers hooked up.

blegat:
Thinking more about this, I'm not sure about the vector of objective.

There is a second alternative, that might also be easy to do: adding an `index` field in the `ObjectiveFunction` attribute and `ObjectiveValue` attribute (but not in the `ObjectiveSense` though, we could say they have the same sense).
This would simplify the job for the objective bridges since each objective can be bridged independently.
It also simplify things when the different objective have different types. Having to store them in the same vector objective can be tricky because of it.
The downside is that most solver will ignore the `index` field and will erase their only objective but we can add tests for that.

odow:
> This would simplify the job for the objective bridges since each objective can be bridged independently

Why does this need to happen?

> Otherwise, how do you bridge a vector of bridges

Regular slack bridge: min f(x) becomes min [t], [t] >= f(x)

> But an entry of a VectorQuadraticFunction might in fact be only affine since they might have been promoted to quadratic because only one is quadratic.

Sure... but we have that problem with VectorQuadraticFunction constraints as well.

> that might also be easy to do: adding an index field in the ObjectiveFunction attribute and ObjectiveValue attribute

I considered this. But then `ObjectiveValue` has `index` and `result`? And if we have different objectives, people are going to ask for different senses.

blegat:
> > This would simplify the job for the objective bridges since each objective can be bridged independently
>
> Why does this need to happen?

We don't need the same bridges for each objective. Suppose you have one affine and one quadratic. The affine don't need to be bridged but the quadratic need to be bridged with a slack bridge. If you add a slack for the affine objective, it's a waste.

> > Otherwise, how do you bridge a vector of bridges
>
> Regular slack bridge: min f(x) becomes min [t], [t] >= f(x)

Yes, but how do you apply different bridges to each entry ? If we implement a bridge doing that, it would be very complicated and would be equivalent to what the bridge optimizer would do if we had several scalar bridges.

> > But an entry of a VectorQuadraticFunction might in fact be only affine since they might have been promoted to quadratic because only one is quadratic.
>
> Sure... but we have that problem with VectorQuadraticFunction constraints as well.

Indeed, but we don't force the user to use it, they can use scalar constraints with LessThan instead of Vector with Nonnegatives. You only use VectorQuadraticFunction if you have a conic set with no equivalent in which case it would not make sense to bridge them independently.

> > that might also be easy to do: adding an index field in the ObjectiveFunction attribute and ObjectiveValue attribute
>
> I considered this. But then `ObjectiveValue` has `index` and `result`? And if we have different objectives, people are going to ask for different senses. 

Yes, `ObjectiveValue` would both have an `index` and a `result`, `ObjectiveValue(2, 3)` might indeed not be so easy to read. Since in multi-objective optimization, you always care about multiple solution, the first should be the result index and the second one the objective index so that `ObjectiveValue(2)` is the first objective but second result. At the JuMP level, it would be with keyword anyway.

For different senses, I don't think we should bother. Some optimization framework even force minimization for instance.
One concern is how to delete an objective. Currently, this is done by setting FeasibilitySense. If we have several objective, it makes sense to use objective indices like variable or constraint indices, i.e., allowing holes in the indices in case of deletion, which would be different than for the Vector approach. To delete, we could use `MOI.delete`.

Also, modification of an objective would be easier. You can modify each objective independently.

Since we are going to have a JuMP interface with scalar objective, if we do vector objective at the MOI level, it would add complexity to JuMP. With scalar objectives at the MOI level, `src/Utilities/objective_container.jl` and `src/Utilities/universal_fallback.jl` will get a bit more complicated but the JuMP side will be straightforward and the bridges will be straightforward as well.

odow:
> We don't need the same bridges for each objective. Suppose you have one affine and one quadratic. The affine don't need to be bridged but the quadratic need to be bridged with a slack bridge. If you add a slack for the affine objective, it's a waste.

You could write a vectorquadratic-> vectorscalar objective bridge that bridged only the needed parts.

But on the flip side, the slack bridge bridge relaxing all terms is still valid. And the user could always write the reformulation if they knew they were using a solver that didn't support quadratic objectives. It's the same with any bridge.

> Yes, but how do you apply different bridges to each entry ?

You don't.

> Since we are going to have a JuMP interface with scalar objective, if we do vector objective at the MOI level, it would add complexity to JuMP

See https://github.com/jump-dev/JuMP.jl/pull/3176. This is already implemented and working.

> One concern is how to delete an objective. Currently, this is done by setting FeasibilitySense. If we have several objective, it makes sense to use objective indices like variable or constraint indices, i.e., allowing holes in the indices in case of deletion, which would be different than for the Vector approach.

You can't delete individual objectives. If you want to delete objective 3, set an entirely new objective. This is unlikely to be a bottleneck. And ending up with a hole in your objective is a bigger problem that seems ill-defined.

odow:
I added a vector-valued objective bridge: https://github.com/jump-dev/MathOptInterface.jl/pull/2070/files#diff-0e215633d1123dcfe004cd241301e4efd132ac083bfc0475160366f99498670d

This `AbstractVectorFunction` approach fits too nicely into our abstraction to have to think about independent objectives.

blegat:
> > Yes, but how do you apply different bridges to each entry ?
>
> You don't.

I'd say that might be an issue

> See https://github.com/jump-dev/JuMP.jl/pull/3176. This is already implemented and working.

Indeed, but this is a vector interface at the JuMP level too

odow:
> I'd say that might be an issue

Do you have an example of a model where it is a problem?

odow:
I've updated the bridge so that it bridges only the scalar objectives that aren't `VariableIndex`. I'm not sure we should keep the commit, but it's a proof that we don't need individual bridges for each objective. I mean, we have a very limited set of function types. It's not that bad if a solver supports only `ObjectiveFunction{VectorOfVariables}` and we slack all objectives.

blegat:
> Do you have an example of a model where it is a problem?

We have `VariableIndex`, `ScalarAffineFunction`, `ScalarQuadraticFunction`, soon `ScalarNonlinearFunction` (maybe one for each type of expression if people one to start playing with different ways of expressing a nonlinear function because one is more efficient for a solver and the solver express the reformulation as a bridge to allow the user to directly give the right expression too) (also maybe different types for different types of oracle (0th order, 1st order, ...)), PolyJuMP defines ScalarPolynomialFunction and will reformulated nonlinear into polynomial with a bridge once `ScalarNonlinearFunction` is done.
Of course, multi-objective is still a niche and affine objective are more common but Vector objective seems like creating a rigid wall that we'll have to get around when trying different objective types.
The attribute mechanism of MOI was built so that it's easy to get multiple solutions and it also makes it easy to set multiple objectives.

> I've updated the bridge so that it bridges only the scalar objectives that aren't `VariableIndex`. I'm not sure we should keep the commit, but it's a proof that we don't need individual bridges for each objective. I mean, we have a very limited set of function types. It's not that bad if a solver supports only `ObjectiveFunction{VectorOfVariables}` and we slack all objectives.

It feels like a hacky way and this hardcoding won't play well with extensions :/

blegat:
Otherwise, another solution is to create a `TupleFunction <: MOI.AbstractVectorFunction`.
When the user writes `@objective(model, Max, (x, y + z))` it creates a `TupleFunction{{MOI.VariableIndex,MOI.ScalarAffineFunction}}`. When he writes `@objective(model, Max, [x, y + z])`, it creates a `MOI.VectorAffineFunction`.
If a solver expects a `MOI.VectorAffineFunction`, we can have a bridge to convert a `TupleFunction` into a ``MOI.VectorAffineFunction`.
We can also convert a `TupleFunction` into a `VectorOfVariables` if it only has `VariableIndex`.
Now, bridging each entry of the `TupleFunction` with the right bridges, then collecting them again into a `TupleFunction` and then converting to `VectorOfVariables` might be tricky. We might need to consider tuples of function types in the hyper-graph, etc...

odow:
So the argument is that in a future multi-objective world, someone might want to write an objective that mixes polynomial and linear objectives, and we wouldn't have a VectorPolynomialFunction support that automatically promoted linear to polynomial? I think we'd be adding complexity for a hypothetical situation, when the most common (>99%) use-case for this is going to be bi-objective (MI)LP.

Even if that's the case, people can easily work around it by introducing epigraph variables for their objectives. 

Nevertheless, I'll start drafting a PR that adds an index to ObjectiveFunction and see how far I get.

odow:
Started here: https://github.com/jump-dev/MathOptInterface.jl/pull/2086.

I'm not convinced. This PR is attractive because it is so simple, and fits so neatly within our existing abstraction. Having a vector-valued objective neatly explains why we support only a single objective sense, it works seamlessly with the bridges, and there are no "hole" issues, like what happens if someone sets `ObjectiveFunction{F}(1)` and `ObjectiveFunction{F}(3)`?

Yes, the bridges are slightly more inflexible for extensions, but simplicity has a quality all of its own.

odow:
Let me summarize our scalar bridging problem:

Suppose I have a new type of scalar function, and I have a bridge for objective
function into another formulation, and this function is not supported as an
epigraph constraint. Then I cannot use it in as the objective in a multi-objective
optimization problem because:

 1) there is no vector form of the function
 2) the function does not support an epigraph formulation
 3) the scalar bridge cannot be used because the objective requires a vector
    function.

I think the answer is for the author to implement a vector version of their
function, and a bridge which treats each element as a scalar.

We haven't previously been in this situation because there was no need for the
vector function, but it seems a reasonable way forward, and doesn't introduce
much extra overhead. And certainly less than the overhead of rewriting the
bridging system in #2086.

In PolyJuMP's case, it could just implement a `VectorPolynomialFunction`, which
seems reasonable if it wants to support multi-objective polynomial optimization.

blegat:
Yes, the user can always do it with epigraph as well. Let's take an example.
A user want to compute the Markowitz Efficient Frontier, he will model (https://jump.dev/Convex.jl/stable/examples/portfolio_optimization/portfolio_optimization2/) with is a variation of https://jump.dev/JuMP.jl/stable/tutorials/nonlinear/portfolio/,
He will use `@constraint(model, Min, [w' * Σ * w, w'μ])` so the second objective will be promoted to quadratic and will have to go through the `QuadToSOC` bridge.
Indeed, this efficient frontier has infinitely many points so I don't know whether it makes much sense to consider a nonlinear objective.
The main practical issue might be mixing `VariableIndex` and `ScalarAffineFunction` in the same objective in which case you will create a slack variable for the objective that are already variables if the solver expects a `VectorOfVariables` objective. For that case, the `try`-`catch` bridge you showed might do the trick.
Let's summarize our analysis of the pros and cons at the JuMP-dev call today and decide.

odow:
Friday's call was in favor of this PR. I still need to shop it around various people interested in multi-objective support in JuMP.

The best way to try this out is with the JuMP PR: https://github.com/jump-dev/JuMP.jl/pull/3176

And by reading the documentation: https://jump.dev/JuMP.jl/previews/PR3176/manual/objective/#Set-a-vector-valued-objective

Here's a simple solver that implements this end-to-end: https://github.com/odow/MOO.jl

odow:
I think @xgandibleux is happy with this, https://github.com/jump-dev/JuMP.jl/issues/2099#issuecomment-1408962911, so I'll leave open for a few more days and then merge, unless there are any review comments that arise before then.

odow:
I talked to Xavier, who shopped this around a few colleagues to positive feedback, so I think we're good.

I'll merge this, but we're not in a rush to tag a new MOI release, so we can always address comments once it's on the `master` branch before a release.

PR #2074, opened 2023-01-10
[docs] Document the two-argument version of optimize

### LINKED ISSUE #2073: Document two-arg version of `optimize!`
There is a two-arg version of [`optimize!`](https://github.com/jump-dev/MathOptInterface.jl/blob/60709256ab029656f1ad0466e93aadff7928d8b2/src/MathOptInterface.jl#L84) that is not documented.
It is used in SCS.jl to implement a one-shot optimizer, so maybe it should be mentioned in https://jump.dev/MathOptInterface.jl/stable/tutorials/implementing/ ?

odow:
Closes #2073

### DIFF
--- a/docs/src/reference/models.md
+++ b/docs/src/reference/models.md
@@ -59,6 +59,7 @@ ListOfConstraintAttributesSet
 AbstractOptimizer
 OptimizerWithAttributes
 optimize!
+optimize!(::ModelLike, ::ModelLike)
 instantiate
 default_cache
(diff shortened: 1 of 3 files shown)


odow:
So this is weird. The two-argument has a docstring, but it doesn't show up:

image

odow:
Issue is upstream: https://github.com/JuliaDocs/Documenter.jl/issues/1390

PR #2080, opened 2023-01-22
Add SnoopPrecompile directives

odow:
x-ref https://github.com/jump-dev/JuMP.jl/issues/3192


src/instantiate.jl
```diff
@@ -143,7 +143,7 @@ problem incrementally (see [`supports_incremental_interface`](@ref)), then a
 model.
 """
 function instantiate(
-    optimizer_constructor;
+   (@nospecialize optimizer_constructor);
```
odow, line 146:
```suggestion
    (@nospecialize optimizer_constructor);
```

PR #2082, opened 2023-01-24
[FileFormats.LP] fix writing of off-diag quadratic elements

odow:
Closes https://github.com/jump-dev/MathOptInterface.jl/issues/2081

cc @chriscoey does this fix it?


chriscoey:
that does fix it! thanks @odow!

PR #2086, opened 2023-01-25
Add support for multiple objective functions

odow:
An alternative to #2070

Instead of adding vector-valued objectives, this PR adds
```Julia
struct ObjectiveFunction{F<:AbstractScalarFunction} <: AbstractModelAttribute
    objective_index::Int
    function ObjectiveFunction{F}(
        objective_index::Int = 1,
    ) where {F<:AbstractScalarFunction}
        return new{F}(objective_index)
    end
end
```

This has a number of implications:

 * Other attributes like `ObjectiveValue` need an index
 * Most solver's current implementation of `supports(::Optimizer, ::MOI.ObjectiveFunction{F})` will be incorrect, because they must **also** check the value of the `objective_index` field, that is, every solver will currently lie and tell JuMP that they support multiobjective optimization.
 * We probably need a way to delete individual objectives, otherwise if you want to drop from 3 to 2 objectives, you'll need to set `FEASIBILITY_SENSE`
 * The objective bridging system needs to be rewritten to support multiple objectives
 * We support multiple objectives but only a single objective sense

Still TODOs:

 - [ ] Setting `FEASIBILITY_SENSE` erases all objectives
 - [ ] Bridges
 - [ ] Introduce an `ObjectiveIndex` object instead of using integers?

### DIFF
--- a/src/Test/test_multiobjective.jl
+++ b/src/Test/test_multiobjective.jl
@@ -1,0 +1,353 @@
+# Copyright (c) 2017: Miles Lubin and contributors
+#
+# Use of this source code is governed by an MIT-style license that can be found
+# in the LICENSE.md file or at https://opensource.org/licenses/MIT.
+
+function test_multiobjective_vector_of_variables(
+    model::MOI.ModelLike,
+    ::Config{T},
+) where {T}
+    F = MOI.VariableIndex
+    @requires MOI.supports(model, MOI.ObjectiveFunction{F}(2))
+    x = MOI.add_variables(model, 2)
+    MOI.set(model, MOI.ObjectiveSense(), MOI.MIN_SENSE)
+    for i in 1:2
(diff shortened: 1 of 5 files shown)


blegat:
> The objective bridging system needs to be rewritten to support multiple objectives

Not sure why you say "rewritten", we almost don't need to change anything.

> Do the `objective_index` map to the solver's list of objectives? How does one interact with multiple objectives in `direct_model`?

It might not, the `CachingOptimizer` can map it with the `index_map`.

> Introduce an `ObjectiveIndex` object instead of using integers?

Might be needed for using it with the `CachingOptimizer` index map. The solvers don't have to support deletion of objective, only `MOI.Utilities.Model` does. So the user will be able to delete an objective but the solvers can always assume having their objective functions added in contiguous order.

odow:
Closing in favor of #2070

PR #2091, opened 2023-02-06
[Bridges] add NumberConversionBridge

odow:
Closes https://github.com/jump-dev/MathOptInterface.jl/issues/1694

@blegat is this what you had in mind?

### DIFF
--- a/src/Bridges/Constraint/Constraint.jl
+++ b/src/Bridges/Constraint/Constraint.jl
@@ -42,6 +42,7 @@ include("bridges/ltgt_to_interval.jl")
 include("bridges/norm_infinity.jl")
 include("bridges/norm_one.jl")
 include("bridges/norm_spec_nuc_to_psd.jl")
+include("bridges/number_conversion.jl")
 include("bridges/quad_to_soc.jl")
 include("bridges/relentr_to_exp.jl")
 include("bridges/rsoc_soc.jl")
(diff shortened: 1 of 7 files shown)


blegat:
Yes, this is what I had in mind, thanks!

src/Bridges/Constraint/Constraint.jl
```diff
@@ -112,6 +113,8 @@ function add_all_bridges(bridged_model, ::Type{T}) where {T}
     MOI.Bridges.add_bridge(bridged_model, IndicatorGreaterToLessThanBridge{T})
     MOI.Bridges.add_bridge(bridged_model, SemiToBinaryBridge{T})
     MOI.Bridges.add_bridge(bridged_model, ZeroOneBridge{T})
+    # Do not add by default
+    # MOI.Bridges.add_bridge(bridged_model, NumberConversionBridge{T})
```
odow, line 117: I guess a question: should we turn it on automatically?

odow: Or should MiniZinc add it as a `ListOfNonStandardBridges`?

odow: I think the conclusion is `ListOfNonStandardBridges`: https://github.com/jump-dev/MathOptInterface.jl/issues/1694#issue-1067285249

We can always enable later, but it's breaking to disable.

odow:
Happy?

src/Bridges/Constraint/Constraint.jl
blegat: Yes, a solver wanting constraints in a type `S` other than `T` should add the bridge to `S` in `ListOfNonStandardBridges{T}`. Since JuMP only creates constraints of type `T`, I don't know how it would be useful to have the bridge that converts everything to `T` by default. We could wait and see.

blegat (approved):

Issue #2094, opened 2023-02-14
Add Parameter{T} <: AbstractScalarSet


odow:
Today I had a call with @sstroemer and @jd-lara, and we discussed the logistics of adding a `Parameter{T} <: AbstractScalarSet` to MOI.

## Motivation

The `Parameter{T}` set is similar to the `EqualTo{T}` set, but with the special case that solvers could support `add_constrained_variable(model, ::Parameter)` and not add the index as a decision variable to the underlying solver object.

The goal is to support JuMP syntax like:
```julia
model = Model()
@variable(model, x)
@variable(model, p in MOI.Parameter(1.0))
# or
@variable(model, p == 1.0, Parameter)
@constraint(model, p * x <= p + 1)
@objective(model, Min, p * x + 3 * p)
optimize!(model)
fix(p, 2.0)
optimize!(model)
```

The benefits of an explicit parameter are that, compared to `EqualTo`, they can reduce the problem size and allow solvers to interpret `p * x` as an affine function instead of quadratic.

## Backstory

We've tried quite a few ways to achieve this over the years...

### EqualTo

The easiest approach is just to use fixed variables a parameters and let the solvers presolve them out. This works for additive parameters, but fails for `p * x` if the solver does not support quadratic constraints.

Another downside is that it leads to more decision variables in the problem because `p` is added as a decision variable.

### Nonlinear

JuMP already has nonlinear parameters, which are added via `@NLparameter`. If we add a `NonlinearFunction`, https://github.com/jump-dev/MathOptInterface.jl/pull/2059, then we'll need some way to add parameters to MOI anyway.

The current parameter implementation is very specialized inside `MOI.Nonlinear.Model`.

### ParameterJuMP

@joaquimg started [ParameterJuMP.jl](https://github.com/JuliaStochOpt/ParameterJuMP.jl) to support right-hand side parameters in JuMP.

The core premise is that it is a JuMP extension which defines a new `ParameterRef <: AbstractVariableRef`, and then stores parameterized JuMP expressions as a double expression:
```Julia
struct ParameterizedAffExpr
    lhs::GenericAffExpr{V,VariableRef}
    rhs::GenericAffExpr{V,ParameterRef}
end
```

There are two main downsides to this approach:

 * Two expressions, coupled with Julia's compiler optimizations which can sometimes put an `AffExpr` on the stack instead of heap, means that memory overhead can be 2-3X larger than a JuMP model with the `EqualTo` approach.
[...]

PR #2095, opened 2023-02-15
Add Parameter set

odow:
Part of #2094

### DIFF
--- a/src/Bridges/Variable/bridges/parameter.jl
+++ b/src/Bridges/Variable/bridges/parameter.jl
@@ -1,0 +1,117 @@
+# Copyright (c) 2017: Miles Lubin and contributors
+# Copyright (c) 2017: Google Inc.
+#
+# Use of this source code is governed by an MIT-style license that can be found
+# in the LICENSE.md file or at https://opensource.org/licenses/MIT.
+
+"""
+    ParameterToEqualToBridge{T} <: Bridges.Variable.AbstractBridge
+
+`ParameterToEqualToBridge` implements the following reformulation:
+
+* ``x \\in Parameter(v)`` into ``x == v``
+
+## Source node
+
+`ParameterToEqualToBridge` supports:
+
+ * [`MOI.VariableIndex`](@ref) in [`MOI.Parameter`](@ref)
+
+## Target nodes
+
+`ParameterToEqualToBridge` creates:
+
+ * One variable node: [`MOI.VariableIndex`](@ref) in [`MOI.EqualTo{T}`](@ref)
+"""
+struct ParameterToEqualToBridge{T} <: AbstractBridge
+    x::MOI.VariableIndex
+    ci::MOI.ConstraintIndex{MOI.VariableIndex,MOI.EqualTo{T}}
+end
+
+const ParameterToEqualTo{T,OT<:MOI.ModelLike} =
+    SingleBridgeOptimizer{ParameterToEqualToBridge{T},OT}
+
+function bridge_constrained_variable(
+    ::Type{ParameterToEqualToBridge{T}},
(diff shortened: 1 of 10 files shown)


blegat:
I wouldn't bridge it to `EqualTo` but rather to nothing and substitute it for its value. I guess both bridges are useful and we can discuss which one should be the default. It seems to me that substitution would be a good default. Otherwise mutliplicative parameters won't work. Having to remove a bridge in order to get the dual (or use POI if multiplicative parameters are used) seems reasonable.

odow:
> I wouldn't bridge it to EqualTo but rather to nothing and substitute it for its value.

The whole point of parameters is that we can efficiently update them. The substitution doesn't preserve enough information to later update the parameters. We do need to bridge to `EqualTo`.

src/sets.jl
```diff
@@ -245,10 +245,26 @@ struct EqualTo{T<:Number} <: AbstractScalarSet
...
 
+"""
+    Parameter{T<:Number}(value::T)
+
+The set containing the single point ``x \\in \\mathbb{R}`` where ``x`` is given
+by `value`.
+
+This set is conceptually similar to the [`EqualTo`](@ref) set, except that a
+variable constrained to the `Parameter` set cannot have other constraints added
+to it, and the `Parameter` set can never be deleted. Thus, solvers are free to
```
joaquimg, line 256: Should it always be used with `add_constrained_variable`? so that it would be ok to "delete" de variable, which will take the set with it...

odow: I guess the answer is "it depends." It'd be fine to add the variable then the constraint in `MOI.Utilities.Model`, but if it's a solver (or POI), it might not like getting the constraint after it has already created a variable.

So yes, you should probably always use `add_constrained_variable`, which also gets around the deletion issue, because you can't delete the constraint of a constrained variable.

src/Bridges/Variable/bridges/parameter.jl
```diff
@@ -0,0 +1,106 @@
...
+
+function MOI.get(
+    model::MOI.ModelLike,
+    attr::Union{MOI.ConstraintFunction,MOI.ConstraintPrimal,MOI.ConstraintDual},
+    bridge::ParameterToEqualToBridge,
+)
+    return MOI.get(model, attr, bridge.ci)
+end
+
+MOI.Bridges.bridged_function(bridge::ParameterToEqualToBridge) = bridge.x
```
blegat, line 102: I think `bridged_function` and `unbridged_map` should always return `ScalarAffineFunction`. We could make this more flexible to avoid extra allocations here but I would double check that this PR works with substitutions, I'm surprised it does

odow: It seemed to work fine either way.  The functions are getting called in the tests, so I guess it's working.

src/sets.jl
```diff
@@ -245,10 +245,47 @@ struct EqualTo{T<:Number} <: AbstractScalarSet
...
+by `value`.
+
+The `Parameter` set is conceptually similar to the [`EqualTo`](@ref) set, except
+that a variable constrained to the `Parameter` set cannot have other constraints
+added to it, and the `Parameter` set can never be deleted. Thus, solvers are
+free to treat the variable as a constant, and they need not add it as a decision
+variable to the model.
+
+Because of this behavior, you must add parameters using [`add_constrained_variable`](@ref),
+and solvers should declare [`supports_constrained_variable`](@ref) and not
```
odow, line 261:
```suggestion
and solvers should declare [`supports_add_constrained_variable`](@ref) and not
```

blegat (approved):

joaquimg (approved):

PR #2096, opened 2023-02-16
Use import MathOptInterface as MOI

odow:
After using `import MultiObjectiveAlgorithms as MOA`, I realized that we've never exploited this Julia syntax in MOI (because most of the code pre-dates Julia introducing it).

I think we can and should be recommend that people use `import MathOptInterface as MOI` instead of `import MathOptInterface; const MOI = MathOptInterface`. The latter leads to a mixture of `MathOptInterface.` and `MOI.` in the code, gets around the `using` / `import` confusion, and `MOI` doesn't export anything.

The diff touches a lot of files, but the changes are trivial.

### DIFF
--- a/test/Bridges/Variable/free.jl
+++ b/test/Bridges/Variable/free.jl
@@ -8,8 +8,7 @@ module TestVariableFree
 
 using Test
 
-using MathOptInterface
-const MOI = MathOptInterface
+import MathOptInterface as MOI
 
 function runtests()
     for name in names(@__MODULE__; all = true)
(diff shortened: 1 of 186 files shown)


mlubin (approved):

joaquimg (approved):

test/Bridges/Variable/free.jl
```diff
@@ -85,16 +84,16 @@ function test_modification_multirow_vectoraffine_nonpos()
     )
     s = """
     variables: xpos, xneg
-    nonneg: [xpos, xneg] in MathOptInterface.Nonnegatives(2)
-    c: [4.0xpos + -4.0xneg + -1.0, 3.0xpos + -3.0xneg + -1.0] in MathOptInterface.Nonpositives(2)
+    nonneg: [xpos, xneg] in Nonnegatives(2)
```
blegat, line 87: This is nice

PR #2097, opened 2023-02-22
[Utilities] add with_cache_type to instantiate

odow:
This is useful when creating optimizers which do not support the incremental interface, or which only partially implement it.

This change is motivated by some recent examples in which it was necessary for the user to hard-code the caching logic to work-around the lack of deletion in a solver, and the caching logic is confusing to even advanced users. See:

 * https://discourse.julialang.org/t/jump-scip-error/93958
 * https://github.com/jump-dev/MultiObjectiveAlgorithms.jl/pull/46

### DIFF
--- a/src/instantiate.jl
+++ b/src/instantiate.jl
@@ -125,36 +125,68 @@ end
 """
     instantiate(
         optimizer_constructor,
-        with_bridge_type::Union{Nothing, Type} = nothing,
+        with_cache_type::Union{Nothing,Type} = nothing,
+        with_bridge_type::Union{Nothing,Type} = nothing,
     )
 
-Creates an instance of optimizer by either:
+Create an instance of an optimizer by either:
+
  * calling `optimizer_constructor.optimizer_constructor()` and setting the
    parameters in `optimizer_constructor.params` if `optimizer_constructor` is a
    [`OptimizerWithAttributes`](@ref)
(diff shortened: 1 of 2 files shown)


src/instantiate.jl
```diff
@@ -125,36 +125,68 @@ end
...
-problem incrementally (see [`supports_incremental_interface`](@ref)), then a
-[`Utilities.CachingOptimizer`](@ref) is added to store a cache of the bridged
-model.
+If `with_cache_type` and `with_bridge_type` are both not `nothing`, then they
+must be the same type.
 """
 function instantiate(
     (@nospecialize optimizer_constructor);
     with_bridge_type::Union{Nothing,Type} = nothing,
+    with_cache_type::Union{Nothing,Type} = nothing,
```
blegat, line 163:
```suggestion
    with_cache_type::Union{Nothing,Type} = supports_incremental_interface(optimizer) ? nothing : with_bridge_type,
```

blegat: That would allow instantiating solver not supporting incremental interface, e.g., if you know you are using `copy_to`

odow: I don't understand the benefit of the change? And we can't call it here because `optimizer` isn't in scope yet.

blegat (approved):

PR #2099, opened 2023-02-23
Improve docstrings and methods in src/sets.jl

odow:
Given that the set definitions are perhaps the most important part of MOI, the docstrings have always been a little sparse.

This PR adds examples to them and re-orders some of the methods to break-apart Unions and place the definition directly beneath each set. (The `==` definition for `Interval` was 1000 lines away beneath `Semiinteger`.)

### DIFF
--- a/src/sets.jl
+++ b/src/sets.jl
@@ -4,8 +4,6 @@
 # Use of this source code is governed by an MIT-style license that can be found
 # in the LICENSE.md file or at https://opensource.org/licenses/MIT.
 
-# Sets
-
 # Note: When adding a new set, also add it to Utilities.Model.
 
 """
@@ -18,9 +16,10 @@ as the set is passed unmodifed during [`copy_to`](@ref).
 abstract type AbstractSet end
 
 """
-    dimension(s::AbstractSet)
+    dimension(set::AbstractSet)
 
-Return the [`output_dimension`](@ref) that an [`AbstractFunction`](@ref) should have to be used with the set `s`.
+Return the [`output_dimension`](@ref) that an [`AbstractFunction`](@ref) should
+have to be used with the set `set`.
 
 ### Examples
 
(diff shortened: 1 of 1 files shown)


src/sets.jl
```diff
@@ -4,25 +4,51 @@
...
 
-Abstract supertype for set objects used to encode constraints. A set object
-should not contain any [`VariableIndex`](@ref) or [`ConstraintIndex`](@ref)
-as the set is passed unmodifed during [`copy_to`](@ref).
+Abstract supertype for set objects used to encode constraints.
+
+!!! note
+    When developing new sets, the set struct must not contain any
+    [`VariableIndex`](@ref) or [`ConstraintIndex`](@ref) objects, because the
+    set is passed unmodifed during [`copy_to`](@ref).
```
blegat, line 17: I don't think the reader will understand my a set needs to be modified during `copy_to` unless he's a core dev

src/sets.jl
```diff
@@ -4,25 +4,51 @@
...
+ * `Base.:(==)(x::S, y::S)`
+
+Subtypes of `AbstractSet` such as [`AbstractScalarSet`](@ref) and
+[`AbstractVectorSet`](@ref) may prescribe additional required methods.
+
+## Optional methods
+
+You may optionally implement:
+
+ * [`dual_set`](@ref)
```
blegat, line 33: I would add `set_dot`

blegat: Saying to only implement it in case the dot product is not `LinearAlgebra.dot` between the vectors

src/sets.jl
```diff
@@ -112,16 +133,325 @@ dimension(s::AbstractScalarSet) = 1
     AbstractVectorSet
 
 Abstract supertype for subsets of ``\\mathbb{R}^n`` for some ``n``.
+
+## Required methods
+
+All `AbstractVectorSet`s of type `S` must implement:
+
+ * [`dimension`](@ref), although by convention, you do not need to implement
```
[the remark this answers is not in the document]
odow: Done

src/sets.jl
odow: Removed the `because ... copy_to` clause, and moved the note to a `## Note for developers` since it doesn't apply to users.

PR #2101, opened 2023-02-27
[Utilities] Fix for empty product of sets

blegat:
I got this error when working on https://github.com/jump-dev/DiffOpt.jl/pull/231. For a program with a quadratic objective and no constraints except variable bounds, the DiffOpt's product of sets don't record any set type hence the product of sets has no set type.

### DIFF
--- a/src/Utilities/product_of_sets.jl
+++ b/src/Utilities/product_of_sets.jl
@@ -206,7 +206,12 @@ end
 
 function MOI.dimension(sets::OrderedProductOfSets)
     @assert sets.final_touch
-    return sets.num_rows[end]
+    if isempty(sets.num_rows)
+        # There is not set type
+        return 0
+    else
+        return sets.num_rows[end]
+    end
 end
 
 function rows(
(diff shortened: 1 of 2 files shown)


src/Utilities/product_of_sets.jl
```diff
@@ -206,7 +206,12 @@ end
 
 function MOI.dimension(sets::OrderedProductOfSets)
     @assert sets.final_touch
-    return sets.num_rows[end]
+    if isempty(sets.num_rows)
+        # There is not set type
```
odow, line 210:
```suggestion
        # There is no set type
```

test/Utilities/matrix_of_constraints.jl
```diff
@@ -610,6 +610,19 @@ function test_set_with_dimension()
...
+
+function test_empty_product_of_sets(T = Int)
+    model = _new_VectorSets(EmptyProductOfSets)
+    x = MOI.add_variable(model)
+    MOI.Utilities.final_touch(model, nothing)
+    A = convert(
+        SparseArrays.SparseMatrixCSC{T,Int},
+        model.constraints.coefficients,
+    )
+    @test size(A) == (0, 1)
```
odow, line 623:
```suggestion
    @test size(A) == (0, 1)
    return
```

PR #2108, opened 2023-03-01
Tidy functions.jl and improve docstrings

odow:
A follow-up to #2099 

Preview: https://jump.dev/MathOptInterface.jl/previews/PR2108/reference/standard_form/

### DIFF
--- a/src/functions.jl
+++ b/src/functions.jl
@@ -279,6 +279,19 @@ function Base.copy(f::ScalarQuadraticFunction)
 end
 
 """
+    abstract type AbstractVectorFunction <: AbstractFunction
+
+Abstract supertype for vector-valued [`AbstractFunction`](@ref)s.
+
+## Required methods
+
+All subtypes of `AbstractVectorFunction` must implement:
+
+ * [`output_dimension`](@ref)
+"""
+abstract type AbstractVectorFunction <: AbstractFunction end
+
+"""
     VectorOfVariables(variables::Vector{VariableIndex}) <: AbstractVectorFunction
 
 The vector-valued function `f(x) = variables`, where `variables` is a subset of
--- a/src/functions.jl
+++ b/src/functions.jl
@@ -282,7 +282,7 @@ end
     VectorOfVariables(variables::Vector{VariableIndex}) <: AbstractVectorFunction
 
 The vector-valued function `f(x) = variables`, where `variables` is a subset of
-[`MOI.VariableIndex`](@ref)es in the model.
+[`VariableIndex`](@ref)es in the model.
 
 The list of `variables` may contain duplicates.
 
--- a/src/functions.jl
+++ b/src/functions.jl
@@ -383,7 +383,7 @@ term_pair(t::VectorAffineTerm) = term_indices(t) => coefficient(t)
 
 The vector-valued affine function ``A x + b``, where:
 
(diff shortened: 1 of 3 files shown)


src/functions.jl
```diff
@@ -173,21 +463,64 @@ function VectorQuadraticTerm(
...
+           MOI.VectorAffineTerm(Int64(2), MOI.ScalarAffineTerm(3.0, x)),
+       ];
+
+julia> quad_terms = [
+        MOI.VectorQuadraticTerm(Int64(1), MOI.ScalarQuadraticTerm(2.0, x, x)),
+        MOI.VectorQuadraticTerm(Int64(2), MOI.ScalarQuadraticTerm(3.0, x, y)),
+           ];
+
+julia> f = MOI.VectorQuadraticFunction(quad_terms, affine_terms, constants)
+MathOptInterface.VectorQuadraticFunction{Float64}(MathOptInterface.VectorQuadraticTerm{Float64}[MathOptInterface.VectorQuadraticTerm{Float64}(1, MathOptInterface.ScalarQuadraticTerm{Float64}(2.0, MathOptInterface.VariableIndex(1), MathOptInterface.VariableIndex(1))), MathOptInterface.VectorQuadraticTerm{Float64}(2, MathOptInterface.ScalarQuadraticTerm{Float64}(3.0, MathOptInterface.VariableIndex(1), MathOptInterface.VariableIndex(2)))], MathOptInterface.VectorAffineTerm{Float64}[MathOptInterface.VectorAffineTerm{Float64}(1, MathOptInterface.ScalarAffineTerm{Float64}(2.0, MathOptInterface.VariableIndex(1))), MathOptInterface.VectorAffineTerm{Float64}(2, MathOptInterface.ScalarAffineTerm{Float64}(3.0, MathOptInterface.VariableIndex(1)))], [4.0, 5.0])
```
blegat, line 519: We should really improve the printing of functions, debugging it awful atm

src/functions.jl
```diff
@@ -86,80 +367,89 @@ function VectorAffineTerm(
...
+    VectorAffineFunction{T}(
+        terms::Vector{VectorAffineTerm{T}},
+        constants::Vector{T},
+    ) where {T}
 
 The vector-valued affine function ``A x + b``, where:
-* ``A`` is a sparse matrix specified by a list of `VectorAffineTerm` objects.
-* ``b`` is a vector specified by `constants`
+
+ * ``A c`` is the sparse matrix given by the vector of [`VectorAffineTerm`](@ref)s
```
blegat, line 386:
```suggestion
 * ``A`` is the sparse matrix given by the vector of [`VectorAffineTerm`](@ref)s
```

src/functions.jl
odow: Yes. But that can be a separate PR.

I even wonder about something like:
```Julia
ScalarAffineFunction:
    constant: 1.0
    terms:
      VariableIndex(1) => 2.0
      VariableIndex(2) => 2.0
      VariableIndex(3) => 2.0
      VariableIndex(4) => 2.0
```
Rather than trying to pretty-print it all.

blegat: I would prefer pretty printing for the oneline version as it is sometimes coefficients of a multivariate polynomials that is trying to print itself in oneline.

blegat: For the long version, your suggestion seems nice

odow: Open an issue? What is the "long version"?

PR #2110, opened 2023-03-01
[docs] Add Google style guide for Vale and fix errors and warnings

odow:
Styles taken from https://github.com/errata-ai/Google. They are MIT licensed.

I haven't included all the rules because some generate too many warnings or false positives. 

 * All "suggestion" level rules
 * Google.Headings : false positives with links in headers
 * Google.Spacing : false positives with Julia syntax
 * Google.We : I violate this one all the time. It'd be a large PR to change, but we could do it in future
 * Google.Will : I violate this one all the time. It'd be a large PR to change, but we could do it in future 
 * Google.WordList : a few pedantic suggestions that don't apply well to MOI or JuMP

### DIFF
--- a/docs/src/changelog.md
+++ b/docs/src/changelog.md
@@ -392,7 +392,7 @@ removed, similar to how Julia 1.0 was Julia 0.7 with deprecations removed.
 ### Troubleshooting problems when updating
 
 If you experience problems when updating, you are likely using previously
-deprecated functionality. (By default, Julia does not warn when you use
+deprecated features. (By default, Julia does not warn when you use
 deprecated features.)
 
 To find the deprecated features you are using, start Julia with `--depwarn=yes`:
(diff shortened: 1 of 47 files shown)


docs/src/changelog.md
```diff
@@ -1180,10 +1180,10 @@ This release contains backports from the ongoing development of the v0.10 releas
 - Add test for empty rows in vector linear constraint.
 - Rework errors: `CannotError` has been renamed `NotAllowedError` and
   the distinction between `UnsupportedError` and `NotAllowedError` is now
-  about whether the element is not supported (i.e. it cannot be copied a
+  about whether the element is not supported (that is, it cannot be copied a
```
odow, line 1183:
```suggestion
  about whether the element is not supported (for example, it cannot be copied a
```

blegat (approved):

PR #2112, opened 2023-03-02
Print algebraic form of AbstractFunction by default

### LINKED ISSUE #2111: Better printing of functions
It would be nice if we would have `sprint(show, f)` print the same as `JuMP.function_string` with MIME"text/plain" and `sprint(show, "text/latex", f)` print the same as `JuMP.function_string` with MIME"text/plain". We could also have `sprint(print, f)` print the suggestion of @odow in https://github.com/jump-dev/MathOptInterface.jl/pull/2108#discussion_r1122236516.
Note that we already have all the code needed to do that in `MOI.Utilities._to_string`. We can just pass it a dummy
```julia
struct NoVariableNameModel <: MOI.ModelLike end
MOI.get(::NoVariableNameModel, ::MOI.VariableName, ::MOI.VariableIndex) = ""
```

odow:
Closes #2111 

Open to ideas for what we should print, exactly.

Docstrings can all be updated using `--fix` when building the docs:

```julia
julia> ARGS
1-element Vector{String}:
 "--fix"

julia> include("docs/make.jl")
WARNING: redefinition of constant _PAGES. This may fail, cause incorrect answers, or produce other errors.
[ Info: SetupBuildDirectory: setting up build directory.
[ Info: Doctest: running doctests.
[ Info: Skipped ExpandTemplates step (doctest only).
[ Info: Skipped CrossReferences step (doctest only).
[ Info: Skipped CheckDocument step (doctest only).
[ Info: Skipped Populate step (doctest only).
[ Info: Skipped RenderDocument step (doctest only).
  3.325270 seconds (4.70 M allocations: 319.401 MiB, 3.95% gc time)
┌ Warning: Documenter could not auto-detect the building environment Skipping deployment.
└ @ Documenter ~/.julia/packages/Documenter/H5y27/src/deployconfig.jl:75
```

### DIFF
--- a/src/Utilities/print.jl
+++ b/src/Utilities/print.jl
@@ -637,22 +637,32 @@ function Base.print(io::IO, model::MOI.ModelLike; kwargs...)
     return _print_model(io, _PrintOptions(MIME("text/plain"); kwargs...), model)
 end
 
-struct _NoVariableNameModel <: MOI.ModelLike end
+struct _NoVariableNameModel{M} <: MOI.ModelLike end
 
 function MOI.get(
-    ::_NoVariableNameModel,
+    ::_NoVariableNameModel{MIME"text/plain"},
     ::MOI.VariableName,
     x::MOI.VariableIndex,
 )
     return "MOI.VariableIndex($(x.value))"
 end
 
+function MOI.get(
+    ::_NoVariableNameModel{MIME"text/latex"},
+    ::MOI.VariableName,
+    ::MOI.VariableIndex,
+)
+    return ""  # Leave as default
+end
+
 function Base.show(io::IO, mime::MIME"text/plain", f::MOI.AbstractFunction)
-    print(io, _to_string(_PrintOptions(mime), _NoVariableNameModel(), f))
+    model = _NoVariableNameModel{MIME"text/plain"}()
+    print(io, _to_string(_PrintOptions(mime), model, f))
     return
 end
 
 function Base.show(io::IO, mime::MIME"text/latex", f::MOI.AbstractFunction)
-    print(io, _to_string(_PrintOptions(mime), _NoVariableNameModel(), f))
+    model = _NoVariableNameModel{MIME"text/latex"}()
+    print(io, _to_string(_PrintOptions(mime), model, f))
     return
 end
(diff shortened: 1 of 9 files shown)


src/Utilities/print.jl
```diff
@@ -636,3 +636,23 @@ end
...
+    return "MOI.VariableIndex($(x.value))"
+end
+
+function Base.show(io::IO, mime::MIME"text/plain", f::MOI.AbstractFunction)
+    print(io, _to_string(_PrintOptions(mime), _NoVariableNameModel(), f))
+    return
+end
+
+function Base.show(io::IO, mime::MIME"text/latex", f::MOI.AbstractFunction)
+    print(io, _to_string(_PrintOptions(mime), _NoVariableNameModel(), f))
```
blegat, line 656: Let's give `"v"` as `default_name` in the `_PrintOptions` here

blegat:
Let's do https://github.com/jump-dev/MathOptInterface.jl/issues/2111#issuecomment-1453979907 for LaTeX

odow:
Done

src/Utilities/print.jl
```diff
@@ -636,3 +636,33 @@ end
...
+function MOI.get(
+    ::_NoVariableNameModel{MIME"text/latex"},
+    ::MOI.VariableName,
+    ::MOI.VariableIndex,
+)
+    return ""  # Leave as default
+end
+
+function Base.show(io::IO, mime::MIME"text/plain", f::MOI.AbstractFunction)
+    model = _NoVariableNameModel{MIME"text/plain"}()
```
blegat, line 659:
```suggestion
    model = _NoVariableNameModel(mime)
```

src/Utilities/print.jl
```diff
@@ -636,3 +636,33 @@ end
...
+    print(io, _to_string(_PrintOptions(mime), model, f))
+    return
+end
+
+function Base.show(io::IO, mime::MIME"text/latex", f::MOI.AbstractFunction)
+    model = _NoVariableNameModel{MIME"text/latex"}()
```
blegat, line 665:
```suggestion
    model = _NoVariableNameModel(mime)
```

blegat (approved):
Looks good, thanks! I have a small suggestion but it's more of a why not so up to you

PR #2115, opened 2023-03-06
Implement starting values for Constraint.QuadtoSOCBridge

blegat:
Needed for https://github.com/jump-dev/DiffOpt.jl/pull/231

- [x] Tests

### DIFF
--- a/src/Bridges/Constraint/bridges/quad_to_soc.jl
+++ b/src/Bridges/Constraint/bridges/quad_to_soc.jl
@@ -9,7 +9,7 @@
 
 `QuadtoSOCBridge` converts quadratic inequalities
 ```math
-\\frac{1}{2}x^T Q x + a^T x + b \\le 0
+\\frac{1}{2}x^T Q x + a^T x \\le ub
 ```
 into [`MOI.RotatedSecondOrderCone`](@ref) constraints, but it only applies when
 ``Q`` is positive definite.
@@ -17,13 +17,17 @@ into [`MOI.RotatedSecondOrderCone`](@ref) constraints, but it only applies when
 This is because, if `Q` is positive definite, there exists `U` such that
 ``Q = U^T U``, and so the inequality can then be rewritten as;
 ```math
-\\|U x\\|_2^2 \\le 2 (-a^T x - b)
+\\|U x\\|_2^2 \\le 2 (-a^T x + ub)
 ```
 
-Therefore, `QuadtoSOCBridge` implements the following reformulation:
+Therefore, `QuadtoSOCBridge` implements the following reformulations:
 
-  * ``\\frac{1}{2}x^T Q x + a^T x + b \\le 0`` into
-    ``(1, -a^T x - b, Ux) \\in RotatedSecondOrderCone``
+  * ``\\frac{1}{2}x^T Q x + a^T x \\le ub`` into
+    ``(1, -a^T x + ub, Ux) \\in RotatedSecondOrderCone``
+    where ``Q = U^T U``
+  * ``\\frac{1}{2}x^T Q x + a^T x \\ge lb`` into
+    ``(1, a^T x - lb, Ux) \\in RotatedSecondOrderCone``
+    where ``-Q = U^T U``
 
 ## Source node
 
@@ -85,9 +89,9 @@ function bridge_constraint(
         )
     end
     # Construct the VectorAffineFunction. We're aiming for:
-    #  |          1 |
-    #  | -a^T x - b | ∈ RotatedSecondOrderCone()
-    #  |     Lx + 0 |
+    #  |          1  |
+    #  | -a^T x + ub | ∈ RotatedSecondOrderCone()
+    #  |     Ux + 0  |
     # Start with the -a^T x terms...
     vector_terms = MOI.VectorAffineTerm{T}[
         MOI.VectorAffineTerm(
@@ -95,7 +99,7 @@ function bridge_constraint(
             MOI.ScalarAffineTerm(scale * term.coefficient, term.variable),
         ) for term in func.affine_terms
     ]
-    # Add the Lx terms...
+    # Note that `L = U'` so we add the terms of L'x
     L, p = SparseArrays.sparse(F.L), F.p
     I, J, V = SparseArrays.findnz(L)
     for i in 1:length(V)
@@ -109,11 +113,12 @@ function bridge_constraint(
             MOI.VectorAffineTerm(J[i] + 2, MOI.ScalarAffineTerm(V[i], xi)),
         )
     end
-    # This is the [1, b, 0] vector...
+    # This is the [1, ub, 0] vector...
     set_constant = MOI.constant(set)
+    MOI.throw_if_scalar_and_constant_not_zero(func, typeof(set))
     vector_constant = vcat(
         one(T),
-        scale * (func.constant - set_constant),
+        -scale * set_constant,
         zeros(T, size(L, 1)),
     )
     f = MOI.VectorAffineFunction(vector_terms, vector_constant)
@@ -230,12 +235,20 @@ function MOI.get(
     attr::Union{MOI.ConstraintPrimal,MOI.ConstraintPrimalStart},
     bridge::QuadtoSOCBridge,
 )
+    # The constraint primal is x'Qx/2 + a'x
+    # If `less_than` then `Q = U'U` and we have the value of
+    # `Ux` and `-a'x + ub`, so we get it with
+    # `(Ux)'Ux / 2 - (-a'x + ub) + ub`
+    # Otherwise, `Q = -U'U` and we have the value of
+    # `Ux` and `a'x - ub`, so we get it with
+    # `-(Ux)'Ux / 2 + (a'x - lb) + ub`
     soc = MOI.get(model, attr, bridge.soc)
-    if soc == nothing
+    if soc === nothing
         return nothing
     end
     output = sum(soc[i]^2 for i in 3:bridge.dimension)
     output /= 2
+    @show output
     output -= soc[1] * soc[2]
     if !bridge.less_than
         output = -output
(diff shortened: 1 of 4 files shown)


src/Bridges/Constraint/bridges/quad_to_soc.jl
```diff
@@ -268,6 +286,26 @@ function MOI.get(
...
+    model::MOI.ModelLike,
+    attr::MOI.ConstraintDualStart,
+    bridge::QuadtoSOCBridge,
+    λ,
+)
+    u = bridge.less_than ? -λ : λ
+    # This assumes that the variable primal start was set first
+    # If it was not, we cannot do anything anyway so let's hope
+    x = MOI.get(model, MOI.VariablePrimalStart(), bridge.index_to_variable_map)
+    v = u * sum(x.^2) / 2
```
odow, line 303:
```suggestion
    v = u * sum(x .^ 2) / 2
```

src/Bridges/Constraint/bridges/quad_to_soc.jl
```diff
@@ -118,21 +119,19 @@ function bridge_constraint(
     f = MOI.VectorAffineFunction(vector_terms, vector_constant)
     dimension = MOI.output_dimension(f)
     soc = MOI.add_constraint(model, f, MOI.RotatedSecondOrderCone(dimension))
-    return QuadtoSOCBridge(soc, dimension, less_than, set_constant)
+    return QuadtoSOCBridge(soc, dimension, less_than, set_constant, index_to_variable_map)
```
odow, line 122:
```suggestion
    return QuadtoSOCBridge(
        soc,
        dimension,
        less_than,
        set_constant,
        index_to_variable_map,
    )
```

odow:
Tests should be automatic if you define `MOI.supports` for `MOI.ConstraintPrimalStart` and `MOI.ConstraintDualStart`

src/Bridges/Constraint/bridges/quad_to_soc.jl
```diff
@@ -226,6 +244,35 @@ function MOI.get(
...
+    soc = MOI.get(model, MOI.ConstraintFunction(), bridge.soc)
+    s = MOI.Utilities.eval_variables(soc) do v
+        # If not set, we assume x = 0.
+        return something(MOI.get(model, MOI.VariablePrimalStart(), v), zero(T))
+    end
+    if !bridge.less_than
+        value = -value
+    end
+    s[2] += value
+    MOI.set(model, attr, bridge.soc, s)
```
odow, line 262: I'm not sure this method is right? It doesn't preserve a round trip.

src/Bridges/Constraint/bridges/quad_to_soc.jl
```diff
@@ -226,6 +244,35 @@ function MOI.get(
...
 
+function MOI.set(
+    model::MOI.ModelLike,
+    attr::MOI.ConstraintPrimalStart,
+    bridge::QuadtoSOCBridge{T},
+    value,
+) where {T}
+    soc = MOI.get(model, MOI.ConstraintFunction(), bridge.soc)
+    s = MOI.Utilities.eval_variables(soc) do v
+        # If not set, we assume x = 0.
```
blegat, line 255: Not sure about setting zero here. If the user sets the `VariablePrimalStart` after `ConstraintPrimalStart`, I prefer an error than incorrect result that would be a nightmare to debug. If it's just for making tests pass, let's just add something to the tests

odow: > I prefer an error

Shouldn't the start values silently skip if unsupported? Also, I don't know if we have any precedence that says variable primal must be set before constraint primal.

odow: This is also a kinda pointless exercise. Because setting the constraint primal for the quadratic does not give enough information to set the constraint primal for the RotatedSecondOrderCone.

And if VariablePrimalStart is not set, then isn't x = 0 a reasonable starting point?

odow: Maybe this is another case for https://github.com/jump-dev/JuMP.jl/pull/3271#issuecomment-1458711074

blegat: > Shouldn't the start values silently skip if unsupported?

If the solver does not support it yes but here it does, it's just the user did not set it correctly so an error is expected so that users fixes his code.

If the `VariablePrimalStart` is not set, I don't think that it's so useful to have this work. If we assume it to be zero, we could as well have the user set it to zero explicitly. At least he's making this sloppy choice.

It didn't work before so having it work in the case variable starting points are set before is already an improvement. If we get complains of users wanting to set it without setting as variable starting value of setting the variable starting value after, we could try to fix that case but I prefer not silently doing weird like having a zero variable starting value.

blegat: If the user sets the constraint start first and then the variable start, the result he will get is completely wrong and will be so awful to debug. Compared to that, an error message is so much better

src/Bridges/Constraint/bridges/quad_to_soc.jl
```diff
@@ -209,13 +215,25 @@ function MOI.delete(model::MOI.ModelLike, bridge::QuadtoSOCBridge)
...
+end
+
 function MOI.get(
     model::MOI.ModelLike,
-    attr::MOI.ConstraintPrimal,
+    attr::Union{MOI.ConstraintPrimal,MOI.ConstraintPrimalStart},
     bridge::QuadtoSOCBridge,
 )
     soc = MOI.get(model, attr, bridge.soc)
+    if soc == nothing
```
matbesancon, line 234:
```suggestion
    if soc === nothing
```
usually the convention for identity equality with nothing

src/Bridges/Constraint/bridges/quad_to_soc.jl
```diff
@@ -261,13 +308,50 @@ end
 # is exactly the same. Q.E.D.
 function MOI.get(
     model::MOI.ModelLike,
-    attr::MOI.ConstraintDual,
+    attr::Union{MOI.ConstraintDual,MOI.ConstraintDualStart},
     bridge::QuadtoSOCBridge,
 )
-    λ = MOI.get(model, attr, bridge.soc)[2]
+    dual = MOI.get(model, attr, bridge.soc)
+    if dual == nothing
```
matbesancon, line 315:
```suggestion
    if dual === nothing
```

src/Bridges/Constraint/bridges/quad_to_soc.jl
blegat: We can even use `isnothing` now that Julia v1.6 is the lower bound

matbesancon: potentially yes, I never saw the interest of that function compared to equality check

blegat: If it was better to do `=== nothing`, I guess the function wouldn't exist but IMO both are fine

src/Bridges/Constraint/bridges/quad_to_soc.jl
```diff
@@ -226,6 +244,48 @@ function MOI.get(
     return output
 end
 
+function _primal_start_or_error(model, attr, v)
+    var_attr = MOI.VariablePrimalStart()
+    value = MOI.get(model, MOI.VariablePrimalStart(), v)
+    if isnothing(value)
```
odow, line 250:
```suggestion
    if value === nothing
```

src/Bridges/Constraint/bridges/quad_to_soc.jl
```diff
@@ -209,15 +216,35 @@ function MOI.delete(model::MOI.ModelLike, bridge::QuadtoSOCBridge)
...
+    # Otherwise, `Q = -U'U` and we have the value of
+    # `Ux` and `a'x - ub`, so we get it with
+    # `-(Ux)'Ux / 2 + (a'x - lb) + ub`
     soc = MOI.get(model, attr, bridge.soc)
+    if soc === nothing
+        return nothing
+    end
     output = sum(soc[i]^2 for i in 3:bridge.dimension)
     output /= 2
+    @show output
```
odow, line 247:
```suggestion
```

src/Bridges/Constraint/bridges/quad_to_soc.jl
```diff
@@ -226,6 +253,69 @@ function MOI.get(
...
+    # which is obtained as
+    # |          1          |
+    # | value - x'Qx/2 - lb |
+    # |        U * x        |
+    # where we compute `x'Qx/2` and `U * x` using the starting values of the variable.
+    soc = MOI.get(model, MOI.ConstraintFunction(), bridge.soc)
+    Ux = MOI.Utilities.eval_variables(MOI.Utilities.eachscalar(soc)[3:end]) do v
+        return _primal_start_or_error(model, attr, v)
+    end
+    @show bridge.set_constant
```
odow, line 299:
```suggestion
```

src/attributes.jl
```diff
@@ -1583,6 +1583,12 @@ A constraint attribute for the initial assignment to some constraint's
 
 May be `nothing` (unset), a number for [`AbstractScalarFunction`](@ref), or a
 vector for [`AbstractVectorFunction`](@ref).
+
+!!! note
+    In order to set the `ConstraintDualStart` for a constraints that is
+    transformed by the [`Bridges.Constraint.QuadtoSOCBridge`](@ref), the
+    [`VariablePrimalStart`](@ref) of all variables involved in the constraints
+    should be set first. Otherwise, an error will be thrown by the bridge.
```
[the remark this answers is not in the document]
odow: If variable starts haven't been provide, the bridge should silently skip them. Erroring is wrong.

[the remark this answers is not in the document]
blegat: We have that in the context of `copy_to`:
https://github.com/jump-dev/MathOptInterface.jl/blob/7e42e0c2bf0b304b1d5d24a06d9ead9c2882d218/src/Utilities/copy.jl#L68-L70
In here, we don't know whether we are in the context of `copy_to` though.

src/attributes.jl
```diff
@@ -1583,6 +1583,12 @@ A constraint attribute for the initial assignment to some constraint's
 
 May be `nothing` (unset), a number for [`AbstractScalarFunction`](@ref), or a
 vector for [`AbstractVectorFunction`](@ref).
+
+!!! note
+    In order to set the `ConstraintDualStart` for a constraints that is
```
matbesancon, line 1588:
```suggestion
    In order to set the `ConstraintDualStart` for a constraint that is
```

src/attributes.jl
```diff
@@ -1594,6 +1600,12 @@ A constraint attribute for the initial assignment to some constraint's
...
```
matbesancon, line 1605:
```suggestion
    In order to set the `ConstraintDualStart` for a constraint that is
```

src/attributes.jl
```diff
@@ -1594,6 +1600,12 @@ A constraint attribute for the initial assignment to some constraint's
...
+    transformed by the [`Bridges.Constraint.QuadtoSOCBridge`](@ref), the
+    [`VariablePrimalStart`](@ref) of all variables involved in the constraints
```
matbesancon, line 1607:
```suggestion
    [`VariablePrimalStart`](@ref) of all variables involved in the constraint
```

blegat:
I believe it's blocked by https://github.com/jump-dev/MathOptInterface.jl/pull/2115#discussion_r1139998271

odow:
Yes, I still think it is wrong to have an exception for a specific bridge in the docstring of `Constraint(Primal|Dual)Start`. Can discuss tomorrow.

odow (approved):

PR #2116, opened 2023-03-09
Test support for VariablePrimalStart in Bridges.Variable

odow:
x-ref #2115 

This is actually a rabbit hole that I'm yet to make my way out of...

### DIFF
--- a/src/Bridges/Variable/bridges/rsoc_soc.jl
+++ b/src/Bridges/Variable/bridges/rsoc_soc.jl
@@ -93,11 +93,3 @@ function MOI.Bridges.inverse_adjoint_map_function(
 )
     return MOI.Bridges.map_function(BT, func)
 end
-
-function MOI.supports(
-    ::MOI.ModelLike,
-    ::MOI.VariablePrimalStart,
-    ::Type{<:RSOCtoSOCBridge},
-)
-    return false
-end
--- a/src/Bridges/Variable/bridges/rsoc_soc.jl
+++ b/src/Bridges/Variable/bridges/rsoc_soc.jl
@@ -93,3 +93,11 @@ function MOI.Bridges.inverse_adjoint_map_function(
 )
     return MOI.Bridges.map_function(BT, func)
 end
+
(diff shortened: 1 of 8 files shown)


src/Bridges/Variable/bridges/rsoc_soc.jl
```diff
@@ -93,3 +93,11 @@ function MOI.Bridges.inverse_adjoint_map_function(
...
     return MOI.Bridges.map_function(BT, func)
 end
+
+function MOI.supports(
+    ::MOI.ModelLike,
+    ::MOI.VariablePrimalStart,
+    ::Type{<:RSOCtoSOCBridge},
+)
+    return false
+end
```
odow, line 103: I've opted-out here, since setting or getting the start can't be done in isolation.

blegat: Yes, this is part of the optional API. By the way, this should be default as in
https://github.com/jump-dev/MathOptInterface.jl/blob/5025b50de1d6cb29e6bbb0d26adfcf1b57b6a51c/src/Bridges/bridge.jl#L105-L149

blegat: Bridges should have to implement it to return `false`

src/Bridges/bridge_optimizer.jl
```diff
@@ -1149,7 +1149,12 @@ function MOI.supports(
     attr::MOI.AbstractVariableAttribute,
     ::Type{MOI.VariableIndex},
 )
-    return MOI.supports(b.model, attr, MOI.VariableIndex)
+    if !MOI.supports(b.model, attr, MOI.VariableIndex)
```
blegat, line 1152: We `supports`, we prefer false positive than false negative. Previously, there were false positive but here, there might be false negative in case not all variables are set.

odow:
Happy?

blegat:
Yes, I just wanted to let you take a look

PR #2119, opened 2023-03-13
Add NormCone for representing the epigraph of a p-norm

### LINKED ISSUE #2118: Add NormPCone
We have `NormOneCone` and `NormInfCone` (and, I guess, `SecondOrderCone`). We should add a `NormPCone`:

x-ref: https://github.com/jump-dev/JuMP.jl/pull/3282#issuecomment-1465442016

See also https://discourse.julialang.org/t/nonlinear-optimization-using-jump-the-solver-does-not-support-nonlinear-problems-i-e-nlobjective-and-nlconstraint/95978

We should add a `NormPToPowerBridge`, but we could also add a `NormPToSecondOrderConeBridge`, following the rational norm atom in Convex.jl:

https://github.com/jump-dev/Convex.jl/blob/c6a21df2979dc1eaf60674f7f4b1324a4f30f8b8/src/atoms/second_order_cone/rationalnorm.jl#L1-L130

I think it would error if it got bridged and the `p` wasn't rational.

odow:
Closes #2118 

The bike shed is open for naming. 

`PNormCone` doesn't match `NormOneCone`, `NormInfinityCone`, `NormSpectralCone`, and `NormPCone` seemed superfluous, so I went with `NormCone`. Happy to change back to `NormPCone` though.

I haven't added the rational norm bridge yet.


src/Bridges/Constraint/bridges/norm_special_case.jl
```diff
@@ -0,0 +1,123 @@
...
+# in the LICENSE.md file or at https://opensource.org/licenses/MIT.
+
+"""
+    NormSpecialCaseBridge{T,F} <: Bridges.Constraint.AbstractBridge
+
+`NormSpecialCaseBridge` implements the following reformulations:
+
+  * ``(t, x) in NormOneCone(d)`` into ``(t, x) in NormCone(1, d)``
+  * ``(t, x) in SecondOrderCone(d)`` into ``(t, x) in NormCone(2, d)``
+  * ``(t, x) in NormInfinityCone(d)`` into ``(t, x) in NormCone(Inf, d)``
```
odow, line 14: I've added this bridge for now, mainly to test out, because I'm not sure if this would be used by an actual solver. (I can't see a solver supporting `NormCone` or `PowerCone` but not `SecondOrderCone` or linear constraints.)

The other option I considered was the reverse. `NormCone(1, d)` creates `NormOneCone(d)`, etc.

That's a bit weird though, because the `NormCone` constructor wouldn't necessarily return a `::NormCone`. And some solver might support a general `NormCone` but not the specialized `NormOne` and `NormInfinity` cones.

It's a little messy.

blegat: Yes, I wouldn't have the constructor do that. At least not at the MOI level.

src/Bridges/Constraint/bridges/norm_to_power.jl
```diff
@@ -0,0 +1,160 @@
...
+
+`NormToPowerBridge` supports:
+
+  * `F` in [`MOI.NormCone`](@ref)
+
+## Target nodes
+
+`NormToPowerBridge` creates:
+
+  * `F` in [`MOI.PowerCone{T}`](@ref)
```
blegat, line 25: Maybe we can cite the same book as https://github.com/jump-dev/Convex.jl/blob/c6a21df2979dc1eaf60674f7f4b1324a4f30f8b8/src/atoms/second_order_cone/rationalnorm.jl#L1-L130

odow: Yes, I guess I need to add a rational norm bridge. But the implementation is always a little messy.

src/Bridges/Constraint/bridges/norm_to_power.jl
```diff
@@ -0,0 +1,160 @@
...
+    ::Type{<:MOI.AbstractVectorFunction},
+    ::Type{MOI.NormCone},
+) where {T}
+    return true
+end
+
+function MOI.Bridges.added_constrained_variable_types(
+    ::Type{<:NormToPowerBridge},
+)
+    return Tuple{Type}[]
```
odow, line 74:
```suggestion
    return Tuple{Type}[(MOI.Reals,)]
```

src/Bridges/Constraint/bridges/norm_to_power.jl
```diff
@@ -0,0 +1,160 @@
...
+#
+# Use of this source code is governed by an MIT-style license that can be found
+# in the LICENSE.md file or at https://opensource.org/licenses/MIT.
+
+"""
+    NormToPowerBridge{T,F} <: Bridges.Constraint.AbstractBridge
+
+`NormToPowerBridge` implements the following reformulation:
+
+  * ``(t, x...) \\in NormCone(p, 1+d)`` into ``(r_i, t, x_i) \\in PowerCone(1 / p)``
```
odow, line 12:
```suggestion
  * ``(t, x) \\in NormCone(p, 1+d)`` into ``(r_i, t, x_i) \\in PowerCone(1 / p)``
```

odow:
We should merge https://github.com/jump-dev/MathOptInterface.jl/pull/2128 and then rebase this, because a few things will need to change.

src/Bridges/Constraint/bridges/norm_special_case.jl
```diff
@@ -0,0 +1,123 @@
...
+  * `F` in [`MOI.NormInfinityCone`](@ref)
+  * `F` in [`MOI.SecondOrderCone`](@ref)
+
+## Target nodes
+
+`NormSpecialCaseBridge` creates:
+
+  * `F` in [`MOI.NormCone`](@ref)
+"""
+struct NormSpecialCaseBridge{T,F} <: AbstractBridge
```
blegat, line 30: It would be easier to implement it as a `SetMapBridge`

odow: Refactored in to SetMapBridge

src/Bridges/Constraint/bridges/norm_special_case.jl
```diff
@@ -0,0 +1,144 @@
...
+
+  * `F` in [`MOI.NormCone`](@ref)
+"""
+const NormInfinityConeToNormConeBridge{T,F} =
+    AbstractNormSpecialCaseBridge{T,MOI.NormInfinityCone,F}
+
+const NormInfinityConeToNormCone{T,OT<:MOI.ModelLike} =
+    SingleBridgeOptimizer{NormInfinityConeToNormConeBridge{T},OT}
+
+function concrete_bridge_type(
```
blegat, line 138: This method does not have to be defined for each alias, it can be defined just one for `AbstractNormSpecialCaseBridge`

src/Bridges/Constraint/bridges/norm_special_case.jl
```diff
@@ -0,0 +1,144 @@
+# Copyright (c) 2017: Miles Lubin and contributors
+# Copyright (c) 2017: Google Inc.
+#
+# Use of this source code is governed by an MIT-style license that can be found
+# in the LICENSE.md file or at https://opensource.org/licenses/MIT.
+
+struct AbstractNormSpecialCaseBridge{T,S,F} <:
```
blegat, line 7: It's confusing to name if `Abstract...` if it's not an abstract type

odow:
Updated.

blegat (approved):

PR #2120, opened 2023-03-14
[Nonlinear] fix splatting when collection does not support reverse

odow:
A common example of such a type is a DenseAxisArray.

Raised on Discourse: https://discourse.julialang.org/t/issue-with-memoization-with-jump-for-nlp-when-adding-nlobjective/96033

### DIFF
--- a/src/Nonlinear/parse.jl
+++ b/src/Nonlinear/parse.jl
@@ -87,11 +87,8 @@ function _parse_splat_expression(stack, data, expr, x, parent_index)
             "`(x + 1)...`, `[x; y]...` and `g(f(y)...)` are not.",
         )
     end
-    # It isn't sufficient to call `reverse` here, because the collection might
-    # not support it. Iterate over the reverse of the collected indices instead,
-    # which _does_ work in all(?) cases.
-    for i in reverse(collect(eachindex(x.args[1])))
-        push!(stack, (parent_index, x.args[1][i]))
+    for arg in Iterators.Reverse(x.args[1])
+        push!(stack, (parent_index, arg))
     end
     return
 end
--- a/src/Nonlinear/parse.jl
+++ b/src/Nonlinear/parse.jl
@@ -87,8 +87,11 @@ function _parse_splat_expression(stack, data, expr, x, parent_index)
             "`(x + 1)...`, `[x; y]...` and `g(f(y)...)` are not.",
         )
     end
-    for xi in reverse(x.args[1])
-        push!(stack, (parent_index, xi))
+    # It isn't sufficient to call `reverse` here, because the collection might
+    # not support it. Iterate over the reverse of the collected indices instead,
+    # which _does_ work in all(?) cases.
+    for i in reverse(collect(eachindex(x.args[1])))
+        push!(stack, (parent_index, x.args[1][i]))
     end
     return
 end
--- a/test/Nonlinear/Nonlinear.jl
+++ b/test/Nonlinear/Nonlinear.jl
@@ -973,6 +973,22 @@ function test_pow_complex_result()
     return
 end
 
+struct _NoReverse{T} <: AbstractArray{T,1}
+    data::Vector{T}
+end
+
+Base.eachindex(x::_NoReverse) = eachindex(x.data)
+Base.getindex(x::_NoReverse, args...) = getindex(x.data, args...)
+
+function test_parse_splat_no_reverse()
+    model = Nonlinear.Model()
+    x = MOI.VariableIndex.(1:2)
+    y = _NoReverse(x)
+    expr = Nonlinear.parse_expression(model, :(+($y...)))
+    @test expr == Nonlinear.parse_expression(model, :(+($(x[1]), $(x[2]))))
+    return
+end
+
 end
 
 TestNonlinear.runtests()


src/Nonlinear/parse.jl
```diff
@@ -87,8 +87,11 @@ function _parse_splat_expression(stack, data, expr, x, parent_index)
             "`(x + 1)...`, `[x; y]...` and `g(f(y)...)` are not.",
         )
     end
-    for xi in reverse(x.args[1])
-        push!(stack, (parent_index, xi))
+    # It isn't sufficient to call `reverse` here, because the collection might
+    # not support it. Iterate over the reverse of the collected indices instead,
+    # which _does_ work in all(?) cases.
+    for i in reverse(collect(eachindex(x.args[1])))
```
blegat, line 93: That starts getting hacky, wouldn't `Base.Iterators.Reverse` work ?
https://github.com/JuliaLang/julia/blob/17cfb8e65ead377bf1b4598d8a9869144142c84e/base/iterators.jl#L70-L94
It is using `reverse(eachindex(...))` which is almost what you are doing but I'm not sure that the collect is necessary.

odow: I think I'd prefer to leave it as this. It's more explicit as to what is going on. And the collect is needed. Consider something like:
```Julia
julia> d = Dict(1 => 2)
kDict{Int64, Int64} with 1 entry:
  1 => 2

julia> keys(d)
KeySet for a Dict{Int64, Int64} with 1 entry. Keys:
  1

julia> reverse(keys(d))
ERROR: MethodError: no method matching reverse(::Base.KeySet{Int64, Dict{Int64, Int64}})
Closest candidates are:
  reverse(::Union{SubString{String}, String}) at strings/substring.jl:166
  reverse(::CartesianIndices) at multidimensional.jl:490
  reverse(::ChainRulesCore.AbstractZero, ::Any...; kwargs...) at /Users/oscar/.julia/packages/ChainRulesCore/C73ay/src/tangent_types/abstract_zero.jl:42
  ...
Stacktrace:
 [1] top-level scope
   @ REPL[4]:1

julia> reverse(eachindex(d))
ERROR: MethodError: no method matching reverse(::Base.KeySet{Int64, Dict{Int64, Int64}})
Closest candidates are:
  reverse(::Union{SubString{String}, String}) at strings/substring.jl:166
  reverse(::CartesianIndices) at multidimensional.jl:490
  reverse(::ChainRulesCore.AbstractZero, ::Any...; kwargs...) at /Users/oscar/.julia/packages/ChainRulesCore/C73ay/src/tangent_types/abstract_zero.jl:42
  ...
Stacktrace:
 [1] top-level scope
   @ REPL[5]:1

julia> reverse(collect(eachindex(d)))
1-element Vector{Int64}:
 1

julia> Iterators.Reverse(eachindex(d))
Base.Iterators.Reverse{Base.KeySet{Int64, Dict{Int64, Int64}}}([1])

julia> collect(Iterators.Reverse(eachindex(d)))
ERROR: MethodError: no method matching iterate(::Base.Iterators.Reverse{Base.KeySet{Int64, Dict{Int64, Int64}}})
Closest candidates are:
  iterate(::Union{LinRange, StepRangeLen}) at range.jl:664
  iterate(::Union{LinRange, StepRangeLen}, ::Int64) at range.jl:664
  iterate(::T) where T<:Union{Base.KeySet{var"#s77", var"#s76"} where {var"#s77", var"#s76"<:Dict}, Base.ValueIterator{var"#s75"} where var"#s75"<:Dict} at dict.jl:693
  ...
Stacktrace:
 [1] copyto!(dest::Vector{Int64}, src::Base.Iterators.Reverse{Base.KeySet{Int64, Dict{Int64, Int64}}})
   @ Base ./abstractarray.jl:843
 [2] _collect(cont::UnitRange{Int64}, itr::Base.Iterators.Reverse{Base.KeySet{Int64, Dict{Int64, Int64}}}, #unused#::Base.HasEltype, isz::Base.HasLength)
   @ Base ./array.jl:608
 [3] collect(itr::Base.Iterators.Reverse{Base.KeySet{Int64, Dict{Int64, Int64}}})
   @ Base ./array.jl:602
 [4] top-level scope
   @ REPL[8]:1
```

It'd be weird to splat a dictionary, but it wouldn't be weird to splat an ordered dictionary, or some other type that returns an iterator for `eachindex` that doesn't implement `reverse`.

blegat: For dictionaries, it wouldn't do the right thing:
```julia
julia> d = Dict(1 => 2)
Dict{Int64, Int64} with 1 entry:
  1 => 2

julia> tuple(d...)
(1 => 2,)

julia> [getindex(d, i) for i in collect(eachindex(d))]
1-element Vector{Int64}:
 2
```
That's why I feel, it's getting hacky. At least with `Iterators.Reverse`, if there is an issue with a custom array, we can reply that it does not work with the standard `Iterators.Reverse` so it's not on JuMP/MOI

odow: Reverse wouldn't work for `DenseAxisArray`: https://github.com/JuliaLang/julia/blob/17cfb8e65ead377bf1b4598d8a9869144142c84e/base/iterators.jl#L107-L108

And dictionaries don't work at the moment, because:
```julia
julia> d = Dict(1 => 2)
Dict{Int64, Int64} with 1 entry:
  1 => 2

julia> reverse(d)
ERROR: MethodError: no method matching reverse(::Dict{Int64, Int64})
Closest candidates are:
  reverse(::Union{SubString{String}, String}) at strings/substring.jl:166
  reverse(::CartesianIndices) at multidimensional.jl:490
  reverse(::ChainRulesCore.AbstractZero, ::Any...; kwargs...) at /Users/oscar/.julia/packages/ChainRulesCore/a4mIA/src/tangent_types/abstract_zero.jl:42
  ...
Stacktrace:
```

This is strictly a bug-fix for currently broken code in JuMP. It doesn't solve the general case (and I don't know if we can).

odow:
Objections to merging? This is a bug fix.

src/Nonlinear/parse.jl
blegat: To clarify: At the moment, dictionaries give an error which is fine. With `Iterators.Revers`, it still gives an error which is fine. With your PR, it may give an incorrect result which is an issue.

odow:
Okay, I've updated

PR #2123, opened 2023-03-16
Add Bridges.Constraint.HermitiantoSymmetricPSDBridge by default

blegat:
Closes https://github.com/jump-dev/MathOptInterface.jl/issues/2124


odow:
Ah, yes, seems like we missed it here: https://github.com/jump-dev/MathOptInterface.jl/pull/1977

odow (approved):

PR #2128, opened 2023-03-24
[Bridges] change print_active_bridges to use types instead of runtime objects

odow:
Working on https://github.com/jump-dev/JuMP.jl/issues/3301

But it required a little more work than I initially expected. 

To test out different functions and sets, we need to do it with the type information only. 

This mostly works, except that some bridges _may_ add some constraints depending on runtime values, whereas we must now report the union of all possible things added. This could be confusing, because it may say some things are added when the real model won't have them, but it clarifies that a solver must support all possible runtime constraints to use a bridge.

The main example of this is the objective slack bridge, which adds a `>=` or `<=` constraint, depending on the objective sense.

We also need bridges to report that they add `x in Reals` if they add variables.

TODO:

 - [x] more tests

### DIFF
--- a/src/Bridges/Constraint/bridges/det.jl
+++ b/src/Bridges/Constraint/bridges/det.jl
@@ -144,7 +144,7 @@ function MOI.supports_constraint(
 end
 
 function MOI.Bridges.added_constrained_variable_types(::Type{<:LogDetBridge})
-    return Tuple{Type}[]
+    return Tuple{Type}[(MOI.Reals,)]
 end
 
 function MOI.Bridges.added_constraint_types(
--- a/src/Bridges/Constraint/bridges/det.jl
+++ b/src/Bridges/Constraint/bridges/det.jl
@@ -366,7 +366,7 @@ function MOI.supports_constraint(
 end
 
 function MOI.Bridges.added_constrained_variable_types(::Type{<:RootDetBridge})
-    return Tuple{Type}[]
+    return Tuple{Type}[(MOI.Reals,)]
 end
 
 function MOI.Bridges.added_constraint_types(
--- a/src/Bridges/debug.jl
+++ b/src/Bridges/debug.jl
@@ -623,7 +623,8 @@ end
     print_active_bridges(
         [io::IO=stdout,]
         b::MOI.Bridges.LazyBridgeOptimizer,
-        F::Type{<:MOI.AbstractFunction}
+        F::Type{<:MOI.AbstractFunction},
+        S::Type{<:MOI.AbstractSet},
     )
 
 Print the set of bridges required for a constraint of type `F`-in-`S`.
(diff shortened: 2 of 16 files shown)


test/Bridges/lazy_bridge_optimizer.jl
```diff
@@ -993,6 +996,8 @@ Objective function of type `MOI.ScalarQuadraticFunction{$T}` is not supported an
               MOI.Utilities.replace_acronym(
             """
 Objective function of type `MOI.ScalarQuadraticFunction{$T}` is not supported and cannot be bridged into a supported objective function by adding only supported constrained variables and constraints. See details below:
+ [1] constrained variables in `MOI.Reals` are not supported because no added bridge supports bridging it.
+   Cannot add free variables and then constrain them because free variables are bridged but no functionize bridge was added.
```
blegat, line 1000: This is wrong, you cannot bridge free variables with the `Constraint.Functionize` bridge

src/Bridges/Constraint/bridges/det.jl
```diff
@@ -144,7 +144,7 @@ function MOI.supports_constraint(
 end
 
 function MOI.Bridges.added_constrained_variable_types(::Type{<:LogDetBridge})
-    return Tuple{Type}[]
+    return Tuple{Type}[(MOI.Reals,)]
```
blegat, line 147: Were any bridge already returning `MOI.Reals` here ?

blegat:
> We also need bridges to report that they add `x in Reals` if they add variables.

Indeed, I don't remember why it was not included.

src/Bridges/Constraint/bridges/det.jl
odow: Nope. I think it was just an oversight.

test/Bridges/lazy_bridge_optimizer.jl
odow: Hmm yeah. I haven't changed the error messages. This is just what was printed. I think I have a much better handle on the graph system now. It was a pretty easy mistake to omit these from being printed when we wrote this debugger!

odow:
> We should clarify it in the docstring

Clarified.

blegat (approved):

src/Bridges/debug.jl
```diff
@@ -561,119 +564,171 @@ function _print_unsupported(io, arg...)
...
+    _print_bridge(io, b, BT, offset)
     return
 end
 
-function _print_constraint_tree(io, b, F, S, offset)
+"""
+    print_active_bridges(
+        [io::IO=stdout,]
+        b::MOI.Bridges.LazyBridgeOptimizer,
+        F::Type{<:MOI.AbstractFunction}
```
odow, line 626:
```suggestion
        F::Type{<:MOI.AbstractFunction},
        S::Type{<:MOI.AbstractSet},
```

PR #2130, opened 2023-03-30
[FileFormats.NL] make ResultCount more permissive in when it returns 1

odow:
Closes https://github.com/jump-dev/AmplNLWriter.jl/issues/166

### DIFF
--- a/test/FileFormats/NL/sol.jl
+++ b/test/FileFormats/NL/sol.jl
@@ -279,6 +279,8 @@ function test_statuses()
             (MOI.DUAL_INFEASIBLE, MOI.UNKNOWN_RESULT_STATUS),
         (400, "norm limit exceeded") =>
             (MOI.OTHER_LIMIT, MOI.UNKNOWN_RESULT_STATUS),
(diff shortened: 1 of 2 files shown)


test/FileFormats/NL/sol.jl
```diff
@@ -307,6 +309,12 @@ function test_sol_empty()
     return
 end
 
+function test_result_count()
+    results = SolFileResults("raw", MOI.OTHER_LIMIT)
```
odow, line 313:
```suggestion
    results = NL.SolFileResults("raw", MOI.OTHER_LIMIT)
```

test/FileFormats/NL/sol.jl
```diff
@@ -307,6 +309,12 @@ function test_sol_empty()
...
+    @test MOI.get(results, MOI.ResultCount()) == 0
+    results.variable_primal[MOI.VariableIndex(1)] = 0.0
+    @test MOI.get(results, MOI.ResultCount()) == 1
+    return
```
odow, line 317:
```suggestion
    return
end

```

PR #2132, opened 2023-04-01
Prep for v1.14.0

odow:
Checking https://github.com/jump-dev/SolverTests/actions/runs/4602400287

### DIFF
--- a/docs/src/changelog.md
+++ b/docs/src/changelog.md
@@ -7,7 +7,7 @@ CurrentModule = MathOptInterface
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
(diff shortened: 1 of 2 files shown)


odow:
So this release will break the JuMP tests: https://github.com/jump-dev/SolverTests/actions/runs/4602400287/jobs/8131315687

But only the printing of `print_active_bridges`. We probably shouldn't rely on the exact text in the JuMP tests.

docs/src/changelog.md
```diff
@@ -7,6 +7,22 @@ CurrentModule = MathOptInterface
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## v1.14.0 (April 2, 2023)
```
odow, line 10:
```suggestion
## v1.14.0 (April 4, 2023)
```

PR #2133, opened 2023-04-03
Add test for infinite bounds on variable

blegat:
That would have caught https://github.com/jump-dev/MosekTools.jl/issues/122

### DIFF
--- a/src/Test/test_linear.jl
+++ b/src/Test/test_linear.jl
@@ -4078,11 +4078,7 @@ function test_linear_variable_open_intervals(
     config::Config{T},
 ) where {T}
     @requires _supports(config, MOI.optimize!)
-    @requires MOI.supports_constraint(
-        model,
-        MOI.VariableIndex,
-        MOI.Interval{T},
-    )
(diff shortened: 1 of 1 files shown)


src/Test/test_linear.jl
```diff
@@ -4064,6 +4064,47 @@ end
...
+
+Test that the solver supports open-intervals like `(-Inf, u]`, `[l, Inf)`, and
+`(-Inf, Inf)`.
+"""
+function test_linear_variable_open_intervals(
+    model::MOI.ModelLike,
+    config::Config{T},
+) where {T}
+    @requires _supports(config, MOI.optimize!)
+    @requires MOI.supports_constraint(model, MOI.VariableIndex, MOI.Interval{T})
```
odow, line 4081: Wasn't there an issue with this and solvers which bridged via SplitInterval? I forget where we left things with that.

odow: We updated splitinterval to not pass on the infinite bounds.

PR #2135, opened 2023-04-04
[Bridges] fix bug in print_active_bridges

odow:
I missed an edge case: if a variable is unsupported and is bridged by a constraint instead.

x-ref https://github.com/jump-dev/JuMP.jl/issues/3301

### DIFF
--- a/test/Bridges/debug.jl
+++ b/test/Bridges/debug.jl
@@ -290,7 +290,8 @@ MOI.Utilities.@model(
 )
 
 function test_print_active_bridges_variable_bridged_with_constraint()
-    model = MOI.Bridges.full_bridge_optimizer(ConstraintModel{Float64}(), Float64)
+    model =
+        MOI.Bridges.full_bridge_optimizer(ConstraintModel{Float64}(), Float64)
     content = """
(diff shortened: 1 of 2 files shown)


test/Bridges/debug.jl
```diff
@@ -277,6 +277,42 @@ function test_print_active_bridges_variable_unsupported()
...
+    (),
+    (),
+    (),
+    (MOI.ScalarAffineFunction,),
+    (),
+    ()
+)
+
+function test_print_active_bridges_variable_bridged_with_constraint()
+    model = MOI.Bridges.full_bridge_optimizer(ConstraintModel{Float64}(), Float64)
```
odow, line 293:
```suggestion
    model =
        MOI.Bridges.full_bridge_optimizer(ConstraintModel{Float64}(), Float64)
```

blegat:
Yes, that's always a tricky one. There is also the case where it is **not** added with `add_constrained_variables`  so you'll have to apply the constraint bridge even if it's not the shortest path but that doesn't apply to this printing function.

blegat (approved):

PR #2147, opened 2023-04-17
[Bridges] fix deleting variable in bridged objective

odow:
This is a bit of a weird case: if you delete a variable that was an objective, does the type of the objective change, or do we remove one of the objective rows?

I obviously thought so when writing the tests

https://github.com/jump-dev/MathOptInterface.jl/blob/403bc583628682ea6f11b4343fa91b621f2862cf/src/Test/test_multiobjective.jl#L21-L35

but I missed the case that this could also be bridged.

### DIFF
--- a/src/Bridges/bridge_optimizer.jl
+++ b/src/Bridges/bridge_optimizer.jl
@@ -587,7 +587,7 @@ function MOI.delete(b::AbstractBridgeOptimizer, vi::MOI.VariableIndex)
     F = MOI.get(b, MOI.ObjectiveFunctionType())
     if is_objective_bridged(b) && F == MOI.VectorOfVariables
         f = MOI.get(b, MOI.ObjectiveFunction{MOI.VectorOfVariables}())
-        if any(isequal(v1), f.variables)
+        if any(isequal(vi), f.variables)
             g = MOI.VectorOfVariables(filter(!isequal(vi), f.variables))
             MOI.set(b, MOI.ObjectiveFunction{MOI.VectorOfVariables}(), g)
         end
--- a/src/Bridges/bridge_optimizer.jl
+++ b/src/Bridges/bridge_optimizer.jl
@@ -546,11 +546,14 @@ function _delete_variables_in_variables_constraints(
 end
 
 function MOI.delete(b::AbstractBridgeOptimizer, vis::Vector{MOI.VariableIndex})
-    if is_objective_bridged(b)
-        F = MOI.get(b, MOI.ObjectiveFunctionType())
-        f = MOI.get(b, MOI.ObjectiveFunction{F}())
-        g = MOI.Utilities.filter_variables(v -> !Base.Fix2(in, vis), f)
-        MOI.set(b, MOI.ObjectiveFunction{typeof(g)}(), g)
+    F = MOI.get(b, MOI.ObjectiveFunctionType())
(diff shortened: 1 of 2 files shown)


blegat:
> This is a bit of a weird case: if you delete a variable that was an objective, does the type of the objective change, or do we remove one of the objective rows?

You remove the row, that's consistent with what happens in a `VectorOfVariables`-in-`Nonnegatives` constraint

src/Bridges/bridge_optimizer.jl
```diff
@@ -546,6 +546,15 @@ function _delete_variables_in_variables_constraints(
 end
 
 function MOI.delete(b::AbstractBridgeOptimizer, vis::Vector{MOI.VariableIndex})
+    F = MOI.get(b, MOI.ObjectiveFunctionType())
+    if is_objective_bridged(b) && F == MOI.VectorOfVariables
+        f = MOI.get(b, MOI.ObjectiveFunction{MOI.VectorOfVariables}())
+        discard = Base.Fix2(in, vis)
+        if any(discard, f.variables)
+            g = MOI.VectorOfVariables([x for x in f.variables if !discard(x)])
```
blegat, line 554:
```suggestion
            g = MOI.VectorOfVariables(filter(!discard, f.variables))
```

src/Bridges/bridge_optimizer.jl
```diff
@@ -575,6 +584,14 @@ function MOI.delete(b::AbstractBridgeOptimizer, vis::Vector{MOI.VariableIndex})
 end
 
 function MOI.delete(b::AbstractBridgeOptimizer, vi::MOI.VariableIndex)
+    F = MOI.get(b, MOI.ObjectiveFunctionType())
+    if is_objective_bridged(b) && F == MOI.VectorOfVariables
+        f = MOI.get(b, MOI.ObjectiveFunction{MOI.VectorOfVariables}())
+        if any(Base.Fix1(==, vi), f.variables)
```
blegat, line 590:
```suggestion
        if any(isequal(v1), f.variables)
```

src/Bridges/bridge_optimizer.jl
```diff
@@ -575,6 +584,14 @@ function MOI.delete(b::AbstractBridgeOptimizer, vis::Vector{MOI.VariableIndex})
...
+            g = MOI.VectorOfVariables([x for x in f.variables if x != vi])
```
blegat, line 591:
```suggestion
            g = MOI.VectorOfVariables(filter(!isequal(vi), f.variables))
```

blegat:
Maybe create a utility function that is called with `discard` and `isequal(vi)` ?

src/Bridges/bridge_optimizer.jl
```diff
@@ -575,6 +584,14 @@ function MOI.delete(b::AbstractBridgeOptimizer, vis::Vector{MOI.VariableIndex})
 end
 
 function MOI.delete(b::AbstractBridgeOptimizer, vi::MOI.VariableIndex)
+    F = MOI.get(b, MOI.ObjectiveFunctionType())
+    if is_objective_bridged(b) && F == MOI.VectorOfVariables
+        f = MOI.get(b, MOI.ObjectiveFunction{MOI.VectorOfVariables}())
+        if any(isequal(v1), f.variables)
```
odow, line 590:
```suggestion
        if any(isequal(vi), f.variables)
```

PR #2149, opened 2023-04-17
Add solver-tests.yml GitHub action

odow:
This is a workflow_dispatch only action (for now) that tests most downstream solvers.

The intention is that changes in MOI can be tested across solvers on-demand by running the action from a chosen branch.

For now this is a workflow_dispatch only, and I've commented out the commercial solvers so we don't need to add their keys to this repo's secrets. Can fix in future.

### DIFF
--- a/.github/workflows/solver-tests.yml
+++ b/.github/workflows/solver-tests.yml
@@ -1,0 +1,106 @@
+name: solver-tests
+on:
+  workflow_dispatch:
+jobs:
+  test-ubuntu:
+    name: ${{ matrix.package }}
+    runs-on: ubuntu-latest
+    env:
+      PACKAGE: ${{ matrix.package }}
+    strategy:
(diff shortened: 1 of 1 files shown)


joaquimg:
Maybe commercial ones should even be in a separate workflow...

.github/workflows/solver-tests.yml
```diff
@@ -0,0 +1,106 @@
...
+          - package: 'PATHSolver'
+          - package: 'Pavito'
+          - package: 'ProxSDP'
+          - package: 'SCIP'
+          - package: 'SCS'
+          - package: 'SDPA'
+          - package: 'SDPAFamily'
+          - package: 'Tulip'
+    steps:
+      - uses: actions/checkout@v2
```
blegat, line 45:
```suggestion
      - uses: actions/checkout@v3
```

PR #2154, opened 2023-04-25
Add ScaledPositiveSemidefiniteConeTriangle

blegat:
Closes https://github.com/jump-dev/MathOptInterface.jl/issues/531

Checklist from https://github.com/jump-dev/MathOptInterface.jl/pull/2167

## Basic

 - [x] Implement `Base.copy(set::S)`
 - [x] Implement `Base.:(==)(x::S, y::S)`
 - [x] If an `AbstractVectorSet`, implement `dimension(set::S)`

## Utilities

 - [x] If an `AbstractVectorSet`, implement `Utilities.set_dot`,
       `LinearAlgebra.dot`
 - [x] If an `AbstractVectorSet`, implement `Utilities.set_with_dimension` in

## Documentation

## Tests

## MathOptFormat

 - [x] Open an issue at `https://github.com/jump-dev/MathOptFormat` to add
       support for the new set https://github.com/jump-dev/MathOptFormat/issues/27

## Optional

 - [x] Implement `dual_set(::S)` and `dual_set_type(::Type{S})`
 - [x] Add new bridges to convert your set into more commonly used sets


src/sets.jl
```diff
@@ -1091,6 +1091,38 @@ dual_set_type(::Type{NormNuclearCone}) = NormSpectralCone
...
+
+Compared to the [`PositiveSemidefiniteConeTriangle`](@ref), the off-diagonal entries
+are scaled by `√2`. Thanks to this scaling, [`MOI.Utilities.set_dot`](@ref) is the
+simply the sum of the pairwise product while for
+[`PositiveSemidefiniteConeTriangle`](@ref), the pairwise product additionally have
+to be multiplied by 2.
+"""
+struct ScaledPositiveSemidefiniteConeTriangle <: AbstractVectorSet
+    side_dimension::Int
+    function ScaledPositiveSemidefiniteConeTriangle(side_dimension::Base.Integer)
```
odow, line 1108:
```suggestion
    function ScaledPositiveSemidefiniteConeTriangle(
        side_dimension::Base.Integer,
    )
```

src/sets.jl
```diff
@@ -1091,6 +1091,38 @@ dual_set_type(::Type{NormNuclearCone}) = NormSpectralCone
...
+                    "Side dimension of PositiveSemidefiniteConeTriangle must " *
+                    "be >= 0, not $(side_dimension).",
+                ),
+            )
+        end
+        return new(side_dimension)
+    end
+end
+
+dual_set(s::ScaledPositiveSemidefiniteConeTriangle) = copy(s)
```
odow, line 1121:
```suggestion
dual_set(s::ScaledPositiveSemidefiniteConeTriangle) = copy(s)

```

src/sets.jl
```diff
@@ -1188,7 +1220,12 @@ products we have
 """
 abstract type AbstractSymmetricMatrixSetTriangle <: AbstractVectorSet end
 
-function dimension(set::AbstractSymmetricMatrixSetTriangle)
+function dimension(
+    set::Union{
+        AbstractSymmetricMatrixSetTriangle,
+        ScaledPositiveSemidefiniteConeTriangle,
```
odow, line 1226: Shouldn't `ScaledPositiveSemidefiniteConeTriangle` be `<:AbstractSymmetricMatrixSetTriangle` instead of `<:AbstractVectorSet`?

Or am I forgetting some triviality.

odow (commented):
Before merging, we should:

 * Add to docs
 * Add tests to `MOI.Test`
 * Add a bridge for PSDCone -> ScaledPSDCone

src/Test/test_conic.jl
```diff
@@ -4750,9 +4750,11 @@ function _test_conic_PositiveSemidefiniteCone_helper(
     psdcone,
     config::Config{T},
 ) where {T}
+    @show psdcone
```
odow, line 4753: To be removed

src/Test/test_conic.jl
```diff
@@ -4773,18 +4775,27 @@ function _test_conic_PositiveSemidefiniteCone_helper(
         MOI.ScalarAffineFunction{T},
         MOI.EqualTo{T},
     )
+    println("GO")
```
odow, line 4778: TODO

test/Bridges/Constraint/symmetric_matrix_scaling.jl
```diff
@@ -0,0 +1,50 @@
...
+        if startswith("$(name)", "test_")
+            @testset "$(name)" begin
+                getfield(@__MODULE__, name)()
+            end
+        end
+    end
+    return
+end
+
+include("../utilities.jl")
```
odow, line 24: You don't need `utilities` for this.

test/Bridges/Constraint/symmetric_matrix_scaling.jl
```diff
@@ -0,0 +1,50 @@
...
+ """
+    not_scaled = """
+variables: x, y, z
+[x, 1.0 * y, z] in PositiveSemidefiniteConeTriangle(2)
+ """
+    MOI.Bridges.runtests(
+        MOI.Bridges.Constraint.SymmetricMatrixScalingBridge,
+        not_scaled,
+        scaled,
+    )
```
odow, line 39: I think I'd prefer two explicit tests for this, rather than a single function. The bridges are separate.

odow (commented):
We also need to add this to `test_basic_constraints`:
https://github.com/jump-dev/MathOptInterface.jl/blob/d5c356742031438eb238264d9daa7e86f2ee103d/src/Test/test_basic_constraint.jl#L101-L103

https://github.com/jump-dev/MathOptInterface.jl/blob/d5c356742031438eb238264d9daa7e86f2ee103d/src/Test/test_basic_constraint.jl#L289

src/Test/test_conic.jl
```diff
@@ -4967,6 +4982,41 @@ function setup_test(
...
+) where {T}
+    MOIU.set_mock_optimize!(
+        model,
+        (mock::MOIU.MockOptimizer) -> MOIU.mock_optimize!(
+            mock,
+            ones(T, 3),
+            (
+                MOI.VectorAffineFunction{T},
+                MOI.ScaledPositiveSemidefiniteConeTriangle,
+            ) => [T[1, -√2, 1]],
```
odow, line 5013:
```suggestion
            ) => [T[1, -sqrt(T(2)), 1]],
```

src/Utilities/sets.jl
```diff
@@ -158,6 +158,47 @@ function trimap(row::Integer, column::Integer)
...
+    else
+        return s.scaling
+    end
+end
+
+# `LinearAlgebra.Diagonal` checks that the first index is 1 so we need
+# to implement this
+Base.firstindex(::SymmetricMatrixScalingVector) = 1
+
+Base.IteratorSize(::SymmetricMatrixScalingVector) = Base.IsInfinite()
```
odow, line 200: It looks like we need to implement `Base.show(::IO, ::SymmetricMatrixScalingVector)`.

blegat: Yes, just did it in latest commit

odow:
I changed to use an explicit length. Makes things simpler.

src/Utilities/functions.jl
```diff
@@ -3409,3 +3409,87 @@ function MA.promote_operation(
...
+
+function operate_term(::typeof(*), D::Diagonal, term::MOI.VectorAffineTerm)
+    return MOI.VectorAffineTerm(
+        term.output_index,
+        operate_term(*, D.diag[term.output_index], term.scalar_term),
+    )
+end
+
+function operate_term(::typeof(*), D::Diagonal, term::MOI.VectorQuadraticTerm)
+    return MOI.VectorQuadraticTerm(
```
odow, line 3434: If we're going to add these, they need tests?

blegat:
> I changed to use an explicit length. Makes things simpler.

Why does it make things simpler ? It's quite convenient to be able to use the same scaling for whatever length. By the way, it's only possible because we chose upper-triangular instead of lower-triangular vectorization

src/Utilities/functions.jl
blegat: At the moment, yes

odow (approved):

PR #2156, opened 2023-04-26
[Bridges] add tests for deleting variable in bridged objective

### LINKED ISSUE #2153: Variable deletion with bridges
All bridges in MOI are affine transformations except the QuadtoSOC bridge.
As deletion commutes with affine transformations, except for the QuadtoSOC bridge, there should be no issue just deleting the variable in the bridged solver model.
Deleting a variable part of a quadratic term of a quadratic function bridged by the QuadtoSOC bridged might be an issue though.
We should check that and add tests (with quadratic objective and quadratic constraints).

See https://github.com/jump-dev/MathOptInterface.jl/pull/2150/files#r1173436839

odow:
Closes #2153

@blegat everything looks okay to me.

### DIFF
--- a/test/Bridges/Constraint/quad_to_soc.jl
+++ b/test/Bridges/Constraint/quad_to_soc.jl
@@ -278,6 +278,50 @@ function test_copy_to_start()
     @test MOI.get(dest, MOI.VariablePrimalStart(), index_map[x]) == -1.0
 end
 
+function test_deletion_of_variable_in_bridged_constraint()
+    inner = MOI.Utilities.Model{Float64}()
+    model = MOI.Bridges.Constraint.QuadtoSOC{Float64}(inner)
+    x = MOI.add_variables(model, 2)
+    MOI.set(model, MOI.VariableName(), x, ["x", "y"])
+    f = 1.0 * x[1] * x[1] + 1.0 * x[2] * x[2]
+    c = MOI.add_constraint(model, f, MOI.LessThan(1.0))
+    MOI.delete(model, x[1])
+    @test MOI.get(model, MOI.ConstraintFunction(), c) ≈ 1.0 * x[2] * x[2]
+    return
+end
+
+MOI.Utilities.@model(
+    Model2153,
+    (),
+    (MOI.EqualTo,),
+    (MOI.RotatedSecondOrderCone,),
+    (),
+    (),
+    (MOI.ScalarAffineFunction,),
+    (MOI.VectorOfVariables,),
+    (),
+)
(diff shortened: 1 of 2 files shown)


test/Bridges/Constraint/quad_to_soc.jl
```diff
@@ -278,6 +278,50 @@ function test_copy_to_start()
...
+function MOI.supports(
+    ::Model2153{T},
+    ::MOI.ObjectiveFunction{MOI.ScalarQuadraticFunction{T}},
+) where {T}
+    return false
+end
+
+function test_deletion_of_variable_in_bridged_slacked_objective()
+    model = MOI.Bridges.full_bridge_optimizer(Model2153{Float64}(), Float64)
+    MOI.Bridges.add_bridge(model, MOI.Bridges.Constraint.QuadtoSOCBridge{Float64})
```
odow, line 314:
```suggestion
    MOI.Bridges.add_bridge(
        model,
        MOI.Bridges.Constraint.QuadtoSOCBridge{Float64},
    )
```

blegat:
So suppose you have `x'*U'*U*x + a'x <= ub` and you delete `x_1`.
Before, the RSOC constraint was `(1, -a'x + ub, U*x)` and after you should have `(1, -a'x[2:end] + ub, U[:,2:end] * x[2:end])` with the current implementation.
However, each row `r` of `U` for which `U[r,:]` is now zero should be dropped if you want to match the transformation that you would have if you restart the bridging after deleting the variable. This mismatch might be confusing for the user.

odow:
I dunno. Having to delete the row of a constraint if it goes to zero because you deleted a variable is a much bigger change. 

> This mismatch might be confusing for the user.

But equally, there was a row in the constraint matrix. We deleted the variable so the row stills stays. If it was the only variable, the row is now zero. That seems straightforward.

odow:
An example is that if you had `1.0 * x in GreaterThan(0.0)`, you wouldn't expect Gurobi to delete the row just because you deleted the `x` variable.

blegat:
I doesn't make sense to delete the row if you delete the variable that's why it won't work if you don't delete the quadratic term and bridge again. For the bridge, since the transformation is internal, it does make sense to simplify if the user remove quadratic terms.

odow:
> For the bridge, since the transformation is internal, it does make sense to simplify if the user remove quadratic terms.

That just seems like added complications for minimal benefit. Can we merge these tests, and wait for someone (other than you) to complain? I bet no one even notices.

PR #2162, opened 2023-05-01
Add four functions in the NLP API

amontoison:
When we interface solvers dedicated to NLP problems, the Hessian of augmented Lagrangian is generally only required.
However, as a user, It's sometimes relevant to only determine the Hessian of the objective or a specific constraint.
I implemented the following functions for the NLP API:
- hessian_objective_structure
- hessian_constraint_structure
- eval_hessian_objective
- eval_hessian_constraint


src/Nonlinear/evaluator.jl
```diff
@@ -143,10 +145,32 @@ function MOI.eval_constraint_jacobian(evaluator::Evaluator, J, x)
     return
 end
 
+function MOI.hessian_objective_structure(evaluator::Evaluator)
+    return MOI.hessian_objective_structure(evaluator.backend)
+end
+
+function MOI.hessian_constraint_structure(evaluator::Evaluator, i)
+    return MOI.hessian_constraint_structure(evaluator.backend)
```
odow, line 153:
```suggestion
    return MOI.hessian_constraint_structure(evaluator.backend, i)
```

src/Nonlinear/ReverseAD/mathoptinterface_api.jl
```diff
@@ -250,22 +256,41 @@ function MOI.eval_constraint_jacobian_transpose_product(
     return
 end
 
+function MOI.hessian_objective_structure(d::NLPEvaluator)
+    @assert d.want_hess
+    H = Tuple{Int64,Int64}[(row, col) for (row, col) in zip(d.objective.hess_I, d.objective.hess_J)]
+    return H
+end
+
+function MOI.hessian_constraint_structure(d::NLPEvaluator, i::Int64)
```
odow, line 265:
```suggestion
function MOI.hessian_constraint_structure(d::NLPEvaluator, i::Integer)
```

src/Nonlinear/ReverseAD/precompile.jl
```diff
@@ -42,6 +59,13 @@ function _precompile_()
             Vector{Float64},
         },
     )
+    Base.precompile(Tuple{typeof(MOI.hessian_objective_structure),NLPEvaluator})
+    Base.precompile(
+        Tuple{typeof(MOI.hessian_constraint_structure),NLPEvaluator, Int64},
```
odow, line 64:
```suggestion
        Tuple{typeof(MOI.hessian_constraint_structure),NLPEvaluator,Int64},
```

amontoison:
Maybe we should add \sigma and \mu as optional argument with default value 1.0 for the functions `eval_hessian_objective` and `eval_hessian_constraint`.
What do you think @odow?

odow:
> Maybe we should add \sigma and \mu as optional argument with default value 1.0 for the functions

Nah. Not if it's just a premultiplier. The user can always do that afterwards. 

Before merging we also need to add the docstrings to the docs

docs/src/reference/nonlinear.md
```diff
@@ -31,9 +31,13 @@ features_available
 eval_objective
 eval_constraint
 eval_objective_gradient
+hessian_objective_structure
+eval_hessian_objective
```
odow, line 35: So we have `eval_objective_gradient`, but `eval_hessian_objective`. 

Perhaps it should be `eval_objective_hessian`?

docs/src/reference/nonlinear.md
```diff
@@ -31,9 +31,13 @@ features_available
...
 jacobian_structure
 hessian_lagrangian_structure
 eval_constraint_jacobian
+hessian_constraint_structure
+eval_hessian_constraint
```
odow, line 40: Ditto with `eval_constraint_hessian`

amontoison: Yes, I agree with you  but I wanted to be consistent with `eval_hessian_lagrangian` that is here since MPB.
Maybe we should add `eval_lagrangian_hessian` and deprecated `eval_hessian_lagrangian`?

odow: Ah. Perhaps this is okay then. I'll re-order the docs to group things together. I don't really want to start deprecating existing methods.

docs/src/reference/nonlinear.md
```diff
@@ -32,10 +32,14 @@ eval_objective
...
 jacobian_structure
-hessian_lagrangian_structure
 eval_constraint_jacobian
 eval_constraint_jacobian_product
 eval_constraint_jacobian_transpose_product
+hessian_lagrangian_structure
+hessian_objective_structure
+hessian_constraint_structure
+eval_hessian_objective
+eval_hessian_constraint
```
odow, line 42: Putting it in this order makes sense. So perhaps we'll just keep it as-is.

odow (approved):
LGTM. We'll leave this open for a while to give @blegat and @mlubin a chance to take a look.

mlubin:
> However, as a user, It's sometimes relevant to only determine the Hessian of the objective or a specific constraint.

Do you mean in an exploratory context or in a performance-sensitive context? You *can* access the hessian of the objective or of a specific constraint by setting the `σ` and `μ` parameters of `eval_hessian_lagrangian` appropriately, although it's not the most efficient approach. (I'm not opposing this change, just want to make sure I understand the motivation.)

odow:
I viewed the addition as: we already have the underlying functionality implemented. So yes, you could call the Lagrangian function to get particular Hessians, but that is a bit painful. The code addition is minimal, even if it is four new functions.

mlubin (approved):

PR #2171, opened 2023-05-13
[Bridges] fix bug in HermitianToSymmetricPSDBridge

### LINKED ISSUE #2170: Bug in HermitianToSymmetricPSDBridge
Some recent change, probably in JuMP 1.10 or 1.11, made the Hermitian PSD bridge stop working. Now an SDP that needs those constraints returns variables that are not necessarily PSD, and usually gives the wrong answer. The exception is of course Hypatia, as it doesn't use the bridge, but instead implements the Hermitian PSD cone natively.

An example code to see the issue is the state discrimination code I have posted many times before. It only shows up if d=3 or larger. Often the discrepancy between the value returned by the SDP and the analytical answer is small and the eigenvalues are non-negative, so you should try it a couple of times to see a big difference.
```julia
function random_state(d)
	x = randn(ComplexF64, (d,d))
	y = x*x'
	rho = Hermitian(y/tr(y))
	return rho
end


function discriminate(d)
#    model = Model(Hypatia.Optimizer)
    model = Model(SCS.Optimizer)
    
    rho1 = random_state(d)
    rho2 = random_state(d)
    E1 = @variable(model, [1:d, 1:d] in HermitianPSDCone())
    E2 = @variable(model, [1:d, 1:d] in HermitianPSDCone())
    @constraint(model, E1+E2 == I)
    obj = real(dot(rho1,E1) + dot(rho2,E2))/2
    @objective(model, Max, obj)
    optimize!(model)
    println(value(obj))

	#we print the analytical answer as well for debugging
    println(0.5 + 0.25*sum(svdvals(rho1-rho2)))

    #the eigenvalues are not always non-negative
    display(eigvals(value.(E1)))
    display(eigvals(value.(E2)))
end
```

odow:
Closes #2170

### DIFF
--- a/src/Bridges/Constraint/bridges/hermitian.jl
+++ b/src/Bridges/Constraint/bridges/hermitian.jl
@@ -99,11 +99,6 @@ function MOI.Bridges.inverse_map_set(
     return MOI.HermitianPositiveSemidefiniteConeTriangle(div(dim, 2))
 end
 
-function _to_y_offset(row, col, n)
-    @assert row < col
-    return div(n * (n + 1), 2) + MOI.Utilities.trimap(row, col - 1)
-end
-
 function MOI.Bridges.map_function(
     ::Type{<:HermitianToSymmetricPSDBridge{T}},
     func,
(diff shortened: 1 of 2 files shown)


odow:
@blegat, I didn't really understand your indexing logic, so I just started from scratch, working my way through the result vector in a single pass.

`inverse_map_function` is correct, but you should take a look at `adjoint_map_function`. I don't really understand what it's doing or if it's correct.

odow:
cc @araujoms

You can try this branch with:
```julia
] add MathOptInterface#od/fix-hermitian-bridge
```

araujoms:
Just tested it, works like a charm.

blegat:
Arf, I messed up the adjoint of the imaginary part, that's the issue when coding without a pen and paper. It would be quite unreadable to maintain the index as it is moving horizontally and not vertically (because of the transpose) so I switched to using `trimap`

PR #2172, opened 2023-05-14
Prep for v1.16.0

odow:
## Basic

 - [x] `version` field of `Project.toml` has been updated
       - If a breaking change, increment the MAJOR field and reset others to 0
       - If adding new features, increment the MINOR field and reset PATCH to 0
       - If adding bug fixes or documentation changes, increment the PATCH field

## Documentation

 - [x] Add a new entry to `docs/src/changelog.md`, following existing style

## Tests

 - [x] The `solver-tests.yml` GitHub action does not have unexpected failures.
         https://github.com/jump-dev/MathOptInterface.jl/actions/runs/4974400735

## TODO

 - [x] https://github.com/jump-dev/MathOptInterface.jl/pull/2173 fix for MosekTools failure

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "MathOptInterface"
 uuid = "b8f27783-ece8-5eb3-8dc8-9495eed66fee"
-version = "1.15.1"
+version = "1.16.0"
 
 [deps]
 BenchmarkTools = "6e4b80f9-dd63-53aa-95a3-0cdb28fa8baf"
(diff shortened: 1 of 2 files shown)


odow:
@blegat MosekTools is failing the `NormCone` tests with

image

I see this shows up in other excluded tests
https://github.com/jump-dev/MosekTools.jl/blob/cf691731a84e56eaef8317181f9c3dd8e1c00449/test/runtests.jl#L179-L181
so perhaps it's okay to ignore?

blegat:
Yes, the only way to fix this is to stop supporting `VectorOfVariables` constraints and only support constrained variables but Mosek will drop support for this in v10 so that should just go away eventually.

odow:
Good to go then?

PR #2173, opened 2023-05-14
Fix convert(::VariableIndex, ::ScalarAffineFunction) with zero coefficients

odow:
Showed up in https://github.com/jump-dev/MathOptInterface.jl/pull/2172

MosekTools fails tests because of:

image

It's pretty reasonable that we should skip zero terms when converting.

Check that it fixes Mosek
 - [x] https://github.com/jump-dev/MathOptInterface.jl/actions/runs/4974798432

### DIFF
--- a/.github/workflows/solver-tests.yml
+++ b/.github/workflows/solver-tests.yml
@@ -60,6 +60,7 @@ jobs:
         shell: julia --color=yes {0}
         run: |
           import Pkg
+          Pkg.develop(PackageSpec(path=pwd()))
           Pkg.develop(ENV["PACKAGE"])
           Pkg.test(ENV["PACKAGE"])
   test-cplex:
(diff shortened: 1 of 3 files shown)


odow:
Hmm. I don't think I have `solver-tests` setup correctly. MosekTools doesn't seem to have installed the branch of MOI.

odow:
https://github.com/jump-dev/MathOptInterface.jl/actions/runs/4975115966

odow:
Okay, this fixes that particular issue. There's an unrelated problem with `NormCone`: https://github.com/jump-dev/MathOptInterface.jl/actions/runs/4975191716/jobs/8902130828

blegat (approved):

odow:
I could expect `convert_approx` to handle something like `0.5x + 0.5x`. But skipping zeros here seems like a low bar. It still requires exactly one non-zero with a coefficient of `1.0`.

PR #2177, opened 2023-05-24
Fix support for external sets in loadfromstring

### LINKED ISSUE #2175: Bridge tests outside of MOI
Using `MOI.Bridges.runtests` is not feasible with a set defined outside MOI because of
https://github.com/jump-dev/MathOptInterface.jl/blob/9a9188a15e0ccdf7d9e7a44dfd7a9f6bdf33229a/src/Utilities/parser.jl#L336
and
https://github.com/jump-dev/MathOptInterface.jl/blob/9a9188a15e0ccdf7d9e7a44dfd7a9f6bdf33229a/src/Utilities/parser.jl#L365

odow:
Closes #2175

### DIFF
--- a/test/FileFormats/SDPA/SDPA.jl
+++ b/test/FileFormats/SDPA/SDPA.jl
@@ -10,7 +10,6 @@ using Test
 
 import MathOptInterface as MOI
 import MathOptInterface.Utilities as MOIU
-
 const SDPA = MOI.FileFormats.SDPA
 const SDPA_TEST_FILE = "test.sdpa"
 const SDPA_MODELS_DIR = joinpath(@__DIR__, "models")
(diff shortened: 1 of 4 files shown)


test/FileFormats/SDPA/SDPA.jl
```diff
@@ -10,6 +10,7 @@ using Test
 
 import MathOptInterface as MOI
 import MathOptInterface.Utilities as MOIU
+
```
odow, line 13:
```suggestion
```

odow:
This is technically breaking if the is given as `MathOptInterface.XYZ()` and MathOptInterface is not a defined symbol in the current scope (e.g., because it was `import MathOptInterface as MOI`). 

Let's see if `solver-tests` complains: https://github.com/jump-dev/MathOptInterface.jl/actions/runs/5075528679

odow:
HiGHS failed, https://github.com/jump-dev/MathOptInterface.jl/actions/runs/5075528679/jobs/9116842487, so I guess I need to fix this properly.

blegat (approved):

PR #2179, opened 2023-05-29
Fix promote_operation for ScalarNonlinearFunction

odow:
I guess I didn't run `solver-tests.yml` on the ScalarNonlinearFunction PR recently.

 https://github.com/jump-dev/MathOptInterface.jl/actions/runs/5108546474

### DIFF
--- a/src/Utilities/functions.jl
+++ b/src/Utilities/functions.jl
@@ -1820,6 +1820,14 @@ function promote_operation(
     ::typeof(-),
     ::Type{T},
     ::Type{MOI.ScalarNonlinearFunction},
+) where {T}
+    return MOI.ScalarNonlinearFunction
+end
+
+function promote_operation(
+    ::typeof(-),
+    ::Type{T},
+    ::Type{MOI.ScalarNonlinearFunction},
     ::Type{T},
 ) where {T}
     return MOI.ScalarNonlinearFunction
--- a/src/Utilities/functions.jl
+++ b/src/Utilities/functions.jl
@@ -1824,33 +1824,6 @@ function promote_operation(
     return MOI.ScalarNonlinearFunction
 end
 
-function promote_operation(
-    ::typeof(-),
-    ::Type{T},
-    ::Type{MOI.ScalarNonlinearFunction},
-    ::Type{T},
-) where {T<:Number}
-    return MOI.ScalarNonlinearFunction
-end
-
-function promote_operation(
-    ::typeof(-),
-    ::Type{<:Number},
-    ::Type{MOI.ScalarNonlinearFunction},
-    ::Type{MOI.VariableIndex},
-)
-    return MOI.ScalarNonlinearFunction
-end
-
-function operate(
-    op::Union{typeof(+),typeof(-)},
-    ::Type{T},
-    f::MOI.ScalarNonlinearFunction,
-    g::ScalarQuadraticLike{T},
-) where {T}
-    return MOI.ScalarNonlinearFunction(Symbol(op), Any[f, g])
-end
-
 function operate(
     ::typeof(-),
     ::Type{T},
@@ -1863,6 +1836,39 @@ function operate(
         end
     end
     return MOI.ScalarNonlinearFunction(:-, Any[f])
+end
+
+function promote_operation(
+    ::Union{typeof(+),typeof(-),typeof(*),typeof(/)},
+    ::Type{T},
+    ::Type{MOI.ScalarNonlinearFunction},
+    ::Type{S},
+) where {
+    T<:Number,
+    S<:Union{
+        T,
+        MOI.VariableIndex,
+        MOI.ScalarAffineFunction{T},
+        MOI.ScalarQuadraticFunction{T},
+        MOI.ScalarNonlinearFunction,
+    }
+}
+    return MOI.ScalarNonlinearFunction
+end
+
+function operate(
+    op::Union{typeof(+),typeof(-),typeof(*),typeof(/)},
+    ::Type{T},
+    f::MOI.ScalarNonlinearFunction,
+    g::Union{
+        T,
+        MOI.VariableIndex,
+        MOI.ScalarAffineFunction{T},
+        MOI.ScalarQuadraticFunction{T},
+        MOI.ScalarNonlinearFunction,
+    }
+) where {T}
+    return MOI.ScalarNonlinearFunction(Symbol(op), Any[f, g])
 end
 
 ### Base methods
--- a/src/Utilities/functions.jl
+++ b/src/Utilities/functions.jl
@@ -1857,8 +1857,11 @@ function operate(
     ::Type{T},
     f::MOI.ScalarNonlinearFunction,
 ) where {T}
-    if f.head == :- && length(f.args) == 1 && f.args[1] isa MOI.ScalarNonlinearFunction
-        return f.args[1]
+    if f.head == :- && length(f.args) == 1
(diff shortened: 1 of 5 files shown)


test/Bridges/Constraint/vectorize.jl
```diff
@@ -238,6 +238,27 @@ function test_runtests()
...
+    ()
+)
+
+function test_unsupported_ScalarNonnlinearFunction()
+    model = MOI.instantiate(Model2179{Float64}; with_bridge_type = Float64)
+    MOI.supports_constraint(
+        model,
+        MOI.ScalarNonlinearFunction,
+        MOI.GreaterThan{Float64},
+    )
```
odow, line 259:
```suggestion
    )
    return
```

test/Bridges/Constraint/vectorize.jl
```diff
@@ -238,6 +238,27 @@ function test_runtests()
...
+    (MOI.GreaterThan, MOI.LessThan),
+    (),
+    (),
+    (),
+    (MOI.ScalarAffineFunction,),
+    (),
+    ()
+)
+
+function test_unsupported_ScalarNonnlinearFunction()
```
chriscoey, line 253:
```suggestion
function test_unsupported_ScalarNonlinearFunction()
```

odow:
Something weird is going on with this commit and Revise.

If I run the test, edit, and then re-run the test I can trigger a stack overflow:
```Julia
julia> MOI.Bridges.runtests(
               MOI.Bridges.Constraint.LessToGreaterBridge,
               """
               variables: x
               ScalarNonlinearFunction(1.5 * x) <= 1.0
               """,
               """
               variables: x
               ScalarNonlinearFunction(-(1.5 * x)) >= -1.0
               """,
           )
┌ Error: Failed to revise /Users/oscar/.julia/dev/MathOptInterface/src/Utilities/functions.jl
│   exception =
│    StackOverflowError:
│    Stacktrace:
│     [1] top-level scope
│       @ ~/.julia/dev/MathOptInterface/src/Utilities/functions.jl:1612
│    Revise evaluation error at /Users/oscar/.julia/dev/MathOptInterface/src/Utilities/functions.jl:1612
│    
└ @ Revise ~/.julia/packages/Revise/WHZdV/src/packagedef.jl:714
┌ Warning: The running code does not match the saved version for the following files:
│ 
│   /Users/oscar/.julia/dev/MathOptInterface/src/Utilities/functions.jl
│ 
│ If the error was due to evaluation order, it can sometimes be resolved by calling `Revise.retry()`.
│ Use Revise.errors() to report errors again. Only the first error in each file is shown.
│ Your prompt color may be yellow until the errors are resolved.
└ @ Revise ~/.julia/packages/Revise/WHZdV/src/packagedef.jl:824
```
but that doesn't happen if I just run without Revise.

I can't figure out the cause though... I'll re-run solver tests to see what happens.

src/Utilities/functions.jl
```diff
@@ -1816,6 +1816,15 @@ end
...
 ### ScalarNonlinearFunction
 
+# Needed for LessToGreaterBridge
+function promote_operation(
+    ::Union{typeof(+),typeof(-)},
+    ::Type{<:Real},
+    ::Type{MOI.ScalarNonlinearFunction},
+)
+    return MOI.ScalarNonlinearFunction
+end
```
odow, line 1826: @blegat: so one issue. A lot of the bridges are very permissive in the functions that accept. And then they use `promote_operation` to figure out what the resultant function is. That makes it hard for new functions, because we need to add a bunch of these `promote_operation` methods. An alternative would be to make the bridge explicitly declare what functions they support, and external packages would need to opt-into them. Not something we necessarily need to change now, but another MOI 2.0 list.

odow:
https://github.com/jump-dev/MathOptInterface.jl/actions/runs/5115502094

odow:
The type inference segfault in x86 looks like a problem. But I can't figure out why it's happening. The amount of dispatch going on with `promote_operation` makes it really hard to trace or reason about. There's lots of `Union` types that pick of different bits.

src/Utilities/functions.jl
blegat: Having to opt-in for each bridge would loose a lot of the bridges advantages. Right now, the fact that we will be able to no nonlinear conic with the bridges just by implementing a few `MOI.Utilities` is quite appealing.
I can understand that when you create a new function you would like to be able to test it without having to do all these things.
One thing you could do is to use `MOI.Bridges.remove_bridge` with the bridges that are annoying you.
We could have a way to say that an operation is not supported as well so that you could do that and then not have to implement `MOI.Utilities.operate`, etc...

src/Utilities/functions.jl
```diff
@@ -1843,6 +1853,20 @@ function operate(
...
 
+function operate(
+    ::typeof(-),
+    ::Type{T},
+    f::MOI.ScalarNonlinearFunction,
+) where {T}
+    if f.head == :- && length(f.args) == 1
+        # A simplification for -(-(f)) into f, but only if f is an SNF.
+        if f.args[1] isa MOI.ScalarNonlinearFunction
+            return f.args[1]
```
blegat, line 1864: This does not seem to be covered by the tests

odow: It is in the `GreaterToLess` bridge tests. Because it tries the `g = map_function(f)` and then `inverse_map_function(g)` to see if it's equal to `f`.

src/Utilities/functions.jl
```diff
@@ -1816,6 +1816,16 @@ end
 
 ### ScalarNonlinearFunction
 
+@static if Sys.WORD_SIZE == 64
```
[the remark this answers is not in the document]
odow: We need to remove this before merging. It's incorrect. Now the test fails because this method is missing. But at least it doesn't segfault...

src/Utilities/functions.jl
blegat: All good then :)

odow:
This PR also triggered it: https://github.com/jump-dev/MathOptInterface.jl/actions/runs/5126479287/jobs/9221017061. So it turns out it's not this PR...

src/Utilities/functions.jl
```diff
@@ -1817,28 +1817,56 @@ end
...
     ::Type{T},
     f::MOI.ScalarNonlinearFunction,
-    g::ScalarQuadraticLike{T},
+    g::Union{
+        T,
+        MOI.VariableIndex,
+        MOI.ScalarAffineFunction{T},
+        MOI.ScalarQuadraticFunction{T},
+        MOI.ScalarNonlinearFunction,
+    }
```
odow, line 1869:
```suggestion
    },
```

src/Utilities/functions.jl
```diff
@@ -1817,28 +1817,56 @@ end
...
+    ::Type{S},
+) where {
+    T<:Number,
+    S<:Union{
+        T,
+        MOI.VariableIndex,
+        MOI.ScalarAffineFunction{T},
+        MOI.ScalarQuadraticFunction{T},
+        MOI.ScalarNonlinearFunction,
+    }
```
odow, line 1854:
```suggestion
    },
```

odow:
https://github.com/jump-dev/MathOptInterface.jl/actions/runs/5130375265

odow:
@blegat happy? The x86 error is sporadic. Restarting CI once (or twice) seems to get it passed. I wonder if it's an issue with the available memory?

blegat (approved):
Excited to have the SNF propagate through the bridge graph

PR #2186, opened 2023-06-01
Add documentation for enum instances

odow:
I've been continuing to look into https://github.com/jump-dev/JuMP.jl/pull/3386, and constants like `MOI.MIN_SENSE` currently have no docstrings attached. This makes it harder to automatically document that `JuMP.MIN_SENSE` is part of JuMP's public API.

Adding docstrings to enum instances is cumbersome at the moment, because you need to manually use `@doc("my docs", instance)`. This PR introduces a new macro that let's us write inline docstrings.


odow:
Preview: https://jump.dev/MathOptInterface.jl/previews/PR2186/reference/models/#ResultStatusCode

I think this will look good if we update MOI to use the automatic reference documentation.

PR #2187, opened 2023-06-01
WIP: explore better API documentation

odow:
I started to looking documenting our public API in more detail, and got blocked by the number of methods in MOI that are private but do not start with `_`.

This PR tidies up the docstring and methods in `src/constraints.jl`. It also removes a couple of methods that should be private (and they aren't used by anyone in https://juliahub.com/ui/Search?q=throw_add_constraint_error_fallback&type=code). I can split the PR in two if that'd be better.


src/constraints.jl
```diff
@@ -79,137 +89,80 @@ function Base.showerror(
...
+   and a constraint was already added to this variable that sets a lower bound
+ * An [`UpperBoundAlreadySet`](@ref) error is thrown if `F` is a [`VariableIndex`](@ref)
+   and a constraint was already added to this variable that sets an upper bound.
 """
 function add_constraint(
     model::ModelLike,
-    func::AbstractFunction,
-    set::AbstractSet,
-)
-    return throw_add_constraint_error_fallback(model, func, set)
```
odow, line 148: Previous fallback was very convoluted. There was `throw_add_constraint_error_fallback`

src/constraints.jl
```diff
@@ -79,137 +89,80 @@ function Base.showerror(
...
-# throw_add_constraint_error_fallback checks whether func and set are both
-# scalar or both vector. If it is the case, it calls
-# `correct_throw_add_constraint_error_fallback`
-function throw_add_constraint_error_fallback(
-    model::ModelLike,
-    func::AbstractScalarFunction,
-    set::AbstractScalarSet;
-    kwargs...,
-)
-    return correct_throw_add_constraint_error_fallback(
```
odow, line 160: And then `correct_throw_add_constraint_error_fallback`, this time with unused `kwargs`

src/constraints.jl
```diff
@@ -79,137 +89,80 @@ function Base.showerror(
...
-        "Cannot add a constraint of the form `VectorFunction`-in-`ScalarSet`",
-    )
-end
-
-# func and set are both scalar or both vector
-function correct_throw_add_constraint_error_fallback(
-    model::ModelLike,
-    func::AbstractFunction,
-    set::AbstractSet;
-    error_if_supported = AddConstraintNotAllowed{typeof(func),typeof(set)}(),
```
odow, line 206: And then this method with the kwarg that was never anything different.

odow:
I changed my mind. I'm going to try out some API documentation things. And we can break this PR apart if we decide it's worth pursuing.

odow:
Here's a preview

https://jump.dev/MathOptInterface.jl/previews/PR2187/api/MathOptInterface/

Pros:

 * I like having an explicit list of the public API

Cons:

 * We currently don't export any of the API, which goes against JuMP's style guide. We could start exporting, but it'll muck up anyone who has written `using MathOptInterface` because it'll bring in symbols like `get` and `delete`. (This was one reason we didn't export in the first place.) There are a _lot_ of people with `using MathOptInterface`, so this isn't a niche consideration: https://juliahub.com/ui/Search?q=using%20MathOptInterface&type=code
 * The API list is ordered by name, not by type or category. For example, it's hard to find all of the AbstractVectorSets, or all of the nonlinear API. 

If we don't want to export, and we don't want an alphabetically sorted list, then we're going to need some manual curation. That's not difficult (now that I understand Documenter's internals), but it is a pain to check that we're documenting all of the public API. There's a lot to be said for automatic generation.

odow:
Closing for now as stale. I'd still like to revisit this though.

PR #2194, opened 2023-06-08
[Bridges] add SlackBridgePrimalDualStart

odow:
x-ref https://github.com/jump-dev/JuMP.jl/pull/3271

@blegat, it seemed to me that we would never want to set this to something other than `-1` for the dual and `0` for the slack primal, so I opted to consolidate everything into a single attribute.

Does this seem right? Or am I missing something?

### DIFF
--- a/src/Bridges/Objective/bridges/slack.jl
+++ b/src/Bridges/Objective/bridges/slack.jl
@@ -199,6 +199,13 @@ end
 # Pretend that every model supports, and silently skip in set if unsupported
 MOI.supports(::MOI.ModelLike, ::SlackBridgePrimalDualStart) = true
 
+function MOI.supports(
+    ::MOI.Bridges.AbstractBridgeOptimizer,
+    ::SlackBridgePrimalDualStart,
+)
+    return true
+end
+
 function MOI.set(
     model::MOI.ModelLike,
     ::SlackBridgePrimalDualStart,
--- a/src/Bridges/Objective/bridges/slack.jl
+++ b/src/Bridges/Objective/bridges/slack.jl
@@ -197,14 +197,7 @@ function MOI.throw_set_error_fallback(
 end
 
 # Pretend that every model supports, and silently skip in set if unsupported
-MOI.supports(::MOI.ModelLike, ::SlackBridgePrimalDualStart) = true
-
-function MOI.supports(
-    ::MOI.Bridges.AbstractBridgeOptimizer,
-    ::SlackBridgePrimalDualStart,
-)
-    return true
-end
+MOI.supports_fallback(::MOI.ModelLike, ::SlackBridgePrimalDualStart) = true
 
 function MOI.set(
     model::MOI.ModelLike,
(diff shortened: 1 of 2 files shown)


src/Bridges/Objective/bridges/slack.jl
```diff
@@ -174,3 +174,76 @@ function MOI.get(
...
+    return true
+end
+
+function MOI.set(
+    model::MOI.ModelLike,
+    ::SlackBridgePrimalDualStart,
+    b::SlackBridge{T},
+    value::Bool,
+) where {T}
+    @assert value
```
blegat, line 213: Maybe we could just pass `nothing` ?

src/Bridges/Objective/bridges/slack.jl
```diff
@@ -174,3 +174,76 @@ function MOI.get(
...
+    # ConstraintDual: if the objective function had a dual, it would be `-1` for
+    # the Lagrangian function to be the same.
+    if MOI.supports(model, MOI.ConstraintDualStart(), typeof(b.constraint))
```
blegat, line 216: This does not seem right, we should get an `UnsupportedAttribute` error if it's not supported.
We should implement `supports` for the bridge so that it returns `false` if the constraint type does not support it. That way, it would be ignored by `copy_to` ?

src/Bridges/Objective/bridges/slack.jl
```diff
@@ -174,3 +174,76 @@ function MOI.get(
...
+    MOI.set(model, MOI.ConstraintPrimalStart(), b.constraint, zero(T))
+    return
+end
+
+function MOI.set(
+    b::MOI.Bridges.AbstractBridgeOptimizer,
+    attr::SlackBridgePrimalDualStart,
+    value,
+)
+    if MOI.Bridges.is_objective_bridged(b)
```
blegat, line 241: This will only work if the `SlackBridge` is the first objective bridge of the chain but it's enough for now if you add a comment saying we would do as complicated as `ObjectiveFunctionValue`/`ObjectiveValue` but we simplify since we know the slack bridge is usually the first one of the chain.

blegat:
I would also assume it to always be `-1`, unless with some kind of homogeneization but if it's not, we can still add the two attributes of https://github.com/jump-dev/JuMP.jl/pull/3271/ for a more fine-graine access without being breaking

src/Bridges/Objective/bridges/slack.jl
odow: We only ignore the `Name` attribute. This is a special case because we want to pretend that every solver supports it.

odow:
Yeah. Or support passing something other than `nothing`, like `(primal_slack, dual)`.

src/Bridges/Objective/bridges/slack.jl
```diff
@@ -174,3 +174,79 @@ function MOI.get(
...
+    ::SlackBridgePrimalDualStart,
+    ::Nothing,
+)
+    return # Silently ignore if the model does not support.
+end
+
+function MOI.supports(
+    ::MOI.ModelLike,
+    ::SlackBridgePrimalDualStart,
+    ::Type{<:SlackBridge},
```
odow, line 202:
```suggestion
```

src/Bridges/Objective/bridges/slack.jl
blegat: Indeed, maybe drop a comment here so that people see it's not the common way of handling starting values and it's not copy-pasted out of context

odow: Done

PR #2196, opened 2023-06-08
Require list of model attributes in tests

blegat:
I want to exclude this attribute in https://github.com/JuliaManifolds/Manopt.jl/pull/264


odow:
Why do you want to exclude it?

odow:
The change is okay. But excluding is a bad sign.

blegat:
If you are in writing the wrapper and want to test if everything works without this attribute. Once you have the green light you can add things gradually. You exclude it explicitly in the config so I find it a clean workflow

PR #2198, opened 2023-06-08
Add set for low-rank constrained SDP

blegat:
Closes https://github.com/jump-dev/MathOptInterface.jl/issues/2197

I moved the content of this comment to https://github.com/blegat/LowRankOpt.jl/issues/2

This is now just a follow up for https://github.com/jump-dev/MathOptInterface.jl/pull/2509 + it make `ScalarFunctionIterator` be an `AbstractVector`


odow:
This is getting a little out of hand. `PositiveSemidefiniteConeSquare`, `PositiveSemidefiniteConeTriangle`, `ScaledPositiveSemidefiniteConeTriangle`, `HermitianPositiveSemidefiniteConeTriangle`...

Do we really need yet another PSD cone? Why isn't this just a solver enhancement where they detect and exploit low-rank structures in the general matrix?

blegat:
You're right, maybe this isn't really a PSD think, we could generalize it for any cone.
```julia
model = Model()
@variable(model, (AX[1:m], X[1:n, 1:n]) in ConeWithInnerProducts(A, PSDCone()))
@constraint(model, AX == b)
@objective(model, Min, dot(X, C))
```
and
```julia
model = Model()
@variable(model, y[1:m])
@constraint(model, (y, C) in AffineSpanInCone(A, PSDCone()))
@objective(model, Max, dot(y, b))
```

odow:
Do you have a numerical example of what the `A` matrix would be?

Also, `(y, C) in` isn't going to work because we don't support tuples as arguments.

blegat:
> Also, `(y, C) in` isn't going to work because we don't support tuples as arguments.

No, the JuMP syntax won't work, it's just to give an idea.

odow:
Other questions:

 * Clarify low-rank property of the input data vs low-rank constraints on the output variables
 * How can the set be serialized to MOF?
 * What solvers support this?

blegat:
> What solvers support this?

I try to keep the list in https://github.com/jump-dev/MathOptInterface.jl/issues/2197

> How can the set be serialized to MOF?

I think some special case might not be serializable but that's already the case for other parametrized sets like `Scaled` and `Indicator`. Same with `MOI.Utililities.Model`, we cannot aim to every every set there, other will be handled a bit less efficiently by `UniversalFallback` anyway, what's important there is to have the LP sets since LP solvers can handle millions of them

odow:
@mlubin's concern around the serialization wasn't parameterized sets, but if we allow
```
struct ConeWithInnerProducts{A<:AbstractMatrix,S<:AbstractVectorSet}
    matrix::A
    set::S
end
```
How would we serialize `matrix::A` to JSON if we don't know the matrix type?

We need to think about doing something like:
```
struct ConeWithInnerProducts{T,S<:AbstractVectorSet}
    matrix::LinearAlgebra.LowerTriangular{Matrix{T}}
    set::S
end
```

blegat:
If we have your former example, what prevents serializing `ConeWithInnerProducts{LinearAlgebra.LowerTriangular{Matrix{T}}}` ?

odow:
Nothing. We could do that. But the point is that we'd need to pick the matrix type. We couldn't support arbitrary user-defined matrices.

docs/src/manual/standard_form.md
```diff
@@ -82,6 +82,8 @@ The vector-valued set types implemented in MathOptInterface.jl are:
 | [`RelativeEntropyCone(d)`](@ref MathOptInterface.RelativeEntropyCone) | ``\{ (u, v, w) \in \mathbb{R}^{d} : u \ge \sum_i w_i \log (\frac{w_i}{v_i}), v_i \ge 0, w_i \ge 0 \}`` |
 | [`HyperRectangle(l, u)`](@ref MathOptInterface.HyperRectangle)        | ``\{x \in \bar{\mathbb{R}}^d: x_i \in [l_i, u_i] \forall i=1,\ldots,d\}`` |
 | [`NormCone(p, d)`](@ref MathOptInterface.NormCone)       | ``\{ (t,x) \in \mathbb{R}^{d} : t \ge \left(\sum\limits_i \lvert x_i \rvert^p\right)^{\frac{1}{p}} \}`` |
+| [`SetWithDotProcuts(s, v)`](@ref MathOptInterface.SetWithDotProducts) | The cone `s` with dot products with the fixed vectors `v`. |
```
blegat, line 85:
```suggestion
| [`SetWithDotProducts(s, v)`](@ref MathOptInterface.SetWithDotProducts) | The cone `s` with dot products with the fixed vectors `v`. |
```

blegat:
This is probably best explored as a separate MOI extension. That would allow us making breaking releases while we explore the space of possible interfaces. I moved the code to https://github.com/blegat/LowRankOpt.jl. This PR now only contains the changes necessary to make in MOI for LowRankOpt to work.

odow (approved):
Need any specific tests?

blegat:
Yes, I'll add some tests

PR #2199, opened 2023-06-08
Remove unused argument names

blegat:


odow:
Yeah I tidy these up when I see them. But I haven't been systematic about it.

PR #2200, opened 2023-06-09
Add constraint_gradient_structure and eval_constraint_gradient

odow:
Closes https://github.com/jump-dev/MathOptInterface.jl/issues/2193.

@blegat is this what you had in mind?

- [x] Needs tests etc.

### DIFF
--- a/docs/src/reference/nonlinear.md
+++ b/docs/src/reference/nonlinear.md
@@ -32,6 +32,8 @@ eval_objective
 eval_constraint
 eval_objective_gradient
 jacobian_structure
+eval_constraint_gradient
+constraint_gradient_structure
 eval_constraint_jacobian
 eval_constraint_jacobian_product
 eval_constraint_jacobian_transpose_product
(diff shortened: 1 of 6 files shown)


blegat (approved):
I thought the internals were more tightly coupled with getting the full jacobian but it seems it's the opposite and the jacobian just concatenate the individual gradients

amontoison:
Nice! I'm wondering if we should not add the gradient of the Lagragian too.
We could easily check the first-order conditions of the optimization problem with it.

odow:
The Lagrangian gradient is pretty simple to implement though. It doesn't require you to poke the internals (or compute excess stuff like the full Jacobian just to get a single gradient).

```julia
function eval_lagrangian_gradient(d::AbbstractNLPEvaluator, x, σ, μ)
    ∇L = zeros(length(x))
    eval_objective_gradient(d, ∇L, x)
    ∇L .*= σ
    y = zeros(length(x))
    eval_constraint_jacobian_transpose_product(d, y, x, μ)
    ∇L .+= y
    return ∇L
end
```

amontoison:
I forgot that we can use `eval_constraint_jacobian_transpose_product`...

A little bit off-topic but I have a comment related to the function `time()`. I recently observed in my research that it's not precise at all on Windows.
I switched to `time_ns()`: https://github.com/JuliaSmoothOptimizers/Krylov.jl/pull/730#issuecomment-1538454646

PR #2201, opened 2023-06-09
Add VectorNonlinearFunction

odow:
Resurrecting `VectorNonlinearFunction` from the depths of https://github.com/jump-dev/MathOptInterface.jl/pull/2059.

There are two main motivating problems for this:

 1. `VectorNonlinearFunction-in-Complements` constrains for nonlinear complementarity problems
     - https://github.com/chkwon/PATHSolver.jl/pull/82
 2. `ObjectiveFunction{VectorNonlinearFunction}` for nonlinear multi-objective problems
     - https://github.com/jump-dev/MultiObjectiveAlgorithms.jl/pull/68

Documentation

https://jump.dev/MathOptInterface.jl/previews/PR2201/reference/standard_form/#MathOptInterface.VectorNonlinearFunction

I also removed this from #2059 because there we a lot of issues with the bridges, but I have a better handle on what to do now, and we can easily run `solver-tests` to catch the culprits.

https://github.com/jump-dev/MathOptInterface.jl/actions/runs/5218217869

### DIFF
--- a/src/Bridges/Constraint/bridges/scalarize.jl
+++ b/src/Bridges/Constraint/bridges/scalarize.jl
@@ -263,7 +263,7 @@ end
 
 function MOI.set(
     model::MOI.ModelLike,
-   ::MOI.ConstraintFunction,
+    ::MOI.ConstraintFunction,
     bridge::ScalarizeBridge{T,F,S},
     func,
 ) where {T,F,S}
--- a/src/Bridges/Constraint/bridges/scalarize.jl
+++ b/src/Bridges/Constraint/bridges/scalarize.jl
@@ -263,7 +263,7 @@ end
 
 function MOI.set(
     model::MOI.ModelLike,
-    attr::MOI.ConstraintFunction,
+   ::MOI.ConstraintFunction,
     bridge::ScalarizeBridge{T,F,S},
     func,
 ) where {T,F,S}
--- a/src/Bridges/Constraint/bridges/scalarize.jl
+++ b/src/Bridges/Constraint/bridges/scalarize.jl
@@ -263,7 +263,7 @@ end
 
 function MOI.set(
     model::MOI.ModelLike,
-    ::MOI.ConstraintFunction,
+    attr::MOI.ConstraintFunction,
     bridge::ScalarizeBridge{T,F,S},
     func,
 ) where {T,F,S}
--- a/src/Bridges/Constraint/bridges/vectorize.jl
+++ b/src/Bridges/Constraint/bridges/vectorize.jl
@@ -67,14 +67,6 @@ function MOI.supports_constraint(
     return true
 end
 
-function MOI.supports_constraint(
-    ::Type{VectorizeBridge{T}},
-    ::Type{MOI.ScalarNonlinearFunction},
-    ::Type{<:MOI.Utilities.ScalarLinearSet{T}},
-) where {T}
-    return false
-end
-
 function MOI.Bridges.added_constrained_variable_types(::Type{<:VectorizeBridge})
     return Tuple{Type}[]
 end
--- a/src/Utilities/functions.jl
+++ b/src/Utilities/functions.jl
@@ -197,6 +197,16 @@ end
 # The `eval_variables(::F, ::MOI.ModelLike, ::MOI.ScalarNonlinearFunction)`
 # method is defined in the MOI.Nonlinear submodule.
 
+function eval_variables(
+    value_fn::F,
+    model::MOI.ModelLike,
+    f::MOI.GenericVectorFunction,
+) where {F}
+    return map(f.rows) do row
+        return eval_variables(value_fn, model, row)
+    end
+end
+
 """
     map_indices(index_map::Function, attr::MOI.AnyAttribute, x::X)::X where {X}
 
@@ -339,7 +349,7 @@ function map_indices(
     f::MOI.GenericVectorFunction{T},
 ) where {F<:Function,T}
     return MOI.GenericVectorFunction{T}(
-        convert(Vector{T}, map_indices(index_map, f.rows))
+        convert(Vector{T}, map_indices(index_map, f.rows)),
     )
 end
 
@@ -522,6 +532,14 @@ function substitute_variables(
         operate_output_index!(+, T, term.output_index, g, sub)
     end
     return g
+end
+
+function substitute_variables(
+    variable_map::F,
+    f::MOI.GenericVectorFunction{T},
+) where {T,F<:Function}
+    new_rows = map(Base.Fix1(substitute_variables, variable_map), f.rows)
+    return MOI.GenericVectorFunction{T}(convert(Vector{T}, new_rows))
 end
 
 """
@@ -922,6 +940,8 @@ function is_canonical(
     return _is_strictly_sorted(f.affine_terms) &&
            _is_strictly_sorted(f.quadratic_terms)
 end
+
+is_canonical(f::MOI.GenericVectorFunction) = all(is_canonical, f.rows)
 
 function _is_strictly_sorted(x::Vector)
     if isempty(x)
--- a/src/Utilities/functions.jl
+++ b/src/Utilities/functions.jl
@@ -486,28 +486,19 @@ function substitute_variables(
     return g
 end
 
-function _unstable_substitute_variables(
-    variable_map::F,
-    x::MOI.VariableIndex,
-) where {F<:Function}
-    return variable_map(x)
-end
-
-function _unstable_substitute_variables(
-    variable_map::F,
-    x::Any,
-) where {F<:Function}
-    return substitute_variables(variable_map, x)
-end
-
 function substitute_variables(
     variable_map::F,
     f::MOI.ScalarNonlinearFunction,
 ) where {F<:Function}
-    # TODO(odow): this uses recursion. We should remove at some point.
-    new_args =
-        map(Base.Fix1(_unstable_substitute_variables, variable_map), f.args)
-    return MOI.ScalarNonlinearFunction(f.head, convert(Vector{Any}, new_args))
+    new_args = Any[]
+    for arg in f.args
+        if arg isa MOI.VariableIndex
+            push!(new_args, variable_map(arg))
+        else
+            push!(new_args, substitute_variables(variable_map, arg))
+        end
+    end
+    return MOI.ScalarNonlinearFunction(f.head, new_args)
 end
 
 function substitute_variables(
@@ -549,9 +540,8 @@ function substitute_variables(
     variable_map::F,
     f::MOI.VectorNonlinearFunction,
 ) where {F<:Function}
-    return MOI.VectorNonlinearFunction(
-        map(Base.Fix1(_unstable_substitute_variables, variable_map), f.rows),
-    )
+    rows = [substitute_variables(variable_map, row) for row in f.rows]
+    return MOI.VectorNonlinearFunction(rows)
 end
 
 """
--- a/src/Utilities/functions.jl
+++ b/src/Utilities/functions.jl
@@ -488,15 +488,30 @@ function substitute_variables(
     return g
 end
 
+function _unstable_substitute_variables(
+    variable_map::F,
+    x::MOI.VariableIndex,
+) where {F<:Function}
+    return variable_map(x)
+end
+
+function _unstable_substitute_variables(
+    variable_map::F,
+    x::Any,
+) where {F<:Function}
+    return substitute_variables(variable_map, x)
+end
+
 function substitute_variables(
     variable_map::F,
     f::MOI.ScalarNonlinearFunction,
 ) where {F<:Function}
     # TODO(odow): this uses recursion. We should remove at some point.
-    return MOI.ScalarNonlinearFunction(
-        f.head,
-        Any[substitute_variables(variable_map, a) for a in f.args],
+    new_args = map(
+        Base.Fix1(_unstable_substitute_variables, variable_map),
+        f.args,
     )
+    return MOI.ScalarNonlinearFunction(f.head, convert(Vector{Any}, new_args))
 end
 
 function substitute_variables(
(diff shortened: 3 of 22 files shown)


odow:
Marking this as blocked by #2203

odow:
This is an interesting suggestion. We'd need backward compat for `(::VectorOfVariables).variables`, but a generic `AbstractVectorFunction` that is a vector of scalar terms would be useful.

odow:
This is not fun. I think I'm going to revert the `GenericVectorFunction` changes to `MOI.VectorOfVariables`. There are too many little issues that crop up. Perhaps we can revisit once this PR is merged, and once we have a solution for the `operate` mess.

blegat:
> There are too many little issues that crop up.

The tests seem to pass except for some printing issues though

odow:
> The tests seem to pass except for some printing issues though

The tests don't cover the code enough. If you try running the tests with Clp/SCS etc lots of things break.

odow:
This is still blocked by the `operate` issue, but we're getting closer: https://github.com/jump-dev/MathOptInterface.jl/pull/2215

src/Bridges/Constraint/bridges/soc_rsoc.jl
```diff
@@ -72,7 +72,13 @@ function MOI.Bridges.map_function(::Type{<:SOCtoRSOCBridge{T}}, func) where {T}
...
     ts = MOI.Utilities.operate!(/, T, t, sqrt(T(2)))
     us = MOI.Utilities.operate!(/, T, u, sqrt(T(2)))
-    return MOI.Utilities.operate(vcat, T, ts + us, ts - us, x)
+    return MOI.Utilities.operate(
+        vcat,
+        T,
+        MOI.Utilities.operate(+, T, ts, us),
+        MOI.Utilities.operate(-, T, ts, us),
+        x,
+    )
```
odow, line 81: Fixed separately in https://github.com/jump-dev/MathOptInterface.jl/pull/2216

odow:
https://github.com/jump-dev/MathOptInterface.jl/actions/runs/5341127312

odow:
> If you try running the tests with SCS etc lots of things break.

Okay. One massive source of confusion for me had been failures with SCS. Which were because the bridge in SCS was wrong: https://github.com/jump-dev/SCS.jl/pull/277.

It used to do this:
```julia
julia> MOI.Bridges.print_active_bridges(model, MOI.VectorNonlinearFunction, MOI.PositiveSemidefiniteConeTriangle)
 * Unsupported constraint: MOI.GenericVectorFunction{Any}-in-MOI.PositiveSemidefiniteConeTriangle
 |  bridged by:
 |   SCS.ScaledPSDConeBridge{Float64, G} where G
 |  may introduce:
 |   * Supported constraint: MOI.VectorAffineFunction{Float64}-in-SCS.ScaledPSDCone
```
So no wonder lots of weird unrelated things broke.

But I only dug deep enough once I figured out it wasn't a method error with the various operate methods.

odow:
So I was looking into `VectorNonlinearFunction-in-PositiveSemidefiniteConeTriangle` etc. And there are a number of thorny issues:

 * `Utilities.eachscalar` assumes that each row of the vector function is `scalar_type(F)`. But the row of a VectorNonlinearFunction might not be ScalarNonlinearFunction.
 * Easy fix is to convert every row, but then what's the difference between `Utilities.scalarize` and `Utilities.eachscalar`?
 * Operations don't simplify. So `f = sin(x)` and then `-(f, f)` doesn't simplify to `0`. So we get an expression swelling problem that prevents `Utilities.isapprox_zero` from working. The main problem with this is in the `Bridges.Constraint.SquareBridge`, which checks if off-diagonals are equivalent by checking if their difference is zero.

blegat:
> Easy fix is to convert every row, but then what's the difference between `Utilities.scalarize` and `Utilities.eachscalar`?

Yes, I think every row should be an SNF. The difference of `scalarize` is that it allocates the vector of functions while `eachscalar` does not. If you have a lot or rows but you're only going to get one, `eachscalar` will be faster.
In fact, `scalarize` is really like conversion to a `GenericVectorFunction`

> The main problem with this is in the `Bridges.Constraint.SquareBridge`, which checks if off-diagonals are equivalent by checking if their difference is zero. 

Maybe it should use `isequal_canonical` or something like that

odow:
Should `scalarize` be a copy? Currently it isn't always:

https://github.com/jump-dev/MathOptInterface.jl/blob/5afc08bbd153e3dab3d7f5d60420a3bea6f3da19/src/Utilities/functions.jl#L2110-L2120

Ditto with `vectorize`

https://github.com/jump-dev/MathOptInterface.jl/blob/5afc08bbd153e3dab3d7f5d60420a3bea6f3da19/src/Utilities/functions.jl#L2044-L2057

blegat:
I don't think it has to be a copy. `MA.operate(scalarize, f)` should be a copy :laughing:

odow:
https://github.com/jump-dev/MathOptInterface.jl/actions/runs/5383997349

test/Bridges/Constraint/square.jl
```diff
@@ -240,6 +240,80 @@ function test_square_warning()
     return
 end
 
+function test_VectorNonlinearFunction_symmetric()
+    inner = MOI.Utilities.Model{Float64}()
+    model = MOI.Bridges.Constraint.Square{Float64}(inner)
+    x = MOI.add_variables(model, 3)
+    fis = Any[MOI.ScalarNonlinearFunction(:log, Any[x[i]]) for i in 1:3]
+    f = MOI.VectorNonlinearFunction(Any[fis[1], fis[2], fis[2], fis[3]])
+    c = MOI.add_constraint(model, f, MOI.PositiveSemidefiniteConeSquare(2))
```
odow, line 249: @blegat this is now working.

But if you print `model` you'll get a horrible function, because VectorNonlinear doesn't simplify `(f(x) - g(x)) + g(x) --> f(x)` at the expression level.

odow:
PATHSolver.jl and MultiObjectiveAlgorithms.jl are both passing tests with this branch, so I don't think we're far away. Just need https://github.com/jump-dev/MathOptInterface.jl/actions/runs/5383997349 to finish.

src/Bridges/Constraint/bridges/vectorize.jl
```diff
@@ -49,7 +49,7 @@ function bridge_constraint(
     if !iszero(scalar_const)
         throw(MOI.ScalarFunctionConstantNotZero{T,G,typeof(set)}(scalar_const))
     end
-    vector_f = convert(F, scalar_f)
+    vector_f = convert(F, MOI.Utilities.operate(vcat, T, scalar_f))
```
odow, line 52: So many assumptions. Here's a fun one: we assumed that you could call 
`convert(::AbstractVectorFunction, ::AbstractScalarFunction)`.

odow:
https://github.com/jump-dev/MathOptInterface.jl/actions/runs/5408332786

src/functions.jl
```diff
@@ -951,6 +1019,14 @@ end
 
 # ScalarNonlinearFunction
 
+function Base.convert(::Type{ScalarNonlinearFunction}, x::Real)
+    return ScalarNonlinearFunction(:+, Any[x])
+end
+
+function Base.convert(::Type{ScalarNonlinearFunction}, x::VariableIndex)
+    return ScalarNonlinearFunction(:+, Any[x])
```
odow, line 1027: So I wondered about introducing `:identity` as a special-case nonlinear argument.

Then `VectorNonlinearFunction` would be `Vector{ScalarNonlinearFunction}`, and there wouldn't be this distinction between `scalarize` returning the `Vector{Any}` and `eachscalar` returning an iterator that converted each element to `ScalarNonlinearFunction`.

I guess `identity` is the same as univariate `:+`, but it's more explicit as to what is happening.

blegat: Yes, I also prefer `identity`. Indeed, I can see the issue that `scalar_type` is `ScalarNonlinearFunction` so `eachscalar` has to convert to `ScalarNonlinearFunction which is inconsistent. Maybe that's a good argument against `Vector{Any}`

odow: So I've been experimenting with this. It really doesn't add anything, and only complicates stuff. let's just leave as `:+` for now.

odow:
https://github.com/jump-dev/MathOptInterface.jl/actions/runs/5756841023

odow:
https://github.com/jump-dev/MathOptInterface.jl/actions/runs/5778904348

src/Bridges/Constraint/bridges/scalarize.jl
```diff
@@ -263,7 +263,7 @@ end
 
 function MOI.set(
     model::MOI.ModelLike,
-    ::MOI.ConstraintFunction,
+    attr::MOI.ConstraintFunction,
```
odow, line 266:
```suggestion
   ::MOI.ConstraintFunction,
```

odow:
@blegat @mlubin this could use a review.

src/Bridges/Constraint/bridges/scalarize.jl
```diff
@@ -263,7 +263,7 @@ end
 
 function MOI.set(
     model::MOI.ModelLike,
-    ::MOI.ConstraintFunction,
+   ::MOI.ConstraintFunction,
```
odow, line 266:
```suggestion
    ::MOI.ConstraintFunction,
```

src/functions.jl
```diff
@@ -665,6 +665,69 @@ function Base.copy(f::VectorQuadraticFunction)
...
+
+julia> MOI.VectorNonlinearFunction(Any[g, x])
+┌                                 ┐
+│^(sin(MOI.VariableIndex(1)), 2.0)│
+│MOI.VariableIndex(1)             │
+└                                 ┘
+```
+"""
+struct VectorNonlinearFunction <: AbstractVectorFunction
+    rows::Vector{Any}
```
odow, line 708: Should we make this 
```suggestion
    rows::Vector{ScalarNonlinearFunction}
```

odow: @blegat, take a look at this as-is with `Any`. I was playing around with `ScalarNonlinearFunction`, and it just ends up requiring a bunch of `convert` in annoying places. If we force each row to be `+(arg)`, then the downstream solvers are still going to have to special case a bunch of stuff, and they'll still need support for `ScalarAffineFunction` etc, because they can be arguments to `ScalarNonlinearFunction`.

The only problem with the current `Any` is the subtlety around `eachscalar` and `scalarize`.

blegat: > I was playing around with `ScalarNonlinearFunction`, and it just ends up requiring a bunch of `convert` in annoying places

Could you be more explicit ? Naively, I wouldn't think it would so I would need some examples to convince me ^^

> If we force each row to be `+(arg)`, then the downstream solvers are still going to have to special case a bunch of stuff, and they'll still need support for `ScalarAffineFunction` etc, because they can be arguments to `ScalarNonlinearFunction`.

Not really if they use `MOI.Nonlinear.Model` for AD. If they use the expression graph, they are probably still going to parse the expression and do the obvious simplifications so this simplification might just come as a special case.
It's going to cause a bunch of troubles that accessing the fields, `scalarize`, `eachscalar` and `scalar_type` don't match. I don't see an added value of that design that would outweight the added complexity of having these inconsistencies.

src/Utilities/functions.jl
```diff
@@ -762,6 +815,20 @@ function Base.getindex(
     return MOI.VectorQuadraticFunction(vqt, vat, it.f.constants[output_indices])
 end
 
+function Base.getindex(
+    it::ScalarFunctionIterator{MOI.VectorNonlinearFunction},
+    output_index::Integer,
+)
+    return convert(MOI.ScalarNonlinearFunction, it.f.rows[output_index])
```
odow, line 822:
```suggestion
    return it.f.rows[output_index]
```

src/functions.jl
odow: Okay, take a look at this now. I think I got it to a nice point. The automatic conversion is still a bit weird. But maybe that's okay.

odow: [PATHSolver](https://github.com/chkwon/PATHSolver.jl/pull/82/files#diff-64cebae681efceded214317cd18b678aaefed5658af7a123f9b01433313a2844R231-R238) now needs this

image

where previously it could just require that `xi = f.rows[i+N]::MOI.VariableIndex`.

blegat: It makes sense, it's similar and consistent with Indicator which transforms the `VariableIndex` for having it fit into the `VectorAffineFunction`. It would be nice to have tools to make it easier though.

src/Utilities/functions.jl
```diff
@@ -469,15 +486,28 @@ function substitute_variables(
     return g
 end
 
+function _unstable_substitute_variables(
+    variable_map::F,
+    x::MOI.VariableIndex,
+) where {F<:Function}
+    return variable_map(x)
```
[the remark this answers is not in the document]
blegat: For instance, having
```julia
function substitute_variables(map, f)
    r = substitute(map, f)
    if f isa MOI.VariableIndex && ...
    end
    return r
end
```
and then replace all implementations of `substitute_variables` to implementations for `substitute`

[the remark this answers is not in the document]
blegat: Then you can use `substitute` instead of `_unstable_substitute_variables`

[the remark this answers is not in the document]
odow: Yeah only to avoid the error in the substitute variables for a variable.

src/Utilities/functions.jl
```diff
@@ -515,6 +545,15 @@ function substitute_variables(
     return g
 end
 
+function substitute_variables(
+    variable_map::F,
+    f::MOI.VectorNonlinearFunction,
+) where {F<:Function}
+    return MOI.VectorNonlinearFunction(
+        map(Base.Fix1(_unstable_substitute_variables, variable_map), f.rows),
```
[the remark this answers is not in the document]
blegat: Yes

src/Utilities/functions.jl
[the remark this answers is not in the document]
blegat: Here is a draft, it might be more explicit
https://github.com/jump-dev/MathOptInterface.jl/pull/2250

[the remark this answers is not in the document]
blegat: Ah but then, the `substitute` is unstable for `VariableIndex`, I see what you mean now. But maybe it does not matter now that small unions are fast. I wrote that a while ago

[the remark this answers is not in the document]
odow: I've simplified and removed `_unstable_substitute_variables`

blegat (approved):

mlubin:
I'll take a look over the weekend

mlubin (approved):

PR #2205, opened 2023-06-12
[Bridges] add IntegerToZeroOneBridge

### LINKED ISSUE #2202: IntegerToZeroOne bridge
Alpine doesn't support general integer variables: https://github.com/lanl-ansi/Alpine.jl/issues/233.

We could bridge to ZeroOne with a `final_touch` that checks for finite variable bounds.

odow:
Closes #2202

## Basic

 - [x] Create a new file in `src/Bridges/XXX/bridges`
       struct must end in `Bridge`
       the `Bridge` suffix
 - [x] `include` the file in `src/Bridges/XXX/bridges/XXX.jl`

## Tests

       bridge
       untested lines.

## Documentation

 - [x] Add the docstring to `docs/src/submodules/Bridges/list_of_bridges.md`

## Final touch

the model:

 - [x] Implement `MOI.Utilities.needs_final_touch(::Bridge)`
 - [x] Implement `MOI.Utilities.final_touch(::Bridge, ::MOI.ModelLike)`
 - [x] Ensure that `final_touch` can be called multiple times in a row

### DIFF
--- a/src/Bridges/Constraint/bridges/integer_to_zeroone.jl
+++ b/src/Bridges/Constraint/bridges/integer_to_zeroone.jl
@@ -1,0 +1,174 @@
+# Copyright (c) 2017: Miles Lubin and contributors
+# Copyright (c) 2017: Google Inc.
+#
+# Use of this source code is governed by an MIT-style license that can be found
+# in the LICENSE.md file or at https://opensource.org/licenses/MIT.
+
+"""
+    IntegerToZeroOneBridge{T} <: Bridges.Constraint.AbstractBridge
+
+`IntegerToZeroOneBridge` implements the following reformulation:
+
+  * ``x \\in \\mathbf{Z}`` into ``y_i \\in \\{0, 1\\}``,
+    ``x == lb + \\sum 2^{i-1} y_i``.
+
+## Source node
+
+`IntegerToZeroOneBridge` supports:
+
+  * `VariableIndex` in [`MOI.Integer`](@ref)
+
+## Target nodes
+
+`IntegerToZeroOneBridge` creates:
+
+  * [`MOI.VariableIndex`](@ref) in [`MOI.ZeroOne`](@ref)
+  * [`MOI.ScalarAffineFunction{T}`](@ref) in [`MOI.EqualTo{T}`](@ref)
+"""
+mutable struct IntegerToZeroOneBridge{T} <: AbstractBridge
+    x::MOI.VariableIndex
+    y::Vector{MOI.VariableIndex}
+    ci::MOI.ConstraintIndex{MOI.ScalarAffineFunction{T},MOI.EqualTo{T}}
+    last_bounds::Union{Nothing,NTuple{2,T}}
+
+    function IntegerToZeroOneBridge{T}(x::MOI.VariableIndex) where {T}
+        return new{T}(
+            x,
+            MOI.VariableIndex[],
+            MOI.ConstraintIndex{MOI.ScalarAffineFunction{T},MOI.EqualTo{T}}(0),
+            nothing,
+        )
+    end
+end
+
+const IntegerToZeroOne{T,OT<:MOI.ModelLike} =
+    SingleBridgeOptimizer{IntegerToZeroOneBridge{T},OT}
+
+function bridge_constraint(
+    ::Type{IntegerToZeroOneBridge{T}},
+    ::MOI.ModelLike,
+    x::MOI.VariableIndex,
+    ::MOI.Integer,
+) where {T}
+    # !!! info
+    #     Postpone creation until final_touch.
+    return IntegerToZeroOneBridge{T}(x)
+end
+
+function MOI.supports_constraint(
+    ::Type{IntegerToZeroOneBridge{T}},
+    ::Type{MOI.VariableIndex},
+    ::Type{MOI.Integer},
+) where {T}
+    return true
+end
+
+function MOI.Bridges.added_constrained_variable_types(
+    ::Type{<:IntegerToZeroOneBridge},
+)
+    return Tuple{Type}[(MOI.ZeroOne,)]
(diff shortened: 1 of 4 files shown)


test/Bridges/Constraint/integer_to_zeroone.jl
```diff
@@ -0,0 +1,100 @@
...
+        ErrorException(
+            "Unable to use IntegerToZeroOneBridge because the variable " *
+            "$(x) has a non-finite domain",
+        ),
+        MOI.Bridges.final_touch(model),
+    )
+    return
+end
+
+function test_finite_domain_error()
```
odow, line 88:
```suggestion
function test_final_touch_twice()
```

src/Bridges/Constraint/bridges/integer_to_zeroone.jl
```diff
@@ -0,0 +1,168 @@
...
+function MOI.Bridges.final_touch(
+    bridge::IntegerToZeroOneBridge{T},
+    model::MOI.ModelLike,
+) where {T}
+    ret = MOI.Utilities.get_bounds(model, T, bridge.x)
+    if ret === bridge.last_bounds
+        return nothing  # final_touch already called
+    elseif ret[1] == typemin(T) || ret[2] == typemax(T)
+        error(
+            "Unable to use IntegerToZeroOneBridge because the variable " *
```
Vaibhavdixit02, line 153: I am wondering if this should check against `nothing`s? My reasoning is that one could provide these values as the bounds intentionally. I can see that it might have to be done because of the defaults currently but it seems to be an edge case that might come up

src/Bridges/Constraint/bridges/integer_to_zeroone.jl
```diff
@@ -0,0 +1,168 @@
...
+MOI.Bridges.needs_final_touch(::IntegerToZeroOneBridge) = true
+
+function MOI.Bridges.final_touch(
+    bridge::IntegerToZeroOneBridge{T},
+    model::MOI.ModelLike,
+) where {T}
+    ret = MOI.Utilities.get_bounds(model, T, bridge.x)
+    if ret === bridge.last_bounds
+        return nothing  # final_touch already called
+    elseif ret[1] == typemin(T) || ret[2] == typemax(T)
```
blegat, line 151: We should probably have a utility for that shared with `SplitInterval` and other bridges doing this like `is_infinity`. In `Interval` we take care of `Int` while we don't do it here.

src/Bridges/Constraint/bridges/integer_to_zeroone.jl
blegat: > My reasoning is that one could provide these values as the bounds intentionally.

Which values ? I'm not sure to follow

src/Bridges/Constraint/bridges/integer_to_zeroone.jl
```diff
@@ -0,0 +1,168 @@
...
+    end
+    f = MOI.ScalarAffineFunction([MOI.ScalarAffineTerm(T(1), bridge.x)], T(0))
+    lb, ub = ceil(Int, ret[1]), floor(Int, ret[2])
+    N = floor(Int, log2(ub - lb)) + 1
+    for i in 1:N
+        y, _ = MOI.add_constrained_variable(model, MOI.ZeroOne())
+        push!(bridge.y, y)
+        push!(f.terms, MOI.ScalarAffineTerm(-(T(2)^(i - 1)), y))
+    end
+    bridge.ci = MOI.add_constraint(model, f, MOI.EqualTo{T}(lb))
```
blegat, line 165: The variable bridge spirit would be to substitute `x`. Here it seems to be a bad idea because it `x` is a  big sum.
This is like bridge a constrained variables `x in Int` by first adding the variable free and constraining it, then functionizing it and then slack bridge and then variable bridge for the slack. I'm wondering if there could be a nice way to implement this as a variable bridge but defaulting it to be used like that while allowing the user to have `x` be substituted as well.

blegat: Could you just drop a comment that it would more naturally be implemented as a variable bridge but we don't want to substitute every occurence of `x` so that's why we do it as a constraint bridge ?

src/Bridges/Constraint/bridges/integer_to_zeroone.jl
Vaibhavdixit02: One of `typemax(T)` or `typemin(T)`

odow: Most solvers will treat variables as unbounded if the bounds are > 1e20 or so (sometimes 1e30, sometimes 1e101).

You should never set the bounds as `typemin(T)` or `typemax(T)`. And even if you did, we can't reformulate that as binary.

src/Bridges/Constraint/bridges/integer_to_zeroone.jl
odow: I don't understand the SplitInterval reference. The SplitInterval bridge doesn't rely on a finite domain, it works with one-sided intervals. In comparison, we can't bridge an open set to ZeroOne (unless we pick a big M and do a 20 or 30 bit expansion?).

src/Bridges/Constraint/bridges/integer_to_zeroone.jl
odow: > I'm wondering if there could be a nice way to implement this as a variable bridge

That would only be necessary for a solver supports _only_ binary variables and not real. We don't really have any current examples, although @ccoffrin might be interested for future quantum annealing formulations. (They have some pretty specific structure restrictions though, which mean doing it naively might be the wrong thing to do.)

odow:
Is there anything else to do here?

PR #2206, opened 2023-06-12
[Utilities] refactor promote_operation

### LINKED ISSUE #2203: Re-factor promote_operation and operate
There are a bunch of rules for various `promote_operation` and `operate`. 
https://github.com/jump-dev/MathOptInterface.jl/blob/6b2192a991bab65a087284fcba6493de09bf786d/src/Utilities/functions.jl#L1597-L1610
https://github.com/jump-dev/MathOptInterface.jl/blob/6b2192a991bab65a087284fcba6493de09bf786d/src/Utilities/functions.jl#L1697-L1715

Defining a new function is really hard because it requires whack-a-mole: https://github.com/jump-dev/MathOptInterface.jl/pull/2201

I think we need an explicit write up of what functions are implemented, along with tests, so that we can check everything is working as intended.

odow:
This PR consolidates the `Utilities.promote_operation` methods into a single file with tests.

<s>This is going to be a WIP until I settle on a nice design.</s>

### DIFF
--- a/src/Utilities/promote_operation.jl
+++ b/src/Utilities/promote_operation.jl
@@ -4,50 +4,50 @@
 # Use of this source code is governed by an MIT-style license that can be found
 # in the LICENSE.md file or at https://opensource.org/licenses/MIT.
 
-# There are five methods for which we implement `Utilities.promote_operation`
-# and `Utilities.operate`:
-#
-#  1. `+`
-#     a. `promote_operation(::typeof(+), ::Type{T}, ::Type{F1}, ::Type{F2})`
-#  2. `-`
-#     a. `promote_operation(::typeof(-), ::Type{T}, ::Type{F})`
-#     b. `promote_operation(::typeof(-), ::Type{T}, ::Type{F1}, ::Type{F2})`
-#     c. `promote_operation(::typeof(-), ::Type{T}, ::Type{F1}, ::Type{Vector{T}})`
-#  3. `*`
-#     a. `promote_operation(::typeof(*), ::Type{T}, ::Type{T}, ::Type{F})`
-#     b. `promote_operation(::typeof(*), ::Type{T}, ::Type{F1}, ::Type{F2})`
-#        where `F1` and `F2` are `VariableIndex` or `ScalarAffineFunction`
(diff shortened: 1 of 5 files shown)


odow:
This is looking okay.

I'll tackle the `operate!` issue in a separate PR, which is also blocked by https://github.com/jump-dev/MathOptInterface.jl/pull/2207.

src/Utilities/promote_operation.jl
```diff
@@ -0,0 +1,426 @@
...
+of the arguments `args` are `ArgsTypes`.
+
+There are five methods for which we implement `Utilities.promote_operation`:
+
+ 1. `+`
+    a. `promote_operation(::typeof(+), ::Type{T}, ::Type{F1}, ::Type{F2})`
+ 2. `-`
+    a. `promote_operation(::typeof(-), ::Type{T}, ::Type{F})`
+    b. `promote_operation(::typeof(-), ::Type{T}, ::Type{F1}, ::Type{F2})`
+    c. `promote_operation(::typeof(-), ::Type{T}, ::Type{F1}, ::Type{Vector{T}})`
```
blegat, line 24: Why not 
```julia
`promote_operation(::typeof(-), ::Type{T}, ::Type{F1}, ::Type{T})`
```
then ?

src/Utilities/promote_operation.jl
```diff
@@ -0,0 +1,426 @@
...
+ 3. `*`
+    a. `promote_operation(::typeof(*), ::Type{T}, ::Type{T}, ::Type{F})`
+    b. `promote_operation(::typeof(*), ::Type{T}, ::Type{T}, ::Type{F})`
```
blegat, line 27:
```suggestion
    b. `promote_operation(::typeof(*), ::Type{T}, ::Type{F}, ::Type{T})`
```

src/Utilities/promote_operation.jl
odow: We actually did't need 2c as a case. It was already covered by 2b.

odow:
Merging because this is clearly a win

PR #2207, opened 2023-06-13
Fix incorrect usage of Utilities.operate!

odow:
`operate!` may modify its argument, but it may also return a new object. These usages were buggy and incorrect. (Although, correct for now, but will break with {Vector,Scalar}NonlinearFunction

### DIFF
--- a/src/Bridges/Constraint/bridges/count_belongs.jl
+++ b/src/Bridges/Constraint/bridges/count_belongs.jl
@@ -312,7 +312,7 @@ function MOI.Bridges.final_touch(
             end
         end
     end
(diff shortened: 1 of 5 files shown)


odow:
@blegat so much of the bridging/utilities system is premised on the structure of our Affine/Quadratic functions. There is supposed flexibility there, but we weren't actually testing it because none of the types needed it.

blegat (approved):

odow:
> Indeed, to be consistent with MutableArithmetics and BangBang, we should have called it operate!!

Added to https://github.com/jump-dev/MathOptInterface.jl/issues/2180

PR #2208, opened 2023-06-19
[Utillities] refactor operate, operate!, operate_term, operate_terms, operate_output_index!

odow:
Part of #2203

Still need to work on a bunch of methods


odow:
@blegat I've now moved all of the `Utilities.operate_xxx` methods into operate.jl with documentation and tests. I found quite a few missing methods and quirks, but things are in a nice state now.

There are only two `operate` methods left in `functions.jl`, which are for `ScalarNonlinearFunction`. I'll leave them for now, and come back in a new PR with the new methods. 

We should also revisit the diversity of names. I think `operate_coefficient` could be deprecated in favor of `operate_term`.

And we can potentially remove some methods in favor of the methods in `mutable_arithmetics.jl`. 

But they're all smaller changes for a new PR.

blegat:
> We should also revisit the diversity of names. I think `operate_coefficient` could be deprecated in favor of `operate_term`.

We can't really deprecate before MOI v2

> And we can potentially remove some methods in favor of the methods in `mutable_arithmetics.jl`.

Yes, the only advantage of `operate` is the easy dispatch since we know `T` but it was developed before MA so MA replaces most of it

odow:
> We can't really deprecate before MOI v2

Yeah. But we can forward the methods and add a note to the docstring.

odow:
Merging because this is a bit hard to review, but it's a clear improvement to the code base.

PR #2211, opened 2023-06-20
[Bridges] Add MOI.Bridge.FirstBridge attribute

blegat:
I find this useful when debugging. Getting the second bridge doesn't make much sense since the first bridge might create several constraints.

### DIFF
--- a/src/Bridges/Constraint/bridges/functionize.jl
+++ b/src/Bridges/Constraint/bridges/functionize.jl
@@ -42,6 +42,7 @@ function MOI.get(
     return MOI.Utilities.canonical(f)
 end
 
+# Needed to avoid an ambiguity with the getter for MOI.AbstractConstraintAttribute
 function MOI.get(
     ::MOI.ModelLike,
     ::MOI.Bridges.FirstBridge,
--- a/src/Bridges/Constraint/bridges/functionize.jl
+++ b/src/Bridges/Constraint/bridges/functionize.jl
@@ -42,6 +42,14 @@ function MOI.get(
     return MOI.Utilities.canonical(f)
 end
 
+function MOI.get(
+    ::MOI.ModelLike,
+    ::MOI.Bridges.FirstBridge,
+    bridge::AbstractFunctionConversionBridge,
+)
+    return bridge
+end
+
 function MOI.supports(
     model::MOI.ModelLike,
     attr::MOI.AbstractConstraintAttribute,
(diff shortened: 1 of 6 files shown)


odow:
What's the status of this? I don't know if we need this in MOI

blegat:
This is often useful for debugging but I was waiting for feedback before continuing

odow:
Is it useful though? For debugging it seems like you could just use the various `.model` etc.

blegat:
> Is it useful though? For debugging it seems like you could just use the various `.model` etc.

Yes, but it's always a bit hacky, takes time and depends on how many layers and so on. So you can't use in a tutorial for instance.

odow:
Why would we want to use this in a tutorial? How exactly does it help debugging?

odow:
What do you want to do with this? Shall I fix the test and add? Or shall we close?

src/Utilities/functions.jl
```diff
@@ -221,7 +221,15 @@ constraint functions, attribute values and submittable values. If you define a
...
+function map_indices(
+    variable_map::AbstractDict{T,T},
+    attr::MOI.AnyAttribute,
+    x::X,
+)::X where {T<:MOI.Index,X}
+    return map_indices(Base.Fix1(getindex, variable_map), attr, x)
+end
 
 # RawOptimizerAttribute values are passed through un-changed.
 map_indices(::Any, ::MOI.RawOptimizerAttribute, x) = x
```
odow, line 235:
```suggestion
map_indices(::Function, ::MOI.RawOptimizerAttribute, x) = x
```

odow:
@blegat merge or close?

src/Bridges/Constraint/bridges/functionize.jl
```diff
@@ -42,6 +42,14 @@ function MOI.get(
     return MOI.Utilities.canonical(f)
 end
 
+function MOI.get(
+    ::MOI.ModelLike,
+    ::MOI.Bridges.FirstBridge,
+    bridge::AbstractFunctionConversionBridge,
```
blegat, line 48: This is to avoid an ambiguity ? If so we should add a comment

src/Bridges/Constraint/bridges/functionize.jl
```diff
@@ -42,6 +42,14 @@ function MOI.get(
     return MOI.Utilities.canonical(f)
 end
 
+function MOI.get(
```
odow, line 45:
```suggestion
# Needed to avoid an ambiguity with the getter for MOI.AbstractConstraintAttribute
function MOI.get(
```

PR #2212, opened 2023-06-20
[Utilities] simplify various operate_ methods

odow:
https://github.com/jump-dev/MathOptInterface.jl/pull/2208 moved a bunch of methods around and added tests, but I didn't change any behavior or merge the functionality of different methods.

This PR adds a generic `operate_term` method (which was really `operate_coefficient`), and then leverages that to simplify a number of other methods, in some cases (like univariate -) removing the specialized methods all together.

I've also "deprecated" (in documentation only) `operate_terms!` which was only used in three places in mutable arithmetics. We should aim for fewer simpler methods, instead of a potpourri of methods that we've added for special cases.

### DIFF
--- a/src/Utilities/operate.jl
+++ b/src/Utilities/operate.jl
@@ -271,7 +271,7 @@ function operate(
         MOI.ScalarQuadraticFunction{T},
         MOI.VectorAffineFunction{T},
         MOI.VectorQuadraticFunction{T},
-    }
+    },
 ) where {T}
     return operate_coefficients(-, f)
 end
--- a/src/Utilities/operate.jl
+++ b/src/Utilities/operate.jl
@@ -1061,22 +1061,11 @@ end
 function operate!(
     ::typeof(/),
     ::Type{T},
-    f::Union{MOI.ScalarAffineFunction{T},MOI.ScalarQuadraticFunction{T}},
+    f::MOI.ScalarAffineFunction{T},
     g::T,
 ) where {T}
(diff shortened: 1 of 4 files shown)


src/Utilities/operate.jl
```diff
@@ -1072,23 +1072,23 @@ end
...
     ::typeof(/),
     ::Type{T},
-    f::MOI.ScalarQuadraticFunction{T},
+    f::Union{MOI.VectorAffineFunction{T},MOI.VectorQuadraticFunction{T}},
     g::T,
 ) where {T}
-    f.affine_terms .= operate_term.(/, f.affine_terms, g)
-    f.quadratic_terms .= operate_term.(/, f.quadratic_terms, g)
-    f.constant /= g
+    map_terms!(term -> operate_term(/, term, g), f)
```
odow, line 1090: Oh the methods just keep coming. I've now found `map_terms!`...

src/Utilities/operate.jl
```diff
@@ -286,21 +266,14 @@ end
...
-function operate(
-    ::typeof(-),
-    ::Type{T},
-    f::MOI.VectorQuadraticFunction{T},
+    f::Union{
+        MOI.ScalarAffineFunction{T},
+        MOI.ScalarQuadraticFunction{T},
+        MOI.VectorAffineFunction{T},
+        MOI.VectorQuadraticFunction{T},
+    }
```
odow, line 274:
```suggestion
    },
```

PR #2213, opened 2023-06-21
[Utilities] tidy Base overloads in functions.jl

odow:
@blegat there's some inconsistent behavior with how we deal with mixed coefficient types. What would you like to happen?

Here's the latest release version (NOT this PR):
```julia
(jmp) pkg> st -m MathOptInterface
      Status `/private/tmp/jmp/Manifest.toml`
  [b8f27783] MathOptInterface v1.17.1

MOI.VariableIndex(1)

julia> 2 * x
(0) + (2) MOI.VariableIndex(1)

julia> (2 * x) + 3
(3) + (2) MOI.VariableIndex(1)

julia> (2 * x) + 3.0
ERROR: MethodError: no method matching +(::MathOptInterface.ScalarAffineFunction{Int64}, ::Float64)
[...]

julia> (2 * x) * 3.0
0.0 + 6.0 MOI.VariableIndex(1)

julia> (2 * x) - 3.0
ERROR: MethodError: no method matching -(::MathOptInterface.ScalarAffineFunction{Int64}, ::Float64)
[...]

julia> (2 * x) / 3.0
ERROR: MethodError: no method matching /(::MathOptInterface.ScalarAffineFunction{Int64}, ::Float64)
[...]
```


odow:
Part of the complexity here because of

https://github.com/jump-dev/MathOptInterface.jl/blob/9e3a1eece6df80b327c2f0352db37f8d900fb22f/test/Utilities/functions.jl#L30-L50

Do we really want to define `*` between things that are not numbers? What does that even mean? Do we have any use-cases of a function with non-number coefficients? How do we know an arithmetic exists?

blegat:
I don't know of any use-case. We had so many issues with upstream expecting things to be number, e.g., in https://github.com/jump-dev/MutableArithmetics.jl/issues/47 that I wanted to stay away from it when possible. However for MOI coefficients it's probably fine to assumes the coefficients are numbers

odow:
Do we want `+` and `-` to also work for non-numbers? I'll leave `*` as-is for now, but add a comment about removing it in MOI 2.0.

odow:
Left as-is for now. There hasn't been a need to add more methods for `Base.:+` and `Base.:-`, so let's hold off until there is a reason.

PR #2214, opened 2023-06-21
[Utilities] tidy methods in functions.jl

odow:
Continuing work related to https://github.com/jump-dev/MathOptInterface.jl/issues/2203


blegat:
Where was `constant_vector` actually used ?

odow:
> Where was constant_vector actually used ?

It wasn't :laughing: the things you find when you go spelunking.

odow:
No one else used it either: https://juliahub.com/ui/Search?q=constant_vector&type=code

But StochasticPrograms extended it just for kicks. Likely because they went through and did everything, even though some isn't useful.

blegat:
It was renamed from `constant`. It might be that I renamed it but kept it so as to not be breaking even if I wanted to get rid of it.

PR #2218, opened 2023-06-22
[Utilities] fix various utilities for ScalarNonlinearFunction

odow:
Now that the functions.jl file is a lot cleaner, some things stick out: https://github.com/jump-dev/MathOptInterface.jl/issues/2203.

### DIFF
--- a/src/Utilities/functions.jl
+++ b/src/Utilities/functions.jl
@@ -92,6 +92,8 @@ function value_type(
     return MA.promote_operation(*, C, T)
 end
 
+value_type(::Type{T}, ::Type{MOI.ScalarNonlinearFunction}) where {T} = T
+
 function value_type(
     ::Type{T},
     ::Type{F},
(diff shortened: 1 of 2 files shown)


odow:
We're also going to need `eval_variables`. But this is going to require `eval_variables(value_fn, model, f)`, so I'll leave that for a new PR.

src/Utilities/functions.jl
```diff
@@ -294,6 +296,7 @@ function map_indices(
     index_map::F,
     f::MOI.ScalarNonlinearFunction,
 ) where {F<:Function}
+    # TODO(odow): this uses recursion. We should remove at some point.
```
odow, line 299: I know these use recursion. We can fix in future. Very few models should exceed the recursion depth.

src/Utilities/functions.jl
```diff
@@ -796,6 +810,10 @@ is_canonical(::MOI.AbstractFunction) = false
 
 is_canonical(::Union{MOI.VariableIndex,MOI.VectorOfVariables}) = true
 
+function is_canonical(f::MOI.ScalarNonlinearFunction)
+    return all(is_canonical(arg) for arg in f.args)
```
blegat, line 814: It seems we now need to declare `is_canonical(::Number) = true` ^^

odow: Haha. I never run the full tests locally, just the file I'm working on.

I also switched to a non-recursive implementation, since this will get called for all `add_constraint`.

PR #2219, opened 2023-06-22
[Utilities] add eval_variables support for ScalarNonlinearFunction

odow:
Related to https://github.com/jump-dev/MathOptInterface.jl/pull/2218

Blocker for https://github.com/jump-dev/MathOptInterface.jl/pull/2201

The implementation leaves a bit to be desired (it uses recursion, and every call needs to create an operator registry), but we can address that later.

### DIFF
--- a/src/Utilities/results.jl
+++ b/src/Utilities/results.jl
@@ -26,10 +26,9 @@ function get_fallback(model::MOI.ModelLike, attr::MOI.ObjectiveValue)
     MOI.check_result_index_bounds(model, attr)
     F = MOI.get(model, MOI.ObjectiveFunctionType())
     f = MOI.get(model, MOI.ObjectiveFunction{F}())
-    obj = eval_variables(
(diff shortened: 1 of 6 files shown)


src/Utilities/results.jl
```diff
@@ -26,10 +26,9 @@ function get_fallback(model::MOI.ModelLike, attr::MOI.ObjectiveValue)
     MOI.check_result_index_bounds(model, attr)
     F = MOI.get(model, MOI.ObjectiveFunctionType())
     f = MOI.get(model, MOI.ObjectiveFunction{F}())
-    obj = eval_variables(
-        vi -> MOI.get(model, MOI.VariablePrimal(attr.result_index), vi),
-        f,
-    )
+    obj = eval_variables(model, f) do vi
+        return vi -> MOI.get(model, MOI.VariablePrimal(attr.result_index), vi)
```
odow, line 30:
```suggestion
        return MOI.get(model, MOI.VariablePrimal(attr.result_index), vi)
```

PR #2231, opened 2023-06-28
[FileFormats.NL] fix nested scalar affine and quadratic functions

odow:
Missed a case from https://github.com/jump-dev/MathOptInterface.jl/pull/2228


odow:
With this change:
 * MINLPTests.jl now passing: https://github.com/jump-dev/MINLPTests.jl/pull/38
 * AmplNLWriter.jl now passing: https://github.com/jump-dev/AmplNLWriter.jl/pull/171

PR #2224, opened 2023-06-24
Fix method ambiguities and limit operate to Number subtypes

odow:
Closes https://github.com/jump-dev/MathOptInterface.jl/issues/2222

- `solver-tests`: https://github.com/jump-dev/MathOptInterface.jl/actions/runs/5361630981

PR #2233, opened 2023-06-29
[Bridges] add ScalarQuadraticToScalarNonlinearBridge

odow:
cc @chriscoey 

Necessary precursor to https://github.com/jump-dev/Pavito.jl/issues/67

## Basic

 - [x] Create a new file in `src/Bridges/XXX/bridges`
       struct must end in `Bridge`
       the `Bridge` suffix
 - [x] `include` the file in `src/Bridges/XXX/bridges/XXX.jl`

## Tests

       bridge
       untested lines.

## Documentation

 - [x] Add the docstring to `docs/src/submodules/Bridges/list_of_bridges.md`

## Final touch

the model:

 - [x] <s>Implement `MOI.Utilities.needs_final_touch(::Bridge)`</s>
 - [x] <s>Implement `MOI.Utilities.final_touch(::Bridge, ::MOI.ModelLike)`</s>
 - [x] <s>Ensure that `final_touch` can be called multiple times in a row</s>

### DIFF
--- a/src/Bridges/Constraint/bridges/functionize.jl
+++ b/src/Bridges/Constraint/bridges/functionize.jl
@@ -322,3 +322,111 @@ function MOI.get(
     f = MOI.get(model, attr, b.constraint)
     return MOI.Utilities.convert_approx(MOI.VectorOfVariables, f)
 end
+
+"""
+    ScalarQuadraticToScalarNonlinearBridge{T,S} <: Bridges.Constraint.AbstractBridge
+
+`ScalarQuadraticToScalarNonlinearBridge` implements the following reformulations:
+
+  * ``f(x) \\in S`` into ``g(x) \\in S``
+
+where `f` is a [`MOI.ScalarQuadraticFunction`](@ref) and `g` is a
+[`MOI.ScalarNonlinearFunction{T}`](@ref).
+
+## Source node
+
+`ScalarQuadraticToScalarNonlinearBridge` supports:
+
+  * [`MOI.ScalarQuadraticFunction`](@ref) in `S`
+
+## Target nodes
+
+`ScalarQuadraticToScalarNonlinearBridge` creates:
+
+  * [`MOI.ScalarNonlinearFunction{T}`](@ref) in `S`
+"""
+struct ScalarQuadraticToScalarNonlinearBridge{T,S} <:
+       AbstractFunctionConversionBridge{MOI.ScalarNonlinearFunction,S}
+    constraint::MOI.ConstraintIndex{MOI.ScalarNonlinearFunction,S}
+end
+
+const ScalarQuadraticToScalarNonlinear{T,OT<:MOI.ModelLike} =
+    SingleBridgeOptimizer{ScalarQuadraticToScalarNonlinearBridge{T},OT}
+
+function bridge_constraint(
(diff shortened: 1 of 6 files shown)


src/Bridges/Constraint/bridges/functionize.jl
```diff
@@ -322,3 +322,111 @@ function MOI.get(
...
+
+function concrete_bridge_type(
+    ::Type{<:ScalarQuadraticToScalarNonlinearBridge{T}},
+    ::Type{MOI.ScalarQuadraticFunction{T}},
+    S::Type{<:MOI.AbstractScalarSet},
+) where {T}
+    return ScalarQuadraticToScalarNonlinearBridge{T,S}
+end
+
+function MOI.get(
```
blegat, line 394: This method and the next two should be implemented at the `AbstractFunctionConversion` level

blegat: But this can be done in a separate PR

blegat (approved):

src/Bridges/Constraint/bridges/functionize.jl
odow: Yeah there's the canonical issue as well. But we can come back to it with https://github.com/jump-dev/MathOptInterface.jl/pull/2224

odow:
I just removed the canonical method. It'll get added back generically by #2224

chriscoey:
thanks @odow! exciting. 
here's the code I was using to convert SNF to SAF/SQF previously (which may not work in future if `operate` supports SNF?):
```
# convert SNF to SAF/SQF{T}
function nl_to_aff_or_quad(::Type{T}, f::MOI.ScalarNonlinearFunction) where {T <: Real}
    args = nl_to_aff_or_quad.(T, f.args)
    if !any(Base.Fix2(isa, MOI.ScalarNonlinearFunction), args)
        if f.head == :^
            if length(args) == 2 && args[2] == 2
                return MOI.Utilities.operate(*, T, args[1], args[1])
            end
        else
            h = get(_quad_ops, f.head, nothing)
            isnothing(h) || return MOI.Utilities.operate(h, T, args...)
        end
    end
    return error("function $f cannot be converted to linear or quadratic form")
end

nl_to_aff_or_quad(::Type{<:Real}, f::MOI.VariableIndex) = f
nl_to_aff_or_quad(::Type{T}, f::T) where {T <: Real} = f
nl_to_aff_or_quad(::Type{T}, f::Real) where {T <: Real} = convert(T, f)

_quad_ops = Dict(:+ => +, :- => -, :* => *, :/ => /)
```
do you know off-hand if the SNF to SAF/SQF convert in this PR will handle all the cases the above code does? I'd love to be able to remove my hacky code when this PR lands.

odow:
Yeah that code will now just make a ScalarNonlinearFunction. Let me take another look.

odow:
So I guess one decision is how hard we want to fight to recognize `*(2, *(3, /(x, 2)))` = `2 * (3 * x / 2) = 3x` as an affine expression. Also things like `+(x, sin(1.0))`. Should we see `:sin` and error? Or recognize that it's actually a constant.

Perhaps this would be easier if we had a smart `canonicalize` function which grouped like terms and lifted nested `+` and `*` etc.

odow:
That'll do for now. We can revisit `convert` is a separate PR if needed.

PR #2235, opened 2023-06-29
[Bridges] add Constraint.FunctionConversionBridge

odow:
x-ref #2233


odow:
So I was simplifying things, when I realized that there was a better design (`FunctionConversionBridge`). It seems pointless to have both, but the new bridge makes it easy to cover every transform.

I guess our options are:

 * Don't add. Keep what we have
 * Keep old and add new
 * Change the signature of `AbstractFunctionConversionBridge` to accept `G` so that we can convert both ways.

docs/src/submodules/Bridges/list_of_bridges.md
```diff
@@ -28,7 +28,7 @@ Bridges.Constraint.ScalarSlackBridge
 Bridges.Constraint.VectorSlackBridge
 Bridges.Constraint.ScalarFunctionizeBridge
 Bridges.Constraint.VectorFunctionizeBridge
-Bridges.Constraint.ScalarQuadraticToScalarNonlinearBridge
+Bridges.Cosntrainnt.FunctionConversionBridge
```
odow, line 31:
```suggestion
Bridges.Constraint.FunctionConversionBridge
```

odow:
I've moved some stuff around. We could leave removing `ScalarFunctionizeBridge` and `VectorFunctionizeBridge` for MOI 2.0.

odow:
https://github.com/jump-dev/MathOptInterface.jl/actions/runs/5417795253

src/Bridges/Constraint/bridges/functionize.jl
```diff
@@ -63,6 +63,71 @@ function MOI.set(
...
+    return
+end
+
+function MOI.get(
+    model::MOI.ModelLike,
+    attr::MOI.CanonicalConstraintFunction,
+    b::AbstractFunctionConversionBridge,
+)
+    f = MOI.get(model, MOI.ConstraintFunction(), b)
+    return MOI.Utilities.canonicalize!(f)
```
blegat, line 103: It's not allowed to modify the output of `ConstraintFunction`

blegat:
> Change the signature of `AbstractFunctionConversionBridge` to accept `G` so that we can convert both ways.

We can add a new subtype in between to make it non-breaking.

It still make sense to have one different bridge by output function. Here, we have a chain: variable -> affine -> quadratic -> nl
so it makes sense to have a bridge that converts to the next one of the chain but it's not so composable.
For instance, PolyJuMP defines the polynomial function type which sits in between quadratic and nl.
If things are hardcoded then it's not so easy.
Maybe we could have something like
```julia
struct FunctionConversionBridge{T,F,G} ... end
const ScalarFunctionizeBridge{T} = FunctionConversionBridge{T,ScalarAffineFunction}
...
``
so that implementing a bridge is one-line but the final function is still specified by the bridge.
We would have to add a `convertible` function to the API to make this work though.

File: src/Bridges/Constraint/bridges/functionize.jl (hunk 2 above)
Comment (odow) [inline src/Bridges/Constraint/bridges/functionize.jl:103]:
```suggestion
    return MOI.Utilities.canonical(f)
```

odow:
It should work if you define:
```julia
function MOI.supports_constraint(
    ::Type{<:FunctionConversionBridge{T}},
    ::Type{PolyJuMP.Polynomial{T}},
    ::Type{<:MOI.AbstractScalarSet},
) where {T}
    return true
end

function concrete_bridge_type(
    ::Type{<:FunctionConversionBridge{T}},
    ::Type{PolyJuMP.Polynomial{T}},
    ::Type{S},
) where {T,S<:MOI.AbstractScalarSet}
    return FunctionConversionBridge{
        T,
        PolyJuMP.Polynomial{T},
        MOI.ScalarNonlinearFunction,
        S,
    }
end
```

blegat:
That converts a polynomial to nonlinear but how do you create a bridge from nonlinear to polynomial ?
Also, since it's only one bridge, it makes it difficult to use `remove_bridge` to disable one particular conversion.

odow:
> but how do you create a bridge from nonlinear to polynomial

You can't, because that's not possible in general?

> since it's only one bridge, it makes it difficult to use remove_bridge to disable one particular conversion.

This is a reasonable point. But it's pretty niche...

blegat:
> > but how do you create a bridge from nonlinear to polynomial
>
> You can't, because that's not possible in general?

This is exactly the goal of quadratic to RSOC. Because a conic solver does not support quadratic and this bridge **might** then you better try


> > since it's only one bridge, it makes it difficult to use remove_bridge to disable one particular conversion.
>
> This is a reasonable point. But it's pretty niche...

My point with these two examples is that these are the tip of the iceberg pointing towards the fact that we're loosing some of the nice extensibility features of the bridges here. And this comes from the fact that we hardcode the next function to convert to in the chain of conversions.

odow:
You could write a different bridge for SNF to polynomial. This bridge is a simple attempt at covering common use cases. I can't think of a situation in which you might like to disable ScalarAffine -> ScalarNonlinear, for example.

blegat:
> This bridge is a simple attempt at covering common use cases

It's actually as simple to cover all use cases.
It's also that hardcoding the conversion like that is quite unnatural for bridges. What is a bridge doing "all conversions" ? It's quite ill defined. It's much more natural to write a bridge "conversion to affine" or "conversion to quadratic" or "conversion to nonlinear".
We are actually disabling the "conversion from SOC to quadratic" by default and the argument was that SOC is convex while the quadratic formulation could be solved with a nonlinear program that may not guarantee global optimality and it thought this could be counter intuitive. The same argument could apply to the "linear -> nonlinear" bridge actually.

You PR is a big improvement over the current bridges and you does not need much change to become `FunctionConvertionBridge{T,F}; const ScalarAffineFunctionConvertionBridge{T} = FunctionConversionBridge{T,ScalarAffineFunction{T}}; ...`

blegat:
Once we have a function conversion like that, it's also easy to add something like function conversion from SNF to a `DisciplineConvexFunction` that have meta information about convexity like Convex.jl.
Then, a solver could choose to support `DCF` and it will be bridged for them.
Then, we can add bridges from `DCF` -> separate conic constraints by adding slacks as well.
With black box functions, if we go with three functions: 0th order black box, 1st order black box and 2nd order black box, there are also trivial conversions there.
I'm not saying we will define these functions in the future, I'm just saying I don't want to bet that we will define none. And the design of this PR is already closing many doors I want leave open (already for PolyJuMP). Of course we could then create a new bridge but then that will be a bit confusing and inconsistent and the goal of trying to make things simpler of this PR (which I understand) would have an opposite effect.

odow:
> You PR is a big improvement over the current bridges and you does not need much change to become

This doesn't fix the deletion issue though?

odow:
Shall we hold off on this PR for a bit then? I originally wanted to get it in for v1.19, but if `ScalarQuadraticToScalarNonlinearBridge` is going to stay, then there is no rush.

blegat:
Which deletion issue ? It's not going to stay, it's going to be replaced by ToScalarNonlinearBridge that converts from anything to SNF, not only quadratic

odow:
Maybe we should call about this. I thought you wanted separate bridges for each change, rather than one that tried everything.

blegat:
IIUC, this PR is "Any-to-Any", the current status is "fixed function-to-fixed function" and I am suggesting "Any-to-fixed function"

odow:
> IIUC, this PR is "Any-to-Any", the current status is "fixed function-to-fixed function" and I am suggesting "Any-to-fixed function"

I'd say this PR is "F-to-G" for an explicit list of functions. The current status is "scalarquadratic to scalarnonlinear." I don't think we want an any-to-fix because it just leads to weird errors that are hard to interpret.

src/Bridges/Constraint/bridges/functionize.jl
```diff
@@ -236,188 +263,144 @@ function MOI.supports_constraint(
...
+
 function MOI.supports_constraint(
-    ::Type{ScalarQuadraticToScalarNonlinearBridge{T}},
-    ::Type{MOI.ScalarQuadraticFunction{T}},
+    ::Type{<:FunctionConversionBridge{T}},
+    ::Type{MOI.ScalarAffineFunction{T}},
     ::Type{<:MOI.AbstractScalarSet},
 ) where {T}
     return true
 end
```
odow, line 345: @blegat, functions for this bridge are opt-in, you need to explicitly implement `supports_constraint` and `concrete_bridge_type`.

src/Bridges/Constraint/Constraint.jl
```diff
@@ -84,10 +84,7 @@ function add_all_bridges(bridged_model, ::Type{T}) where {T}
     MOI.Bridges.add_bridge(bridged_model, VectorSlackBridge{T})
     MOI.Bridges.add_bridge(bridged_model, ScalarFunctionizeBridge{T})
     MOI.Bridges.add_bridge(bridged_model, VectorFunctionizeBridge{T})
-    MOI.Bridges.add_bridge(
-        bridged_model,
-        ScalarQuadraticToScalarNonlinearBridge{T},
-    )
+    MOI.Bridges.add_bridge(bridged_model, FunctionConversionBridge{T})
```
odow, line 87: @blegat, don't you need to explicitly add the bridges here?

docs/src/submodules/Utilities/reference.md
```diff
@@ -154,6 +154,7 @@ Utilities.get_fallback
 The following utilities are available for functions:
 
 ```@docs
+supports_convert
```
odow, line 157: This seems like it is in the wrong place.

odow:
https://github.com/jump-dev/MathOptInterface.jl/actions/runs/5791304222

odow:
https://github.com/jump-dev/MathOptInterface.jl/actions/runs/5792030974

odow:
Hypatia's failure in solver-tests is...confusing. 
```julia
import Hypatia

model = MOI.instantiate(
    () -> Hypatia.Optimizer{Float64}(; default_tol_relax = 4);
    with_bridge_type = Float64,
)
MOI.set(model, MOI.Silent(), true)
MOI.Test.runtests(
    model,
    MOI.Test.Config(
        Float64,
        atol = 1e-4,
        rtol = 1e-4,
        exclude = Any[MOI.SolverVersion],
    ),
)
```

leads to

```Julia
test_conic_Exponential_hard: Error During Test at /Users/oscar/.julia/dev/MathOptInterface/src/Test/Test.jl:229
  AssertionError: dim >= 3
  Stacktrace:
    [1] EpiPerSquare
      @ ~/.julia/dev/Hypatia/src/Cones/epipersquare.jl:42 [inlined]
    [2] cone_from_moi(#unused#::Type{Float64}, cone::MathOptInterface.RotatedSecondOrderCone)
      @ Hypatia ~/.julia/dev/Hypatia/src/MathOptInterface/cones.jl:50
    [3] copy_to(opt::Hypatia.Optimizer{Float64}, src::MathOptInterface.Utilities.UniversalFallback{MathOptInterface.Utilities.Model{Float64}})
      @ Hypatia ~/.julia/dev/Hypatia/src/MathOptInterface/wrapper.jl:203
    [4] optimize!
      @ ~/.julia/dev/MathOptInterface/src/MathOptInterface.jl:84 [inlined]
    [5] optimize!(m::MathOptInterface.Utilities.CachingOptimizer{Hypatia.Optimizer{Float64}, MathOptInterface.Utilities.UniversalFallback{MathOptInterface.Utilities.Model{Float64}}})
      @ MathOptInterface.Utilities ~/.julia/dev/MathOptInterface/src/Utilities/cachingoptimizer.jl:316
    [6] optimize!
      @ ~/.julia/dev/MathOptInterface/src/Bridges/bridge_optimizer.jl:376 [inlined]
    [7] test_conic_Exponential_hard(model::MathOptInterface.Bridges.LazyBridgeOptimizer{MathOptInterface.Utilities.CachingOptimizer{Hypatia.Optimizer{Float64}, MathOptInterface.Utilities.UniversalFallback{MathOptInterface.Utilities.Model{Float64}}}}, config::MathOptInterface.Test.Config{Float64})
      @ MathOptInterface.Test ~/.julia/dev/MathOptInterface/src/Test/test_conic.jl:3667
```

where did that `MathOptInterface.RotatedSecondOrderCone` come from?!!?!

cc @chriscoey @lkapelevich

blegat:
This is scary, it might be that the bridges decided to go `SAF`-in-`LessThan` -> `SQF`-in-`LessThan` -> `VAF`-in-`RSOC` instead of `SAF`-in-`LessThan` -> `VAF`-in-`Nonpositives` -> ...
We probably need to introduce bridge weights...

odow:
https://github.com/jump-dev/MathOptInterface.jl/actions/runs/5801988551

This PR is getting a little out of hand... at minimum, we need to split out the bridge cost into a separate PR.

odow:
I actually wonder if we even need `ToScalarQuadraticBridge`. It just destroys structure. 

The only real use-cases we have are for `ToScalarNonlinear` and `ToScalarAffine`. There is no reasonable use-case for `ToScalarQuadratic`.

odow:
Adding bridge weights needs a lot more thought than as a hack (even a good one) to make this particular case work.

blegat:
We could also not add `ToScalarQuadraticBridge` by default indeed but then also `ToVectorQuadraticBridge` since it can be bridged with `Scalarize` then `QuadToSOC`.
For `ToScalarNonlinearBridge`, there is also a risk.
Suppose the solver supports `VectorAffineFunction`-in-`Nonnegatives` and `ScalarNonlinearFunction`-in-`LessThan`. Then given a `ScalarAffineFunction`-in-`LessThan`, it will bridge it to the wrong one.

odow:
> Suppose the solver supports VectorAffineFunction-in-Nonnegatives and ScalarNonlinearFunction-in-LessThan

We don't have such a solver

odow:
Maybe we should just close this for now. I only did the refactoring because I thought it'd make things simpler. But there are unintended consequences, and I don't know that we need to open this box just yet.

blegat:
> > Suppose the solver supports VectorAffineFunction-in-Nonnegatives and ScalarNonlinearFunction-in-LessThan
> 
> We don't have such a solver

Not that we know of but that means it is breaking.

blegat:
> Maybe we should just close this for now. I only did the refactoring because I thought it'd make things simpler. But there are unintended consequences, and I don't know that we need to open this box just yet.

I think this is useful and the PR is passing all tests.

I don't feel the bridges cost are used as a hack. It make sense to have higher xost when you destroy structure. I would simply say that the cost are subject to changes

odow:
> Then given a ScalarAffineFunction-in-LessThan, it will bridge it to the wrong one

It's also not "wrong" it's just less efficient than the VAF-in-nonnegatives. I don't think that adding a bridge cost and getting a different formulation is breaking.

odow:
@mlubin you might need to take a look.

Currently, we have separate bridges for
   * VariableIndex->ScalarAffineFunction
   * VectorOfVariables->VectorAffineFunction
   * ScalarQuadraticFunction->ScalarNonlinearFunction

The last bridge is unreleased, so we can change without breaking changes. If we tag a new release, then we're stuck with it.

The original goal of this PR was to unify the function conversions, to end up with bridges for
  * VariableIndex->ScalarAffineFunction
  * ScalarAffineFunction->ScalarQuadraticFunction
  * ScalarQuadraticFunction->ScalarNonlinearFunction

and matching ones for the Vector functions.

We then changed to a generic "ToXXXFunction" bridges, which 

  * {VariableIndex}->ScalarAffineFunction
  * {VariableIndex,ScalarAffineFunction}->ScalarQuadraticFunction
  * {VariableIndex,ScalarAffineFunctionScalarQuadraticFunction}->ScalarNonlinearFunction

Conceptually simple, these bridges expose a problem. For conic solvers, VariableIndex-in-LessThan used to get bridged in a two-step process to VectorAffineFunction-in-Nonpositives and then VectorAffineFunction-in-Nonnegatives.

But with the new ToScalarQuadraticBridge, it can also go to ScalarQuadraticFunction-in-LessThan and then VectorAffineFunction-in-RotatedSecondOrderCone. Both are two bridges, so we pick arbitrarily.

Options are:

 * Close this PR and leave the three special-case function conversions
 * Merge only the ToScalarNonlinearFunction, which would be useful for solvers like Patio (and other nonlinear solvers which didn't exploit linear/quadratic structure or could infer from graph). Doesn't resolve the equivalent bridging problem, but fixes most common issues.
 * Introduce weights into the bridge graph for the first time.

src/Bridges/debug.jl
```diff
@@ -39,7 +48,7 @@ function print_node_info(
 )
     print(io, " ")
     print_node(io, b, node)
-    d = _dist(b.graph, node)
+    d = _cost(b.graph, node)
```
odow, line 51: I'd like to revert all of these naming changes. Let's keep the minimal set of changes for this PR. `_dist` and `_cost` are synonymous anyway.

src/Bridges/graph.jl
```diff
@@ -4,7 +4,7 @@
 # Use of this source code is governed by an MIT-style license that can be found
 # in the LICENSE.md file or at https://opensource.org/licenses/MIT.
 
-const INFINITY = typemax(Int)
+const INFINITY = typemax(Float64)
```
odow, line 7: Why does the distance/cost need to be a `Float64`? Aren't integer weights sufficient?

src/functions.jl
```diff
@@ -836,6 +836,26 @@ end
...
+"""
+    conversion_cost(F::Type{<:MOI.AbstractFunction}, G::Type{<:MOI.AbstractFunction})
+
+Return a `Float64` returning the *cost* of converting any function of type `G`
+to a function of type `F` with `convert`. This cost is used to compute
+[`Bridges.bridging_cost`](@ref).
+This cost is `Inf` should be used to avoid the
+[`Bridges.Constraint.FunctionConversionBridge`](@ref) to attempt the conversion.
+"""
+function conversion_cost(::Type{<:AbstractFunction}, ::Type{<:AbstractFunction})
```
odow, line 848: This feels like it should be a method in the bridges, not in MOI in general. It's  weird special case that isn't generally applicable, and the values are arbitrary.

src/Bridges/graph.jl
odow: Because @blegat wanted to use `Inf` to mean that bridging shouldn't happen

src/Bridges/lazy_bridge_optimizer.jl
```diff
@@ -152,7 +152,12 @@ Return the `Edge` or `ObjectiveEdge` in the hyper-graph associated with the
 bridge `BT`, where `index` is the index of `BT` in the list of bridges.
 """
 function _edge(b::LazyBridgeOptimizer, index::Int, BT::Type{<:AbstractBridge})
-    return Edge(index, _variable_nodes(b, BT), _constraint_nodes(b, BT))
+    return Edge(
+        index,
+        _variable_nodes(b, BT),
+        _constraint_nodes(b, BT),
+        bridging_cost(BT),
```
odow, line 159: One question is whether this should be `bridging_cost(b, BT)`. Will we ever need to query something in the model?

odow:
I've reverted the name changes and the switch from `Int` to `Float64`. 

This make the PR a lot smaller and easier to digest. The change required to support bridge-specific costs is actually very minor.

Open questions would be:

 * How do we choose weights?
 * Are `Int` weights sufficient?

odow:
https://github.com/jump-dev/MathOptInterface.jl/actions/runs/5851380666

src/Bridges/graph.jl
blegat: `ConstraintBridgingCost` and `VariableBridgingCost` are `Float64` anyway. I also thought it might be useful to use fractional sometimes. I can imagine the case when you want to give priority for a bridge but you cannot make the cost below 1 so you're stuck because it is `Int`. With `Float64`, you won't have that kind of artificial limitations. I don't see any advantage of `Int` so the pros-and-cons are easy to compare.

src/functions.jl
blegat: Yes, we could move it to the bridges

src/Bridges/lazy_bridge_optimizer.jl
blegat: I thought the same. The current approach is to keep the bridging API simple and avoid `added_constraint_types`, `supports_constraint`, ... take the model as argument for to force the bridges to be simple and easier to reason about. The dependence with the model is solely handled by the bridged optimizer and the bridge graph

blegat:
> How do we choose weights?

It seems to be difficult to give rules. Note that in the future we could imagine the user changing the weights of a solver defined in MOI instead of having to do `remove_bridge` but we can wait for it to be useful.

> Are Int weights sufficient?

I can see it being annoying, see https://github.com/jump-dev/MathOptInterface.jl/pull/2235#discussion_r1293105562

odow:
We could change to `Float64`. Alternatively, we could make the default cost `100`, to give people room to go lower.

odow:
We spoke offline and decided to switch to `Float64`

odow:
One last run of `solver-tests`, and then I'm marking this as good to go:

https://github.com/jump-dev/MathOptInterface.jl/actions/runs/5860994743

mlubin:
Can we add tests that certain important combinations like `VariableIndex-in-LessThan` get bridged exactly as we expect with a hypothetical conic solver?

odow:
That's what this test is:
 https://github.com/jump-dev/MathOptInterface.jl/pull/2235/files#diff-0bac7c57a5e8c70eefceef8d3791b77dd904969489fd87b638be82f6f92e9efeR2050-R2092

although it's a little opaque. We could be more explicit.

mlubin:
Ah, I missed that. The test is named `test_bridging_cost` and it checks the numerical values of the bridging costs, so I assumed that was all it covered. It make sense to split this into multiple tests.

odow:
I've tweaked the tests to check that we get the costs expected, and that the internal model is built as expected.

mlubin (approved):

odow:
Merging because I discussed this change with @blegat last night. Once merged, I'll update the prep for v1.19, and he can double check things before we tag the release.

PR #2232, opened 2023-06-28
Prep for v1.19.0

odow:
## Basic

 - [x] `version` field of `Project.toml` has been updated
       - If a breaking change, increment the MAJOR field and reset others to 0
       - If adding new features, increment the MINOR field and reset PATCH to 0
       - If adding bug fixes or documentation changes, increment the PATCH field

## Documentation

 - [x] Add a new entry to `docs/src/changelog.md`, following existing style

## Tests

 - [x] The `solver-tests.yml` GitHub action does not have unexpected failures.
       https://github.com/jump-dev/MathOptInterface.jl/actions/runs/5399018088

PR #2249, opened 2023-08-08
[Bridges] mark ScalarQuadraticToScalarNonlinearBridge as experimental

odow:
#2232 has been sitting open for too long. It is blocked by #2235, which would remove this bridge. If we tagged 1.19 before #2235, then removing it would be breaking. So I've opted to leave the code in for now, but not add it by default. 

That should let us release v1.19 and take a longer look at #2235.


odow:
I don't like cherry-picked releases. We're not trying to conserve version numbers, so let's not add more complications.

odow:
Closing because we should make a decision on #2235

PR #2250, opened 2023-08-09
Add substitute

blegat:
See https://github.com/jump-dev/MathOptInterface.jl/pull/2201/files#r1288670625

### DIFF
--- a/src/Utilities/functions.jl
+++ b/src/Utilities/functions.jl
@@ -362,7 +362,7 @@ end
 # change in `substitute_variables`.
 
 """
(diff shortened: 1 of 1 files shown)


odow:
I'm not in favor of this. It just adds yet another method. We already have too many

odow:
See updated https://github.com/jump-dev/MathOptInterface.jl/pull/2201. I don't think we need this

blegat:
Yes, I think we should rather change the existing one with was a bad design but that's an MOI v2 thing

Issue #2183, opened 2023-05-30
x86 frequently fails in CI


odow:
Various PRs #2179 #2182 have triggered this StackOverflow on 32-bit. I've tried both Julia 1.6 and 1.9. 

<img width="1035" alt="image" src="https://github.com/jump-dev/MathOptInterface.jl/assets/8177701/92291e74-ed33-427d-8dd8-4d88d0c3a893">

odow:
Here's a different one

<img width="1064" alt="image" src="https://github.com/jump-dev/MathOptInterface.jl/assets/8177701/9ed83b18-0770-4d0a-a11e-288cefb322bb">

odow:
I think this settles it

<img width="1064" alt="image" src="https://github.com/jump-dev/MathOptInterface.jl/assets/8177701/31a86fcc-efb5-4337-949a-3b068bfb08b3">

blegat:
Why does the testset make such a difference

odow:
I haven't seen this fail recently, so I'm going to close. We can always re-open.

PR #2253, opened 2023-08-11
Partition tests into multiple CI jobs to reduce runtime and memory pressure

odow:


odow:
#2183 was still a problem. This change has two benefits: it reduces total CI time from ~1 hr to ~30 minutes, and it also reduces the peak memory usage. So I think a win-win.

mlubin (approved):
Nice!

PR #2257, opened 2023-08-17
adding ObjectiveLimit attribute

matbesancon:
This MR introduces an attribute to set an objective limit, after which the solver can stop.
This can be useful e.g. for separation routines, when one requires at least a certain objective value to be reached.

Setting this objective bound as a constraint in a feasibility formulation is often ill-advised and can lead to poor solver behaviour, most MIP solvers provide a dedicated parameter instead:
- Gurobi: https://www.gurobi.com/documentation/current/refman/bestobjstop.html#parameter:BestObjStop
- SCIP: `limits/objectivestop` is coming in the next major release
- HiGHS: objective_bound https://ergo-code.github.io/HiGHS/stable/options/definitions/#objective_bound

CPLEX seems to have [this](https://www.ibm.com/docs/en/icos/20.1.0?topic=parameters-lower-objective-value-limit) but it seems to be for LPs only?
Monotonically improving iterative solvers could also implement the attribute

### DIFF
--- a/src/Test/test_attribute.jl
+++ b/src/Test/test_attribute.jl
@@ -196,6 +196,8 @@ function test_attribute_ObjectiveLimit(model::MOI.AbstractOptimizer, ::Config)
     value = MOI.get(model, MOI.ObjectiveLimit())
     MOI.set(model, MOI.ObjectiveLimit(), 0.0)
     @test MOI.get(model, MOI.ObjectiveLimit()) == 0.0
+    MOI.set(model, MOI.ObjectiveLimit(), nothing)
+    @test MOI.get(model, MOI.ObjectiveLimit()) === nothing
     MOI.set(model, MOI.ObjectiveLimit(), 1.0)
     @test MOI.get(model, MOI.ObjectiveLimit()) == 1.0
     MOI.set(model, MOI.ObjectiveLimit(), value)
--- a/src/Test/test_attribute.jl
+++ b/src/Test/test_attribute.jl
@@ -190,6 +190,30 @@ function setup_test(
     return
 end
 
+function test_attribute_ObjectiveLimit(model::MOI.AbstractOptimizer, ::Config)
+    @requires MOI.supports(model, MOI.ObjectiveLimit())
+    # Get the current value to restore it at the end of the test
+    value = MOI.get(model, MOI.ObjectiveLimit())
+    MOI.set(model, MOI.ObjectiveLimit(), 0.0)
+    @test MOI.get(model, MOI.ObjectiveLimit()) == 0.0
+    MOI.set(model, MOI.ObjectiveLimit(), 1.0)
+    @test MOI.get(model, MOI.ObjectiveLimit()) == 1.0
+    MOI.set(model, MOI.ObjectiveLimit(), value)
+    @test value == MOI.get(model, MOI.ObjectiveLimit()) # Equality should hold
+    _test_attribute_value_type(model, MOI.ObjectiveLimit())
+    return
+end
+test_attribute_ObjectiveLimit(::MOI.ModelLike, ::Config) = nothing
+
+function setup_test(
+    ::typeof(test_attribute_ObjectiveLimit),
+    model::MOIU.MockOptimizer,
+    ::Config,
+)
+    MOI.set(model, MOI.ObjectiveLimit(), nothing)
+    return
+end
+
 """
     test_attribute_AbsoluteGapTolerance(model::MOI.AbstractOptimizer, config::Config)
 
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -851,6 +851,18 @@ struct TimeLimitSec <: AbstractOptimizerAttribute end
 attribute_value_type(::TimeLimitSec) = Union{Nothing,Float64}
 
 """
+    ObjectiveLimit()
+
+
+An optimizer attribute for setting a limit on the objective value.
+The solver may stop when the `ObjectiveValue` is better (lower for minimization, higher for maximization) than the `ObjectiveLimit`.
+When `set` to `nothing`, it removes the solver limit. The default value is `nothing`.
+"""
+struct ObjectiveLimit <: AbstractOptimizerAttribute end
+
+attribute_value_type(::ObjectiveLimit) = Union{Nothing,Float64}
+
+"""
     RawOptimizerAttribute(name::String)
 
 An optimizer attribute for the solver-specific parameter identified by `name`.
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -860,10 +860,10 @@ The provided limit must be a `Union{Float64,Nothing}`.
 When `set` to `nothing`, the limit reverts to the solver's default.
 
 The default value is `nothing`.
+
 The solver may stop when the [`ObjectiveValue`](@ref) is better (lower for
 minimization, higher for maximization) than the `ObjectiveLimit`. If stopped,
 the [`TerminationStatus`](@ref) should be `OBJECTIVE_LIMIT`.
-
 """
 struct ObjectiveLimit <: AbstractOptimizerAttribute end
 
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -853,10 +853,17 @@ attribute_value_type(::TimeLimitSec) = Union{Nothing,Float64}
 """
     ObjectiveLimit()
 
-
 An optimizer attribute for setting a limit on the objective value.
-The solver may stop when the `ObjectiveValue` is better (lower for minimization, higher for maximization) than the `ObjectiveLimit`.
-When `set` to `nothing`, it removes the solver limit. The default value is `nothing`.
+
+The provided limit must be a `Union{Float64,Nothing}`.
+
+When `set` to `nothing`, the limit reverts to the solver's default.
+
+The default value is `nothing`.
+The solver may stop when the [`ObjectiveValue`](@ref) is better (lower for
+minimization, higher for maximization) than the `ObjectiveLimit`. If stopped,
+the [`TerminationStatus`](@ref) should be `OBJECTIVE_LIMIT`.
+
 """
 struct ObjectiveLimit <: AbstractOptimizerAttribute end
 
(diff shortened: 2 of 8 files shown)


src/attributes.jl
```diff
@@ -850,6 +850,18 @@ struct TimeLimitSec <: AbstractOptimizerAttribute end
 
 attribute_value_type(::TimeLimitSec) = Union{Nothing,Float64}
 
+"""
+    ObjectiveLimit()
+
```
odow, line 855:
```suggestion
```

src/Test/test_attribute.jl
```diff
@@ -190,6 +190,30 @@ function setup_test(
     return
 end
 
+function test_attribute_ObjectiveLimit(model::MOI.AbstractOptimizer, ::Config)
+    @requires MOI.supports(model, MOI.ObjectiveLimit())
+    # Get the current value to restore it at the end of the test
+    value = MOI.get(model, MOI.ObjectiveLimit())
+    MOI.set(model, MOI.ObjectiveLimit(), 0.0)
+    @test MOI.get(model, MOI.ObjectiveLimit()) == 0.0
+    MOI.set(model, MOI.ObjectiveLimit(), 1.0)
```
odow, line 199: We should explicitly test setting `nothing` here as well

src/attributes.jl
```diff
@@ -850,6 +850,18 @@ struct TimeLimitSec <: AbstractOptimizerAttribute end
...
+
+An optimizer attribute for setting a limit on the objective value.
```
odow, line 857:
```suggestion
An optimizer attribute for setting a limit on the objective value.

```

src/attributes.jl
```diff
@@ -850,6 +850,18 @@ struct TimeLimitSec <: AbstractOptimizerAttribute end
...
+The solver may stop when the `ObjectiveValue` is better (lower for minimization, higher for maximization) than the `ObjectiveLimit`.
```
odow, line 858:
```suggestion
The solver may stop when the [`ObjectiveValue`](@ref) is better (lower for
minimization, higher for maximization) than the `ObjectiveLimit`. If stopped,
the [`TerminationStatus`](@ref) should be `OBJECTIVE_LIMIT`.

```

src/attributes.jl
```diff
@@ -850,6 +850,18 @@ struct TimeLimitSec <: AbstractOptimizerAttribute end
...
+When `set` to `nothing`, it removes the solver limit. The default value is `nothing`.
```
odow, line 859:
```suggestion
When `set` to `nothing`, the limit reverts to the solver's default.

The default value is `nothing`.
```

src/attributes.jl
```diff
@@ -850,6 +850,18 @@ struct TimeLimitSec <: AbstractOptimizerAttribute end
...
+"""
+struct ObjectiveLimit <: AbstractOptimizerAttribute end
+
+attribute_value_type(::ObjectiveLimit) = Union{Nothing,Float64}
```
odow, line 863: An implicit assumption is that this applies only to scalar-valued objective functions, and that the objective is `Float64`. What if the optimizer is `BigFloat`, or `Rational{BigFloat}`? And what about vector-valued objectives?

src/precompile.jl
```diff
@@ -79,6 +79,7 @@ function precompile_model(model, constraints)
         ObjectiveValue,
         Silent,
         TimeLimitSec,
+        ObjectiveBound,
```
odow, line 82: Unrelated?

src/Test/test_attribute.jl
```diff
@@ -190,6 +190,30 @@ function setup_test(
     return
 end
 
+function test_attribute_ObjectiveLimit(model::MOI.AbstractOptimizer, ::Config)
+    @requires MOI.supports(model, MOI.ObjectiveLimit())
+    # Get the current value to restore it at the end of the test
+    value = MOI.get(model, MOI.ObjectiveLimit())
+    MOI.set(model, MOI.ObjectiveLimit(), 0.0)
+    @test MOI.get(model, MOI.ObjectiveLimit()) == 0.0
```
odow, line 198:
```suggestion
    @test MOI.get(model, MOI.ObjectiveLimit()) == 0.0
    MOI.set(model, MOI.ObjectiveLimit(), nothing)
    @test MOI.get(model, MOI.ObjectiveLimit()) === nothing
```

src/attributes.jl
```diff
@@ -850,6 +850,25 @@ struct TimeLimitSec <: AbstractOptimizerAttribute end
...
+"""
+    ObjectiveLimit()
+
+An optimizer attribute for setting a limit on the objective value.
+
+The provided limit must be a `Union{Float64,Nothing}`.
+
+When `set` to `nothing`, the limit reverts to the solver's default.
+
+The default value is `nothing`.
```
odow, line 862:
```suggestion
The default value is `nothing`.

```

src/attributes.jl
```diff
@@ -850,6 +850,25 @@ struct TimeLimitSec <: AbstractOptimizerAttribute end
...
+The solver may stop when the [`ObjectiveValue`](@ref) is better (lower for
+minimization, higher for maximization) than the `ObjectiveLimit`. If stopped,
+the [`TerminationStatus`](@ref) should be `OBJECTIVE_LIMIT`.
+
```
odow, line 866:
```suggestion
```

src/precompile.jl
odow, line 82:
```suggestion
```

src/attributes.jl
matbesancon: vetor-valued objectives is another story I wouldn't touch at first. I would be okay upgrading this to:
```
Union{Nothing,Real}
```

matbesancon:
good to go?

src/attributes.jl
odow: Maybe we should just remove this for now. It's only implemented for a subset of attributes, and the fallback is `Any`.

matbesancon: Sure yes

src/attributes.jl
```diff
@@ -850,6 +850,25 @@ struct TimeLimitSec <: AbstractOptimizerAttribute end
...
+The default value is `nothing`.
+
+The solver may stop when the [`ObjectiveValue`](@ref) is better (lower for
+minimization, higher for maximization) than the `ObjectiveLimit`. If stopped,
+the [`TerminationStatus`](@ref) should be `OBJECTIVE_LIMIT`.
+"""
+struct ObjectiveLimit <: AbstractOptimizerAttribute end
+
+attribute_value_type(::ObjectiveLimit) = Union{Nothing,Float64}
```
matbesancon, line 870:
```suggestion
```

src/attributes.jl
```diff
@@ -850,6 +850,24 @@ struct TimeLimitSec <: AbstractOptimizerAttribute end
...
+
+The default value is `nothing`.
+
+The solver may stop when the [`ObjectiveValue`](@ref) is better (lower for
+minimization, higher for maximization) than the `ObjectiveLimit`. If stopped,
+the [`TerminationStatus`](@ref) should be `OBJECTIVE_LIMIT`.
+"""
+struct ObjectiveLimit <: AbstractOptimizerAttribute end
+
+
```
matbesancon, line 870:
```suggestion
```

odow (approved):
I'm okay with this, but it probably needs @blegat to take a look.

odow:
Bump @blegat

odow:
Bump again @blegat

matbesancon:
ping @joaquimg too in case

src/attributes.jl
```diff
@@ -850,6 +850,23 @@ struct TimeLimitSec <: AbstractOptimizerAttribute end
 
 attribute_value_type(::TimeLimitSec) = Union{Nothing,Float64}
 
+"""
+    ObjectiveLimit()
+
+An optimizer attribute for setting a limit on the objective value.
+
+The provided limit must be a `Union{Float64,Nothing}`.
```
blegat, line 858: It's not Float64. For a solver using BigFloat, it can be BigFloat. Let's just say it's Real

matbesancon: done

blegat (approved):

jajhall:
> * HiGHS: objective_bound https://ergo-code.github.io/HiGHS/stable/options/definitions/#objective_bound

This is applied to the dual objective in dual simplex, since it's a bound on the optimal primal objective value. So, it's not the option that should be set in the MIP context

HiGHS also has an `objective_target` option that was defined so that the primal simplex solver could stop if the target was reached. However this hasn't yet been implemented, and interpreting `objective_target` in the MIP context makes sense - and is easy to implement.

My only hesitation relates to the possibility of losing consistency with Gurobi. Its `BestObjStop` is just for the MIP context. Do you know whether Gurobi has a separate "target" parameter for continuous problems? My immediate search suggests not.

matbesancon:
> My only hesitation relates to the possibility of losing consistency with Gurobi. Its BestObjStop is just for the MIP context. Do you know whether Gurobi has a separate "target" parameter for continuous problems? My immediate search suggests not.

not that I have seen no, I have seen it mostly applicable on MIP (on Gurobi and CPLEX)

matbesancon:
the attribute defined here is independent of the type of problem solved. For an LP, using `objective_target` makes sense for LPs, I don't know whether this should force the use of the primal simplex or throw an error if not used together with the primal simplex?

Would that be okay to have the same parameter used for MIPs?

jajhall:
> For an LP, using `objective_target` makes sense for LPs, I don't know whether this should force the use of the primal simplex or throw an error if not used together with the primal simplex?

Interesting thought. The default simplex strategy is to allow HiGHS to choose - and it always uses dual, since it's generally better and the implementation is better than the primal simplex implementation. I guess that it should throw an error if  `objective_target` has been set and the user specifies dual simplex, as it feeds back useful information.

jajhall:
> Would that be okay to have the same parameter used for MIPs?

Looking at  [Gurobi status codes](https://www.gurobi.com/documentation/current/refman/optimization_status_codes.html) for something else yesterday, I followed the links for `USER_OBJ_LIMIT`, and both refer to MIP. So, it's safe to assume that LP target/bounds aren't available to users

PR #2259, opened 2023-08-22
[Utilities] improve performance of ScalarNonlinearFunction utilities

odow:
x-ref https://github.com/jump-dev/JuMP.jl/pull/3106

### DIFF
--- a/test/Utilities/functions.jl
+++ b/test/Utilities/functions.jl
@@ -1980,7 +1980,7 @@ function test_ScalarNonlinearFunction_map_indices_nested()
     index_map = MOI.copy_to(dest, src)
     new_f = MOI.Utilities.map_indices(index_map, f)
     y = [index_map[xi] for xi in x]
-    @test new_f ≈  MOI.ScalarNonlinearFunction(
+    @test new_f ≈ MOI.ScalarNonlinearFunction(
         :+,
         Any[
             y[1],
--- a/test/Utilities/functions.jl
+++ b/test/Utilities/functions.jl
@@ -1954,10 +1954,62 @@ function test_ScalarNonlinearFunction_map_indices()
     x = MOI.add_variable(src)
     f = MOI.ScalarNonlinearFunction(:log, Any[x])
     c = MOI.add_constraint(src, f, MOI.LessThan(1.0))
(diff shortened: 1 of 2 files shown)


test/Utilities/functions.jl
```diff
@@ -1954,13 +1954,65 @@ function test_ScalarNonlinearFunction_map_indices()
...
+            ),
+            x[5],
+        ],
+    )
+    c = MOI.add_constraint(src, f, MOI.LessThan(1.0))
+    dest = MOI.Utilities.MockOptimizer(MOI.Utilities.Model{Float64}())
+    index_map = MOI.copy_to(dest, src)
+    new_f = MOI.Utilities.map_indices(index_map, f)
+    y = [index_map[xi] for xi in x]
+    @test new_f ≈  MOI.ScalarNonlinearFunction(
```
odow, line 1983:
```suggestion
    @test new_f ≈ MOI.ScalarNonlinearFunction(
```

odow:
It was only a bottleneck because of the type stability problem. For the `::AbstractVector` case, we were passing in a `Vector{Any}`, then trying to infer the return type, and then casting back to `convert(Vector{Any}`. Just lots of unnecessary stuff.  The recursion also prevented deeply nested expressions.

One general improvement could be that if the `CachingOptimizer` knows that the mapping is one-to-one, then it could store `nothing` here:
https://github.com/jump-dev/MathOptInterface.jl/blob/dd155490f9cb07ed6a3dd854c6c8d9458386a9dd/src/Utilities/cachingoptimizer.jl#L82-L83
and we could add a new overload to make all of these calls a no-op:
https://github.com/jump-dev/MathOptInterface.jl/blob/dd155490f9cb07ed6a3dd854c6c8d9458386a9dd/src/Utilities/cachingoptimizer.jl#L525

blegat:
Yes I agree but I would make that logic inside CleverDict

blegat:
The `IndexMap` is holding `CleverDict`'s. `CleverDict` is already recognizing if the indices of `src` are contiguous so it seems natural to also recognize when the mapping is trivial. It's be discussed a few time with @joaquimg and I also reporting that it could improve performance in other situations.

PR #2261, opened 2023-08-27
Remove empty code sample in docstring

blegat:


odow:
Hmm. I was sure I added an example here. Perhaps something prevented it. I'll merge this for now and look at a proper fix.

PR #2264, opened 2023-09-06
[Bridges] fix adjoint functions in SetDotInverseScalingBridge

odow:
Potential fix for https://github.com/jump-dev/MathOptInterface.jl/pull/2262#issuecomment-1709227184.

I don't really understand what's going here, but it makes sense that the adjoint should be the opposite of the forward map? That's what it is for `SetDotScalingBridge`, so I think this is just a typo.

Seemed to fix SDPA, so I'll re-run solver-tests on this branch: https://github.com/jump-dev/MathOptInterface.jl/actions/runs/6103555292

### DIFF
--- a/src/Bridges/Constraint/bridges/set_dot_scaling.jl
+++ b/src/Bridges/Constraint/bridges/set_dot_scaling.jl
@@ -179,14 +179,14 @@ function MOI.Bridges.adjoint_map_function(
     ::Type{<:SetDotInverseScalingBridge{T,S}},
     func,
 ) where {T,S}
-    return _inverse_scale(T, S, func)
+    return _scale(T, S, func)
 end
 
 function MOI.Bridges.inverse_adjoint_map_function(
     ::Type{<:SetDotInverseScalingBridge{T,S}},
     func,
 ) where {T,S}
-    return _scale(T, S, func)
+    return _inverse_scale(T, S, func)
 end
 
 # Since the set type is not defined, the default `MOI.supports_constraint`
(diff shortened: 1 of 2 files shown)


odow:
Mosek still broken, but because you explicitly coded the bridge type into a test: 
https://github.com/jump-dev/MosekTools.jl/blob/6624a98dcef6b4d6619111e9b40b8afb05b40bae/test/runtests.jl#L264

blegat:
This is weird. The adjoint is the transpose and the transpose if a Diagonal should be the matrix itself

blegat:
We should add a test that sets and gets ConstraintDualStart in the bridge test, it should clarify

odow:
> We should add a test that sets and gets ConstraintDualStart in the bridge test, it should clarify

There is a test. But it just tests that the round trip preserves. So it doesn't help if it gets the wrong value... 😆

odow:
> This is weird. The adjoint is the transpose and the transpose if a Diagonal should be the matrix itself

Isn't the `SetDotInverseScalingBridge` just the inverse of this? So it makes sense they'd be swapped.

https://github.com/jump-dev/MathOptInterface.jl/blob/fdf2559261a0697196f7758f78048952c9ffac09/src/Bridges/Constraint/bridges/set_dot_scaling.jl#L90-L108

blegat:
It's good that I left a comment to explain this to my future self 😆 . So this fix is indeed correct because the adjoint is not the transpose. Can you add a test that don't just check the round trip ?

PR #2266, opened 2023-09-16
Update to Documenter v1

blegat:


odow:
Let me do this. A few things need to change

PR #2267, opened 2023-09-18
[Test] rename Test module to avoid conflict with Base.Test

odow:
JET is getting to the point where it can analyze MathOptInterface. (Previously blocked by https://github.com/aviatesk/JET.jl/issues/289 and https://github.com/aviatesk/JET.jl/issues/521.)

Now it's blocked by the fact that `MOI.Test.Test` is not `MOI.Test`. It seems like this could be problematic in general, and it'd be nice to start looking at the warnings given by JET. It seems like static analysis is only going to increase in popularity in Julia.


blegat (approved):
Nice workaround

PR #2269, opened 2023-09-18
Fix various JET errors

odow:
Part of #2268 

Brings count from 211 down to 196. I'll comment in-line.

### DIFF
--- a/src/FileFormats/MPS/MPS.jl
+++ b/src/FileFormats/MPS/MPS.jl
@@ -1319,7 +1319,7 @@ function parse_name_line(data::TempMPSModel, line::String)
     if m === nothing
         error("Malformed NAME line: ", line)
     end
-    data.name = strip(m[1])
+    data.name = strip(m[1]::AbstractString)
     return
 end
 
--- a/src/FileFormats/NL/NL.jl
+++ b/src/FileFormats/NL/NL.jl
@@ -781,11 +781,11 @@ function _assert_has_model(::Nothing, attr)
     )
 end
 
-_assert_has_model(::MOI.Utilities.UniversalFallback, ::Any) = nothing
+_assert_has_model(model::MOI.Utilities.UniversalFallback, ::Any) = model
 
 function MOI.get(model::Model, attr::MOI.AbstractModelAttribute)
-    _assert_has_model(model.model, attr)
-    return MOI.get(model.model, attr)
+    inner = _assert_has_model(model.model, attr)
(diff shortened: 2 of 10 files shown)


src/Utilities/objective_container.jl
```diff
@@ -116,26 +116,58 @@ function MOI.supports(
...
+    return o.vector_nonlinear::MOI.VectorNonlinearFunction
+end
+
 function MOI.get(
     o::ObjectiveContainer{T},
     ::MOI.ObjectiveFunction{F},
 ) where {T,F}
     if o.scalar_affine !== nothing
-        return convert(F, o.scalar_affine)
+        return convert(F, _get_scalar_affine(o))
```
odow, line 156: All the changes in this file deal with the problem that `o.scalar_affine::Union{Nothing,MOI.ScalarAffineFunction{T}}`.

At runtime, the `if` statement handles the `nothing` case, so we never call `convert(F, nothing)`, but it's useful to have a type-stable way to get the non-nothing function.

Although writing this now, I wonder if I could use `_not_nothing(o.scalar_affine)` to disambiguate the union.

src/FileFormats/NL/NL.jl
```diff
@@ -781,29 +781,29 @@ function _assert_has_model(::Nothing, attr)
...
 end
 
-_assert_has_model(::MOI.Utilities.UniversalFallback, ::Any) = nothing
+_assert_has_model(model::MOI.Utilities.UniversalFallback, ::Any) = model
 
 function MOI.get(model::Model, attr::MOI.AbstractModelAttribute)
-    _assert_has_model(model.model, attr)
-    return MOI.get(model.model, attr)
+    inner = _assert_has_model(model.model, attr)
+    return MOI.get(inner, attr)
```
odow, line 788: Previously compiler couldn't prove that `model.model` was not `nothing` because the check happened in `_assert_has_model`. Okay at runtime, not at compilation.

src/FileFormats/MPS/MPS.jl
```diff
@@ -1319,7 +1319,7 @@ function parse_name_line(data::TempMPSModel, line::String)
     if m === nothing
         error("Malformed NAME line: ", line)
     end
-    data.name = strip(m[1])
+    data.name = strip(m[1]::AbstractString)
```
odow, line 1322: There's actually a bunch of problems where we assume that the regex finds something.

PR #2270, opened 2023-09-18
Fix various JET errors in test_conic.jl

odow:
x-ref #2268 

These ones are all needed for calls like
```
ones(T, 4)
```
because JET can't tell if `T` is a type or an `Int`, so it annotate the return type as `ones(T,4)::Union{Vector{T},Matrix{Float64}`.

Locally, I'm down to `═════ 172 possible errors found ═════`

### DIFF
--- a/src/Test/test_conic.jl
+++ b/src/Test/test_conic.jl
@@ -26,7 +26,7 @@ function _test_conic_linear_helper(
     model::MOI.ModelLike,
     config::Config{T},
     use_VectorOfVariables::Bool,
-) where {T}
+) where {T<:Real}
     @requires MOI.supports_incremental_interface(model)
     @requires MOI.supports(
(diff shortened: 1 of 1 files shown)


src/Test/test_conic.jl
```diff
@@ -7079,7 +7079,7 @@ end
 
 Test [`MOI.NormCone`](@ref).
 """
-function test_conic_NormCone(model::MOI.ModelLike, config::Config{T}) where {T}
+function test_conic_NormCone(model::MOI.ModelLike, config::Config{T}) where {T<:Real}
```
odow, line 7082:
```suggestion
function test_conic_NormCone(
    model::MOI.ModelLike,
    config::Config{T},
) where {T<:Real}
```

PR #2271, opened 2023-09-19
Fix various JET errors

odow:
x-ref #2268

This gets things down to `═════ 127 possible errors found ═════`


odow:
And now `═════ 119 possible errors found ═════`

PR #2272, opened 2023-09-19
[Utilities] remove keyword arguments from copy_to

odow:


odow:
I think these ones I just missed removing. If you tried to use them they'd error (`default_copy_to` doesn't accept kwargs).

blegat (approved):

PR #2273, opened 2023-09-19
Add new MipDualBound status

adrien-mogenet:
This is useful when the solver hits a time limit and contains no solution available from the "objective value" or "objective bound" field, but we still want to retrieve the latest dual bound computed in the last iteration.

This requires a version of HiGHS that implements this change: https://github.com/ERGO-Code/HiGHS/pull/1420 , hence the draft status for the time being

### DIFF
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -1288,6 +1288,15 @@ A model attribute for the best known bound on the optimal objective value.
 """
 struct ObjectiveBound <: AbstractModelAttribute end
 
+
+"""
+    MipDualBound()
+
+A model attribute for the best known dual bound on the objective value.
+"""
+struct MipDualBound <: AbstractModelAttribute end
+
+
 """
     RelativeGap()
 
@@ -2192,6 +2201,7 @@ function is_set_by_optimize(
         ConstraintDual,
         ConstraintBasisStatus,
         VariableBasisStatus,
+        MipDualBound,
     },
 )
     return true


adrien-mogenet:
Will open a revised version following the last insights from the HiGHS team.

odow:
See https://github.com/jump-dev/HiGHS.jl/blob/6a39882e8a4f5ed8c1770109bb0aceaed1081955/src/MOI_wrapper.jl#L1941-L1950

adrien-mogenet:
> See https://github.com/jump-dev/HiGHS.jl/blob/6a39882e8a4f5ed8c1770109bb0aceaed1081955/src/MOI_wrapper.jl#L1941-L1950

Thanks, I recently noticed your contribution when undertaking that change, actually.
However, I'm wondering if this has the expected behavior, or if I misunderstand the behavior, or if it's not applicable for what I'm trying to achieve.

```
return sense == -1 ? max(primal, -p[]) : min(primal, p[])
```

Basically, the function returns the primal value for my maximization problem (MIP), whereas I'm truly interested in getting the "Best Bound" as displayed by HiGHS, thus `mip_dual_bound`, i.e., `p[]`.

odow:
Can you open an issue at HiGHS.jl https://github.com/jump-dev/HiGHS.jl/issues/new with a reproducible example of the behavior and what you want to happen? I don't remember why I added this. It might be that HiGHS has subsequently fixed the original bug.

PR #2276, opened 2023-09-20
Fix various JET errors

odow:
Part of #2268 

Now down to `═════ 70 possible errors found ═════`

https://github.com/jump-dev/MathOptInterface.jl/actions/runs/6242312138

### DIFF
--- a/src/Utilities/model.jl
+++ b/src/Utilities/model.jl
@@ -34,10 +34,6 @@ function MOI.add_variable(model::AbstractModel)
     return x
 end
 
-function MOI.add_variables(model::AbstractModel, n::Integer)
-    return [MOI.add_variable(model) for i in 1:n]
-end
-
 """
     remove_variable(f::MOI.AbstractFunction, s::MOI.AbstractSet, vi::MOI.VariableIndex)
 
--- a/src/Utilities/objective_container.jl
+++ b/src/Utilities/objective_container.jl
@@ -123,12 +123,12 @@ function MOI.get(
     T,
     F<:Union{
         MOI.VariableIndex,
-        MOI.ScalarAffineFunction{T},
-        MOI.ScalarQuadraticFunction{T},
+        MOI.ScalarAffineFunction,
+        MOI.ScalarQuadraticFunction,
         MOI.ScalarNonlinearFunction,
         MOI.VectorOfVariables,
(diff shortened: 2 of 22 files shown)


src/Utilities/objective_container.jl
```diff
@@ -118,8 +118,20 @@ end
...
+        MOI.VariableIndex,
+        MOI.ScalarAffineFunction{T},
+        MOI.ScalarQuadraticFunction{T},
+        MOI.ScalarNonlinearFunction,
+        MOI.VectorOfVariables,
+        MOI.VectorAffineFunction{T},
+        MOI.VectorQuadraticFunction{T},
+        MOI.VectorNonlinearFunction,
+    },
+}
```
odow, line 134: Hmm. This breaks JuMP:
https://github.com/jump-dev/JuMP.jl/blob/1ee5faee4335386435e3211b8758fba13799234e/test/test_objective.jl#L168-L180

I'm not sure whether we intended this:
```julia
julia> model = GenericModel{Float32}();

julia> @variable(model, x)
x

julia> @objective(model, Min, 2 * x + 1)
2 x + 1

julia> f = objective_function(model, GenericAffExpr{Float64,VariableRef})
2 x + 1

julia> typeof(f)
GenericAffExpr{Float64, GenericVariableRef{Float32}}

julia> g = objective_function(model)
2 x + 1

julia> typeof(g)
GenericAffExpr{Float32, GenericVariableRef{Float32}}
```

odow: https://github.com/jump-dev/MathOptInterface.jl/actions/runs/6244194786

blegat: No I think we meant to use `T` instead of `Float64` in the test

src/Utilities/model.jl
```diff
@@ -34,10 +34,6 @@ function MOI.add_variable(model::AbstractModel)
     return x
 end
 
-function MOI.add_variables(model::AbstractModel, n::Integer)
```
blegat, line 37: Where did that one go ?

odow: It's the same as the default fallback

blegat: Ah yes, it's in `Utilities/model.jl`

blegat (approved):

PR #2277, opened 2023-09-20
Fix JET errors in bridge_optimizer.jl

odow:
This one deserved a separate commit because the changes are a little subtle.

The various Bridge submodules define EmptyMap as a fallback for when an AbstractBridgeOptimizer doesn't have an associated set of arcs (like a Variable.SingleBridgeOptimizer). But this makes all calls of .bridges(b) type unstable. And sometimes we really do want either, but sometimes (like the ones annnotated here), we want only the explicit Map.

For MOI v2, we could consider refactoring this code to remove EmptyMap in favor of a more explicit fallback like nothing.

This PR brings us down to `═════ 42 possible errors found ═════`

### DIFF
--- a/src/Bridges/bridge_optimizer.jl
+++ b/src/Bridges/bridge_optimizer.jl
@@ -527,9 +527,8 @@ function _delete_variables_in_variables_constraints(
     # We reverse for the same reason as for `VariableIndex` below.
     # As the iterators are lazy, when the inner bridge constraint is deleted,
     # it won't be part of the iteration.
-    for ci in Iterators.reverse(
-        Constraint.vector_of_variables_constraints(c_map),
-    )
+    for ci in
+        Iterators.reverse(Constraint.vector_of_variables_constraints(c_map))
         _delete_variables_in_vector_of_variables_constraint(b, vis, ci)
     end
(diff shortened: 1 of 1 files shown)


blegat (approved):
Not sure why `nothing` would be more convenient than `EmptyMap` but the changes of this PR make sense. I'm starting to like JET more and more

odow:
> Not sure why nothing would be more convenient than EmptyMap

I think it'd just make it clearer what's actually going on. We don't really want to use `EmptyMap` for anything other than to signal that there isn't a map.

blegat:
> We don't really want to use `EmptyMap` for anything other than to signal that there isn't a map.

IIRC, it was helpful because it implements the `Map` API so instead of doing `if isnothing` all over the place in `bridge_optimizer.jl`, you can just call the API and it will work either way.

odow:
> it was helpful because it implements the Map API so instead of doing if isnothing all over the place in bridge_optimizer.jl, you can just call the API and it will work either way.

Only in some places though. It doesn't implement any of these call sites which need annotations (because they wouldn't get called for a `Variable.SingleBridgeOptimizer`).

blegat:
It's possible, I'd have to take another look, it's been a long time

PR #2279, opened 2023-09-20
Fix various JET errors

odow:
More changes for #2268

Most of these changes are due to the inferrability of a generator. I'm not super keen on the changes, but I think removing JET warnings outweighs the cost of a few extra lines.


blegat:
Are there generators that work with JET ? Or do we need to stop using generators from now on in JuMP and MOI ?

odow:
I don't know. Let's hold off merging this and I'll dig into the underlying causes. I think it's the mix of generator with `if`.

odow:
Closing as won't-fix.

PR #2281, opened 2023-09-21
[Utilities] remove unecessary acronyms

odow:
We should check solver-tests, but `MOI.Utilities.VI` and `MOI.Utilities.SQF` should not be public symbols.


blegat (approved):

odow:
https://github.com/jump-dev/MathOptInterface.jl/actions/runs/6258609625

PR #2291, opened 2023-09-24
Add SolutionLimit attribute

### LINKED ISSUE #2282: Optimizer option for solution limit
I see there is a solver status code for solution limit: https://jump.dev/MathOptInterface.jl/stable/reference/models/#MathOptInterface.SOLUTION_LIMIT

But I didn't see an option for collecting up to N solutions. I think this would be useful to abstract over at least for MIP solvers and constraint solvers like MiniZinc? e.g. MiniZinc has https://www.minizinc.org/doc-2.7.6/en/command_line.html?highlight=max%20solutions#cmdoption-n

odow:
Closes #2282


blegat (approved):

chriscoey:
thanks!!

Issue #1993, opened 2022-09-11
Reconsider supports of constraint attributes in bridges


odow:
See discussion in #1992, and these tests:
https://github.com/jump-dev/MathOptInterface.jl/blob/d9c998ac49a80ed8f740d5009133320e982af977/test/Bridges/bridge_optimizer.jl#L898-L933

The issue is that the bridge/model may support an attribute for a variable-in-set constraint if it was added via `add_constrained_variable(s)` but not via `add_constraint` (or vice versa). However, we have no way of telling which way the constraint was added based on the type alone.

We can either default to `&&`, in which case we may return a false negative (the model supports the attribute but `supports` returns `false`) or `||`, in which case we may return a false positive (the model supports the attribute for one of the ways, but we added it the other way). It's not obvious which is better, but our tests are passing with the current implementation, so we should wait for a solver to complain before reconsidering.

odow:
I think we should close this. I can't see a way to fix this in the current form. There's a note in #2180, but that requires a much larger discussion that's more than this specific issue. (And I don't really want to go there.)

It's also a very minor issue that, from what I can tell, has never impacted a user in the wild.

blegat:
We should at least document what it does return then. As per the doc, `supports` normally only has false positive.
From reading https://github.com/jump-dev/MathOptInterface.jl/pull/1992, it seems, in order to fix an issue with Ipopt, it's now a mix of false positive and false negative which is in contradiction with `supports`. I should take a closer look, it's not urgent but I'd rather keep that issue open.

odow:
Trying https://github.com/jump-dev/MathOptInterface.jl/pull/2294 just caused a bunch of solvers to fail tests:

![image](https://github.com/jump-dev/MathOptInterface.jl/assets/8177701/31338a3f-6d54-4af4-a256-60d6fb837e86)

PR #2294, opened 2023-09-27
[Bridges] change supports for AbstractConstraintAttribute to OR formulation

odow:
x-ref #1993

@blegat, I still don't really understand the subtlety here. Is this what you were thinking?

solver-tests: https://github.com/jump-dev/MathOptInterface.jl/actions/runs/6320227080


odow:
Closing for now. I don't understand this.

PR #2293, opened 2023-09-27
[FileFormats.MOF] add MathOptFormat@1.6 support

odow:
Closes #2292 

## Basic

 - [x] The file at `src/FileFormats/MOF/mof.X.Y.schema.json` is updated
 - [x] The constants `SCHEMA_PATH`, `VERSION`, and `SUPPORTED_VERSIONS` are
       updated in `src/FileFormats/MOF/MOF.jl`

## New sets

 - [x] New sets are added to the `@model` in `src/FileFormats/MOF/MOF.jl`
 - [x] New sets are added to the `@enum` in `src/FileFormats/MOF/read.jl`
 - [x] `set_to_moi` is defined for each set in `src/FileFormats/MOF/read.jl`
 - [x] `head_name` is defined for each set in `src/FileFormats/MOF/write.jl`
 - [x] A new unit test calling `_test_model_equality` is aded to
       `test/FileFormats/MOF/MOF.jl`

## Tests

 - [x] The version field in `test/FileFormats/MOF/nlp.mof.json` is updated

## Documentation

 - [x] The version fields are updated in `docs/src/submodules/FileFormats/overview.md`

PR #2295, opened 2023-09-27
DNMY: [FileFormats.MOF] change parse_as_nlpblock to false by default

odow:
I've opened this PR, not because I want to merge it, but because I think it needs some discussion.

Currently, if you have a `.mof.json` file that contains nonlinear functions, they are parsed into an `NLPBlock`.

#2293 added support for parsing them as regular `ScalarNonlinearFunction`s, but made it opt-in behind a `parse_as_nlpblock` keyword.

To read in JuMP, you'd need to write:
```julia
using JuMP
m = JuMP.read_from_file("test.mof.json"; parse_as_nlpblock = false)
```

This PR proposes swapping the default, which is a breaking change if people expect the `NLPBlock` to exist, or if they're passing to a solver that doesn't support the new syntax yet.

We could potentially reduce the breakage by changing `default_copy_to`, so that, if a `src` model has `ScalarNonlinearFunction` but the `dest` model doesn't support them, we could convert to a `NLPBlock`. This would then only impact people using `MOF.Model` directly, not not some reading from a file in JuMP.

Or we could say that the nonlinear support in `.mof.json` was always experimental. It seems likely that very few people are actually using `.mof.json` files to store models, particularly nonlinear ones.

Or we could keep the default as-is and add a warning telling people to use the keyword argument, and that it might change the default in a future release.

Thoughts?

### DIFF
--- a/src/FileFormats/MOF/MOF.jl
+++ b/src/FileFormats/MOF/MOF.jl
@@ -123,7 +123,7 @@ function get_options(m::Model)
     return get(
         m.model.ext,
         :MOF_OPTIONS,
-        Options(false, false, MOI.Nonlinear.SparseReverseMode(), true),
+        Options(false, false, MOI.Nonlinear.SparseReverseMode(), false),
     )
 end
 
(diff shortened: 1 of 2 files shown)


odow:
Change to `use_nlp_block` to match NL files, and document this, and document it at the JuMP level.

odow:
Discussed on call. Closing in favor of #2299

PR #2296, opened 2023-09-28
Add ScalarQuadraticCoefficientChange

### LINKED ISSUE #1208: `MOI.modify` for quadratic terms
Background: https://github.com/jump-dev/Dualization.jl/issues/90 and a [discourse discussion](https://discourse.julialang.org/t/modifying-coefficient-of-quadratic-term-in-objective-function/51622)

`MOI.modify` makes in-place modifications of the coefficients of a function. Through `MOI.ScalarCoefficientChange` it is possible to modify the linear terms of a `ScalarQuadraticFunction`, but as for now the only way to modify the quadratic components is by defining the function again.

I think it would be beneficial to extend `MOI.ScalarCoefficientChange` or add a new type to support the in-place modification of the quadratic coefficients.

odow:
Closes #1208

### DIFF
--- a/src/Bridges/bridge_optimizer.jl
+++ b/src/Bridges/bridge_optimizer.jl
@@ -1160,7 +1160,11 @@ function _modify_bridged_function(
     change::MOI.AbstractFunctionModification,
 )
     if is_bridged(b, ci_or_obj)
-        MOI.modify(recursive_model(b), bridge(b, ci_or_obj), change)
+        try
+            MOI.modify(recursive_model(b), bridge(b, ci_or_obj), change)
+        catch
+            MOI.throw_modify_not_allowed(ci_or_obj, change)
+        end
     else
         MOI.modify(b.model, ci_or_obj, change)
     end
@@ -1877,6 +1881,18 @@ function modify_bridged_change(
     return
 end
 
+function modify_bridged_change(
+    b::AbstractBridgeOptimizer,
+    ci_or_obj,
+    change::MOI.ScalarQuadraticCoefficientChange,
+)
+    return MOI.throw_modify_not_allowed(
+        ci_or_obj,
+        change,
+        "Cannot bridge `ScalarQuadraticCoefficientChange`.",
+    )
+end
+
 function MOI.modify(
     b::AbstractBridgeOptimizer,
     ci::MOI.ConstraintIndex,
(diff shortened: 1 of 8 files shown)


src/Bridges/bridge_optimizer.jl
```diff
@@ -1160,7 +1160,11 @@ function _modify_bridged_function(
     change::MOI.AbstractFunctionModification,
 )
     if is_bridged(b, ci_or_obj)
-        MOI.modify(recursive_model(b), bridge(b, ci_or_obj), change)
+        try
+            MOI.modify(recursive_model(b), bridge(b, ci_or_obj), change)
+        catch
+            MOI.throw_modify_not_allowed(ci_or_obj, change)
+        end
```
odow, line 1167: @blegat do you have any ideas for this? We can't add a generic fallback for the `::Bridge` case of the second argument because then we don't know whether we're changing a constraint or an objective.

blegat: Not sure what you mean. I would add a fallback in `Bridges/bridge.jl`. If you want to differentiate, you have `Bridges.Objective.AbstractBridge` and `Bridges.Constraint.AbstractBridge`.

odow: But how do we know which `ConstraintIndex` it refers to?

blegat: Ah yes, I guess this is an annoying part of this error. This is also why you had to do a `try`-`catch` in https://github.com/jump-dev/MathOptInterface.jl/pull/2300.
I think we should allow throwing this error without knowing what is the constraint index is.
At the end of the day, after the index maps, this might not be the index of the user anyway.

odow: Yeah a generic `ModifyNotAllowed` would be simpler. But that could be a separate PR after this is merged?

blegat: Yes but then open an issue or leave a comment here to keep not of that

odow: Opened an issue: https://github.com/jump-dev/MathOptInterface.jl/issues/2306. I'll work on it after this PR is merged

blegat (approved):

PR #2309, opened 2023-10-13
Use conversion cost for Objective bridge

blegat:
Current, the added test fails which means that converting to a quadratic objective is preferred over using several bridges. The bridge cost fixes this issue the same way we fixed it for the constraint conversion bridge.

### DIFF
--- a/test/Bridges/lazy_bridge_optimizer.jl
+++ b/test/Bridges/lazy_bridge_optimizer.jl
@@ -2140,7 +2140,7 @@ MOI.Utilities.@model(
     ModelQuadObj,
     (),
     (),
-    (MOI.Nonnegatives,MOI.Zeros),
+    (MOI.Nonnegatives, MOI.Zeros),
     (),
     (),
     (),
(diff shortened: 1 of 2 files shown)


test/Bridges/lazy_bridge_optimizer.jl
```diff
@@ -2136,6 +2136,34 @@ function test_ToScalarQuadraticBridge_variable_bounds()
...
+    return false
+end
+
+function test_objective_conversion_cost(T = Float64)
+    model = ModelQuadObj{T}()
+    bridged = MOI.Bridges.full_bridge_optimizer(model, T)
+    x = MOI.add_variable(bridged)
+    MOI.set(bridged, MOI.ObjectiveSense(), MOI.MIN_SENSE)
+    MOI.set(bridged, MOI.ObjectiveFunction{typeof(x)}(), one(T) * x)
+    @test MOI.get(model, MOI.ObjectiveFunctionType()) == MOI.VariableIndex
```
odow, line 2164:
```suggestion
    @test MOI.get(model, MOI.ObjectiveFunctionType()) == MOI.VariableIndex
    return
```

PR #2314, opened 2023-10-22
[Utilities] add more distance_to_set implementations

### LINKED ISSUE #2033: Implement Utilities.distance_to_set for more sets

odow:
Closes #2033 

These are really painful to test and check that they are correct. It's easy to make a mistake.


odow:
I haven't added the matrix cones (or the constraint programming ones, or things like indicator and complementarity), but this is a reasonably complete list for 95% of the use-cases.

src/Utilities/distance_to_set.jl
```diff
@@ -213,3 +213,174 @@ function distance_to_set(
...
+
+function distance_to_set(
+    ::ProjectionUpperBoundDistance,
+    x::AbstractVector{T},
+    set::MOI.SOS2,
+) where {T<:Real}
+    _check_dimension(x, set)
+    p = sortperm(set.weights)
+    pairs = collect(zip(p[1:end-1], p[2:end]))
+    _, k = findmax(abs(x[i]) + abs(x[j]) for (i, j) in pairs)
```
odow, line 384:
```suggestion
    _, k = findmax([abs(x[i]) + abs(x[j]) for (i, j) in pairs])
```

odow:
Thoughts @matbesancon?

blegat:
> I haven't added the matrix cones (or the constraint programming ones, or things like indicator and complementarity), but this is a reasonably complete list for 95% of the use-cases.

I don't think we should require having an implementation for any set or should aim as covering "most" use case. It's more like, if someone has an short and simple distance computation without any dependency and makes a PR that is well tested then we'll accept it in MOI and it does not have to be in MOSD.

blegat (approved):

PR #2315, opened 2023-10-22
[Utilities] add support for VariableIndex in ModelFilter

### LINKED ISSUE #2308: [Utilities] Add support for filtering variables in ModelFilter
Needs to be added here:

https://github.com/jump-dev/MathOptInterface.jl/blob/eb0d3c1a0687eb55adecfa35a7f583654b9d212c/src/Utilities/copy.jl#L512-L584

Once added, we should also add to JuMP's `copy_model`.

cc @sstroemer

odow:
Closes #2308. But I don't really like this. I think it's better just to delete variables after the copy---that will ensure we delete the appropriate constraint coefficients, etc.

### DIFF
--- a/src/Test/test_model.jl
+++ b/src/Test/test_model.jl
@@ -1159,6 +1159,50 @@ function test_model_ModelFilter_ListOfConstraintIndices(
 end
 
 """
+    test_model_ModelFilter_ListOfConstraintIndices(
+        src::MOI.ModelLike,
+        ::Config{T},
+    ) where {T}
+
+Tests `Utilities.ModelFilter` of `ListOfVariableIndices`.
+"""
+function test_model_ModelFilter_ListOfVariableIndices(
+    src::MOI.ModelLike,
+    ::Config{T},
+) where {T}
+    @requires(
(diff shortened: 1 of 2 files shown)


odow:
The more I think about this, the less I like it. Deletion is the way to go.

blegat:
The issue is that it's also unclear what to do with constraints that have these variables, remove them or replace the variable with zero ?
In the first case, we could have to alter `ListOfConstraintIndices` as well but we don't want to do that unless we know the filter is removing variables. So I'd add an option specifying `SET_VARIABLE_TO_ZERO`, `REMOVE_CONSTRAINTS_WITH_VARIABLE` (in that case, it cannot be in the objective) or `KEEP_ALL_VARIABLES`.
By default, we have `KEEP_ALL_VARIABLES` and we error if the filter remove any variables from the list (to let the user know he also need to specify that option).

odow:
> The issue is that it's also unclear what to do with constraints that have these variables, remove them or replace the variable with zero ?

Yes. But there are just too many complications. We don't need to support this because the user can just delete variables after the fact.

Issue #2316, opened 2023-10-23
Indicator and SOS to MILP bridges


odow:
I talked to @fdabrandao (AMPL) at INFORMS. They have a lot of new stuff for reformulating indicator constraints to MILP. We can't offer the same full flexibility as AMPL, but we could write a few XXXtoMILPBridges, just like we've done for constraint programming.

In all cases, `M` is based on bounds of `x`, and we error if bounds are not present.

 * `z --> {a'x <= b}` => `a'x <= b + M(1 - z)`
 * `!z --> {a'x <= b}` => `a'x <= b + Mz`
 * `z --> {a'x >= b}` => `a'x >= b - M(1 - z)`
 * `!z --> {a'x >= b}` => `a'x >= b - Mz`
 * `z --> {a'x == b}` => `a'x == b + y`, `-M(1-z) <= y <= M(1 - z)`
 * `!z --> {a'x == b}` => `a'x == b + y`, `-Mz <= y <= Mz`
 * `x in SOS1` => `-l_i z_i <= x_i <= u_i z_i`, `sum z_i == 1`
 * `x in SOS2` => `-l_i (z_i+z_j) <= x_i <= u_i (z_i + z_j)`, `sum z_i == 1`

cc @glebbelov

odow:
We have `IndicatorSOS1Bridge`:
https://github.com/jump-dev/MathOptInterface.jl/blob/d9417aed53b4c4e56dca8d4f78859aa44606ce7b/src/Bridges/Constraint/bridges/indicator_sos.jl#L7-L33
So it might be sufficient to have `SOS1ToMILPBridge`, and then optionally add bounds to the slack indicator variable if possible.

odow:
Ah, but this would need `final_touch`, and so would `SOS1toMILPBridge`, which would require https://github.com/jump-dev/MathOptInterface.jl/issues/1980

odow:
Closed by #2318

PR #2318, opened 2023-10-24
[Bridges] add {SOS1,SOS2,Indicator}ToMILPBridge

odow:
Part of #2316 

I'll need to run the solver-tests, because this should let HiGHS support SOS constraints, but it will probably also start supporting Indicator because of IndicatorToSOS, except that IndicatorToSOS introduces an unbounded variable in the SOS.

https://github.com/jump-dev/MathOptInterface.jl/actions/runs/6620504066

### DIFF
--- a/src/Bridges/Bridges.jl
+++ b/src/Bridges/Bridges.jl
@@ -256,6 +256,7 @@ function runtests(
     variable_start = 1.2,
     constraint_start = 1.2,
     eltype = Float64,
+    print_inner_model::Bool = false,
 )
     # Load model and bridge it
     inner = MOI.Utilities.UniversalFallback(MOI.Utilities.Model{eltype}())
@@ -264,6 +265,9 @@ function runtests(
     final_touch(model)
     # Should be able to call final_touch multiple times.
     final_touch(model)
+    if print_inner_model
+        print(inner)
+    end
     # Load a non-bridged input model, and check that getters are the same.
     test = MOI.Utilities.UniversalFallback(MOI.Utilities.Model{eltype}())
     MOI.Utilities.loadfromstring!(test, input)
--- a/src/Bridges/Constraint/bridges/sos1_to_milp.jl
+++ b/src/Bridges/Constraint/bridges/sos1_to_milp.jl
@@ -184,43 +184,6 @@ end
 
 MOI.Bridges.needs_final_touch(::SOS1ToMILPBridge) = true
 
-# We use the bridge as the first argument to avoid type piracy of other methods.
-function _get_bounds(
-    bridge::SOS1ToMILPBridge{T},
-    model::MOI.ModelLike,
-    bounds::Dict{MOI.VariableIndex,NTuple{2,T}},
-    f::MOI.ScalarAffineFunction{T},
-) where {T}
-    lb = ub = f.constant
-    for term in f.terms
-        ret = _get_bounds(bridge, model, bounds, term.variable)
-        if ret === nothing
-            return nothing
-        end
-        lb += term.coefficient * ret[1]
-        ub += term.coefficient * ret[2]
-    end
-    return lb, ub
-end
-
-# We use the bridge as the first argument to avoid type piracy of other methods.
-function _get_bounds(
-    ::SOS1ToMILPBridge{T},
-    model::MOI.ModelLike,
-    bounds::Dict{MOI.VariableIndex,NTuple{2,T}},
-    x::MOI.VariableIndex,
-) where {T}
-    if haskey(bounds, x)
-        return bounds[x]
-    end
-    ret = MOI.Utilities.get_bounds(model, T, x)
-    if ret == (typemin(T), typemax(T))
-        return nothing
-    end
-    bounds[x] = ret
-    return ret
-end
-
 function MOI.Bridges.final_touch(
     bridge::SOS1ToMILPBridge{T,F},
     model::MOI.ModelLike,
@@ -229,7 +192,7 @@ function MOI.Bridges.final_touch(
     scalars = collect(MOI.Utilities.eachscalar(bridge.f))
     new_bounds = false
     for (i, x) in enumerate(scalars)
-        ret = _get_bounds(bridge, model, bounds, x)
+        ret = MOI.Utilities.get_bounds(bridge, model, bounds, x)
         if ret === nothing
             error(
                 "Unable to use SOS1ToMILPBridge because element $i " *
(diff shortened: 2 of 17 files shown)


src/Bridges/Bridges.jl
```diff
@@ -264,6 +265,9 @@ function runtests(
     final_touch(model)
     # Should be able to call final_touch multiple times.
     final_touch(model)
+    if print_inner_model
+        print(inner)
```
blegat, line 269: That's useful

src/Utilities/variables.jl
```diff
@@ -13,45 +13,107 @@ Return a tuple `(lb, ub)` of type `Tuple{T, T}`, where `lb` and `ub` are lower
...
+function get_bounds(
+    model::MOI.ModelLike,
+    bounds_cache::Dict{MOI.VariableIndex,NTuple{2,T}},
+    f::MOI.ScalarAffineFunction{T},
+) where {T}
+    lb = ub = f.constant
+    for term in f.terms
+        ret = get_bounds(model, bounds_cache, term.variable)
+        if ret === nothing
+            return nothing
```
blegat, line 79: `f` may not be canonicalized here so we might actually have bounds if all the terms with that variable cancel out.

src/Utilities/variables.jl
```diff
@@ -13,45 +13,107 @@ Return a tuple `(lb, ub)` of type `Tuple{T, T}`, where `lb` and `ub` are lower
...
         end
-    elseif MOI.is_valid(model, c_gt)
-        gt_2::MOI.GreaterThan{T} = MOI.get(model, MOI.ConstraintSet(), c_gt)
-        return gt_2.lower, typemax(T)
-    else
-        return typemin(T), typemax(T)
+        lb += term.coefficient * ret[1]
```
blegat, line 81: If `x in [-1, 2]` then you will say that `-x` is in `[1, -2]`. If the sign of `term.coefficient` is negative, you should reverse the bounds

blegat: Again if you have `x - 2x` then you'll get as lb `-1 - 2 = -3` and ub `2 - 2 * (-1) = 4` while if you canonicalize first you have `-2, 1` which is tighter. So I would canonicalize first.

src/Bridges/Bridges.jl
odow: Yeah I always use it when writing tests, but I normally just manually edit the file to put the `print` in.

src/Utilities/variables.jl
odow: So we could just canonicalize `f` at the top of this function?

src/Utilities/variables.jl
odow: Hmm. I obviously don't have tests for this

blegat: By the way, PolyJuMP.QCQP would be a user of a `get_bounds` for `ScalarQuadraticTerm`: https://github.com/jump-dev/PolyJuMP.jl/blob/aee87cda22f76545b0a611e87cb753cbc1d786f0/src/QCQP/MOI_wrapper.jl#L264-L271

odow: One argument for centralizing the implementation, like I've done in this PR

src/Utilities/variables.jl
blegat: Yes

src/Utilities/variables.jl
blegat: Yes, we reached the threshold that it's useful to centralize

src/Utilities/variables.jl
odow: Fixed

src/Utilities/variables.jl
odow: Fixed

odow:
So this is pretty breaking (to tests, not in practice) because it opens up IndicatorToSOS1 bridge. So I think we need to add an IndicatorToMILPBridge to fix it.

odow:
https://github.com/jump-dev/MathOptInterface.jl/actions/runs/6634509879

odow:
Found a few bugs in https://github.com/jump-dev/HiGHS.jl/pull/187, https://github.com/jump-dev/HiGHS.jl/pull/188.

odow:
https://github.com/jump-dev/MathOptInterface.jl/actions/runs/6647030587

odow:
https://github.com/jump-dev/MathOptInterface.jl/actions/runs/6647326291

odow:
Failures are:

 - [x] COSMO: test_attribute_TimeLimitSec
 - [ ] EAGO: test_attribute_TimeLimitSec, and some out-of-bound accesses?
 - [ ] Pavito: test_linear_Indicator_constant_term because not canonicalized?
 - [x] ProxSDP: test_attribute_TimeLimitSec
 - [ ] SCIP: bug in SCIP checking `is_valid`

odow:
https://github.com/jump-dev/MathOptInterface.jl/actions/runs/6659265716

odow:
It's pretty cool that stuff like this works now:
```julia
using JuMP, HiGHS
xp = 0.0:0.1:pi
yp = sin.(xp)
n = length(xp)
@variable(model, 0 <= x <= π)
@variable(model, -1 <= y <= 1)
@variable(model, 0 <= λ[1:n] <= 1)
@variable(model, z, Bin)
@constraint(model, λ in SOS2())
@constraint(model, sum(λ) == 1)
@constraint(model, xp' * λ == x)
@constraint(model, yp' * λ == y)
@constraint(model, z --> {x <= 0.45})
@constraint(model, !z --> {x == y})
@objective(model, Max, x)
optimize!(model)
```
This will let us use `SOS1` and `SOS2` and indicator constraints in the documentation tutorials.

blegat:
Yes, it's also helpful for teaching since then you can add complexity incrementally. The students can use indicator constraints without having learned the big-M formulation as requirement.

src/Bridges/Constraint/bridges/sos1_to_milp.jl
```diff
@@ -0,0 +1,254 @@
...
+) where {T,F<:Union{MOI.VectorOfVariables,MOI.VectorAffineFunction{T}}}
+    return SOS1ToMILPBridge{T,F}
+end
+
+function MOI.get(
+    ::MOI.ModelLike,
+    ::MOI.ConstraintFunction,
+    bridge::SOS1ToMILPBridge,
+)
+    return bridge.f
```
blegat, line 111: You might need to copy the function here for the same reason `MOI.Utilities.Model` does a copy. IIRC, the user is allowed to modify the function provided by `add_constraint`. If you don't make a copy, it will modify the function stored in the bridge.

blegat (approved):

PR #2325, opened 2023-10-26
[Test] allow skipping ScalarFunctionConstantNotZero errors

odow:
This one is up for debate. We've said that solvers can optionally support this, so some tests fail, even though it's known and okay.

Here's one from Pavito because of https://github.com/jump-dev/MathOptInterface.jl/pull/2318#issuecomment-1780291375

image

### DIFF
--- a/src/Test/Test.jl
+++ b/src/Test/Test.jl
@@ -309,7 +309,12 @@ macro requires(x)
 end
 
 function _error_handler(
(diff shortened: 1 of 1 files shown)


blegat:
Yes I think it's fine. Especially if they have no getter for `ConstraintFunction` and `ConstraintSet`, it's not too hard to support.

PR #2329, opened 2023-10-29
Improve error message unsupported attribute for bridges

blegat:
<s>Closes</s> <s>`https://github.com/jump-dev/JuMP.jl/issues/3550`</s>

### DIFF
--- a/src/Bridges/bridge.jl
+++ b/src/Bridges/bridge.jl
@@ -145,7 +145,7 @@ function MOI.get(
     attr::MOI.AbstractConstraintAttribute,
     bridge::AbstractBridge,
 )
-    message = _attribute_error_message(attr, typepof(bridge), "accessing")
+    message = _attribute_error_message(attr, typeof(bridge), "accessing")
     return throw(ArgumentError(message))
 end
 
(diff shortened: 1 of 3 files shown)


odow:
I can fix the rest of this

odow:
This isn't what I had in mind for JuMP. We need a better error message at the JuMP level.

blegat:
> This isn't what I had in mind for JuMP. We need a better error message at the JuMP level.

Then we can keep the issue open at the JuMP level but I think it's good to improve at the MOI level at least

src/Bridges/bridge.jl
```diff
@@ -134,18 +145,8 @@ function MOI.get(
...
-            "the attribute `$attr`. If you encountered this error " *
-            "unexpectedly, it probably means your model has been " *
-            "reformulated using the bridge, and you are attempting to query " *
-            "an attribute that we haven't implemented yet for this bridge. " *
-            "Please open an issue at https://github.com/jump-dev/MathOptInterface.jl/issues/new " *
-            "and provide a reproducible example explaining what you were " *
-            "trying to do.",
-        ),
-    )
+    message = _attribute_error_message(attr, typepof(bridge), "accessing")
```
odow, line 148:
```suggestion
    message = _attribute_error_message(attr, typeof(bridge), "accessing")
```

odow (approved):

PR #2331, opened 2023-10-30
Add ListOfVariablesWithAttributeSet and ListOfConstraintsWithAttributeSet

### LINKED ISSUE #2317: Get list of indices with non-default attribute values
@bernalde hit me up at INFORMS with this problem:

 * Given `MOI.VariablePrimalStart` appears in `MOI.ListOfVariableAttributesSet`, how do I tell which variables have a non-default start?

Some possible solutions:

 1. You don't. Iterate over every variable
 2. Add new attributes:
     * `MOI.ListOfVariablesWithAttributeSet{A<:MOI.AbstractVariableAttribute}`
     * `MOI.ListOfConstraintsWithAttributeSet{F,S,A<:MOI.AbstractConstraintAttribute}`

We could potentially implement 2. with the default that falls back to `MOI.get(model, MOI.ListOfVariableIndices())`. This would let solvers opt-in to more efficient implementations. But it might also confuse users if they start using `MOI.ListOfVariablesWithAttributeSet` and find that it is a super-set of indices, rather than the exact set.

odow:
Closes #2317


odow:
Thoughts @blegat?

src/Utilities/universalfallback.jl
```diff
@@ -939,3 +939,38 @@ end
 MOI.add_variable(uf::UniversalFallback) = MOI.add_variable(uf.model)
 
 MOI.add_variables(uf::UniversalFallback, n) = MOI.add_variables(uf.model, n)
+
+function MOI.get(
+    uf::UniversalFallback,
+    attr::MOI.ListOfVariablesWithAttributeSet,
+)
```
blegat, line 946: The rule is to check `if !MOI.is_copyable(attr) || MOI.supports(uf.model, attr, MOI.VariableIndex)`. If this is true, we query the inner model. Otherwise, we have to handle it with the fallback. Branching on whether it is a key is not correct since we might redirect the call to the inner model for attributes that it doesn't support.

odow: Fixed, and added tests

blegat (approved):

odow:
@bernalde and @pedromxavier is this what you were after?

pedromxavier:
I will be trying it in the next step of psrenergy/ToQUBO.jl#77.

odow:
Okay, I'll merge this into `master`, but we might hold off on a release until you've had time to take a look.

pedromxavier:
@odow @blegat I tested it and it works like a charm! Thanks for the great work!

PR #2336, opened 2023-11-02
[Utilities] print the name of scalar constraints in print(::ModelLike)

### LINKED ISSUE #2335: Print a constraint with Name
While building a model with Gurobi.jl and MOI, variable names is visiable from `print(model)`, but constraint names are not. I think this is inconsistent.
```julia
julia> import Gurobi
julia> o = Gurobi.Optimizer();
julia> x = MOI.add_variable(o);
julia> ci = MOI.add_constraint(o, 1. * x, MOI.LessThan(0.))
MathOptInterface.ConstraintIndex{MathOptInterface.ScalarAffineFunction{Float64}, MathOptInterface.LessThan{Float64}}(1)

julia> print(o) ### model_print_1
Feasibility

Subject to:

ScalarAffineFunction{Float64}-in-LessThan{Float64}
 0.0 + 1.0 v[1] <= 0.0

julia> MOI.set(o,MOI.VariableName(),x,"x") ### set Variable Name

julia> print(o) ### model_print_2
Feasibility

Subject to:

ScalarAffineFunction{Float64}-in-LessThan{Float64}
 0.0 + 1.0 x <= 0.0 ### Variable Name updated from "v[1]" to "x", compared with model_print_1

julia> MOI.set(o,MOI.ConstraintName(),ci,"x_le_0") ### set Constraint Name

julia> print(o) ### model_print_3, which is the same as model_print_2
Feasibility

Subject to:

ScalarAffineFunction{Float64}-in-LessThan{Float64}
 0.0 + 1.0 x <= 0.0

julia> 
```
I think an ideal case in model_print_3 should be something like
```julia
julia> print(o)
Feasibility

Subject to:

ScalarAffineFunction{Float64}-in-LessThan{Float64}
x_le_0: 0.0 + 1.0 x <= 0.0
```
Do you think this is a good proposal, because when users model physical problems, they need to label constraints of different types (e.g., ramp_up rates, demand_balance,...)
It will be convenient to check them in the prints.

Or maybe this is a question with MOI, generally?

Thanks!

odow:
Closes #2335 

cc @WalterMadelim

I need to think about what to do for vector constraints.

### DIFF
--- a/src/Utilities/print.jl
+++ b/src/Utilities/print.jl
@@ -374,14 +374,8 @@ end
 function _name_prefix(
     options::_PrintOptions{MIME"text/plain"},
     model::MOI.ModelLike,
-    cref::MOI.ConstraintIndex{
-        <:Union{
-            MOI.ScalarAffineFunction,
-            MOI.ScalarQuadraticFunction,
-            MOI.ScalarNonlinearFunction,
-        },
-    },
-)
+    cref::MOI.ConstraintIndex{F},
+) where {F}
     if !MOI.supports(model, MOI.ConstraintName(), typeof(cref))
         return ""
     end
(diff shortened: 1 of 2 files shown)


blegat (approved):

blegat:
Maybe we can print it after the set, like `in PSDCone() (name)`. It is consistent with the `(1)` you have at the end of the equations in LaTeX's environment `equation` or `align`. Then we could do the same for scalar constraint for consistency but then the names are not aligned

odow:
That's a reasonable solution. Thoughts, @WalterMadelim?

WalterMadelim:
Good morning, @odow. I think it’s also pretty to print at the end of a line after a constraint. It’s consistent with paper publications. And it seems to save the space of a line on screen for those Vector constraints.

odow:
Updated

blegat (approved):

PR #2337, opened 2023-11-02
[Bridges] add VariablePrimalStart support to ParameterToEqualTo and RSOCtoPSD

odow:

### DIFF
--- a/src/Bridges/Variable/bridges/rsoc_to_psd.jl
+++ b/src/Bridges/Variable/bridges/rsoc_to_psd.jl
@@ -269,17 +269,29 @@ function MOI.get(
     return value
 end
 
+_u_start_values(::Nothing) = nothing, nothing
+_u_start_values(value::T) where {T} = 2 * value, zero(T)
+
 function MOI.set(
     model::MOI.ModelLike,
     attr::MOI.VariablePrimalStart,
-    bridge::RSOCtoPSDBridge,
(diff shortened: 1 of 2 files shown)


blegat:
The first two indices are multiplied by a 2x2 matrix (rotation by 90°) which makes this tricky since the starting value for one depend on the one for the other one. For `get`, it's not too bad, you can just query both. For `set`, it's a bit more tricky because when you get `set` the first one, you need to cache it it the bridge and way for the other one. Once you have both, you can set it to the model. We could throw an error in `final_touch` if we get `set` only one of them

src/Bridges/Variable/bridges/rsoc_to_psd.jl
```diff
@@ -248,19 +248,41 @@ function MOI.get(
...
+    model::MOI.ModelLike,
+    attr::MOI.VariablePrimalStart,
+    bridge::RSOCtoPSDBridge,
+    value,
+    i::MOI.Bridges.IndexInVector,
+)
+    if value !== nothing && i.value == 2
+        value = 2 * value
+    end
+    MOI.set(model, attr, _variable(bridge, i), value)
```
blegat, line 282: If `i.value == 2`, you need to set it for the whole diagonal. `_variable` only give you the first one since it is enough for `get`.

odow: Fixed

PR #2338, opened 2023-11-02
[Bridges] add support for start values in NumberConversionBridge

odow:

<s>This isn't very obvious what the return types should be, so I've opted to just leave them as-is and let the solvers deal with them.</s> I used `value_type`.

### DIFF
--- a/src/Bridges/Constraint/bridges/number_conversion.jl
+++ b/src/Bridges/Constraint/bridges/number_conversion.jl
@@ -139,7 +139,11 @@ function MOI.get(
     },
     bridge::NumberConversionBridge{T,F1,S1,F2,S2},
 ) where {T,F1,S1,F2,S2}
-    return MOI.get(model, attr, bridge.constraint)
+    ret = MOI.get(model, attr, bridge.constraint)
+    if ret === nothing
+        return nothing
+    end
+    return convert(MOI.Utilities.value_type(T,F1), ret)
 end
 
 function MOI.set(
--- a/src/Bridges/Constraint/bridges/number_conversion.jl
+++ b/src/Bridges/Constraint/bridges/number_conversion.jl
@@ -143,7 +143,7 @@ function MOI.get(
     if ret === nothing
         return nothing
     end
-    return convert(MOI.Utilities.value_type(T,F1), ret)
+    return convert(MOI.Utilities.value_type(T, F1), ret)
 end
 
 function MOI.set(
--- a/src/Bridges/Constraint/bridges/number_conversion.jl
+++ b/src/Bridges/Constraint/bridges/number_conversion.jl
@@ -124,9 +124,9 @@ end
 function MOI.supports(
     model::MOI.ModelLike,
     attr::Union{MOI.ConstraintPrimalStart,MOI.ConstraintDualStart},
-    ::Type{NumberConversionBridge{T,F1,S1,F2,S2}},
-) where {T,F1,S1,F2,S2}
-    return MOI.supports(model, attr, MOI.ConstraintIndex{F2,S2})
+    ::Type{<:NumberConversionBridge},
+)
+    return true
 end
 
 function MOI.get(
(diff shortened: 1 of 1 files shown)


src/Bridges/Constraint/bridges/number_conversion.jl
```diff
@@ -120,3 +120,43 @@ function MOI.get(
...
+        MOI.ConstraintPrimal,
+        MOI.ConstraintDual,
+    },
+    bridge::NumberConversionBridge{T,F1,S1,F2,S2},
+) where {T,F1,S1,F2,S2}
+    ret = MOI.get(model, attr, bridge.constraint)
+    if ret === nothing
+        return nothing
+    end
+    return convert(MOI.Utilities.value_type(T,F1), ret)
```
odow, line 146:
```suggestion
    return convert(MOI.Utilities.value_type(T, F1), ret)
```

src/Bridges/Constraint/bridges/number_conversion.jl
```diff
@@ -120,3 +120,43 @@ function MOI.get(
     s = MOI.get(model, MOI.ConstraintSet(), bridge.constraint)
     return MOI.Utilities.convert_approx(S1, s)
 end
+
+function MOI.supports(
+    model::MOI.ModelLike,
+    attr::Union{MOI.ConstraintPrimalStart,MOI.ConstraintDualStart},
+    ::Type{NumberConversionBridge{T,F1,S1,F2,S2}},
+) where {T,F1,S1,F2,S2}
+    return MOI.supports(model, attr, MOI.ConstraintIndex{F2,S2})
```
blegat, line 129: You should just return `true`, the `AbstractBridgeOptimizer` takes care of also asking `supports` downstream

src/Bridges/Constraint/bridges/number_conversion.jl
```diff
@@ -120,3 +120,43 @@ function MOI.get(
...
+        MOI.ConstraintPrimal,
+        MOI.ConstraintDual,
+    },
+    bridge::NumberConversionBridge{T,F1,S1,F2,S2},
+) where {T,F1,S1,F2,S2}
+    ret = MOI.get(model, attr, bridge.constraint)
+    if ret === nothing
+        return nothing
+    end
+    return convert(MOI.Utilities.value_type(T, F1), ret)
```
blegat, line 146: This won't work for vector functions

src/Bridges/Constraint/bridges/number_conversion.jl
```diff
@@ -120,3 +120,43 @@ function MOI.get(
...
+    attr::Union{
+        MOI.ConstraintPrimalStart,
+        MOI.ConstraintDualStart,
+        MOI.ConstraintPrimal,
+        MOI.ConstraintDual,
+    },
+    bridge::NumberConversionBridge{T,F1,S1,F2,S2},
+    value,
+) where {T,F1,S1,F2,S2}
+    MOI.set(model, attr, bridge.constraint, value)
```
blegat, line 160: Don't you need to convert to `T` / `Vector{T}` ?

blegat:
Yes, `value_type` seems to be the way to go

src/Bridges/Constraint/bridges/number_conversion.jl
odow: Yes it will?
```Julia
julia> convert(Vector{Float64}, [1, 2])
2-element Vector{Float64}:
 1.0
 2.0
```

src/Bridges/Constraint/bridges/number_conversion.jl
odow: I've typically left that up to the lower-most solver

src/Bridges/Constraint/bridges/number_conversion.jl
odow: Hmm. I guess we're inconsistent about this. I should go through and check.

src/Bridges/Constraint/bridges/number_conversion.jl
blegat: `value_type` is so too good

blegat (approved):

PR #2339, opened 2023-11-02
[Bridges] add support for Constraint{Primal,Dual}Start to SquareBridge

odow:

### DIFF
--- a/src/Bridges/Constraint/bridges/square.jl
+++ b/src/Bridges/Constraint/bridges/square.jl
@@ -279,18 +279,17 @@ end
 function MOI.supports(
     model::MOI.ModelLike,
     attr::Union{MOI.ConstraintPrimalStart,MOI.ConstraintDualStart},
-    ::Type{SquareBridge{T,F,G,TT,ST}},
-) where {T,F,G,TT,ST}
-    return MOI.supports(model, attr, MOI.ConstraintIndex{F,TT}) &&
-           MOI.supports(model, attr, MOI.ConstraintIndex{G,MOI.EqualTo{T}})
+    ::Type{<:SquareBridge},
+)
+    return true
 end
 
 function MOI.set(
     model::MOI.ModelLike,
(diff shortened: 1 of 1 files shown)


odow:
@blegat please take a look. You understand better than me.

src/Bridges/Constraint/bridges/square.jl
```diff
@@ -356,3 +413,79 @@ function MOI.get(
...
+            # suppose the constraint is
+            #   [0 x; y 0] in PositiveSemidefiniteConeSquare(2).
+            # If the dual is
+            #   [λ1 λ3; λ2 λ4]
+            # then we have `y λ2 + x λ3` in the Lagrangian.
+            #
+            # In the bridged model, the constraint is
+            #   [0, x, 0] in PositiveSemidefiniteConeTriangle(2).
+            #   [x - y] in Zeros(1)
+            # If the dual is
```
blegat, line 455: I don't think you need to copy-past the comment, just say that we are doing the inverse of what in the getter

src/Bridges/Constraint/bridges/square.jl
```diff
@@ -276,12 +276,37 @@ function MOI.get(::MOI.ModelLike, ::MOI.ConstraintSet, bridge::SquareBridge)
     return bridge.square_set
 end
 
+function MOI.supports(
+    model::MOI.ModelLike,
+    attr::Union{MOI.ConstraintPrimalStart,MOI.ConstraintDualStart},
+    ::Type{SquareBridge{T,F,G,TT,ST}},
+) where {T,F,G,TT,ST}
+    return MOI.supports(model, attr, MOI.ConstraintIndex{F,TT}) &&
```
blegat, line 284: Just return `true`, `AbstractBridgeOptimizer` takes care of taking these into account

odow: Actually, I don't think this is true?

PR #2340, opened 2023-11-02
[Bridges] fix supports of Constraint and Variable starts for some bridges

odow:


src/Bridges/Variable/bridges/hermitian.jl
```diff
@@ -332,11 +332,11 @@ function unbridged_map(
...
 
 function MOI.supports(
-    ::MOI.ModelLike,
-    ::MOI.VariablePrimalStart,
+    model::MOI.ModelLike,
+    attr::MOI.VariablePrimalStart,
     ::Type{<:HermitianToSymmetricPSDBridge},
 )
-    return true
+    return MOI.set(model, attr, MOI.VariableIndex)
```
odow, line 339:
```suggestion
    return MOI.supports(model, attr, MOI.VariableIndex)
```

PR #2348, opened 2023-11-07
[Bridges] add ConstraintPrimalStart for GeoMeanBridge

blegat:
I though the tests would call this since `supports` is now `true` but it seems it does not.

See https://github.com/jump-dev/MathOptInterface.jl/issues/684

### DIFF
--- a/src/Bridges/Constraint/bridges/geomean.jl
+++ b/src/Bridges/Constraint/bridges/geomean.jl
@@ -315,7 +315,11 @@ end
 
 function _get_attribute(model, attr, bridge::GeoMeanBridge{T}) where {T}
     output = Vector{T}(undef, bridge.d)
-    output[1] = MOI.get(model, attr, bridge.t_upper_bound_constraint)
+    ret = MOI.get(model, attr, bridge.t_upper_bound_constraint)
+    if ret === nothing
+        return ret
+    end
+    output[1] = ret
     if bridge.d == 2
         output[2] = MOI.get(model, attr, bridge.x_nonnegative_constraint)[1]
     else
@@ -333,12 +337,14 @@ function _get_attribute(model, attr, bridge::GeoMeanBridge{T}) where {T}
 end
 
 function MOI.supports(
-    ::MOI.ModelLike,
-    ::MOI.ConstraintPrimalStart,
-    ::Type{<:GeoMeanBridge},
-)
-    @show @__LINE__
-    return true
+    model::MOI.ModelLike,
+    attr::MOI.ConstraintPrimalStart,
+    ::Type{GeoMeanBridge{T,F,G,H}},
+) where {T,F,G,H}
+    FS, GS = MOI.LessThan{T}, MOI.RotatedSecondOrderCone
+    return MOI.supports(model, attr, MOI.ConstraintIndex{F,FS}) &&
+           MOI.supports(model, attr, MOI.ConstraintIndex{G,GS}) &&
+           MOI.supports(model, attr, MOI.ConstraintIndex{H,MOI.Nonnegatives})
 end
 
 function MOI.get(
@@ -347,6 +353,9 @@ function MOI.get(
     bridge::GeoMeanBridge,
 )
     output = _get_attribute(model, attr, bridge)
+    if output === nothing
+        return nothing
+    end
     N = length(bridge.xij) + 1
     # the constraint is t - x_l1/sqrt(2^l) ≤ 0, we need to add the value of x_l1
     if bridge.d == 2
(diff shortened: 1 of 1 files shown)


src/Bridges/Constraint/bridges/geomean.jl
```diff
@@ -332,9 +332,14 @@ function _get_attribute(model, attr, bridge::GeoMeanBridge{T}) where {T}
     return output
 end
 
+function MOI.supports(::MOI.ModelLike, ::MOI.ConstraintPrimalStart, ::GeoMeanBridge)
```
odow, line 335:
```suggestion
function MOI.supports(
    ::MOI.ModelLike,
    ::MOI.ConstraintPrimalStart,
    ::Type{<:GeoMeanBridge},
)
```

src/Bridges/Constraint/bridges/geomean.jl
```diff
@@ -332,9 +332,14 @@ function _get_attribute(model, attr, bridge::GeoMeanBridge{T}) where {T}
...
+    @show @__LINE__
+    return true
```
odow, line 337: This is not true. It needs to check if the `model` supports it as well

src/Bridges/Constraint/bridges/geomean.jl
```diff
@@ -349,6 +354,53 @@ function MOI.get(
...
+                a = xij[offset_next+2j-1]
+                b = xij[offset_next+2j]
+            end
+            c = sqrt(2a * b)
+            xij[offset+j] = c
+            MOI.set(model, attr, bridge.rsoc_constraints[offset+j], [a, b, c])
+        end
+        offset -= 1 << (i - 2)
+    end
+    @assert offset == 0
```
odow, line 400:
```suggestion
    @assert offset == 0
    return
```

src/Bridges/Constraint/bridges/geomean.jl
```diff
@@ -349,6 +354,53 @@ function MOI.get(
...
 
+function MOI.set(
+    model::MOI.ModelLike,
+    attr::MOI.ConstraintPrimalStart,
+    bridge::GeoMeanBridge,
+    value,
+)
+    @show @__LINE__
+    if d == 2
+        MOI.set(model, attr, bridge.t_upper_bound_constraint, value[1] - value[2])
```
odow, line 365:
```suggestion
        MOI.set(
            model,
            attr,
            bridge.t_upper_bound_constraint,
            value[1] - value[2],
        )
```

src/Bridges/Constraint/bridges/geomean.jl
```diff
@@ -349,6 +354,53 @@ function MOI.get(
...
+        MOI.set(model, attr, bridge.x_nonnegative_constraint, [value[2]])
+        return
+    end
+    t = value[1]
+    n = d - 1
+    l = _ilog2(n)
+    N = 1 << l
+    sN = one(T) / sqrt(N)
+    xij = zeros(N - 1)
+    xl1 = (prod(value[2:end]))^(1/n) / xN
```
odow, line 375:
```suggestion
    xl1 = prod(value[2:end])^(1 / n) / xN
```

src/Bridges/Constraint/bridges/geomean.jl
```diff
@@ -349,6 +354,53 @@ function MOI.get(
...
+            end
+            c = sqrt(2a * b)
+            xij[offset+j] = c
+            MOI.set(model, attr, bridge.rsoc_constraints[offset+j], [a, b, c])
+        end
+        offset -= 1 << (i - 2)
+    end
+    @assert offset == 0
+end
+
```
odow, line 402:
```suggestion
```

src/Bridges/Constraint/bridges/geomean.jl
```diff
@@ -349,6 +358,58 @@ function MOI.get(
     return output
 end
 
+function MOI.set(
+    model::MOI.ModelLike,
+    attr::MOI.ConstraintPrimalStart,
+    bridge::GeoMeanBridge,
+    value,
```
odow, line 365: This is also going to need to handle the `nothing` case

src/Bridges/Constraint/bridges/geomean.jl
```diff
@@ -349,6 +367,65 @@ function MOI.get(
...
+        new_value = value[1] - value[2]
+        MOI.set(model, attr, bridge.t_upper_bound_constraint, new_value)
+        MOI.set(model, attr, bridge.x_nonnegative_constraint, [value[2]])
+        return
+    end
+    n = bridge.d - 1
+    l = _ilog2(n)
+    N = 1 << l
+    sN = one(T) / sqrt(N)
+    xl1 = prod(value[2:end])^(1 / n) / xN
```
odow, line 386: What is `xN`?

src/Bridges/Constraint/bridges/geomean.jl
```diff
@@ -349,6 +367,65 @@ function MOI.get(
...
+        num_lvars = 1 << (i - 1)
+        offset_next = offset + num_lvars
+        for j in 1:num_lvars
+            a, b = if i == l
+                _get_x(2j - 1), _get_x(2j)
+            else
+                xij[offset_next+2j-1], xij[offset_next+2j]
+            end
+            c = sqrt(2a * b)
+            xij[offset+j] = c
```
odow, line 404: This gets an indexing error. I don't think it's right.

src/Bridges/Constraint/bridges/geomean.jl
blegat: We had the reverse discussions in your PRs, the `AbstractBridgeOptimizer` takes care of that

odow: No, I don't think it does: https://github.com/jump-dev/MathOptInterface.jl/pull/2340

Is this the method you're thinking of? It just calls the bridge.
https://github.com/jump-dev/MathOptInterface.jl/blob/8b4b3ca384c22250e7e29bd90685e8d907a7e9f5/src/Bridges/bridge_optimizer.jl#L1474-L1527

It does not loop through the various constraints added.

src/Bridges/Constraint/bridges/geomean.jl
```diff
@@ -332,23 +336,108 @@ function _get_attribute(model, attr, bridge::GeoMeanBridge{T}) where {T}
     return output
 end
 
+function MOI.supports(
+    model::MOI.ModelLike,
+    attr::MOI.ConstraintPrimalStart,
+    ::Type{GeoMeanBridge{T,F,G,H}},
+) where {T,F,G,H}
+    FS, GS = MOI.LessThan{T}, MOI.RotatedSecondOrderCone
+    return MOI.supports(model, attr, MOI.ConstraintIndex{F,FS}) &&
```
odow, line 345:
```suggestion
    return MOI.supports(model, MOI.VariablePrimalStart(), MOI.VariableIndex) &&
           MOI.supports(model, attr, MOI.ConstraintIndex{F,FS}) &&
```

src/Bridges/Constraint/bridges/geomean.jl
```diff
@@ -332,23 +336,108 @@ function _get_attribute(model, attr, bridge::GeoMeanBridge{T}) where {T}
...
+            else
+                xij[offset_next+2j-1], xij[offset_next+2j]
+            end
+            c = sqrt(2a * b)
+            xij[offset+j] = c
+            MOI.set(model, attr, bridge.rsoc_constraints[offset+j], [a, b, c])
+        end
+    end
+    @assert offset == 0
+    MOI.set(model, _variable_attribute(attr), bridge.xij, xij)
```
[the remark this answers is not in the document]
odow: This is also pretty complicated, which I guess is why I didn't implement it when I was going through and implementing a lot of these a while ago 😆

src/Bridges/Constraint/bridges/geomean.jl
blegat: Ah but then I was wrong hence https://github.com/jump-dev/MathOptInterface.jl/pull/2339 etc... should be updated

odow: I already did :smile: https://github.com/jump-dev/MathOptInterface.jl/pull/2340

blegat: I didn't follow that, looks good :)

PR #2350, opened 2023-11-08
[Utilities] improve errors and docstrings for querying attributes

### LINKED ISSUE #2345: Attributes simultaneously set by the user and by the optimizer
I am having some trouble with attributes that are set both by the user and by the optimizer. I think that the UI I am trying to provide collides with the assumptions made by the `supports/is_set_by_optimize/is_copyable` interface.

To be more clear, in psrenergy/ToQUBO.jl#77 we are having a recurrent instability with the `ToQUBO.Attributes.ConstraintEncodingPenalty <: ... <: MOI.AbstractConstraintAttribute` attribute.
This attribute is meant to allow the user to set manually the penalty factor that will multiply the encoding of a variable in a unconstrained setting.
If not provided, our meta-solver will compute it automatically.
There is even a third scenario where the solver finds even tighter bounds and might select computed values regardless.

In the last two cases, the user will likely want to query the computed results after the call to the optimizer.
In our latest CI run, there are two broken tests where the `MOI.get(model, ToQUBO.Attributes.ConstraintEncodingPenalty(), c)` calls reach a _key not found_-like error in an interaction with the `DoubleDict` struct that maps constraints over bridges.
This error appears after setting `is_set_by_optimizer` and `is_copyable` to `true`. `is_copyable` is set to avoid another error related to `support`. `is_set_by_optimizer`, on the other hand, is necessary for allowing the penalty values to be requested by the caching optimizer on top of it.

Is there any way to stick to the interface while still providing the desired interface?

I thought of copying these attributes _after_ `optimize!` is called. This would work if `ToQUBO.Optimizer` is accessed directly by a `MOI`/`JuMP` model but would fail if `MOI.copy_to` or `MOI.optimize!(dst, src)` were called by other solvers.

odow:
Closes #2345

But this PR needs some careful thinking.

https://github.com/jump-dev/MathOptInterface.jl/actions/runs/6805116666

### DIFF
--- a/src/Utilities/cachingoptimizer.jl
+++ b/src/Utilities/cachingoptimizer.jl
@@ -867,7 +867,7 @@ end
 function _throw_if_get_attribute_not_allowed(
     model::CachingOptimizer,
     attr;
-    needs_optimizer_map::Bool = false,
+    needs_optimizer_map::Bool,
 )
     # If the state(model) == EMPTY_OPTIMIZER, then
     # `model.model_to_optimizer_map[index]` might be empty (because copy_to
@@ -887,7 +887,11 @@ function MOI.get(model::CachingOptimizer, attr::MOI.AbstractModelAttribute)
     if !MOI.is_set_by_optimize(attr)
         return MOI.get(model.model_cache, attr)
     end
-    _throw_if_get_attribute_not_allowed(model, attr)
+    _throw_if_get_attribute_not_allowed(
+        model,
+        attr;
+        needs_optimizer_map = false,
+    )
     return _get_model_attribute(model, attr)
 end
 
@@ -1046,7 +1050,11 @@ function MOI.get(model::CachingOptimizer, attr::MOI.AbstractOptimizerAttribute)
     # `NumberOfThreads`) should also be stored in the cache so we could
     # return the value stored in the cache instead. However, for
     # non-copyable attributes( e.g. `SolverName`) the error is appropriate.
-    _throw_if_get_attribute_not_allowed(model, attr)
+    _throw_if_get_attribute_not_allowed(
+        model,
+        attr;
+        needs_optimizer_map = false,
+    )
     return map_indices(
         model.optimizer_to_model_map,
         attr,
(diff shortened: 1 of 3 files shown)


odow:
This would break InfiniteOpt with the expected

image

odow:
cc @pulsipher thoughts on this change? It changes a `ErrorException` to ` MOI.GetAttributeNotAllowed`. 

I guess I could keep the `ErrorException`, but the more specific error is probably what should be thrown.

pulsipher:
I have no objections with updating InfiniteOpt to the new error type.

src/Utilities/cachingoptimizer.jl
```diff
@@ -901,51 +916,33 @@ function MOI.get(
...
     index::MOI.Index,
 )
-    if MOI.is_set_by_optimize(attr)
-        if state(model) == NO_OPTIMIZER
-            error(
-                "Cannot query $(attr) from caching optimizer because no " *
-                "optimizer is attached.",
-            )
-        end
-        return map_indices(
```
odow, line 911: @blegat the problematic error because we haven't checked for `state(model) == EMPTY_OPTIMIZER` here. 

and there's a complication: a call like `index_map, false = optimize!(dest, src)` leaves the model in `EMPTY_OPTIMIZER`, even though `model_to_optimizer_map` is populated.

blegat: It seems to me `needs_optimizer_map::Bool = false` should the trick, I can't of a corner case for which your idea won't work

src/Utilities/cachingoptimizer.jl
```diff
@@ -864,15 +864,30 @@ function _get_model_attribute(
     )
 end
 
+function _throw_if_get_attribute_not_allowed(
+    model::CachingOptimizer,
+    attr;
+    needs_optimizer_map::Bool = false,
```
blegat, line 870: Maybe don't give a default value for this keyword and set it explicitly. I feel there is less chance of bug if we set it explicitly at every call

blegat (approved):

PR #2356, opened 2023-12-10
Simpler side_dimension_for_vectorized_dimension

blegat:
As a double check, there does not seem to be any difference until 100 millions.
```julia
findall(m -> isqrt(2m) != div(isqrt(1 + 8n), 2), 1:100000000)
Int64[]
```

### DIFF
--- a/src/Utilities/sets.jl
+++ b/src/Utilities/sets.jl
@@ -128,19 +128,17 @@ function is_diagonal_vectorized_index(index::Base.Integer)
     return isqrt(perfect_square)^2 == perfect_square
 end
 
+# We have `d*(d+1)/2 = n` so `2n = d^2 + d` hence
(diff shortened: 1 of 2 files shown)


chriscoey:
Fascinating

blegat:
It came up when trying to understand the link between `isqrt(2m)` which is used by SDPLR and the Barvinok-Pataki bound in https://github.com/blegat/SDPLR.jl/pull/11. Once I had this 2-line arguments, it clarified things a lot :smile:

odow:
I guess, just as proof, because @blegat's original example has a typo:
```julia
julia> findall(n -> isqrt(2n) != div(isqrt(1 + 8n), 2), 1:100_000_000)
Int64[]
```

PR #2358, opened 2023-12-13
Fix ConstraintIndex conflicts between variable and constraint bridges

blegat:
Fix one of the bugs detected by https://github.com/jump-dev/MathOptInterface.jl/pull/2357

- [ ] Add comments/docstrings to explain the new design
- [ ] Add a custom test not not only rely on the tests brought up by https://github.com/jump-dev/MathOptInterface.jl/pull/2357
- [ ] Add a check in `add_bridge` and `remove_bridge` that the negation does not change


src/Bridges/Constraint/map.jl
```diff
@@ -123,10 +126,22 @@ function _iterate_sv(
     return ci => bridge, (2, elem_state[2])
 end
 
-_index(index, F, S) = MOI.ConstraintIndex{F,S}(index)
+function register_negated(map::Map, S::Type, negated)
```
odow, line 129:
```suggestion
function register_negated(map::Map, ::Type{S}, negated::Bool) where {S}
```

odow:
We probably need to discuss this more. I didn't have a good look at it. Will do tomorrow

blegat:
Actually this does not work since we may also have the case where constrained variables in `S` are bridged and `VectorOfVariables`-in-`S` are not bridged but we are forced to apply a `FunctionizeBridge` because one of the variable is a bridged variable. Let me think about it

blegat:
I have a simpler idea, hold off reviewing

odow:
Closing in favor of #2362

PR #2359, opened 2023-12-13
Fix RSOCtoPSD bridge for dimension 2

blegat:
With this `MultiSetMapBridge`, it should be easy to fix other similar case, e.g., `SecondOrderCone(1)` -> `Nonnegatives(1)` while `SecondOrderCone(2)` -> `RotatedSecondOrderCone(2)` but we can wait for someone or for a test (like in this case) requesting to fix these corner cases before doing it.

### DIFF
--- a/src/Bridges/Constraint/bridges/soc_to_psd.jl
+++ b/src/Bridges/Constraint/bridges/soc_to_psd.jl
@@ -177,7 +177,10 @@ struct RSOCtoPSDBridge{T,F,G} <: SetMapBridge{
     F,
     G,
 }
-    constraint::MOI.ConstraintIndex{F,MOI.PositiveSemidefiniteConeTriangle}
+    constraint::Union{
+        MOI.ConstraintIndex{F,MOI.PositiveSemidefiniteConeTriangle},
+        MOI.ConstraintIndex{G,MOI.Nonnegatives},
+    }
 end
 
 const RSOCtoPSD{T,OT<:MOI.ModelLike} =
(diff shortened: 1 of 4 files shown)


odow:
I've hit this a few times, and always seen this choice as intentional.

test/Bridges/Constraint/soc_to_psd.jl
```diff
@@ -146,28 +146,31 @@ function test_RSOCtoPSD()
...
+                MOI.PositiveSemidefiniteConeTriangle,
+                0,
+            ),
+            (MOI.VectorAffineFunction{Float64}, MOI.Nonnegatives, 0),
+        ),
     )
     return
 end
 
-# TODO(odow): we should fix this so it does not error.
```
odow, line 158: I guess perhaps not intentional. I like seeing TODOs from the past.

blegat:
Yes I'm also hitting this often. When you have a code generating socs of various dimensions and one if them is t^2 >= x^2, it is bridged to RSOC and it tries to bridge from RSOC to PSD and you get this error. Anothet use case of this MultiSetMap is to have faster reformulations. For instance, in Sumofsquares, I reformulate 2x2 psd into socs instead of psd since it's faster

src/Bridges/Constraint/bridges/soc_to_psd.jl
```diff
@@ -208,13 +237,27 @@ function MOI.Bridges.inverse_map_set(
...
 end
 
+function MOI.Bridges.inverse_map_set(
+    ::Type{<:RSOCtoPSDBridge},
+    ::MOI.Nonnegatives,
+)
+    return MOI.RotatedSecondOrderCone(2)
+end
+
+function _rsoc_to_psd_too_small_message(n)
```
odow, line 247: Remove

src/Bridges/Constraint/set_map.jl
```diff
@@ -5,91 +5,98 @@
...
-end
-
 # Attributes, Bridge acting as a model
 
-function MOI.get(
-    ::SetMapBridge{T,S2,S1,F},
-    ::MOI.NumberOfConstraints{F,S2},
-)::Int64 where {T,S2,S1,F}
-    return 1
+_get(::MOI.ConstraintIndex, ::MOI.NumberOfConstraints) = 0
```
odow, line 71: I've found that these "helper" functions just make things more difficult in the long run. Let's just explicitly write out the implementation for both constraint types.

odow:
Are you up late working on this? Otherwise I will push forward this afternoon

odow:
https://github.com/jump-dev/MathOptInterface.jl/actions/runs/7204367333

odow (approved):

PR #2357, opened 2023-12-13
Run all tests against bridged SDPA model

blegat:
Closes #2375

When looking over the failing tests of https://github.com/blegat/SDPLR.jl/, I noticed some were due to MOI bugs. When looking at how to reproduce them, I found out that we could have detected them by running against all the tests. Looking back at the history, it seems I initially just called `nametest` because it wasn't so easy to run all the tests. When @odow migrated to the new test infrastructure, he used `include = ["ConstraintName", "VariableName"],` so that's its equivalent to `nametest`. This did not take the opportunity brought by this new test infrastructure to run it against all the tests. These tests are usually failing for all solvers so fixing them should help many solver wrappers at the same time.

This brings up a few bugs that I suggest we first fix in separate PRs with their own tests:

`GeometricSDPA` (so it should help SCS, ECOS, COSMO, SeDuMi, SDPT3, CDCS, Clarabel, ...)

- [x] `test_conic_SecondOrderCone_INFEASIBLE` https://github.com/jump-dev/MathOptInterface.jl/pull/2359 
- [x] `test_model_LowerBoundAlreadySet` https://github.com/jump-dev/MathOptInterface.jl/pull/2376
- [x] `test_model_UpperBoundAlreadySet` https://github.com/jump-dev/MathOptInterface.jl/pull/2376

`StandardSDPA` (so it should help CSDP, SDPA, SDPLR, DSDP, ...)

- [x] `test_basic_VectorOfVariables_Nonpositives` https://github.com/jump-dev/MathOptInterface.jl/pull/2362
- [x] `test_basic_VectorOfVariables_SecondOrderCone` https://github.com/jump-dev/MathOptInterface.jl/pull/2360
- [x] `test_conic_SecondOrderCone_INFEASIBLE` https://github.com/jump-dev/MathOptInterface.jl/pull/2359 
- [ ] `test_model_ListOfVariablesWithAttributeSet`
- [ ] `test_model_LowerBoundAlreadySet`
- [ ] `test_model_UpperBoundAlreadySet`
- [ ] `test_model_ScalarFunctionConstantNotZero`
- [x] `test_model_delete` https://github.com/jump-dev/MathOptInterface.jl/pull/2377
[...]

PR #2362, opened 2023-12-14
[Bridges] Fix ConstraintIndex conflicts between variable and constraint bridges

blegat:
Closes https://github.com/jump-dev/MathOptInterface.jl/pull/2358

### DIFF
--- a/src/Bridges/Constraint/map.jl
+++ b/src/Bridges/Constraint/map.jl
@@ -280,13 +280,29 @@ Return a new constraint index `ci` and store the mapping `ci => bridge`.
 """
 function add_key_for_bridge end
 
+_ensure_available(::Map, ::Type, ::Type, ::Function) = nothing
+function _ensure_available(
+    map::Map,
+    F::Type{MOI.VectorOfVariables},
+    ::Type{S},
+    is_available::Function,
+) where {S}
+    while !is_available(MOI.ConstraintIndex{F,S}(length(map.bridges) + 1))
+        push!(map.bridges, nothing)
+        push!(map.constraint_types, (F, S))
+    end
+    return
+end
+
 function add_key_for_bridge(
     map::Map,
     bridge::AbstractBridge,
     ::F,
     ::S,
+    is_available::Function,
 ) where {F<:MOI.AbstractFunction,S<:MOI.AbstractSet}
     _register_for_final_touch(map, bridge)
+    _ensure_available(map, F, S, is_available)
     push!(map.bridges, bridge)
     push!(map.constraint_types, (F, S))
     return _index(length(map.bridges), F, S)
--- a/src/Bridges/Constraint/map.jl
+++ b/src/Bridges/Constraint/map.jl
@@ -274,6 +274,7 @@ end
         bridge::AbstractBridge,
         func::MOI.AbstractFunction,
         set::MOI.AbstractSet,
+        is_available::Function,
     )
 
 Return a new constraint index `ci` and store the mapping `ci => bridge`.
@@ -313,6 +314,7 @@ function add_key_for_bridge(
     bridge::AbstractBridge,
     func::MOI.VariableIndex,
     ::S,
+    ::Function,
 ) where {S<:MOI.AbstractScalarSet}
     _register_for_final_touch(map, bridge)
     map.single_variable_constraints[(func.value, S)] = bridge
--- a/src/Bridges/Constraint/map.jl
+++ b/src/Bridges/Constraint/map.jl
@@ -288,7 +288,7 @@ function _ensure_available(
     ::Type{S},
     is_available::Function,
 ) where {S}
-    while !is_available(MOI.ConstraintIndex{F,S}(length(map.bridges) + 1))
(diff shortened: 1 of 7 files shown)


odow:
Let me know when you get this to a point that you want me to take a look.

blegat:
Ready for review

src/Bridges/Constraint/map.jl
```diff
@@ -274,19 +274,39 @@ end
...
 function add_key_for_bridge end
 
+_ensure_available(::Map, ::Type, ::Type, ::Function) = nothing
+function _ensure_available(
+    map::Map,
+    F::Type{MOI.VectorOfVariables},
+    ::Type{S},
+    is_available::Function,
+) where {S}
+    while !is_available(MOI.ConstraintIndex{F,S}(-length(map.bridges) - 1))
```
odow, line 294: So this is a bit confusing. We're just going to push in a bunch of `nothing` until the index becomes available?

What if this ends up adding thousands of elements?

src/Bridges/Variable/map.jl
```diff
@@ -338,26 +382,42 @@ function add_key_for_bridge(
...
     push!(map.parent_index, map.current_context)
     bridge_index = Int64(length(map.parent_index))
-    push!(map.info, -MOI.dimension(set))
+    F = MOI.VectorOfVariables
+    S = typeof(set)
+    while !is_available(
+        MOI.ConstraintIndex{F,S}(-length(map.vector_of_variables_map) - 1),
+    )
+        push!(map.vector_of_variables_map, 0)
+        push!(map.vector_of_variables_length, 0)
```
odow, line 416: This needs tests

src/Bridges/Variable/map.jl
```diff
@@ -338,26 +382,42 @@ function add_key_for_bridge(
...
-`typeof(set)`. It returns a tuple with `variables` and the constraint index
-`MOI.ConstraintIndex{MOI.VectorOfVariables, typeof(set)}(first(variables).value)`.
+`typeof(set)`. It returns a tuple with `variables` and a constraint index
+`ci::MOI.ConstraintIndex{MOI.VectorOfVariables, typeof(set)}` such that
+`is_available(ci)`.
 """
 function add_keys_for_bridge(
     map::Map,
     bridge_fun::Function,
     set::MOI.AbstractVectorSet,
```
odow, line 401:
```suggestion
    set::S,
```

src/Bridges/Variable/map.jl
```diff
@@ -338,26 +382,42 @@ function add_key_for_bridge(
...
+    is_available::Function,
 )
```
odow, line 403:
```suggestion
) where {S<:MOI.AbstractVectorSet}
```

src/Bridges/Variable/map.jl
```diff
@@ -338,26 +382,42 @@ function add_key_for_bridge(
...
     if iszero(MOI.dimension(set))
         return MOI.VariableIndex[],
         MOI.ConstraintIndex{MOI.VectorOfVariables,typeof(set)}(0)
     end
     push!(map.parent_index, map.current_context)
     bridge_index = Int64(length(map.parent_index))
-    push!(map.info, -MOI.dimension(set))
+    F = MOI.VectorOfVariables
+    S = typeof(set)
```
odow, line 411:
```suggestion
```

odow (commented):
I feel like this needs far more tests. I always get a bit nervous with these changes.

https://github.com/jump-dev/MathOptInterface.jl/actions/runs/7268307681

odow:
This looks like it broke a bunch of solver tests, including Gurobi and CPLEX. So as well as the fix, we obviously need more tests in MOI proper.

src/Bridges/Constraint/map.jl
blegat: It means you also have thousands of constrained vector of variables. Another solution would be to have one common vector shared between `Constraint.Map` and `Variable.Map`

blegat: It just kind of break encapsulation so it's a tradeoff

blegat:
> we obviously need more tests in MOI proper.

Yes, that's also the motivation for https://github.com/jump-dev/MathOptInterface.jl/pull/2357

blegat:
New run: https://github.com/jump-dev/MathOptInterface.jl/actions/runs/7273606634

blegat:
New run: https://github.com/jump-dev/MathOptInterface.jl/actions/runs/7277998597

src/Bridges/Constraint/map.jl
odow: Fair enough. Let's wait until it's a problem.

odow:
I don't really have an opinion on the design. But I also don't have a suggestion for anything better.

I do think we need 100% coverage before merging though.

blegat:
> I do think we need 100% coverage before merging though.

We do now

odow (approved):
I'll merge so we can re-trigger #2357.

I'll also pay careful attention to `solver-tests` before tagging a new MOI release.

PR #2367, opened 2023-12-19
Fix map_indices with attributes

blegat:
This change from https://github.com/jump-dev/MathOptInterface.jl/pull/2211 broke DiffOpt (see failure in https://github.com/jump-dev/DiffOpt.jl/pull/254).


odow:
We should add DiffOpt to `solver-tests`

PR #2369, opened 2023-12-20
Fix Base.isless for sorting AbstractModelAttribute during MOI.copy_to

### LINKED ISSUE #2366: Order of model attributes during copy
As reported in https://github.com/jump-dev/MathOptInterface.jl/issues/2363#issuecomment-1862620395, the `SlackBridgePrimalDualStart` is ignored if it is copied before the objective is set which depends on the order of the model attributes.

odow:
Closes #2366

Thoughts on something like this?

An alternative would be to expand on the `_sort_priority(::Any) = 2` to have `copy_to_sort_priority(::ModelLike, ::AbstractModelAttribute)`, but then this gets a bit messy if we have to have extensions working together.

### DIFF
--- a/src/Bridges/Objective/bridges/slack.jl
+++ b/src/Bridges/Objective/bridges/slack.jl
@@ -199,6 +199,10 @@ end
 # Pretend that every model supports, and silently skip in set if unsupported
 MOI.supports_fallback(::MOI.ModelLike, ::SlackBridgePrimalDualStart) = true
 
+# Sort priority: SlackBridgePrimalDualStart after ObjectiveFunction
+Base.isless(::SlackBridgePrimalDualStart, ::MOI.ObjectiveFunction) = false
+Base.isless(::MOI.ObjectiveFunction, ::SlackBridgePrimalDualStart) = true
+
 function MOI.throw_set_error_fallback(
     ::MOI.ModelLike,
     ::SlackBridgePrimalDualStart,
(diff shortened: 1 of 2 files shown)


joaquimg:
Would a change in priorities be considered breaking?

joaquimg:
Also,
Do you envision different solvers having different priorities?
In this case the should have their own copy_to sorting with their priorities, and this default copy_to would be fine.

odow:
I think so. #2366 is really asking for a first-class priority order of attributes.

Perhaps `Base.isless` should be instead some official `MOI.CopyToPriorityFactor` or something.

But open questions are whether it should depend on the `dest::MOI.ModelLike`, or whether there is some solver-independent ordering.

The attributes really form a DAG, and we need a topological sort of them... It'd be nice to say that `ObjectiveFunction` depends on `ObjectiveSense`, and then have the order fall out of that.

blegat:
We could maybe also have `SortedListOfModelAttributes` that could be used for solvers so that they don't need to worry about the sort. `MOI.Utilities.Model` could also implement a getter for this attribute quite easily

odow:
`SortedListOfModelAttributes` is reasonable. But it means that it works only for attributes defined in MOI proper. Everything else would come at the end in an arbitrary order.

odow:
https://github.com/jump-dev/MathOptInterface.jl/pull/2371 is a much cleaner approach.

odow:
Closing in favor of #2371

PR #2371, opened 2023-12-20
Add Utilities.sorted_list_of_model_attributes

### LINKED ISSUE #2366: Order of model attributes during copy

odow:
Closes #2366
Closes #2369 

https://github.com/jump-dev/MathOptInterface.jl/actions/runs/7282138662

### DIFF
--- a/src/Utilities/copy.jl
+++ b/src/Utilities/copy.jl
@@ -10,10 +10,10 @@
 include("copy/index_map.jl")
 
 """
-    SortedListOfModelAttributesSet()
+    sorted_list_of_model_attributes(model::MOI.ModelLike)
 
 Return the vector of [`MOI.ListOfModelAttributesSet`](@ref), except that the
-elements are sorted according to [`sort_by_weight`].
+elements are sorted according to [`sorted_list_of_model_attributes`](@ref).
 
 During [`MOI.copy_to`](@ref), we need to copy [`MOI.UserDefinedFunction`](@ref)s
 first, so that they are in the model before we deal with the objective function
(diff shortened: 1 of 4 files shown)


src/Utilities/copy.jl
```diff
@@ -27,12 +59,7 @@ function pass_attributes(
     src::MOI.ModelLike,
     index_map::IndexMap,
 )
-    attrs = MOI.get(src, MOI.ListOfModelAttributesSet())
-    # We need to deal with the UserDefinedFunctions first, so that they are in
-    # the model before we deal with the objective function or the constraints.
-    # We also need `ObjectiveSense` to be set before `ObjectiveFunction`.
-    sort!(attrs; by = _sort_priority)
-    for attr in attrs
+    for attr in MOI.get(src, MOI.SortedListOfModelAttributesSet())
```
odow, line 62:
```suggestion
    for attr in MOI.get(src, SortedListOfModelAttributesSet())
```

odow:
Looking into why Juniper started to fail tests

odow:
So this issue is that Juniper defines a bunch of model filters:
https://github.com/lanl-ansi/Juniper.jl/blob/05ae25efee510ef6d4cbe90bb098460ec728a8f3/src/filter.jl#L1-L29
and we can't define a new `AbstractModelAttribute` like this without breaking existing behavior.

src/Utilities/copy.jl
```diff
@@ -9,9 +9,39 @@
 
 include("copy/index_map.jl")
 
-_sort_priority(::Any) = 2
-_sort_priority(::MOI.UserDefinedFunction) = 0
-_sort_priority(::MOI.ObjectiveSense) = 1
```
odow, line 14: So in retrospect, all this PR really needed was `_sort_priority(::Any) = 3` and `_sort_priority(::MOI.ObjectiveFunction) = 2`.

So we want to make this all public?

odow:
https://github.com/jump-dev/MathOptInterface.jl/actions/runs/7282725778

src/Utilities/copy.jl
joaquimg: I would do this simple fix, and reopen this PR if this keeps coming back and someone actually needs to extend this kind of methods.

odow:
Closing in favor of #2372

PR #2373, opened 2023-12-21
[Utilities.CachingOptimizer] Enable fallback for ConstraintDual of variable indices

blegat:
This is needed for https://github.com/blegat/SDPLR.jl/pull/19 to work.


odow (approved):
I simplified the implementation a little

PR #2376, opened 2023-12-22
[Bridges] fix BoundAlreadySet errors

odow:
x-ref #2357

### DIFF
--- a/src/Test/test_model.jl
+++ b/src/Test/test_model.jl
@@ -805,12 +805,10 @@ function test_model_LowerBoundAlreadySet(
             continue
         end
         ci = MOI.add_constraint(model, x, set1)
-        err = MOI.LowerBoundAlreadySet{typeof(set1),typeof(set2)}(x)
-        @test_throws err MOI.add_constraint(model, x, set2)
+        @test_throws MOI.LowerBoundAlreadySet MOI.add_constraint(model, x, set2)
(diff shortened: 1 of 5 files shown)


src/Test/test_model.jl
```diff
@@ -805,12 +805,10 @@ function test_model_LowerBoundAlreadySet(
             continue
         end
         ci = MOI.add_constraint(model, x, set1)
-        err = MOI.LowerBoundAlreadySet{typeof(set1),typeof(set2)}(x)
-        @test_throws err MOI.add_constraint(model, x, set2)
+        @test_throws MOI.LowerBoundAlreadySet MOI.add_constraint(model, x, set2)
```
odow, line 808: So. 

If a bridge adds x-in-Interval, it bridges to `x-in-LessThan` and `x-in-GreaterThan`, which are then bridged to `VectorAffineFunction-in-Nonnegatives`.

But, it means that `MOI.is_valid` returns `true` for the `x-in-LessThan` constraint (because there exists a bridge mapping that exact constraint, and we could't tell if it was introduced by the user or a bridge).

That means the "wrong" type error is thrown here.

Perhaps it's easiest just to relax this error check. The main point is that the right type is thrown.

odow: I've worked around this for now by changing the order in which we check the sets.

Issue #2384, opened 2024-01-02
Failure with CachingOptimizer


blegat:
This seems like a bug in `CachingOptimizer`
```julia
using Test
import MathOptInterface as MOI
import CSDP

function tests(cache)
    model = MOI.instantiate(CSDP.Optimizer, with_bridge_type = Float64)
    if cache
        model = MOI.Utilities.CachingOptimizer(
            MOI.Utilities.UniversalFallback(MOI.Utilities.Model{Float64}()),
            model,
        )
    end
    MOI.set(model, MOI.Silent(), true)
    MOI.Test.runtests(
      model,
      MOI.Test.Config(
          rtol = 1e-3,
          atol = 1e-3,
          exclude = Any[
              MOI.ConstraintBasisStatus,
              MOI.VariableBasisStatus,
              MOI.ObjectiveBound,
              MOI.SolverVersion,
          ],
      ),
      include = Regex[
          r"test_variable_solve_with_lowerbound$",
          r"test_modification_delete_variables_in_a_batch$",
      ],
    )
end

@testset "$cache" for cache in [false, true]
   tests(cache)
end
```
produces
```
Test Summary: | Pass  Total  Time
false         |   18     18  0.9s
test_modification_delete_variables_in_a_batch: Test Failed at /home/blegat/.julia/packages/MathOptInterface/vrZOD/src/Test/test_modification.jl:827
  Expression: isapprox(MOI.get(model, MOI.ObjectiveValue()), T(2), config)
   Evaluated: isapprox(6.0, 2.0, MathOptInterface.Test.Config{Float64}(0.001, 0.001, MathOptInterface.OPTIMAL, MathOptInterface.INFEASIBLE, Any[MathOptInterface.ConstraintBasisStatus, MathOptInterface.VariableBasisStatus, MathOptInterface.ObjectiveBound, MathOptInterface.SolverVersion]))

Stacktrace:
 [1] macro expansion
   @ ~/.julia/juliaup/julia-1.10.0+0.x64.linux.gnu/share/julia/stdlib/v1.10/Test/src/Test.jl:672 [inlined]
[...]

PR #2393, opened 2024-01-05
[Bridges] fix deleting variable in Bridges.Variable.VectorizeBridge

odow:
One issue from #2384.

It seems simpler to throw here than to carry around a bunch of extra state about which constants we need to modify.

### DIFF
--- a/src/Bridges/Variable/bridges/vectorize.jl
+++ b/src/Bridges/Variable/bridges/vectorize.jl
@@ -140,13 +140,19 @@ function MOI.get(
     attr::MOI.ConstraintPrimal,
     bridge::VectorizeBridge,
 )
-    return throw(
-        MOI.GetAttributeNotAllowed(
-            attr,
-            "The variable `$(bridge.variable)` is bridged by the " *
-            "`VectorizeBridge`.",
-        ),
-    )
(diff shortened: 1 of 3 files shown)


src/Bridges/Variable/bridges/vectorize.jl
```diff
@@ -98,10 +98,14 @@ function MOI.get(
     return [bridge.vector_constraint]
 end
 
-# References
 function MOI.delete(model::MOI.ModelLike, bridge::VectorizeBridge)
-    MOI.delete(model, bridge.variable)
-    return
+    # It isn't safe to delete the variable because the constant may appear in
+    # other parts of the model (like the objective, or right-hand side sets).     Therefore, we don't implement
```
odow, line 103:
```suggestion
    # other parts of the model (like the objective, or right-hand side sets).
```

odow:
@blegat you might want to take a look here. I don't really understand all of the different levels of indirection.

blegat:
What's wrong with `ConstraintPrimal` ?

odow:
> What's wrong with ConstraintPrimal ?

I couldn't figure it out yet. It gets the wrong answer, but the layers of debugging are many...

But I'll remove for now. The delete fix is correct.

PR #2394, opened 2024-01-07
Fix map_indices for AbstractAutomaticDifferentiation

odow:
cc @ccoffrin

### DIFF
--- a/src/Nonlinear/types.jl
+++ b/src/Nonlinear/types.jl
@@ -186,7 +186,7 @@ abstract type AbstractAutomaticDifferentiation end
 
 function MOI.Utilities.map_indices(
     ::F,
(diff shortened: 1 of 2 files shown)


src/Nonlinear/types.jl
```diff
@@ -184,6 +184,13 @@ An abstract type for extending [`Evaluator`](@ref).
 """
 abstract type AbstractAutomaticDifferentiation end
 
+function MOI.Utilities.map_indices(
+    ::F,
+    backend::MOI.Nonlinear.AbstractAutomaticDifferentiation,
```
odow, line 189:
```suggestion
    backend::AbstractAutomaticDifferentiation,
```

blegat (approved):

PR #2401, opened 2024-01-17
[docs] clarify ordering of nonlinear tape

odow:
cc @rluce

The other thing we didn't discuss, but I assume you figured out, is that we don't really need the `CALL_MULTIVARIATE` and `CALL_UNIVARIATE`. You could figure that out based on the number of child nodes.


rluce:
Yes thanks for pointing it out, the parent array tells it all.

PR #2413, opened 2024-01-24
[Utilities] throw correct error if get_fallback fails

odow:
This is a possible fix for https://github.com/jump-dev/MathOptInterface.jl/pull/2410#issuecomment-1907199988

https://github.com/jump-dev/MathOptInterface.jl/actions/runs/7635259102

### DIFF
--- a/src/Utilities/results.jl
+++ b/src/Utilities/results.jl
@@ -37,17 +37,21 @@ the `ObjectiveFunction` value.
 """
 function get_fallback(model::MOI.ModelLike, attr::MOI.ObjectiveValue)
     MOI.check_result_index_bounds(model, attr)
-    F = MOI.get(model, MOI.ObjectiveFunctionType())
-    f = MOI.get(model, MOI.ObjectiveFunction{F}())
-    obj = eval_variables(model, f) do vi
-        return MOI.get(model, MOI.VariablePrimal(attr.result_index), vi)
+    try
+        F = MOI.get(model, MOI.ObjectiveFunctionType())
+        f = MOI.get(model, MOI.ObjectiveFunction{F}())
+        obj = eval_variables(model, f) do vi
+            return MOI.get(model, MOI.VariablePrimal(attr.result_index), vi)
+        end
+        if is_ray(MOI.get(model, MOI.PrimalStatus()))
+            # Dual infeasibility certificates do not include the primal
+            # objective constant.
+            obj -= MOI.constant(f, typeof(obj))
+        end
+        return obj
+    catch
+        throw(MOI.GetAttributeNotAllowed(attr))
     end
-    if is_ray(MOI.get(model, MOI.PrimalStatus()))
-        # Dual infeasibility certificates do not include the primal
-        # objective constant.
-        obj -= MOI.constant(f, typeof(obj))
-    end
-    return obj
 end
 
 # MOI.DualObjectiveValue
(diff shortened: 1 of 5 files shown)


src/Utilities/results.jl
```diff
@@ -37,17 +37,21 @@ the `ObjectiveFunction` value.
...
-    F = MOI.get(model, MOI.ObjectiveFunctionType())
-    f = MOI.get(model, MOI.ObjectiveFunction{F}())
-    obj = eval_variables(model, f) do vi
-        return MOI.get(model, MOI.VariablePrimal(attr.result_index), vi)
-    end
-    if is_ray(MOI.get(model, MOI.PrimalStatus()))
-        # Dual infeasibility certificates do not include the primal
-        # objective constant.
-        obj -= MOI.constant(f, typeof(obj))
+    try
```
blegat, line 40: Why is this needed ?

odow: I added the try defensively to all methods. But the issue is that `get_fallback` might be used by something that doesn't support getting the `ConstraintFunction. That might be because it isn't implemented, or it might be because the model uses a `ZerosBridge` (or similar). We should throw the proper error, instead of throwing whatever error actually threw, like "can't unbridge the Zeros bridge."

In the Cbc case, it's something like `CachingOptimizer(LazyBridgeOptimizer(CachingOptimizer(Cbc))`

The outer cache says "can you get ConstraintPrimal." The bridge says yes. Ends up erroring, falling back to `get_fallback`, and then errors when it tries to unbridge. 

If it thew `GetAttributeNotAllowed`, then the outer cache would fallback to `get_fallback` and this would succeed.

Even if it didn't have an outer cache to fall back do, we should throw the correct `GetAttributeNotAllowed` for consistency.

blegat: Isn't this also fixed by throwing a `GetAttributeNotAllowed` for the ZerosBridge ?

blegat: If we don't see the error of the zerosbridge then the user won't know he just has to remove this bridge, that's an issue

blegat: What we could do is throw a custom error here:
https://github.com/jump-dev/MathOptInterface.jl/blob/0249cfa8400dd550527b41db942e85cf0626d417/src/Bridges/Variable/map.jl#L582-L586
Here we try catch:
https://github.com/jump-dev/MathOptInterface.jl/blob/0249cfa8400dd550527b41db942e85cf0626d417/src/Bridges/bridge_optimizer.jl#L2384
If we see exactly this error, we thro a `GetAttributeNotAllowed` for `ConstraintFunction` with the error message of:
https://github.com/jump-dev/MathOptInterface.jl/blob/0249cfa8400dd550527b41db942e85cf0626d417/src/Bridges/Variable/map.jl#L582-L586
otherwise, we rethrow

odow: This would address this particular issue with ZerosBridge, but it ignores with wider issue. `get_fallback` may fail, and instead of throwing a nice error, we may throw some arbitrary error. But perhaps you're right, we can just disable ZerosBridge for now in Cbc.

blegat: The issue with this error is that it's difficult to debug since it provides no information on the initial error. Also, I'm not sure all the issues with `ZerosBridge` will go through `get_fallback` so I feel it's easier to fix at the source as I suggested in my previous comment

odow:
Closing in favor of #2415

PR #2414, opened 2024-01-28
[Utilities] use appropriate zero type for get_fallback

### LINKED ISSUE #2412: [Utilities] get_fallback assumes Float64
See various hard-coded `0.0`:

https://github.com/jump-dev/MathOptInterface.jl/blob/145e7bcc95181e89aa35e2dab98c1d938e635506/src/Utilities/results.jl#L367-L380

Found by https://github.com/jump-dev/MathOptInterface.jl/pull/2411

odow:
Closes #2412

### DIFF
--- a/src/Utilities/results.jl
+++ b/src/Utilities/results.jl
@@ -492,7 +492,7 @@ function get_fallback(
     model::MOI.ModelLike,
     attr::MOI.ConstraintDual,
     ci::MOI.ConstraintIndex{<:Union{MOI.VariableIndex,MOI.VectorOfVariables}},
-    ::Type{T} = Float64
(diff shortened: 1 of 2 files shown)


src/Utilities/results.jl
```diff
@@ -483,8 +492,9 @@ function get_fallback(
     model::MOI.ModelLike,
     attr::MOI.ConstraintDual,
     ci::MOI.ConstraintIndex{<:Union{MOI.VariableIndex,MOI.VectorOfVariables}},
-)
+    ::Type{T} = Float64
```
odow, line 495:
```suggestion
    ::Type{T} = Float64,
```

PR #2415, opened 2024-01-29
[Bridges] throw better error when variable cannot be unbridged

odow:
Replacement for https://github.com/jump-dev/MathOptInterface.jl/pull/2413

### DIFF
--- a/src/Bridges/Variable/map.jl
+++ b/src/Bridges/Variable/map.jl
@@ -573,28 +573,14 @@ function function_for(map::Map, ci::MOI.ConstraintIndex{MOI.VectorOfVariables})
 end
 
 """
-    struct CannotUnbridgeVariableFunction(msg::String) <: Exception
-
-The error thrown when Variable bridges do not support unmapping the function.
-
-This is usually because the forward mapping is lossy, for example, when the
-`ZerosBridge` converts `x in Zeros()` to `0.0`; given the constant `0.0`, we
-cannot recover the variable `x`.
-"""
-struct CannotUnbridgeVariableFunction <: Exception
-    msg::String
-
-    CannotUnbridgeVariableFunction(msg::String = "") = new(msg)
-end
-
-"""
     throw_if_cannot_unbridge(map::Map)
 
 Throw an error if some bridged variables do not have any reverse mapping.
 """
 function throw_if_cannot_unbridge(map::Map)
     if map.unbridged_function === nothing
-        err = CannotUnbridgeVariableFunction(
+        err = MOI.GetAttributeNotAllowed(
+            MOI.ConstraintFunction(),
             "Cannot unbridge function because some variables are bridged by " *
             "variable bridges that do not support reverse mapping, for " *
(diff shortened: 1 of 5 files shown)


test/Bridges/Variable/zeros.jl
```diff
@@ -84,14 +84,20 @@ function test_zeros()
...
     # With `c1`, the function does not contain any variable so it tests that it
-    # also throws an error even if it never calls `variable_unbridged_function`.
+    # also throws an error even if it never calls `variable_unbridged_function`
+    err = MOI.GetAttributeNotAllowed(
+        MOI.ConstraintFunction(),
+        "Cannot unbridge function because some variables are bridged by" *
+        " variable bridges that do not support reverse mapping, for example," *
+        " `ZerosBridge`.",
+    )
     @test_throws err MOI.get(bridged_mock, MOI.ConstraintFunction(), c1)
```
odow, line 101: @bleat here's an example. For `ConstraintFunction`, we throw `GetAttributeNotAllowed`. But for `ObjectiveFunction`, we throw `CannotUnbridgeVariableFunction`. What do you think should happen in each case?

odow: Bump @blegat

blegat: We could add a `try`-`catch` here as well: https://github.com/jump-dev/MathOptInterface.jl/blob/b64dbe150ce54282718dd8d70d7a52cdfed5965f/src/Bridges/bridge_optimizer.jl#L1128
and throw `GetAttributeNotAllowed`

src/Bridges/Variable/map.jl
```diff
@@ -572,18 +572,34 @@ function function_for(map::Map, ci::MOI.ConstraintIndex{MOI.VectorOfVariables})
...
+"""
+    struct CannotUnbridgeVariableFunction(msg::String) <: Exception
+
+The error thrown when Variable bridges do not support unmapping the function.
+
+This is usually because the forward mapping is lossy, for example, when the
+`ZerosBridge` converts `x in Zeros()` to `0.0`; given the constant `0.0`, we
+cannot recover the variable `x`.
+"""
+struct CannotUnbridgeVariableFunction <: Exception
```
blegat, line 584: Another approach would be to define an attribute `UnbridgedVariableFunction <: MOI.AbstractVariableAttribute` and to throw here `GetAttributeNotAllowed{UnbridgedVariableFunction}` so that we don't have to try catch and change it to a different attribute. It's not obeying the MethodError principle but I don't think that's the only place we have that ^^ Otherwise we'll have to `try`-`catch` every single line of code

odow: The alternative is just to throw `GetAttributeNotAllowed{MOI.ConstraintFunction}`, which is pretty much what we do below anyway.

odow:
Okay this is a lot simpler. I now just throw `GetAttributeNotAllowed`.

PR #2416, opened 2024-02-02
[Bridge] implement special case for x != y in CountDistinctToMILPBridge

### LINKED ISSUE #2405: Add support for NotEqualTo !=
These are perhaps not very high on anyone's priority list,
and models can avoid needing them, but they come up,
so i wonder if these would be welcomed? (if someone contributes them)?
1. Vector Set "monotonic": `@constraint(model, [i=2:length(x)], x[i] <= x[x-i])` (configurable predicate)
2. Bridge (`Int`) `x != 0` -> `c ? (x >= 1) : (x <= -1)`
3. Bridge (`Int`) `x != y` -> `z = x - y; z != 0`
4. Bridge `x != y` -> `(x, y) in MOI.AllDifferent`
5. Bridge `x >= y` -> `x > y  and  x != y` (configurable predicate)

odow:
Closes #2405

This PR implements a much stronger reformulation for the special case of `[2, x ,y] in CountDistinct` which is created by `[x, y] in AllDifferent(2)`, a.k.a. `x != y`.

### DIFF
--- a/src/Bridges/Constraint/bridges/count_distinct.jl
+++ b/src/Bridges/Constraint/bridges/count_distinct.jl
@@ -55,7 +55,7 @@ which is equivalent to (for suitable `M`):
 \\begin{aligned}
 z \\in \\{0, 1\\} \\\\
 x - y - M * z <= -1 \\\\
-y - x - M * (z - 1) <= -1
+y - x - M * (1 - z) <= -1
 \\end{aligned}
 ```
 
@@ -310,7 +310,7 @@ function _final_touch_not_equal_case(
     # <-->
     #   {x - y <= -1} \/ {y - x <= -1}
     # <-->
-    # {x - y - M * z <= -1} /\ {y - x - M * (z - 1) <= -1}, z in {0, 1}
+    # {x - y - M * z <= -1} /\ {y - x - M * (1 - z) <= -1}, z in {0, 1}
     z, _ = MOI.add_constrained_variable(model, MOI.ZeroOne())
     push!(bridge.variables, z)
     x, y = scalars[2], scalars[3]
@@ -327,15 +327,15 @@ function _final_touch_not_equal_case(
             allow_modify_function = true,
         ),
     )
-    # {y - x - M * (z - 1) <= -1}, M = u_x - l_y + 1
+    # {y - x - M * (1 - z) <= -1}, M = u_x - l_y + 1
     M = by[2] - bx[1] + 1
     g = MOI.Utilities.operate(-, T, y, x)
     push!(
         bridge.less_than,
         MOI.Utilities.normalize_and_add_constraint(
             model,
-            MOI.Utilities.operate!(-, T, g, M * z),
-            MOI.LessThan(T(-1 - M));
+            MOI.Utilities.operate!(+, T, g, M * z),
+            MOI.LessThan(T(-1 + M));
             allow_modify_function = true,
         ),
     )
--- a/src/Bridges/Constraint/bridges/count_distinct.jl
+++ b/src/Bridges/Constraint/bridges/count_distinct.jl
@@ -53,7 +53,7 @@ to `x != y`.
 which is equivalent to (for suitable `M`):
 ```math
 \\begin{aligned}
-z \\in \\{0, 1\\}
+z \\in \\{0, 1\\} \\\\
 x - y - M * z <= -1 \\\\
 y - x - M * (z - 1) <= -1
 \\end{aligned}
--- a/src/Bridges/Constraint/bridges/count_distinct.jl
+++ b/src/Bridges/Constraint/bridges/count_distinct.jl
@@ -41,6 +41,24 @@ non-zero:
 n - \\sum\\limits_{j \\in \\bigcup_{i=1,\\ldots,d} S_i} y_{j} = 0
 ```
 
+## Formulation (special case)
+
+In the special case that the constraint is `[2, x, y] in CountDistinct(3)`, then
+the constraint is equivalent to `[x, y] in AllDifferent(2)`, which is equivalent
+to `x != y`.
+
+```math
+(x - y <= -1) \\vee (y - x <= -1)
+```
+which is equivalent to (for suitable `M`):
+```math
+\\begin{aligned}
+z \\in \\{0, 1\\}
+x - y - M * z <= -1 \\\\
+y - x - M * (z - 1) <= -1
+\\end{aligned}
+```
+
 ## Source node
 
 `CountDistinctToMILPBridge` supports:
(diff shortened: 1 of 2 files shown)


src/Bridges/Constraint/bridges/count_distinct.jl
```diff
@@ -41,6 +41,24 @@ non-zero:
...
+the constraint is equivalent to `[x, y] in AllDifferent(2)`, which is equivalent
+to `x != y`.
+
+```math
+(x - y <= -1) \\vee (y - x <= -1)
+```
+which is equivalent to (for suitable `M`):
+```math
+\\begin{aligned}
+z \\in \\{0, 1\\}
```
odow, line 56:
```suggestion
z \\in \\{0, 1\\} \\\\
```

test/Bridges/Constraint/count_distinct.jl
```diff
@@ -146,6 +171,22 @@ function test_runtests_error_affine()
     return
 end
 
+function test_resolve_with_modified_not_equal_to()
```
LebedevRI, line 174: Could you please also add a test where there are no ranges known at all?

odow: See the existing `test_runtests_error_affine` above

odow:
I don't know why you can't see the coverage: https://app.codecov.io/github/JuMP-dev/MathOptInterface.jl/pull/2416

test/Bridges/Constraint/count_distinct.jl
LebedevRI: So this expansion still won't support that case? That's rather not useful,
and means i won't even be able to rely on this expansion doing the right thing.
That's rather unfortunate.

LebedevRI:
@odow FWIW, thank you for looking into this!
It will be unfortunate if this expansion will still not handle the bound-less case, but if that is what you think is best...

odow:
> but if that is what you think is best...

Can you provide a reformulation that would let us write this as a MILP (excluding indicator constraints) without knowing the bounds?

src/Bridges/Constraint/bridges/count_distinct.jl
```diff
@@ -232,9 +250,105 @@ function MOI.Bridges.final_touch(
...
+        bridge.less_than,
+        MOI.Utilities.normalize_and_add_constraint(
+            model,
+            MOI.Utilities.operate!(-, T, f, M * z),
+            MOI.LessThan(T(-1));
+            allow_modify_function = true,
+        ),
+    )
+    # {y - x - M * (z - 1) <= -1}, M = u_x - l_y
+    M = by[2] - bx[1] + 1
```
LebedevRI, line 331: Either the comment or the code is wrong

LebedevRI (commented):
Speaking of proofs, can you show the proof?
Are bounds inclusive or exclusive?
This doesn't quite proof for me: ~https://godbolt.org/z/3nEEdr1hj https://alive2.llvm.org/ce/z/tUeBwc~
https://godbolt.org/z/Mc87q4cjf https://alive2.llvm.org/ce/z/otmnsE

src/Bridges/Constraint/bridges/count_distinct.jl
odow: The comment :smile:

odow:
I don't understand the relevance of the links.

We're looking for a MILP reformulation of
```
| x | >= 1
```
where `x` is free

src/Bridges/Constraint/bridges/count_distinct.jl
```diff
@@ -232,9 +250,105 @@ function MOI.Bridges.final_touch(
...
+    #   {x - y >= 1} \/ {y - x >= 1}
+    # <-->
+    #   {x - y <= -1} \/ {y - x <= -1}
+    # <-->
+    # {x - y - M * z <= -1} /\ {y - x - M * (z - 1) <= -1}, z in {0, 1}
+    z, _ = MOI.add_constrained_variable(model, MOI.ZeroOne())
+    push!(bridge.variables, z)
+    x, y = scalars[2], scalars[3]
+    bx, by = bridge.bounds[1], bridge.bounds[2]
+    # {x - y - M * z <= -1}, M = u_x - l_y
```
odow, line 318:
```suggestion
    # {x - y - M * z <= -1}, M = u_x - l_y + 1
```

src/Bridges/Constraint/bridges/count_distinct.jl
```diff
@@ -232,9 +250,105 @@ function MOI.Bridges.final_touch(
...
+    push!(
+        bridge.less_than,
+        MOI.Utilities.normalize_and_add_constraint(
+            model,
+            MOI.Utilities.operate!(-, T, f, M * z),
+            MOI.LessThan(T(-1));
+            allow_modify_function = true,
+        ),
+    )
+    # {y - x - M * (z - 1) <= -1}, M = u_x - l_y
```
odow, line 330:
```suggestion
    # {y - x - M * (z - 1) <= -1}, M = u_x - l_y + 1
```

LebedevRI:
> I don't understand the relevance of the links.

Link 1 shows the true variant of the constraint (`src`), and naive re-implementation of this constraint in C (`tgt`), 
Link 2 feeds the implementations to an SMT solver, which says that they are not equivalent.

src/Bridges/Constraint/bridges/count_distinct.jl
LebedevRI: I was mainly talking about the fact that the comment says `max(x) - min(y)`,
but you actually do `max(y) - min(x)`, and comparison has different RHS.
It's confusing.

src/Bridges/Constraint/bridges/count_distinct.jl
```diff
@@ -41,6 +41,24 @@ non-zero:
...
+
+```math
+(x - y <= -1) \\vee (y - x <= -1)
+```
+which is equivalent to (for suitable `M`):
+```math
+\\begin{aligned}
+z \\in \\{0, 1\\} \\\\
+x - y - M * z <= -1 \\\\
+y - x - M * (z - 1) <= -1
```
odow, line 58:
```suggestion
y - x - M * (1 - z) <= -1
```

src/Bridges/Constraint/bridges/count_distinct.jl
```diff
@@ -232,9 +250,105 @@ function MOI.Bridges.final_touch(
...
+    end
+    # [2, x, y] in CountDistinct()
+    # <-->
+    #   x != y
+    # <-->
+    #   {x - y >= 1} \/ {y - x >= 1}
+    # <-->
+    #   {x - y <= -1} \/ {y - x <= -1}
+    # <-->
+    # {x - y - M * z <= -1} /\ {y - x - M * (z - 1) <= -1}, z in {0, 1}
```
odow, line 313:
```suggestion
    # {x - y - M * z <= -1} /\ {y - x - M * (1 - z) <= -1}, z in {0, 1}
```

odow:
Ugh. I can't read or type. I typed `z-1` instead of `1-z` at the top and then blindly followed through without thinking.

LebedevRI:
> > but if that is what you think is best...
> 
> Can you provide a reformulation that would let us write this as a MILP (excluding indicator constraints) without knowing the bounds?

Back when writing that comment, i didn't quite noticed that this was using yet another reformulation,
one without indicator constraints. I don't know if this one can be formulated without bounds.

It would be nice to either see either an exhaustive (correctness) test for this, or a non-manual proof.
But please feel free to ignore me on that! :)

odow:
> It would be nice to either see either an exhaustive (correctness) test for this, or a non-manual proof.

This hints at a much larger issue. We don't have any tools to check the correctness of JuMP models, and it is very easy (as just demonstrated) to make small mistakes. I'll open a JuMP issue.

odow:
See https://github.com/jump-dev/JuMP.jl/issues/3664

PR #2419, opened 2024-02-05
WIP: add FileFormats.Optimizer

odow:
x-ref https://github.com/jump-dev/MathOptInterface.jl/issues/2417#issuecomment-1926155817

PR #2418, opened 2024-02-04
[FileFormats.MPS] improve performance of MPS writer

odow:
Using the `Optimizer` from #2417 
```julia
using JuMP
function main(N)
    model = direct_model(Optimizer(MOI.FileFormats.MPS.Model()))
    @variable(model, 0 <= x[i in 1:N] <= 1; integer = isodd(i))
    @constraint(model, [i in 2:N], x[i] + x[i-1] <= 1)
    @objective(model, Max, sum(i * x[i] for i in 1:N))
    MOI.write_to_file(backend(model).inner, "file.mps")
    return
end
```

Before:

```julia
julia> GC.gc(); @time main(100_000);
  2.005077 seconds (18.45 M allocations: 812.105 MiB, 9.98% gc time, 24.17% compilation time: 44% of which was recompilation)

julia> GC.gc(); @time main(100_000);
  1.495755 seconds (17.55 M allocations: 751.348 MiB, 12.09% gc time)

julia> GC.gc(); @time main(100_000);
  1.529073 seconds (17.55 M allocations: 751.348 MiB, 12.28% gc time)
```

After:

```julia
julia> GC.gc(); @time main(100_000);
  1.873707 seconds (11.57 M allocations: 602.868 MiB, 8.18% gc time, 29.04% compilation time: 45% of which was recompilation)

julia> GC.gc(); @time main(100_000);
  1.341565 seconds (10.55 M allocations: 533.148 MiB, 10.52% gc time)

julia> GC.gc(); @time main(100_000);
  1.336950 seconds (10.55 M allocations: 533.148 MiB, 10.17% gc time)
```

Small improvement in time, but big improvement in the number of allocations.

PR #2421, opened 2024-02-06
[FileFormats.MPS] improve performance of MPS writer

odow:
Oops. 

Cherry-picked from #2418 because I'd actually rebased it onto #2419 instead of `master`.


odow:
Hmm. This broke BilevelJuMP `UndefVarError: `SET_TYPES` not defined`

PR #2423, opened 2024-02-08
Update solver-tests.yml with more packages

odow:
The list was not comprehensive. We were missing, e.g., https://github.com/jump-dev/AmplNLWriter.jl/pull/178

I've based this list mostly on https://github.com/jump-dev/JuMP.jl/blob/master/docs/packages.toml


odow:
https://github.com/jump-dev/MathOptInterface.jl/actions/runs/7823304952

PR #2422, opened 2024-02-07
WIP: explore MPS rewrite

odow:
I started experimenting with https://github.com/jump-dev/MathOptInterface.jl/issues/2420

It's a modest improvement:
```Julia
(base) oscar@Oscars-MBP /tmp % julia --project=jmp ~/Downloads/generate-flow-oscar.jl --output_file mcflow.mps.gz --num_commodities 1000 --num_warehouses 20 --num_stores 200 --seed 1
Parsed args:
  output_file  =>  mcflow.mps.gz
  num_warehouses  =>  20
  num_factories_per_commodity  =>  5
  seed  =>  1
  additional_overtime_cost  =>  0.3
  num_stores  =>  200
  num_commodities  =>  1000
building model ...
model built
 33.586420 seconds (129.25 M allocations: 9.650 GiB, 19.94% gc time, 7.69% compilation time)
```
compared to https://github.com/jump-dev/MathOptInterface.jl/pull/2418#issuecomment-1926237257
```
(base) oscar@Oscars-MBP /tmp % julia --project=opt ~/Downloads/generate-flow-oscar.jl --output_file mcflow.mps.gz --num_commodities 1000 --num_warehouses 20 --num_stores 200 --seed 1
Parsed args:
  output_file  =>  mcflow.mps.gz
  num_warehouses  =>  20
  num_factories_per_commodity  =>  5
  seed  =>  1
  additional_overtime_cost  =>  0.3
  num_stores  =>  200
  num_commodities  =>  1000
building model ...
model built
 41.327799 seconds (138.67 M allocations: 11.201 GiB, 15.64% gc time, 10.55% compilation time)
```
But it's a fairly large change, and I haven't added the quadratic, SOS, indicator support yet.

Another option is to use HiGHS instead:
```julia
import HiGHS
model = direct_model(HiGHS.Optimizer())
HiGHS.Highs_writeModel(JuMP.backend(model), parsed_args["output_file"])
```
resulting in
```
(base) oscar@Oscars-MBP /tmp % julia --project=jmp ~/Downloads/generate-flow-oscar.jl --output_file mcflow.mps.gz --num_commodities 1000 --num_warehouses 20 --num_stores 200 --seed 1
Parsed args:
  output_file  =>  mcflow.mps.gz
  num_warehouses  =>  20
  num_factories_per_commodity  =>  5
  seed  =>  1
  additional_overtime_cost  =>  0.3
  num_stores  =>  200
  num_commodities  =>  1000
building model ...
[...]

PR #2426, opened 2024-02-10
[FileFormats.MPS]: avoid creating list of variable names

mlubin:
The MPS writer has quadratic-ish performance because it creates GC-tracked strings for every variable and constraint. This PR shows one way we could skip this: generate the generic names on demand.

I didn't implement it for constraints because they need fancier code to deal with the constraint type parameters efficiently.

Potential alternatives:
- Use fixed-length strings as provided by WeakRefStrings
- Use WeakRefStrings.StringArray

Before:
```
148.861351 seconds (472.59 M allocations: 38.948 GiB, 40.91% gc time, 2.87% compilation time)
```
After:
```
109.157014 seconds (552.84 M allocations: 40.555 GiB, 21.95% gc time, 3.86% compilation time)
```
If we extrapolate, the version that does this approach for constraints would be around 70 sec, so we'd get a 2x speed up on this one example and better asymptotic performance.

### DIFF
--- a/src/FileFormats/MPS/MPS.jl
+++ b/src/FileFormats/MPS/MPS.jl
@@ -245,8 +245,7 @@ function Base.write(io::IO, model::Model)
         flip_obj = MOI.get(model, MOI.ObjectiveSense()) == MOI.MAX_SENSE
     end
     write_rows(io, model)
-    obj_const, indicators =
-        write_columns(io, model, flip_obj, var_to_column)
+    obj_const, indicators = write_columns(io, model, flip_obj, var_to_column)
     write_rhs(io, model, obj_const)
     write_ranges(io, model)
     write_bounds(io, model, var_to_column)
@@ -476,7 +475,12 @@ function _extract_terms_objective(model, var_to_column, coefficients, flip_obj)
     return obj_func.constant
 end
 
-function _var_name(model::Model, variable::MOI.VariableIndex, column::Int, generic_name::Bool)::String
+function _var_name(
+    model::Model,
+    variable::MOI.VariableIndex,
+    column::Int,
+    generic_name::Bool,
+)::String
     if generic_name
         return "C$column"
     else
@@ -484,12 +488,7 @@ function _var_name(model::Model, variable::MOI.VariableIndex, column::Int, gener
     end
 end
 
-function write_columns(
-    io::IO,
-    model::Model,
-    flip_obj,
-    var_to_column,
-)
+function write_columns(io::IO, model::Model, flip_obj, var_to_column)
     options = get_options(model)
     indicators = Tuple{String,String,MOI.ActivationCondition}[]
     coefficients = Vector{Tuple{String,Float64}}[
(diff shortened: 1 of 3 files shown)


odow:
I didn't think to test this. I just assumed that one list of strings would be better than needing to repeatedly query `VariableName`. Julia really doesn't like lots of `String`s...

mlubin:
Querying `VariableName` is a problem too, that still forces you to create one string per variables.

odow:
> that still forces you to create one string per variable.

Yeah, which explains the increase in allocations. But I guess they are short-lived. The GC just doesn't like long-lived GC objects.

mlubin:
To clarify, the issue isn't *creating* one string per variable, the issue is having O(num_variable) strings in existence simultaneously.

odow:
> the issue is having O(num_variable) strings in existence simultaneously.

Yeah. 

> I didn't implement it for constraints

Is there a similar thing to do for the constraints? Perhaps just changing the `generic_names`?

mlubin:
We need a type-stable function mapping a `ConstraintIndex{F,S}` to its generic name.

mlubin:
I hacked at this a bit more, but the constraint side is harder than I expected. The problem is the string identifiers used in the constraint matrix transpose code:
https://github.com/jump-dev/MathOptInterface.jl/blob/b64dbe150ce54282718dd8d70d7a52cdfed5965f/src/FileFormats/MPS/MPS.jl#L479-L481

I tried something like:
```
Dict{Tuple{Type,Type}, Vector{Vector{Tuple{Int64, Float64}}}}()
```
where we first breakdown by (F,S) then list the coefficients by column. This isn't great because requires a fair amount more memory usage and didn't seem to make things faster anyway.

We really just want to have all constraint types to share the same index space and just do a sparse matrix transpose to get the column-wise format, similar to the type of manipulations that solver interfaces do.

That's about as far as I can push for now.

odow:
The flat linear constraint space was what I started to do in https://github.com/jump-dev/MathOptInterface.jl/pull/2422. 

But now I know the big `Vector{String}`s are a bad idea, let me have another go. It's probably why I got only a modest improvement.

odow (commented):
Looks good. I made a small change. You should also try `JuMP.set_string_names_on_creation(model, false)`.

mlubin:
Is the plan to land this instead of https://github.com/jump-dev/MathOptInterface.jl/pull/2422?

odow:
> Is the plan to land this instead of https://github.com/jump-dev/MathOptInterface.jl/pull/2422?

I think yes. For now, this PR is an improvement. We should only move forward with #2422 if it is demonstrably better than this. #2422 is also missing a lot of things like support for quadratics, SOS, indicator, and reading.

mlubin:
Fair enough. Do we know what's going on with the nightly CI failure?

odow:
No, there have been a few failures. Also in MutableArithmetics: https://github.com/jump-dev/MutableArithmetics.jl/actions/runs/7892806791/job/21540090078?pr=257. I'll take a look.

odow (approved):

odow:
Your first commit to MOI in a while!

PR #2431, opened 2024-02-18
[FileFormats.MPS] fix writing duplicate bounds for binary variables

### LINKED ISSUE #2425: [FileFormats.MPS] Problem when reading generated MPS with CPLEX
When a binary variable is fixed, the generated MPS cannot be read with CPLEX due to `'FX' bound type illegal when prior bound given.`

```julia
using JuMP
m = Model()
@variable(m, foo, Bin)
fix(foo, 1)
write_to_file(m, "mwe.mps")
```

`mwe.mps` file content:
```
NAME          
ROWS
 N  OBJ
COLUMNS
    MARKER    'MARKER'                 'INTORG'
    foo       OBJ       0
RHS
RANGES
BOUNDS
 BV bounds    foo
 FX bounds    foo       1
ENDATA
```

Then in CPLEX:
```
CPLEX> read mwe.mps
Selected objective sense:  MINIMIZE
Selected objective  name:  OBJ
Selected bound      name:  bounds
CPLEX Error  1454: Line 11: 'FX' bound type illegal when prior bound given.
No file read.
```

Removing `BV bounds    foo` works. If the variable is fixed, the fact that it is a binary variable should not be stated anymore in the MPS file.

odow:
Closes #2425

### DIFF
--- a/test/FileFormats/MPS/MPS.jl
+++ b/test/FileFormats/MPS/MPS.jl
@@ -11,8 +11,35 @@ using Test
 import MathOptInterface as MOI
 import MathOptInterface.Utilities as MOIU
 import DataStructures: OrderedDict
+
 const MPS = MOI.FileFormats.MPS
+
 const MPS_TEST_FILE = "test.mps"
+
+function runtests()
+    for name in names(@__MODULE__, all = true)
+        if startswith("$(name)", "test_")
+            @testset "$name" begin
+                getfield(@__MODULE__, name)()
+            end
+        end
+    end
+    sleep(1.0)  # Allow time for unlink to happen.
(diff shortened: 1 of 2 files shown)


test/FileFormats/MPS/MPS.jl
```diff
@@ -11,9 +11,36 @@ using Test
...
+            @testset "$name" begin
+                getfield(@__MODULE__, name)()
+            end
+        end
+    end
+    sleep(1.0)  # Allow time for unlink to happen.
+    rm(MPS_TEST_FILE, force = true)
+    return
+end
+
```
odow, line 31:
```suggestion
```

PR #2444, opened 2024-02-25
[Nonlinear] add support for univariate sign

odow:
Closes https://github.com/jump-dev/MathOptInterface.jl/issues/2448

I don't know about this one. We might be better off documenting people to write a `signpower` function at the JuMP level because this still has the problem that `sign(x) * abs(x)^p` is not defined when `x=0` and `p < 1`.

Thoughts @ccoffrin?

### DIFF
--- a/test/Nonlinear/Nonlinear.jl
+++ b/test/Nonlinear/Nonlinear.jl
@@ -1163,6 +1163,32 @@ function test_automatic_differentiation_backend()
     return
 end
 
+function test_univariate_sign()
+    f(y, p) = sign(y) * abs(y)^p
+    ∇f(y, p) = p * abs(y)^(p-1)
+    ∇²f(y, p) = sign(y) * p * (p-1) * abs(y)^(p-2)
+    for p in (-0.5, 0.5, 2.0)
+        x = MOI.VariableIndex(1)
+        model = MOI.Nonlinear.Model()
+        MOI.Nonlinear.set_objective(model, :(sign($x) * abs($x)^$p))
+        evaluator = MOI.Nonlinear.Evaluator(
+            model,
+            MOI.Nonlinear.SparseReverseMode(),
+            [x],
+        )
+        MOI.initialize(evaluator, [:Grad, :Hess])
+        for y in (-10.0, -1.2, 1.2, 10.0)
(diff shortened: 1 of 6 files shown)


test/Nonlinear/Nonlinear.jl
```diff
@@ -1163,6 +1163,32 @@ function test_automatic_differentiation_backend()
     return
 end
 
+function test_univariate_sign()
+    f(y, p) = sign(y) * abs(y)^p
+    ∇f(y, p) = p * abs(y)^(p-1)
```
odow, line 1168:
```suggestion
    ∇f(y, p) = p * abs(y)^(p - 1)
```

test/Nonlinear/Nonlinear.jl
```diff
@@ -1163,6 +1163,32 @@ function test_automatic_differentiation_backend()
...
+    ∇²f(y, p) = sign(y) * p * (p-1) * abs(y)^(p-2)
```
odow, line 1169:
```suggestion
    ∇²f(y, p) = sign(y) * p * (p - 1) * abs(y)^(p - 2)
```

ccoffrin:
My hot take. 

There may be other uses of `sign` we cannot anticipate, which don't have such an obvious work around. 
If the following conditions are met,
1. Adding support for a new function does not add a major burden (e.g., in maintenance) 
2. It is a very common built-in Julia function (ideally also part of C standard functions)
3. there have been multiple request for support (i.e., users want it)

Then I think it is good to add support for it.

I hear your point that this will likely lead to a bunch of non-ideal modeling choices by non-experts. However, that issue is not limited to this specific new feature. I don't think there is an easy work around for helping typical users model better. Current state of the art is basically to get a PhD in NLP to understand how to write better models.

odow:
`sign` meets your criteria for 1, 2, and 3.

It's probably the currently unsupported operator that I see most often. 

And we already support `min`, `max`, `ifelse`, and `abs`, so non-differentiatbility isn't an argument against adding.

@mlubin should weigh in before we decide to merge.

blegat (approved):
I agree with @ccoffrin 's comment

mlubin (approved):

PR #2455, opened 2024-03-07
Add Xpress to solver-tests.yml

odow:
x-ref https://github.com/jump-dev/Xpress.jl/pull/221

https://github.com/jump-dev/MathOptInterface.jl/actions/runs/8181867191


odow:
Niiiice https://github.com/jump-dev/MathOptInterface.jl/actions/runs/8181867191/job/22372302246

PR #2464, opened 2024-04-04
Fix issue #2452

### LINKED ISSUE #2452: Bug get/setting ConstraintSet with variable bridges
x-ref https://github.com/oxfordcontrol/Clarabel.jl/issues/160

```julia
julia> using JuMP, Clarabel

julia> model = Model(Clarabel.Optimizer)
A JuMP Model
Feasibility problem with:
Variables: 0
Model mode: AUTOMATIC
Solver name: Clarabel

julia> set_silent(model)

julia> @variable(model, x >= 1)
x

julia> @constraint(model, c, 2x == 3)
c : 2 x = 3

julia> optimize!(model)

julia> value(x)
1.4999999999999996

julia> b = normalized_rhs(c)
3.0

julia> set_normalized_rhs(c, b)

julia> optimize!(model)

julia> value(x)
0.9999999995620605
```

```Julia
julia> import Clarabel

julia> src = MOI.Utilities.UniversalFallback(MOI.Utilities.Model{Float64}())
MOIU.UniversalFallback{MOIU.Model{Float64}}

julia> x = MOI.add_variable(src)
MOI.VariableIndex(1)

julia> MOI.add_constraint(src, x, MOI.GreaterThan(1.0))
MathOptInterface.ConstraintIndex{MathOptInterface.VariableIndex, MathOptInterface.GreaterThan{Float64}}(1)

julia> c = MOI.add_constraint(src, 2.0 * x, MOI.EqualTo(3.0))
MathOptInterface.ConstraintIndex{MathOptInterface.ScalarAffineFunction{Float64}, MathOptInterface.EqualTo{Float64}}(1)

julia> dest = MOI.instantiate(Clarabel.Optimizer; with_bridge_type = Float64)
MOIB.LazyBridgeOptimizer{MOIU.CachingOptimizer{Clarabel.MOIwrapper.Optimizer{Float64}, MOIU.UniversalFallback{MOIU.Model{Float64}}}}
with 0 variable bridges
with 0 constraint bridges
with 0 objective bridges
with inner model MOIU.CachingOptimizer{Clarabel.MOIwrapper.Optimizer{Float64}, MOIU.UniversalFallback{MOIU.Model{Float64}}}
  in state EMPTY_OPTIMIZER
  in mode AUTOMATIC
  with optimizer Empty Clarabel - Optimizer

julia> index_map = MOI.copy_to(dest, src)
MathOptInterface.Utilities.IndexMap with 3 entries:
  MOI.VariableIndex(1)                                           => MOI.VariableIndex(-1)
  ConstraintIndex{ScalarAffineFunction{Float64}, EqualTo{Float6… => ConstraintIndex{ScalarAffineFunction{Float64}, EqualTo{Float64}}(1)
  ConstraintIndex{VariableIndex, GreaterThan{Float64}}(1)        => ConstraintIndex{VariableIndex, GreaterThan{Float64}}(-1)

julia> MOI.set(dest, MOI.ConstraintSet(), index_map[c], MOI.EqualTo(3.0))

julia> MOI.optimize!(dest)
-------------------------------------------------------------
           Clarabel.jl v0.7.1  -  Clever Acronym              
                   (c) Paul Goulart                          
                University of Oxford, 2022                   
-------------------------------------------------------------

problem:
  variables     = 1
  constraints   = 2
  nnz(P)        = 0
  nnz(A)        = 2
  cones (total) = 2
    : Zero        = 1,  numel = 1
    : Nonnegative = 1,  numel = 1

settings:
  linear algebra: direct / qdldl, precision: Float64
  max iter = 200, time limit = Inf,  max step = 0.990
  tol_feas = 1.0e-08, tol_gap_abs = 1.0e-08, tol_gap_rel = 1.0e-08,
  static reg : on, ϵ1 = 1.0e-08, ϵ2 = 4.9e-32
  dynamic reg: on, ϵ = 1.0e-13, δ = 2.0e-07
  iter refine: on, reltol = 1.0e-13, abstol = 1.0e-12, 
               max iter = 10, stop ratio = 5.0
  equilibrate: on, min_scale = 1.0e-05, max_scale = 1.0e+05
               max iter = 10

iter    pcost        dcost       gap       pres      dres      k/t        μ       step      
---------------------------------------------------------------------------------------------
  0   0.0000e+00  -0.0000e+00  0.00e+00  4.46e-01  8.28e-01  1.00e+00  1.00e+00   ------   
  1   0.0000e+00   1.0946e+01  1.09e+01  2.85e-01  2.73e-02  1.14e+01  3.59e-02  9.66e-01  
  2   0.0000e+00   1.1412e+03  1.14e+03  2.85e-01  2.74e-04  1.14e+03  3.60e-04  9.90e-01  
  3   0.0000e+00   1.1417e+05  1.14e+05  2.85e-01  2.74e-06  1.14e+05  3.60e-06  9.90e-01  
  4   0.0000e+00   1.1417e+07  1.14e+07  2.85e-01  2.74e-08  1.14e+07  3.60e-08  9.90e-01  
  5   0.0000e+00   1.1417e+09  1.14e+09  2.85e-01  2.74e-10  1.14e+09  3.60e-10  9.90e-01  
---------------------------------------------------------------------------------------------
Terminated with status = primal infeasible
solve time = 2.41ms

julia> MOI.get(dest, MOI.VariablePrimal(), index_map[x])
0.9999999995620605
```

odow:
Fixes #2452 

<s>https://github.com/jump-dev/MathOptInterface.jl/actions/runs/8548954452</s>
<s>https://github.com/jump-dev/MathOptInterface.jl/actions/runs/8550055568</s>
<s>https://github.com/jump-dev/MathOptInterface.jl/actions/runs/8550389750</s>
- [x] https://github.com/jump-dev/MathOptInterface.jl/actions/runs/8592758175

### DIFF
--- a/src/Bridges/bridge_optimizer.jl
+++ b/src/Bridges/bridge_optimizer.jl
@@ -1479,8 +1479,8 @@ function MOI.get(
     # f is the "un-bridged" function in terms of the user-variables. We need to
     # substitute the variable bridges:
     g = bridged_function(b, f)
-    # g is now the "bridged" function. Add its constant to recover the user's
-    # original set (since the constant was shifted by -constant(g) when adding).
+    # `g` is now the "bridged" function. Add its constant to recover the user's
+    # original set (since the constant was shifted by `-constant(g)` when adding).
     return MOI.Utilities.shift_constant(set, MOI.constant(g))
 end
 
--- a/src/Bridges/bridge_optimizer.jl
+++ b/src/Bridges/bridge_optimizer.jl
@@ -1456,8 +1456,8 @@ end
 function MOI.get(
     b::AbstractBridgeOptimizer,
     attr::MOI.ConstraintSet,
-    ci::MOI.ConstraintIndex{<:MOI.AbstractScalarFunction},
-)
+    ci::MOI.ConstraintIndex{<:MOI.AbstractScalarFunction,S},
+) where {S}
     set = if is_bridged(b, ci)
         MOI.throw_if_not_valid(b, ci)
         call_in_context(MOI.get, b, ci, attr)
@@ -1466,7 +1466,11 @@ function MOI.get(
     end
     # This is a scalar function, so if there are variable bridges, it might
     # contain constants that have been moved into the set.
-    if !Variable.has_bridges(Variable.bridges(b))
+    if !MOI.Utilities.supports_shift_constant(S)
+        # If it doesn't support shift_constant, then return the set
+        return set
+    elseif !Variable.has_bridges(Variable.bridges(b))
+        # and the same if there are no variable bridges
         return set
     end
     # The function constant of the bridged function was moved to the set,
@@ -1478,6 +1482,7 @@ function MOI.get(
     end
     f = unbridged_function(b, func)
     g = bridged_function(b, f)
+    # But it's really the difference of the bridged and unbridged functions.
     offset = MOI.constant(g) - MOI.constant(f)
     return MOI.Utilities.shift_constant(set, offset)
 end
--- a/src/Bridges/bridge_optimizer.jl
+++ b/src/Bridges/bridge_optimizer.jl
@@ -1466,25 +1466,22 @@ function MOI.get(
     end
     # This is a scalar function, so if there are variable bridges, it might
     # contain constants that have been moved into the set.
-    if !MOI.Utilities.supports_shift_constant(S)
+    if !Variable.has_bridges(Variable.bridges(b))
+        # If there are no variable bridges, return the set.
+        return set
+    elseif !MOI.Utilities.supports_shift_constant(S)
         # If it doesn't support shift_constant, then return the set
         return set
-    elseif !Variable.has_bridges(Variable.bridges(b))
-        # and the same if there are no variable bridges
-        return set
     end
-    # The function constant of the bridged function was moved to the set,
-    # we need to remove it.
-    func = if is_bridged(b, ci)
-        call_in_context(MOI.get, b, ci, MOI.ConstraintFunction())
-    else
-        MOI.get(b.model, MOI.ConstraintFunction(), ci)
-    end
-    f = unbridged_function(b, func)
+    # The function constant of the bridged function (if it exists) was moved to
+    # the set, we need to remove it.
+    f = MOI.get(b, MOI.ConstraintFunction(), ci)
+    # f is the "un-bridged" function in terms of the user-variables. We need to
+    # substitute the variable bridges:
     g = bridged_function(b, f)
-    # But it's really the difference of the bridged and unbridged functions.
-    offset = MOI.constant(g) - MOI.constant(f)
-    return MOI.Utilities.shift_constant(set, offset)
+    # g is now the "bridged" function. Add its constant to recover the user's
+    # original set (since the constant was shifted by -constant(g) when adding).
+    return MOI.Utilities.shift_constant(set, MOI.constant(g))
 end
 
 ## Other constraint attributes
--- a/src/Bridges/bridge_optimizer.jl
+++ b/src/Bridges/bridge_optimizer.jl
@@ -1473,14 +1473,24 @@ function MOI.get(
         # If it doesn't support shift_constant, then return the set
         return set
     end
-    # The function constant of the bridged function (if it exists) was moved to
-    # the set, we need to remove it.
+    # When the constraint is added with function `f` and set `set_f`,
+    # the function is bridged into `g` with set `set` by
+    # `g, set = bridged_constraint_function(b, f, set_f)`.
+    # By doing so, the function constant of the bridged function (if it exists)
+    # was moved to `set`, we need to remove it to recover `set_f`.
+    # The function `f` contains the variables in the context of `ci`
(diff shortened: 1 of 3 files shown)


odow:
So this seemed to do the job, but I'm not sure I can tell you why.

blegat:
What's happening for example in the test you added ?
```
x >= 1 -> y = x - 1 with y nonneg
2x = 3 -> 2y = 1
func = 2y
f = 2x - 2
g = 2y
```
In this example, `MOI.constant(g)` is zero so it should be the same before and after the change you've made. What's the catch ?

odow:
Ah, but it is actually:
```
x >= 1 -> y = x - 1 with y nonneg
2x = 3 -> 2y = 1
func = 2x
f = 2x
g = 2y + 2
```
because when we try to get `func = 2y`, it calls:
https://github.com/jump-dev/MathOptInterface.jl/blob/606ba2ecae8d97eef79cee89a6d14351126b73b0/src/Bridges/Constraint/bridges/vectorize.jl#L229-L240
But `MOI.get(model, attr, bridge.vector_constraint)` returns the "unbridged" version of the function instead of the bridged. (Because it doesn't have a way to indicate that it will handle the variable bridges afterwards.)

The problematic method which does the unbridging is:
https://github.com/jump-dev/MathOptInterface.jl/blob/606ba2ecae8d97eef79cee89a6d14351126b73b0/src/Bridges/bridge_optimizer.jl#L1380-L1404

I tried adding some new `RawConstraintFunction` attribute, but then I got confused trying to plumb it all together.

odow:
Nightly failure is https://github.com/JuliaLang/julia/pull/53896#discussion_r1555087951

odow:
Okay. I think I understand this a lot better now. I've simplified the implementation and everything looks green.

src/Bridges/bridge_optimizer.jl
```diff
@@ -1467,17 +1467,21 @@ function MOI.get(
...
-    end
-    f = unbridged_function(b, func)
-    return MOI.Utilities.shift_constant(set, -MOI.constant(f))
+    # The function constant of the bridged function (if it exists) was moved to
+    # the set, we need to remove it.
+    f = MOI.get(b, MOI.ConstraintFunction(), ci)
+    # f is the "un-bridged" function in terms of the user-variables. We need to
+    # substitute the variable bridges:
+    g = bridged_function(b, f)
+    # g is now the "bridged" function. Add its constant to recover the user's
```
blegat, line 1482:
```suggestion
    # `g` is now the "bridged" function. Add its constant to recover the user's
```

src/Bridges/bridge_optimizer.jl
```diff
@@ -1467,17 +1467,21 @@ function MOI.get(
...
+    # original set (since the constant was shifted by -constant(g) when adding).
```
blegat, line 1483:
```suggestion
    # original set (since the constant was shifted by `-constant(g)` when adding).
```

blegat:
I think we still need the `if`-`else` with the `call_in_context`. Here is why.
You have the `user` model and the `solver` model. But you might also have models in between.
If a user variable `x` gets variable-bridged by a bridge `bxy` into an intermediate variable `y` which is then variable-bridged by a bridge `byz` into a solver variable `z` then you have an intermediate model `inter` for which the variable is `y`.
All constraints created by `bxy` are in the context of `inter` and the variable they use is `y`. So if a constraint created by `bxy` is bridged by a constraint bridge and this constraint bridge tries to get `MOI.ConstraintSet()`, it should get the `ConstraintFunction` in terms of `y`.
Maybe keeping the original way of querying `func` but then doing `g = bridged_function(b, func)` works ?

odow:
Can you construct a test that demonstrates the current behavior is wrong? Otherwise, I vote we keep this as-is. I don't really understand the ins and outs of call-in-context.

odow:
So perhaps if we kept the `call_in_context`, then we need the `MOI.constant(g) - MOI.constant(f)`, but just getting `::MOI.ConstraintFunction` deals with all the different constants.

I've added a test that the user can't add a function with a constant if there are variable bridges.

blegat (approved):
Thinking more about it, your implementation should also handle the tricky cases with nontrivial context, see the comments I have added. In the initial implementation, I thought getting the constraint function was going to give it in the solver variable but it gives it in the variables of the current context so your PR fixes it the best way I think

odow:
The implementation of what is going on at each level is certainly non-trivial.

klamike:
Thank you for your help on this issue!

Unfortunately, I think this PR does not fully fix #2452. It seems that using the latest master branch, the result of the [pure MOI MWE](https://github.com/jump-dev/MathOptInterface.jl/issues/2452#issuecomment-1980000495) is now correct **but the [original MWE](https://github.com/jump-dev/MathOptInterface.jl/issues/2452#issue-2170473491) still gives the same (incorrect) result.** Interestingly, using `remove_bridge` to remove `VectorizeBridge` still seems to fix the original MWE.

odow:
I've re-opened #2452. I'm sure I tested it at some point and it was fixed...

odow:
Okay. I've dug deeper, and now I have no confidence that this was actually the right fix. It just fixed the surface level `get`/`set`, but it incorrectly modified the inner model.

PR #2474, opened 2024-04-10
[docs] expand MatrixOfConstraints example

### LINKED ISSUE #1940: Add example to use MatrixOfConstraints in the documentation
the documentation of the `MatrixOfConstraints` section is API-oriented with each function presented separately.
It could help to add an example to build a `MatrixOfConstraints` from matrices and RHS à la `linprog` and another one to extract matrices and RHS from another model

odow:
Closes #1940 

JuMP now has https://jump.dev/JuMP.jl/stable/api/JuMP/#lp_matrix_data which is much more useful for people.

Anyone using `MatrixOfConstraints` is likely to just copy existing solver wrappers.

### DIFF
--- a/docs/src/submodules/Utilities/overview.md
+++ b/docs/src/submodules/Utilities/overview.md
@@ -379,34 +379,157 @@ julia> f
 ## Utilities.MatrixOfConstraints
 
 The constraints of [`Utilities.Model`](@ref) are stored as a vector of tuples
-of function and set in a `Utilities.VectorOfConstraints`. Other representations
-can be used by parameterizing the type [`Utilities.GenericModel`](@ref)
-(resp. [`Utilities.GenericOptimizer`](@ref)). For instance, if all
-non-`VariableIndex` constraints are affine, the coefficients of all the
-constraints can be stored in a single sparse matrix using
-[`Utilities.MatrixOfConstraints`](@ref). The constraints storage can even be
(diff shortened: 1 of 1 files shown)


docs/src/submodules/Utilities/overview.md
```diff
@@ -424,34 +547,16 @@ function _copy_to(dest::Optimizer, src::Model)
         row_bounds.upper,
     )
     # Set objective sense and constant (omitted)
-    return
-end
-
-function MOI.copy_to(dest::Optimizer, src::Model)
```
blegat, line 430: This is an important bit we need to keep. The whole point of this is that once `default_cache` returns an `OptimizerCache`, JuMP will use this cache so 99.99% of the time, `copy_to` will be called with an `OptimizerCache` as second argument. `copy_to` with an abstract `MOI.ModelLike` is only there for the rare use-case of some using Clp with MOI and not instantiating it with `MOI.instantiate` (then the cost of the copy is solely in building the index map, see https://github.com/jump-dev/MathOptInterface.jl/issues/1772).
If the only `copy_to` method implemented is the one taking `MOI.ModelLike`, it completely defeats the purpose.

odow: Ah I forgot about `default_cache`. Updated with comments.

PR #2477, opened 2024-04-11
[FileFormats.CBF] read functions as VectorOfVariables if possible

odow:
x-ref https://github.com/jump-dev/MathOptInterface.jl/issues/2457#issuecomment-2048784730

@blegat can decide whether this is an idea worth pursuing.

### DIFF
--- a/src/FileFormats/CBF/read.jl
+++ b/src/FileFormats/CBF/read.jl
@@ -303,6 +303,27 @@ function _read_DCOORD(io::IO, data::_CBFReadData)
 end
 
 """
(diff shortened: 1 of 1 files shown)


blegat:
I'm not a fan of the idea, it's quite orthogonal to the way the rest of JuMP works: no magic unless there is a MOI layer explicitly doing that. I can see myself showcasing difference between standard and geometric conic form and then not understanding why there is actually no difference having to dig in to understand that the CBF file changed it.

odow:
Fine by me

Issue #2457, opened 2024-03-22
[FileFormats.CBF] add support for PSDVAR


rbassett3:
I recently wrote a JuMP model which I exported to CBF and solved using a solver that doesn't have a JuMP interface. I was reviewing the file to make sense of the output and was surprised to note that the MathOptInterface completely ignores one part of the CBF specification by omitting PSDVAR. It's even written into the code but commented out. Is there a reason for this? It seems like it potentially could generate CBF files which are not solved as efficiently as they could be.

[Link to the relevant line of the CBF writer](https://github.com/jump-dev/MathOptInterface.jl/blob/d4c88596ce73ff60cca4c08e6bbd021ca25d3ad3/src/FileFormats/CBF/write.jl#L419)

odow:
Related: https://github.com/jump-dev/MathOptInterface.jl/issues/1402

I closed because no one was really asking for it haha.

It's do-able, but a little tricky to get right. Have you verified that it made a difference to the solver? Which solver is it? What conic form does it expect problems in? `Ax + b in K` or `x in K, Ax = b`?

rbassett3:
I am using SCIP-SDP, with SDP subproblems outsourced to Mosek. So it's `x in K, Ax = b`.

For the problem I'm working on, I've got to do a deep dive into the JuMP CBF interface. If you don't mind, I'd like to keep this issue open for a bit. I'll test to see if constraining a symmetric variable using PSDCON yields any difference in performance compared to using PSDVAR directly. 

On one hand, the formulations seem exactly equivalent, and it's hard for me to imagine a solver doing something special because PSDVAR was used instead of PSDCON. But then why does PSDVAR even exist as part of the CBF specification? Just a shorthand for declaring variables subject to a PSD constraint?

Anyways, I'll test and report back.

odow:
> But then why does PSDVAR even exist as part of the CBF specification

Mosek used to (and maybe still does) require that PSD matrices are "special" and declared up front as PSDVAR. If you added `Ax + b in PSD`, then we'd add a new matrix `S in PSD` with the linear constraints `Ax + b == S`.

odow:
I'm tempted to close this, but @blegat can decide. I think `Ax + b in K` is a reasonable thing to write. File formats don't have to be the most solver-efficient format.

odow:
[...]

PR #2478, opened 2024-04-12
[FileFormats.CBF] write VectorOfVariables constraints in VAR block

odow:
Part of #2457

### DIFF
--- a/src/FileFormats/CBF/write.jl
+++ b/src/FileFormats/CBF/write.jl
@@ -172,6 +172,7 @@ function _add_cones(
 }
     # The Exponential cone in MOI and CBF are reversed. Instead of dealing with
     # this complexity, just write them out as `Ax + b in K` constraints.
+    # TODO(odow): we should support this at some point. See #2478
     for ci in MOI.get(model, MOI.ListOfConstraintIndices{F,S}())
         f = MOI.get(model, MOI.ConstraintFunction(), ci)
         _add_function(data, f, S)
--- a/src/FileFormats/CBF/write.jl
+++ b/src/FileFormats/CBF/write.jl
@@ -19,6 +19,9 @@ mutable struct _CBFDataStructure
     bcoord::Vector{Tuple{Int,Float64}}
     hcoord::Vector{Tuple{Int,Int,Int,Int,Float64}}
     dcoord::Vector{Tuple{Int,Int,Int,Float64}}
+    variables_with_domain::Set{MOI.VariableIndex}
+    variable_cones::Vector{Tuple{Vector{MOI.VariableIndex},String}}
+
     function _CBFDataStructure()
         return new(
             0,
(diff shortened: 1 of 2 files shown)


src/FileFormats/CBF/write.jl
```diff
@@ -125,6 +130,56 @@ function _add_cones(
...
+function _add_cones(
+    data::_CBFDataStructure,
+    model::Model,
+    ::Type{F},
+    ::Type{S},
+) where {
+    F<:MOI.VectorOfVariables,
+    S<:Union{MOI.ExponentialCone,MOI.DualExponentialCone},
+}
+    # The Exponential cone in MOI and CBF are reversed. Instead of dealing with
```
blegat, line 173: I'd prefer if we say it's not supported for now. Is it the same permutation as https://github.com/jump-dev/ECOS.jl/blob/master/src/MOI_wrapper/permuted_exponential_cone.jl

blegat:
If the PSD cone is given, do we error ?

src/FileFormats/CBF/write.jl
blegat: It should be the same as the Mosek exponential cone I guess. If you want it to be supported, you can add the set and bridge in this module. Then we will be able to simplify MosekTools by using this set and adding the bridge in the `ListOfNonstandardBridges`

odow:
No change to PSD. There's still written as affine cones

src/FileFormats/CBF/write.jl
odow: It's not quite the same permutation as ECOS. ECOS is `1, 3, 2`, where this is `3, 2, 1`.

src/FileFormats/CBF/write.jl
```diff
@@ -125,6 +130,56 @@ function _add_cones(
...
+    # this complexity, just write them out as `Ax + b in K` constraints.
```
odow, line 174:
```suggestion
    # this complexity, just write them out as `Ax + b in K` constraints.
    # TODO(odow): we should support this at some point. See #2478
```

blegat:
> No change to PSD. There's still written as affine cones

Ah that's the current behavior. So it goes through the scalarize bridge ?

src/FileFormats/CBF/write.jl
blegat: Yes, that's what I remembered, Mosek and ECOS have a different one ^^

odow:
Before this PR, the model is written out as:
```
min c' x
Ax - b in K
x free
```
All `VectorOfVariables` constraints are accepted (no bridging) but are written as affine constraints.

After this PR, things are written as:
```
min c' x
Ax - b in K
x in some cones but not all
```
VectorOfVariables in Exponential, DualExponential, and PositiveSemidefiniteConeTriangle are still supported, but they are written as affine constraints.

odow:
Merging as an improvement for now. I'll look at doing EXP and PSD variables in a separate PR.

PR #2482, opened 2024-04-15
[FileFormats.CBF] write out EXP and EXP* as variable cones

odow:
Follow-up to #2478

Part of #2457


src/FileFormats/CBF/write.jl
```diff
@@ -21,6 +21,7 @@ mutable struct _CBFDataStructure
     dcoord::Vector{Tuple{Int,Int,Int,Float64}}
     variables_with_domain::Set{MOI.VariableIndex}
     variable_cones::Vector{Tuple{Vector{MOI.VariableIndex},String}}
+    scalar_variables::Vector{Int}
```
blegat, line 24: This is the identity except for exponential cones right ? Maybe a comment saying that would be helpful for readability

blegat (approved):

PR #2497, opened 2024-05-09
Add any function for Bridges.runtests

blegat:
In some cases, it's difficult to build the model with `readfromstrings!`, e.g., in SumOfSquares I need to create polynomial variables etc...


odow (approved):
I'm trying to enforce the convention that `::Function` arguments are named `_fn`.

blegat:
I like conventions

PR #2498, opened 2024-05-10
Add Variable bridges to use runtests without unbridged_variable

blegat:
Some variable bridges don't support `unbridged_variable` and that makes `runtests` fail when getting `ConstraintFunction` or when getting `ConstraintSet` for scalar constraints. For example, `ZerosBridge` does not support it so it cannot use `runtests`.
Another example is the SOS cone: https://github.com/jump-dev/SumOfSquares.jl/pull/353 I could support it in the nonweighted version but if it's weighted by arbitrary polynomials, it seems difficult to unbridge. It might be possible but it's not high priority. By using only vector cones in the tests, it turns out that only this small change is enough to pass the tests. This is still testing the inner model completely.

- [x] Explain in docstring why this is used for
- [x] Use it for `ZerosBridge` by adding a tests without scalar sets

### DIFF
--- a/src/Bridges/Bridges.jl
+++ b/src/Bridges/Bridges.jl
@@ -147,7 +147,7 @@ include("precompile.jl")
 function _test_structural_identical(
     a::MOI.ModelLike,
     b::MOI.ModelLike;
-    allow_constraint_function_error::Bool = false,
+    cannot_unbridge::Bool = false,
 )
     # Test that the variables are the same. We make the strong assumption that
     # the variables are added in the same order to both models.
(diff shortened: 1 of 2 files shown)


src/Bridges/Bridges.jl
```diff
@@ -267,7 +281,11 @@ function runtests(
     # Load a non-bridged input model, and check that getters are the same.
     test = MOI.Utilities.UniversalFallback(MOI.Utilities.Model{eltype}())
     input_fn(test)
-    _test_structural_identical(test, model)
+    _test_structural_identical(
+        test,
+        model;
+        allow_constraint_function_error = allow_outer_constraint_function_error,
```
odow, line 287:
```suggestion
        allow_constraint_function_error,
```
This should be sufficient instead of the long line?

blegat: Note the `outer_`. I only want to allow it when comparing `test` with `model`, not with `inner`

blegat: I'm open to better names ^^

odow: I, in fact, did not notice the `_outer_` 😆

odow:
Did you try `ZerosBridge`? It still failed in a bunch of places

blegat:
I did not try, I was just hoping it could serve as a test

odow:
The issue is that a bunch of other places may call `ConstraintFunction`, so it is not so simple

odow:
I don't think this is the right approach. Just write some other tests. You don't need general infrastructure.

blegat:
What about now ?

PR #2499, opened 2024-05-10
Fix Bridges.runtests when model has no variable

blegat:
Needed for https://github.com/jump-dev/SumOfSquares.jl/pull/296

- [x] Needs tests

Without the fix of this PR, the added tests throws
```julia
test_runtests: Error During Test at /home/blegat/.julia/dev/MathOptInterface/src/Bridges/Bridges.jl:286
  Test threw exception
  Expression: MOI.get(model, attr, x) ≈ primal_start
  MethodError: no method matching zero(::Type{Any})
  
  Closest candidates are:
    zero(::Type{Union{Missing, T}}) where T
     @ Base missing.jl:105
    zero(::Type{Union{}}, Any...)
     @ Base number.jl:310
    zero(::Type{Missing})
     @ Base missing.jl:104
    ...
  
  Stacktrace:
   [1] zero(::Type{Any})
     @ Base ./missing.jl:106
   [2] norm(itr::Vector{Any}, p::Int64)
     @ LinearAlgebra ~/.julia/juliaup/julia-1.10.3+0.x64.linux.gnu/share/julia/stdlib/v1.10/LinearAlgebra/src/generic.jl:596
   [3] isapprox(x::Vector{Any}, y::Vector{Float64}; atol::Int64, rtol::Float64, nans::Bool, norm::typeof(LinearAlgebra.norm))
     @ LinearAlgebra ~/.julia/juliaup/julia-1.10.3+0.x64.linux.gnu/share/julia/stdlib/v1.10/LinearAlgebra/src/generic.jl:1789
   [4] isapprox(x::Vector{Any}, y::Vector{Float64})
     @ LinearAlgebra ~/.julia/juliaup/julia-1.10.3+0.x64.linux.gnu/share/julia/stdlib/v1.10/LinearAlgebra/src/generic.jl:1785
   [5] eval_test(evaluated::Expr, quoted::Expr, source::LineNumberNode, negate::Bool)
     @ Test ~/.julia/juliaup/julia-1.10.3+0.x64.linux.gnu/share/julia/stdlib/v1.10/Test/src/Test.jl:355
   [6] macro expansion
     @ ~/.julia/juliaup/julia-1.10.3+0.x64.linux.gnu/share/julia/stdlib/v1.10/Test/src/Test.jl:669 [inlined]
   [7] runtests(Bridge::Type{MathOptInterface.Bridges.Constraint.GreaterToLessBridge}, input_fn::Main.TestConstraintFlipSign.var"#59#61", output_fn::Main.TestConstraintFlipSign.var"#60#62"; variable_start::Float64, constraint_start::Float64, eltype::Type, print_inner_model::Bool)
     @ MathOptInterface.Bridges ~/.julia/dev/MathOptInterface/src/Bridges/Bridges.jl:286
```

### DIFF
--- a/src/Bridges/Bridges.jl
+++ b/src/Bridges/Bridges.jl
@@ -283,7 +283,10 @@ function runtests(
         Test.@test all(isnothing, MOI.get(model, attr, x))
         primal_start = fill(constraint_start, length(x))
         MOI.set(model, attr, x, primal_start)
-        Test.@test MOI.get(model, attr, x) ≈ primal_start
+        if !isempty(x)
+            # ≈ does not work if x is empty because the return of get is Any[]
+            Test.@test MOI.get(model, attr, x) ≈ primal_start
+        end
     end
     # Test ConstraintPrimalStart and ConstraintDualStart
     for (F, S) in MOI.get(model, MOI.ListOfConstraintTypesPresent())
@@ -295,7 +298,11 @@ function runtests(
                     Test.@test MOI.get(model, attr, ci) === nothing
                     start = _fake_start(constraint_start, set)
                     MOI.set(model, attr, ci, start)
-                    Test.@test MOI.get(model, attr, ci) ≈ start
+                    if !isempty(ci)
+                        # ≈ does not work if ci is empty because the return of
+                        # get is Any[]
+                        Test.@test MOI.get(model, attr, ci) ≈ start
+                    end
                 end
             end
         end
--- a/src/Bridges/Bridges.jl
+++ b/src/Bridges/Bridges.jl
@@ -277,8 +277,9 @@ function runtests(
     bridge_supported = all(values(Variable.bridges(model))) do bridge
         return MOI.supports(model, attr, typeof(bridge))
     end
-    if MOI.supports(model, attr, MOI.VariableIndex) && bridge_supported &&
-        MOI.get(model, MOI.NumberOfVariables()) > 0
+    if MOI.supports(model, attr, MOI.VariableIndex) &&
+       bridge_supported &&
+       MOI.get(model, MOI.NumberOfVariables()) > 0
         x = MOI.get(model, MOI.ListOfVariableIndices())
         MOI.set(model, attr, x, fill(nothing, length(x)))
         Test.@test all(isnothing, MOI.get(model, attr, x))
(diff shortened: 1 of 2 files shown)


src/Bridges/Bridges.jl
```diff
@@ -277,7 +277,9 @@ function runtests(
     bridge_supported = all(values(Variable.bridges(model))) do bridge
         return MOI.supports(model, attr, typeof(bridge))
     end
-    if MOI.supports(model, attr, MOI.VariableIndex) && bridge_supported
+    if MOI.supports(model, attr, MOI.VariableIndex) &&
+       bridge_supported &&
+       MOI.get(model, MOI.NumberOfVariables()) > 0
```
odow, line 282: What is the error without this? Shouldn't it all work? The `ListOfVariableIndices` will be `MOI.VariableIndex[]`.

odow: And what is the point of `runtests` if the problem is empty?

blegat: Good questions :) See https://github.com/jump-dev/SumOfSquares.jl/pull/296/files#diff-49972eb0065227814878ef54ce01433e49bf09e18e6af0ea63d80d000e50bbb4R25-R38, we want to check whether a polynomial is a sum of squares. The vector of coefficients of the polynomial is a `VectorAffineFunction` with no affine term. Because the membership to this set is not so simple, solving such a problem has a point :)

blegat: See https://github.com/jump-dev/SumOfSquares.jl/actions/runs/9041630162/job/24847221171?pr=296

blegat: I added the error in the PR description

odow: Okaaaay
```julia

julia> m = MOI.Utilities.Model{Float64}()
MOIU.Model{Float64}

julia> x = MOI.VariableIndex[]
MathOptInterface.VariableIndex[]

julia> MOI.get(m, MOI.VariablePrimalStart(), x)
Union{}[]
```
I think we could fix this somehow.

odow: https://github.com/jump-dev/MathOptInterface.jl/issues/2500

blegat:
I rebased to check, https://github.com/jump-dev/MathOptInterface.jl/pull/2501does not fix the error

odow:
Correct. That PR was to fix the `Union{}[]` return type, but it doesn't fix `norm(::Vector{Any})`. I'll take a look at this.

PR #2502, opened 2024-05-12
[docs] fix links for the link checker

odow:
Current builds of the documentation are failing because the link checker is timing out.


odow:
Seems like Gurobi is updating their documentation to a new system, which is causing time outs for the link checker. I'll just remove the link.

PR #2509, opened 2024-05-28
Allow SetMapBridge to use bridge value

blegat:
When I decided that these functions would only depend on the type, it was for consistency for `map_function` because `map_function` clearly cannot depend on anything else given that it is used in `bridge_constraint`. However, you could also have a bridge that implements its own `bridge_constraint`, storing extra information then in the bridge to be able to implement `inverse_map_function`, etc... This is the case of the Sum-of-Squares bridge.

Needed for https://github.com/jump-dev/SumOfSquares.jl/pull/355

### DIFF
--- a/src/Bridges/Constraint/set_map.jl
+++ b/src/Bridges/Constraint/set_map.jl
@@ -100,12 +100,12 @@ end
 # function. The user doesn't need to know this, only that they cannot get the
 # attribute. Throwing `GetAttributeNotAllowed` allows `CachingOptimizer` to fall
 # back to using the cache.
-function _not_invertible_error(attr, message)
-    s = "Cannot get $attr as the constraint is reformulated through a linear transformation that is not invertible."
+function _not_invertible_error_message(attr, message)
+    s = "Cannot get `$attr` as the constraint is reformulated through a linear transformation that is not invertible."
     if !isempty(message)
         s *= " " * message
     end
-    return MOI.GetAttributeNotAllowed(attr, s)
+    return s
 end
 
 function MOI.get(
@@ -118,7 +118,7 @@ function MOI.get(
         MOI.Bridges.inverse_map_function(bridge, mapped_func)
     catch err
         if err isa MOI.Bridges.MapNotInvertible
-            throw(_not_invertible_error(attr, err.message))
+            throw(MOI.GetAttributeNotAllowed(attr, _not_invertible_error_message(attr, err.message)))
         end
         rethrow(err)
     end
@@ -172,7 +172,7 @@ function MOI.get(
         # the function. The user doesn't need to know this, only that they
         # cannot get the attribute.
         if err isa MOI.Bridges.MapNotInvertible
-            throw(MOI.GetAttributeNotAllowed(attr))
+            throw(MOI.GetAttributeNotAllowed(attr, _not_invertible_error_message(attr, err.message)))
         end
         rethrow(err)
     end
@@ -214,7 +214,14 @@ function MOI.set(
     if value === nothing
         MOI.set(model, attr, bridge.constraint, nothing)
     else
-        mapped_value = MOI.Bridges.inverse_adjoint_map_function(bridge, value)
+        mapped_value = try
+            MOI.Bridges.inverse_adjoint_map_function(bridge, value)
+        catch err
+            if err isa MOI.Bridges.MapNotInvertible
+                throw(MOI.SetAttributeNotAllowed(attr, _not_invertible_error_message(attr, err.message)))
+            end
+            rethrow(err)
+        end
         MOI.set(model, attr, bridge.constraint, mapped_value)
     end
     return
--- a/src/Bridges/Constraint/set_map.jl
+++ b/src/Bridges/Constraint/set_map.jl
@@ -102,7 +102,7 @@ function MOI.get(
     bridge::MultiSetMapBridge{T,S1,G},
 ) where {T,S1,G}
     mapped_func = MOI.get(model, attr, bridge.constraint)
-    func = MOI.Bridges.inverse_map_function(typeof(bridge), mapped_func)
+    func = MOI.Bridges.inverse_map_function(bridge, mapped_func)
     return MOI.Utilities.convert_approx(G, func)
 end
 
@@ -146,7 +146,7 @@ function MOI.get(
     if value === nothing
         return nothing
     end
-    return MOI.Bridges.inverse_map_function(typeof(bridge), value)
+    return MOI.Bridges.inverse_map_function(bridge, value)
 end
 
 function MOI.set(
@@ -173,7 +173,7 @@ function MOI.get(
     if value === nothing
         return nothing
     end
-    return MOI.Bridges.adjoint_map_function(typeof(bridge), value)
+    return MOI.Bridges.adjoint_map_function(bridge, value)
 end
 
 function MOI.set(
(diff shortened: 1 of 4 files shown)


src/Bridges/set_map.jl
```diff
@@ -81,3 +95,7 @@ Return the image of `func` through the inverse of the adjoint of the linear map
 [`MOI.ConstraintDualStart`](@ref) of constraint bridges.
 """
 function inverse_adjoint_map_function end
+
+function inverse_adjoint_map_function(bridge::AbstractBridge, func)
```
odow, line 99: These need the docstring to be updated

src/Bridges/set_map.jl
```diff
@@ -4,6 +4,8 @@
 # Use of this source code is governed by an MIT-style license that can be found
 # in the LICENSE.md file or at https://opensource.org/licenses/MIT.
 
+struct MapNotInvertible <: Exception end
```
odow, line 7: Needs a docstring and explanation of when to throw

src/Bridges/Constraint/set_map.jl
```diff
@@ -146,7 +154,16 @@ function MOI.get(
     if value === nothing
         return nothing
     end
-    return MOI.Bridges.inverse_map_function(typeof(bridge), value)
+    return try
+        MOI.Bridges.inverse_map_function(bridge, value)
+    catch err
+        if err isa MOI.Bridges.MapNotInvertible
+            throw(MOI.GetAttributeNotAllowed(attr))
```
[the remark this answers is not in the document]
blegat: This is thrown in `inverse_map_function` which does not know what the attribute is

src/Bridges/set_map.jl
```diff
@@ -42,42 +61,65 @@ would be returned by `map_function(BT, func)` except that it may compute the
 the [`MOI.VariablePrimal`](@ref) and
 [`MOI.VariablePrimalStart`](@ref) of variable bridges.
 """
-function map_function end
-
 function map_function(::Type{BT}, func, i::IndexInVector) where {BT}
```
odow, line 64: Does this method need the bridge object instead of the type as well?

src/Bridges/Constraint/set_map.jl
```diff
@@ -102,7 +102,17 @@ function MOI.get(
...
     mapped_func = MOI.get(model, attr, bridge.constraint)
-    func = MOI.Bridges.inverse_map_function(typeof(bridge), mapped_func)
+    func = try
+        MOI.Bridges.inverse_map_function(bridge, mapped_func)
+    catch err
+        # MapNotInvertible is thrown if the bridge does not support inverting
+        # the function. The user doesn't need to know this, only that they
+        # cannot get the attribute.
+        if err isa MOI.Bridges.MapNotInvertible
+            throw(MOI.GetAttributeNotAllowed(attr))
```
odow, line 112: How do we test this?

blegat:
For `map_function`, there are two cases:
1. The bridge can implement `bridge_constraint` and `map_function(::AbstractBridge, ...)` in which case in case the map depends on values stored in the bridge. `map_function` is used for setting the `ConstraintPrimalStart` so actually is not a huge priority to implement it
2. The bridge implement `map_function(::Type, ...)` in which case the default fallback for `bridge_constraint` is enough.
Since the second cases is useful as well, we shouldn't say `map_function(::Type, ...)` is legacy

src/Bridges/Constraint/set_map.jl
blegat: We'll have to create a custom bridge for this in the tests I think

src/Bridges/Constraint/set_map.jl
```diff
@@ -96,13 +96,33 @@ end
 
 # Attributes, Bridge acting as a constraint
 
+# MapNotInvertible is thrown if the bridge does not support inverting
+# the function. The user doesn't need to know this, only that they
+# cannot get the attribute. Throwing `GetAttributeNotAllowed` allows
+# `CachingOptimizer` to fall back to using the cache
+function _not_invertible_error(attr, message)
```
odow, line 103:
```suggestion
function _throw_not_invertible_error(attr, message)
```

For clarity

odow: Or return the error and then throw from the catch

test/Bridges/set_map.jl
```diff
@@ -0,0 +1,148 @@
...
+#
+# Use of this source code is governed by an MIT-style license that can be found
+# in the LICENSE.md file or at https://opensource.org/licenses/MIT.
+
+# Test with a bridge for which the map is defined on the bridge
+# value and not the bridge type
+module TestSetMapBridge
+
+import MathOptInterface as MOI
+
```
odow, line 12:
```suggestion
```

test/Bridges/set_map.jl
```diff
@@ -0,0 +1,148 @@
...
+        MOI.VectorOfVariables(swap(func.variables, set.swap)),
+        MOI.Nonnegatives(2),
+    )
+    return SwapBridge{T}(ci, set)
+end
+
+function MOI.Bridges.map_set(bridge::SwapBridge, set::SwapSet)
+    if set.swap != bridge.set.swap
+        error("Cannot change swap set")
+    else
```
odow, line 54:
```suggestion
    end
```

test/Bridges/set_map.jl
```diff
@@ -0,0 +1,148 @@
...
+        return MOI.Nonnegatives(2)
```
odow, line 55:
```suggestion
    return MOI.Nonnegatives(2)
```

test/Bridges/set_map.jl
```diff
@@ -0,0 +1,148 @@
...
+    end
```
odow, line 56:
```suggestion
```

test/Bridges/set_map.jl
```diff
@@ -0,0 +1,148 @@
...
+    set = SwapSet(true, false)
+    ci = MOI.add_constraint(model, func, set)
+    err = ErrorException("Cannot change swap set")
+    @test_throws err MOI.set(model, MOI.ConstraintSet(), ci, SwapSet(false, false))
+    attr = MOI.ConstraintFunction()
+    err = MOI.GetAttributeNotAllowed(
+        attr,
+        "Cannot get MathOptInterface.ConstraintFunction() as the constraint is reformulated through a linear transformation that is not invertible.no luck no luck",
+    )
+    @test_throws err MOI.get(model, attr, ci)
```
odow, line 122:
```suggestion
    @test_throws err MOI.get(model, attr, ci)
    return
```

src/Bridges/set_map.jl
```diff
@@ -5,35 +5,74 @@
...
-the [`MOI.ConstraintSet`](@ref).
+[`Variable.SetMapBridge`](@ref) and [`Constraint.SetMapBridge`](@ref).
+
+This function is used for bridging the constraint and setting the
+[`MOI.ConstraintSet`](@ref).
+
+The default implementation of [`Variable.bridge_constraint`](@ref) uses
+[`map_set`](@ref) with the bridge type so if this function is defined
+on the bridge type, [`Variable.bridge_constraint`](@ref) does not need
+to be implemented.
```
odow, line 34: I don't understand this comment. `Variable.bridge_constraint` does not exist.

odow: Do you mean
https://github.com/jump-dev/MathOptInterface.jl/blob/193aec66d562c37af5e63f842652dc44a31bce9a/src/Bridges/Constraint/set_map.jl#L37-L45

odow: I just removed the comment

Issue #2513, opened 2024-06-13
Revisit PR#2508


odow:
See https://github.com/jump-dev/MathOptInterface.jl/actions/runs/9491359380/job/26156620575

Likely caused by https://github.com/jump-dev/MathOptInterface.jl/pull/2508

Found working on https://github.com/jump-dev/MathOptInterface.jl/pull/2495

odow:
#2508 is wrong, because they are not the same duals:

```Julia
import CSDP
import MathOptInterface as MOI
function main()
    model = MOI.instantiate(CSDP.Optimizer, with_bridge_type = Float64)
    MOI.Bridges.remove_bridge(model, MOI.Bridges.Variable.ZerosBridge{Float64})
    config = MOI.Test.Config(; rtol = 1e-3, atol = 1e-3)
    MOI.Test.test_conic_PositiveSemidefiniteConeTriangle_VectorOfVariables(model, config)
    return model
end
model = main()
```

```julia
julia> model = main()
CSDP 6.2.0
Iter:  0 Ap: 0.00e+00 Pobj:  0.0000000e+00 Ad: 0.00e+00 Dobj:  0.0000000e+00 
Iter:  1 Ap: 1.00e+00 Pobj: -1.3091666e+01 Ad: 9.00e-01 Dobj: -1.4400002e+00 
Iter:  2 Ap: 1.00e+00 Pobj: -6.5001984e+00 Ad: 9.15e-01 Dobj: -8.2898589e-01 
Iter:  3 Ap: 9.50e-01 Pobj: -2.1617676e+00 Ad: 8.66e-01 Dobj: -1.4438101e+00 
Iter:  4 Ap: 1.00e+00 Pobj: -2.0165778e+00 Ad: 9.38e-01 Dobj: -1.9577837e+00 
Iter:  5 Ap: 1.00e+00 Pobj: -2.0005894e+00 Ad: 9.66e-01 Dobj: -1.9980351e+00 
Iter:  6 Ap: 1.00e+00 Pobj: -2.0000291e+00 Ad: 9.82e-01 Dobj: -1.9999373e+00 
Iter:  7 Ap: 1.00e+00 Pobj: -2.0000041e+00 Ad: 1.00e+00 Dobj: -1.9999974e+00 
Iter:  8 Ap: 1.00e+00 Pobj: -2.0000009e+00 Ad: 1.00e+00 Dobj: -2.0000001e+00 
[...]

PR #2514, opened 2024-06-13
Revert "[Bridges] use dual of equality in slack bridge (#2508)"

odow:
This reverts commit 6d9378acf11653a1cb2ac6c969f388fa4be59f1e.

x-ref #2513 

- [x] https://github.com/jump-dev/MathOptInterface.jl/actions/workflows/solver-tests.yml


odow:
@blegat I didn't really follow the details. It looks like there is a factor of `2` slipping in, likely because the PSD matrices have symmetry?

PR #2515, opened 2024-06-14
[Bridges] use dual of equality in slack bridge (take 2)

blegat:
We can take inspiration from https://github.com/jump-dev/MathOptInterface.jl/blob/08d34003cd3d7500379d9440e92445bc4d9a8fa8/src/Bridges/Constraint/bridges/set_dot_scaling.jl#L64-L75

Closes https://github.com/jump-dev/MathOptInterface.jl/issues/2513

### DIFF
--- a/src/Bridges/Constraint/bridges/slack.jl
+++ b/src/Bridges/Constraint/bridges/slack.jl
@@ -128,17 +128,22 @@ function MOI.get(
 ) where {T}
     # The dual constraint on slack (since it is free) is
     # `-dual_slack_in_set + dual_equality = 0` so the two duals are
-    # equal and we can return either one of them.
-    # We decide to use the dual of the equality constraints because
-    # the `slack_in_set` constraint might be bridge by a variable bridge that
-    # does not # support `ConstraintDual`. This is the case if the adjoint of
-    # the linear map on which the bridge is based is not invertible, as for
-    # instance with `Variable.ZerosBridge` or
-    # `SumOfSquares.Bridges.Variable.KernelBridge`.
+    # equal (modulo a rescaling for things like symmetric matrices) and we can
+    # return either one of them.
+    #
+    # We decide to use the dual of the equality constraints because the
+    # `slack_in_set` constraint might be bridged by a variable bridge that does
+    # not support `ConstraintDual`. This is the case if the adjoint of the
+    # linear map on which the bridge is based is not invertible, for example,
+    # `Variable.ZerosBridge` or `SumOfSquares.Bridges.Variable.KernelBridge`.
     dual = MOI.get(model, a, bridge.equality)
     if dual === nothing
         return nothing
     elseif dual isa AbstractVector
+        # The equality constraints gives the term <dual, primal> with the
+        # standard inner product but <dual, primal>_PSD is like scaling each
+        # entry of dual and primal by the entry of SetDotScalingVector. To undo,
+        # we need to divide by the square.
         scale = MOI.Utilities.SetDotScalingVector{T}(bridge.set)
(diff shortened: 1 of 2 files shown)


src/Bridges/Constraint/bridges/slack.jl
```diff
@@ -124,12 +124,20 @@ end
 function MOI.get(
     model::MOI.ModelLike,
     a::Union{MOI.ConstraintDual,MOI.ConstraintDualStart},
-    bridge::_AbstractSlackBridge,
-)
+    bridge::_AbstractSlackBridge{T},
+) whhere {T}
```
odow, line 128:
```suggestion
) where {T}
```

odow:
https://github.com/jump-dev/MathOptInterface.jl/actions/runs/9531770131

odow:
Still the same problem: https://github.com/jump-dev/MathOptInterface.jl/actions/runs/9531770131/job/26273074388

image

There's a suspicious `1 / sqrt(2)` there, so perhaps we need `1 / scale^2` somewhere?

odow:
https://github.com/jump-dev/MathOptInterface.jl/actions/runs/9539455292

src/Bridges/Constraint/bridges/slack.jl
```diff
@@ -124,12 +124,25 @@ end
...
+    # does not # support `ConstraintDual`. This is the case if the adjoint of
+    # the linear map on which the bridge is based is not invertible, as for
+    # instance with `Variable.ZerosBridge` or
+    # `SumOfSquares.Bridges.Variable.KernelBridge`.
+    dual = MOI.get(model, a, bridge.equality)
+    if dual === nothing
+        return nothing
+    elseif dual isa AbstractVector
+        scale = MOI.Utilities.SetDotScalingVector{T}(bridge.set)
+        return dual ./ scale.^2
```
odow, line 143:
```suggestion
        return dual ./ scale .^ 2
```

odow:
The COPT failure is unrelated: https://github.com/COPT-Public/COPT.jl/pull/40

odow:
I don't fully understand the math logic, but the code does something like `set_dot(scale * x, scale * y)`, which is more or less equivalent to `sum(scale.^2 .* x .* y)`, so to reverse, we need to divide by `scale.^2`.

blegat:
Maybe we should add a test in `test/Bridges/Constraint/slack.jl`

odow:
I've added a test that triggers the `scale` and tweaked the comments. I also needed to update the `MOI.set` for `::ConstraintDualStart` because it didn't scale things either.

As a more general comment: our tests just check that `get/set` is a round-trip. We can't easily automatically test that the numeric values are correct.

odow (approved):

odow:
Let's just double check:

https://github.com/jump-dev/MathOptInterface.jl/actions/runs/9556803924

src/Bridges/Constraint/bridges/slack.jl
```diff
@@ -124,26 +124,46 @@ end
...
     value,
-)
-    # As the slack appears `+slack` in `slack_in_set` and `-slack` in equality,
-    # giving `value` to both will cancel it out in the Lagrangian.
-    # Giving `value` to `bridge.equality` will put the function in the
-    # Lagrangian as expected.
+) where {T}
+    # See comments in MOI.get for why we need to rescale, etc.
     MOI.set(model, attr, bridge.slack_in_set, value)
-    MOI.set(model, attr, bridge.equality, value)
```
blegat, line 146: Ouch, this was a bug, good catch

PR #2520, opened 2024-06-24
Revert removal of _try_constrain_variables_on_creation

odow:
These were removed in https://github.com/jump-dev/MathOptInterface.jl/pull/2495, but I didn't check that it didn't break downstream. Closes https://github.com/jump-dev/MathOptInterface.jl/issues/2519


odow:
https://github.com/jump-dev/MathOptInterface.jl/actions/runs/9639001476

PR #2532, opened 2024-08-07
Allow changing coefficient type in FileFormats.Model

blegat:
Needed for https://github.com/jump-dev/JuMP.jl/pull/3801


odow:
I wonder if this should be `value_type`: https://jump.dev/JuMP.jl/stable/api/JuMP/#value_type

odow:
I don't think we should add this. Many file formats do not support non-Float64 number types, and SDPA already has a `number_type` argument:
https://github.com/jump-dev/MathOptInterface.jl/blob/b5659b5cd4a0928db211a08530d4fa612c81aeba/src/FileFormats/SDPA/SDPA.jl#L57

This is also not needed for JuMP: https://github.com/jump-dev/JuMP.jl/pull/3801#issuecomment-2274907561

odow:
The correct solution is probably to standardize on a `value_type` kwarg where file formats support it.

PR #2536, opened 2024-08-24
Add set conversion bridge

blegat:


odow:
This need docs and tests

odow:
I'm still not convinced this is necessary

src/Bridges/Constraint/bridges/set_conversion.jl
```diff
@@ -0,0 +1,99 @@
...
+`SetConversionBridge` implements the following reformulations:
+
+  * ``f(x) \\in S1`` into ``f(x) \\in S2``
+
+In order to add this bridge, you need to create a bridge specific
+for a given type `T` and set `S2`:
+```julia
+MOI.Bridges.add_bridge(model, MOI.Bridges.Constraint.SetConversionBridge{T,S2})
+```
+In order to define a bridge with `S2` specified but `T` unspecified, e.g.,
```
blegat, line 14:
```suggestion
In order to define a bridge with `S2` specified but `T` unspecified, for example
```

odow:
Isn't a better solution just for CBLS to use the MOI sets?

blegat:
CBLS defines some with additional options so if the user uses the MOI set than he can use MiniZinc as well but if he plans to use CBLS only then he can uses sets with addtional options

src/Bridges/Constraint/bridges/set_conversion.jl
```diff
@@ -0,0 +1,99 @@
+"""
```
odow, line 1: Needs a copyright header

odow (approved):
If you add the header I relent.

src/Bridges/Constraint/bridges/set_conversion.jl
odow, line 1:
```suggestion
# Copyright (c) 2017: Miles Lubin and contributors
# Copyright (c) 2017: Google Inc.
#
# Use of this source code is governed by an MIT-style license that can be found
# in the LICENSE.md file or at https://opensource.org/licenses/MIT.

"""
```

PR #2542, opened 2024-09-02
[Nonlinear] allow univariate operators with only gradient information

### LINKED ISSUE #2534: Univariate user-defined functions with the multivariate signature
See https://discourse.julialang.org/t/nested-optimization-matrix-multiplication-and-internal-functions/118098/10?u=odow

odow:
Closes #2534

### DIFF
--- a/src/Nonlinear/operators.jl
+++ b/src/Nonlinear/operators.jl
@@ -54,6 +54,13 @@ struct _UnivariateOperator{F,F′,F′′}
     f::F
     f′::F′
     f′′::F′′
+    function _UnivariateOperator(
+        f::Function,
+        f′::Function,
+        f′′::Union{Nothing,Function} = nothing,
+    )
+        return new{typeof(f),typeof(f′),typeof(f′′)}(f, f′, f′′)
+    end
 end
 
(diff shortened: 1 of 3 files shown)


src/Nonlinear/operators.jl
```diff
@@ -339,14 +346,17 @@ end
...
 end
 
 function _UnivariateOperator(op::Symbol, f::Function, f′::Function)
-    _validate_register_assumptions(f′, op, 1)
-    f′′ = _checked_derivative(f′, op)
-    return _UnivariateOperator(f, f′, f′′)
+    try
+        _validate_register_assumptions(f′, op, 1)
+        f′′ = _checked_derivative(f′, op)
+        return _UnivariateOperator(f, f′, f′′)
```
odow, line 356: I guess one open question is this behavior: do we always try to ForwardDiff `f′`?

odow:
I'll run https://github.com/jump-dev/MathOptInterface.jl/actions/runs/10764818835 to double check this is okay.

odow:
D'oh. This will break a test in JuMP, because now univariate operators do not need Hessians. I should remove that test before we release this.

PR #2548, opened 2024-10-07
[FileFormats] read double-sided variable bounds separately

### LINKED ISSUE #2547: Inconsistent "interval" constraints when reading models from file
Interval constraints in file readers seem to be handled differently when reading back a model (c.f. #2121, #2351). This happens for `LP` and `MPS` files. It "works" with `MOF`, and is weird with `NL`. I'm not sure right now if making these consistent is even possible with the file formats (it's 05:18am at my place so this might be wrong in many ways...).

**MWE:**

```julia
using JuMP

m = Model()
@variable(m, 0 <= x <= 1)
has_lower_bound(x)              # true

write_to_file(m, "test.lp")
m = read_from_file("test.lp")

x = all_variables(m)[1]
has_lower_bound(x)              # false
```

This may, e.g., lead to

```julia
has_lower_bound(x) || set_lower_bound(x, 0)

# ERROR: MathOptInterface.LowerBoundAlreadySet{MathOptInterface.Interval{Float64},
# MathOptInterface.GreaterThan{Float64}}: Cannot add `VariableIndex`-in-`MathOptInterface.GreaterThan{Float64}`
# constraint for variable MOI.VariableIndex(1) as a `VariableIndex`-in-`MathOptInterface.Interval{Float64}`
# constraint was already set for this variable and both constraints set a lower bound.
```

---

**Additional information:**

A model looking like:

```console
Subject to
 x ≥ 0
 x ≤ 1
```

leads to:

- `LP`: `0 <= x <= 1` in the bounds section
- `MPS`: `LO 0` and `UP 1` in the bounds section
- a correct `NL` file (as far as I can decipher...)

One that is constructed using an explicit interval:

```console
Subject to
 x ∈ [0, 1]
```

leads to: _exactly the same files for `LP`, `MPS`, and `NL` as with the non-interval model_

> I can't really say if the last point with the interval `NL` file is a bug or intended... I kinda think that AMPL supports lower- and upper-bounds on non-linear constraints - there is a type for that (?) and the NL writer seems to handle these [here](https://github.com/jump-dev/MathOptInterface.jl/blob/master/src/FileFormats/NL/NL.jl#L404). But... this format always confuses me, so this is just guess work. But if that were correct, then the `NL` file is kinda inconsistent too.

odow:
Closes #2547


src/FileFormats/MPS/MPS.jl
```diff
@@ -1287,7 +1287,10 @@ function _add_variable(model, data, variable_map, i, name)
     variable_map[name] = x
     MOI.set(model, MOI.VariableName(), x, name)
     set = bounds_to_set(data.col_lower[i], data.col_upper[i])
-    if set !== nothing
+    if set isa MOI.Interval
+        MOI.add_constraint(model, x, MOI.GreaterThan(set.lower::Float64))
+        MOI.add_constraint(model, x, MOI.LessThan(set.upper::Float64))
+    elseif set !== nothing
         MOI.add_constraint(model, x, set)
     end
```
sstroemer, line 1295: Sorry if this is too nit-picky, but: Could creating the "helper" `Interval` set be an overhead for large files?

Something like this would be possible here (but makes it more convoluted):

```julia
if isfinite(data.col_lower[i])
    MOI.add_constraint(model, x, MOI.GreaterThan(data.col_lower[i]::Float64))
end

if isfinite(data.col_upper[i])
    MOI.add_constraint(model, x, MOI.LessThan(data.col_upper[i]::Float64))
end
```

This would also handle two edge cases differently:

1. If `data.col_lower[i] > data.col_upper[i]`, it adds the bounds keeping the model infeasible (`bounds_to_set` would currently remove the bounds and create a free variable)
2. If `data.col_lower[i] == data.col_upper[i]`, it sticks to adding bounds, and does not convert those into a single equality

(1.) is maybe not really important, but (2.) could lead to similar problems as the initial version: 

- I create a variable that somewhere along the way gets equal bounds, then
- write that to a file, read it back in, and
- the bounds are gone, `has_*_bound` will return `false`, and I can't change the bounds anymore.

Further, this also impacts how `fix(x, value; force=false)` behaves. With bounds it errors, with the `EqualTo` it overwrites.

test/FileFormats/MPS/MPS.jl
```diff
@@ -193,8 +195,12 @@ function test_integer_default_bounds()
     model = MPS.Model()
     MOI.read_from_file(model, joinpath(@__DIR__, "integer_default_bounds.mps"))
     x = only(MOI.get(model, MOI.ListOfVariableIndices()))
+    ci = MOI.ConstraintIndex{MOI.VariableIndex,MOI.GreaterThan{Float64}}(x.value)
```
odow, line 198:
```suggestion
    ci =
        MOI.ConstraintIndex{MOI.VariableIndex,MOI.GreaterThan{Float64}}(x.value)
```

src/FileFormats/MPS/MPS.jl
odow: > Could creating the "helper" Interval set be an overhead for large files?

Nope. It's a `struct` that is on the stack.

> Something like this would be possible here

`bounds_to_set` already does this :smile: it will return an `Interval` only if both sides are finite.

> the bounds are gone, has_*_bound will return false, and I can't change the bounds anymore.

You'd need to `unfix(x); set_lower_bound(x, lb)`.

The issue is that JuMP doesn't offer a nice way to interact with interval bounds.

I agree that lower/upper/fixed bounds are more convoluted than a plain lower/upper bound. You'll need to write some JuMP code to deal with the different situations.

PR #2552, opened 2024-10-08
Add new attribute NodeLimit

matbesancon:
This adds a node limit attribute, which seems to make sense given that there is the corresponding termination status. Most MIP solvers have a limit on the number of nodes in the BNB tree. Note that this is in general different from the number of solved LPs, since each node contains its own solve-and-cut loop (at least for SCIP)

### DIFF
--- a/docs/src/tutorials/implementing.md
+++ b/docs/src/tutorials/implementing.md
@@ -335,7 +335,7 @@ method for each attribute.
 | [`TimeLimitSec`](@ref) | Yes           | Yes           | Yes                |
 | [`ObjectiveLimit`](@ref) | Yes         | Yes           | Yes                |
 | [`SolutionLimit`](@ref) | Yes          | Yes           | Yes                |
-| [`NodeLimit`](@ref) | Yes          | Yes           | Yes                |
+| [`NodeLimit`](@ref)     | Yes          | Yes           | Yes                |
 | [`RawOptimizerAttribute`](@ref) | Yes  | Yes           | Yes                |
 | [`NumberOfThreads`](@ref) | Yes        | Yes           | Yes                |
 | [`AbsoluteGapTolerance`](@ref) | Yes   | Yes           | Yes                |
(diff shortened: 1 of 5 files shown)


matbesancon:
this also makes for a clean deterministic limit if one wants to solve a MIP up to the root node only (e.g. to get a dual bound from a tightened LP and solutions from a bunch of heuristics)

src/attributes.jl
```diff
@@ -899,6 +899,29 @@ struct SolutionLimit <: AbstractOptimizerAttribute end
 
 attribute_value_type(::SolutionLimit) = Union{Nothing,Int}
 
+"""
+    NodeLimit()
+
+An optimizer attribute for setting a limit on the number of branch-and-bound nodes explored by a MIP solver.
```
blegat, line 905:
```suggestion
An optimizer attribute for setting a limit on the number of branch-and-bound nodes explored by a mixed-integer program (MIP) solver.
```

src/attributes.jl
```diff
@@ -899,6 +899,29 @@ struct SolutionLimit <: AbstractOptimizerAttribute end
...
+
+## Default values
+
+The provided limit must be a `Union{Nothing,Int}`.
```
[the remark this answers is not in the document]
matbesancon: it's the same issue for many other attributes no?

docs/src/tutorials/implementing.md
```diff
@@ -335,6 +335,7 @@ method for each attribute.
 | [`TimeLimitSec`](@ref) | Yes           | Yes           | Yes                |
 | [`ObjectiveLimit`](@ref) | Yes         | Yes           | Yes                |
 | [`SolutionLimit`](@ref) | Yes          | Yes           | Yes                |
+| [`NodeLimit`](@ref) | Yes          | Yes           | Yes                |
```
odow, line 338:
```suggestion
| [`NodeLimit`](@ref)     | Yes          | Yes           | Yes                |
```

src/attributes.jl
```diff
@@ -899,6 +899,29 @@ struct SolutionLimit <: AbstractOptimizerAttribute end
 
 attribute_value_type(::SolutionLimit) = Union{Nothing,Int}
 
+"""
+    NodeLimit()
+
+An optimizer attribute for setting a limit on the number of branch-and-bound nodes explored by a mixed-integer program (MIP) solver.
```
odow, line 905:
```suggestion
An optimizer attribute for setting a limit on the number of branch-and-bound
nodes explored by a mixed-integer program (MIP) solver.
```

src/attributes.jl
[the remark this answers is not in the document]
odow: I don't think it's common to explore more than `2_147_483_647` nodes on a 32-bit machine :smile:

PR #2555, opened 2024-10-09
Fix dimension of nonlinear function in basic tests

blegat:
EAGO was failing this test because it was recoving the dimension of the set from the dimension of the function instead of writing the dimension of the set in the model.

Part of https://github.com/jump-dev/MathOptInterface.jl/issues/2553


blegat:
I get the same coverage decrease on the two new PRs so it might be due to Julia v1.10 -> v1.11

PR #2559, opened 2024-10-17
Use eachindex in Coloring

blegat:
Fixes VS code warnings


odow:
I'm ambivalent about these changes. But sure.

PR #2561, opened 2024-10-18
[Coloring] avoid index conflict in variable name

blegat:
`i` is already used inside the loop so it's best to use `k` for the iteration just like in the previous `for` loop.


mlubin:
Nice catch

PR #2565, opened 2024-10-25
[DNM] add variables and bound in a single shot

joaquimg:
possible fix for: https://github.com/jump-dev/MathOptInterface.jl/issues/2564


joaquimg:
#2574 is a strict improvement over this.

PR #2572, opened 2024-10-30
[Utilities] fix adding unsupported constraints to AbstractModel

odow:
Closes #2750
Cloess #2751

This PR is an alternative to #2751

### DIFF
--- a/src/Utilities/model.jl
+++ b/src/Utilities/model.jl
@@ -312,9 +312,16 @@ end
 
 function MOI.add_constraint(
     model::AbstractModel,
-    func::MOI.AbstractFunction,
-    set::MOI.AbstractSet,
-)
+    func::F,
+    set::S,
+) where {F<:MOI.AbstractFunction,S<:MOI.AbstractSet}
+    # We check supports_constraint here because it is a common practice for
+    # AbstractModels to declare that they do not support particular constraints,
+    # evenn though the underlying `.constraints` object does. See, e.g., the
+    # various models in MOI.FileFormats.
+    if !MOI.supports_constraint(model, F, S)
+        throw(MOI.UnsupportedConstraint{F,S}())
+    end
     return MOI.add_constraint(model.constraints, func, set)
 end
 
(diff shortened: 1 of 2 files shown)


src/Utilities/model.jl
```diff
@@ -328,10 +335,17 @@ end
...
-    f::MOI.VariableIndex,
-    s::MOI.AbstractScalarSet,
-)
-    return MOI.add_constraint(model.variables, f, s)
+    func::MOI.VariableIndex,
+    set::S,
+) where {S<:MOI.AbstractScalarSet}
+    # We check supports_constraint here because it is a common practice for
+    # AbstractModels to declare that they do not support particular constraints,
+    # evenn though the underlying `.constraints` object does. See, for example,
```
blegat, line 343:
```suggestion
    # even though the underlying `.constraints` object does. See, for example,
```

blegat:
Ah, I understand better your other PR now, it's not such a nasty bug. But I wouldn't say it was expected behavior that we were permissive by letting `add_constraint` work even when `supports_constraint` was defined. I think this PR gives the expected behavior.

blegat (approved):

odow:
Yeah it's a case of "things worked when they shouldn't have." It also never came up in practice at the JuMP level, because we always copy to a file format with bridges, and they checked appropriately.

src/Utilities/model.jl
```diff
@@ -312,9 +312,16 @@ end
...
     model::AbstractModel,
-    func::MOI.AbstractFunction,
-    set::MOI.AbstractSet,
-)
+    func::F,
+    set::S,
+) where {F<:MOI.AbstractFunction,S<:MOI.AbstractSet}
+    # We check supports_constraint here because it is a common practice for
+    # AbstractModels to declare that they do not support particular constraints,
+    # evenn though the underlying `.constraints` object does. See, for example,
```
odow, line 320:
```suggestion
    # even though the underlying `.constraints` object does. See, for example,
```

PR #2574, opened 2024-10-31
add_constrained_variable with 2 sets

blegat:
Alternative fix for https://github.com/jump-dev/MathOptInterface.jl/issues/2564
Actually, this also has the potential to solve another issue.
If you have a variable that belongs to 2 sets, you never know which one should be added at creation as it's solver specific.
When you have a caching optimizer and then a bridge layer it's fine because it will get reordered in `copy_to` but otherwise you have to manually check the bridging cost.
The implementation of this PR could be improved to handle this case as well by doing something like:
```julia
function _cost_of_bridging( # copy pasted from `Utilities/copy.jl`
    dest::MOI.ModelLike,
    ::Type{S},
) where {S<:MOI.AbstractScalarSet}
    x = MOI.get(dest, MOI.VariableBridgingCost{S}())
    y = MOI.get(dest, MOI.ConstraintBridgingCost{MOI.VariableIndex,S}())
    return x - y
end

# ...
cost1 = _cost_of_bridging(model, typeof(set1)
cost2 = _cost_of_bridging(model, typeof(set2)
if cost1 < cost2
    vi, ci1 = MOI.add_constrained_variable(model, set1)
    ci2 = MOI.add_constraint(model, vi, set2)
else
    vi, ci2 = MOI.add_constrained_variable(model, set2)
    ci1 = MOI.add_constraint(model, vi, set1)
end
return vi, ci1, ci2
#...
```

Closes https://github.com/jump-dev/MathOptInterface.jl/pull/2565

### DIFF
--- a/src/Test/test_model.jl
+++ b/src/Test/test_model.jl
@@ -1226,3 +1226,20 @@ function test_model_show(model::MOI.ModelLike, ::Config{T}) where {T}
     @test sprint(show, model) isa String
     return
 end
+
+function test_model_add_constrained_variale_tuple(
+    model::MOI.ModelLike,
+    ::Config{T},
+) where {T}
+    F = MOI.VariableIndex
+    set = (MOI.GreaterThan(zero(T)), MOI.LessThan(one(T)))
+    @requires MOI.supports_add_constrained_variable(model, typeof(set))
+    x, (c_l, c_u) = MOI.add_constrained_variable(model, set)
+    @test c_l == MOI.ConstraintIndex{F,MOI.GreaterThan{T}}(x.value)
+    @test c_u == MOI.ConstraintIndex{F,MOI.LessThan{T}}(x.value)
+    @test MOI.get(model, MOI.ConstraintFunction(), c_l) == x
+    @test MOI.get(model, MOI.ConstraintSet(), c_l) == set[1]
+    @test MOI.get(model, MOI.ConstraintFunction(), c_u) == x
+    @test MOI.get(model, MOI.ConstraintSet(), c_u) == set[2]
+    return
+end
--- a/src/Test/test_model.jl
+++ b/src/Test/test_model.jl
@@ -1227,7 +1227,7 @@ function test_model_show(model::MOI.ModelLike, ::Config{T}) where {T}
     return
 end
 
-function test_model_add_constrained_variale_tuple(
+function test_model_add_constrained_variable_tuple(
     model::MOI.ModelLike,
     ::Config{T},
 ) where {T}
--- a/src/variables.jl
+++ b/src/variables.jl
@@ -124,6 +124,12 @@ function add_constrained_variable(model::ModelLike, set::AbstractScalarSet)
     return variable, constraint
 end
 
+function add_constrained_variable(model::ModelLike, set1::AbstractScalarSet, set2::AbstractScalarSet)
+    variable, constraint1 = add_constrained_variable(model, set1)
+    constraint2 = add_constraint(model, variable, set2)
+    return variable, constraint1, constraint2
+end
+
 """
     supports_add_constrained_variables(
         model::ModelLike,
--- a/src/variables.jl
+++ b/src/variables.jl
@@ -124,10 +124,63 @@ function add_constrained_variable(model::ModelLike, set::AbstractScalarSet)
     return variable, constraint
 end
 
-function add_constrained_variable(model::ModelLike, set1::AbstractScalarSet, set2::AbstractScalarSet)
-    variable, constraint1 = add_constrained_variable(model, set1)
-    constraint2 = add_constraint(model, variable, set2)
-    return variable, constraint1, constraint2
+"""
+    add_constrained_variable(
+        model::ModelLike,
+        set::Tuple{<:GreaterThan,<:LessThan},
+    )
+
+A special-case method to add a scalar variable with a lower and upper bound.
+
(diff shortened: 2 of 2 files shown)


odow:
I really really don't want to make our logic even more complicated than it already is. See https://github.com/jump-dev/JuMP.jl/pull/3863, which involves no changes to MOI, and is "good but not great" and probably sufficient.

blegat:
If it's sufficient then this PR is not necessary. If it's not sufficient, I thought for a while after the JuMP-dev call and came to the conclusion that this PR might be a better idea than adding the ability to modify individual bounds of a `MOI.Interval` constraint.

joaquimg:
I also gave some extra thought about this after the call.
This is by far, my favorite solution.
It seems clean.
We will not need to worry about the cost of bridging for a long time.

odow:
I would approve if this was just:

```julia
function add_constrained_variable(model::ModelLike, set_gt::GreaterThan, set_lt::LessThan)
    x, _ = add_constrained_variable(model, set_gt)
    add_constraint(model, x, set_lt)
    return x, (set_gt, set_lt)
end
```

Are there any other instances where someone might want to add constrained variable to two sets (and it makes a performance difference)? I can't think of any.

We also lock in the order of the two sets, otherwise, we need solvers to support permutations, etc.

This single function has a clear use-case up a level (JuMP setting bounds), and it has a clear use-case down a level (HiGHS adding both bounds). This would only find its way to the solver in direct mode, but that's okay.

odow:
If I add these methods to JuMP and HiGHS, it works as expected:
```julia
julia> using JuMP, HiGHS

julia> function main()
           model = direct_model(HiGHS.Optimizer())
           set_silent(model)
           set_string_names_on_creation(model, false)
           @variable(model, 1 <= x[i in 1:10_000] <= 2)
           return model
       end

julia> @time main();
  0.009286 seconds (30.06 k allocations: 2.307 MiB)

julia> using BenchmarkTools

julia> @benchmark main()
BenchmarkTools.Trial: 479 samples with 1 evaluation.
 Range (min … max):   9.022 ms … 278.281 ms  ┊ GC (min … max): 0.00% … 85.28%
 Time  (median):      9.306 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):   10.433 ms ±  13.956 ms  ┊ GC (mean ± σ):  8.59% ±  6.75%

   ▆▆▇█▅▁                                                       
  ▅███████▅▅▅▂▃▄▃▂▃▃▂▃▃▂▂▃▁▃▃▁▁▂▁▂▁▁▂▂▁▂▁▂▁▁▁▁▁▁▁▂▁▁▁▁▁▁▁▁▁▁▁▂ ▃
  9.02 ms         Histogram: frequency by time         12.6 ms <

 Memory estimate: 2.31 MiB, allocs estimate: 30056.
```

odow:
How about:
```Julia
function add_constrained_variable(
    model::ModelLike,
    set::Tuple{<:GreaterThan,<:LessThan},
)
    set_greater_than, set_less_than = set
    x, c_greater_than = add_constrained_variable(model, set_greater_than)
    c_less_than = add_constraint(model, x, set_less_than)
    return x, (c_greater_than, c_less_than)
end
```

odow:
This is one very particular special case because it comes up often.

odow:
I'll update this PR.

src/Test/test_model.jl
```diff
@@ -1226,3 +1226,20 @@ function test_model_show(model::MOI.ModelLike, ::Config{T}) where {T}
     @test sprint(show, model) isa String
     return
 end
+
+function test_model_add_constrained_variale_tuple(
```
odow, line 1230:
```suggestion
function test_model_add_constrained_variable_tuple(
```

src/variables.jl
```diff
@@ -124,6 +124,65 @@ function add_constrained_variable(model::ModelLike, set::AbstractScalarSet)
...
+VariableIndex-in-GreaterThan{Float64}
+ v[1] >= 1.0
+
+VariableIndex-in-LessThan{Float64}
+ v[1] <= 2.0
+```
+"""
+function add_constrained_variable(
+    model::ModelLike,
+    set::Tuple{<:GreaterThan,<:LessThan},
```
odow, line 170: @blegat are you okay with this signature?

joaquimg: I have a small preference for two separate arguments rather than the tuple. But if you both prefer the Tuple, let's do it!

odow: I feel like the two separate arguments just opens the door too much.

This feels more like a "set" composed of a tuple. Rather than a completely new method. It's more logically consistent with the rest of the API. And the output of the constraint index matches the input of the set. Tuple->Tuple.

odow: I also think restricting to this one special case is a good start. We can always relax later if it turns out to be a good idea. But this is really the only case where we've had the need, and it's been years since we started MOI...

joaquimg: Fair enough, lets go for it.

odow: I continued to think about this all day, and I didn't come up with a reason not to do this.

blegat: > And the output of the constraint index matches the input of the set.

I agree, otherwise `MOI.add_constrained_variable(model, set)` should return `vi, (ci,)` but that would be breaking in addition to being annoying

blegat:
I can't review my own PR so I mark my validation by this comment

PR #2575, opened 2024-11-01
[Utilities] ignore attributes in UniversalFallback if set to nothing

odow:
x-ref https://github.com/jump-dev/Gurobi.jl/pull/584

We don't have a way to remove attributes once they've been set. We've mostly settled on `nothing` as the resetting to default.

For example
https://github.com/jump-dev/MathOptInterface.jl/blob/278fba5831af37d89bb0faed656e182f260343ff/src/attributes.jl#L329-L331

https://github.com/jump-dev/MathOptInterface.jl/blob/278fba5831af37d89bb0faed656e182f260343ff/src/attributes.jl#L848-L850

https://github.com/jump-dev/MathOptInterface.jl/blob/278fba5831af37d89bb0faed656e182f260343ff/src/attributes.jl#L861-L863

### DIFF
--- a/src/Utilities/universalfallback.jl
+++ b/src/Utilities/universalfallback.jl
@@ -486,8 +486,10 @@ function MOI.get(
     listattr::MOI.ListOfVariableAttributesSet,
 )
     list = MOI.get(uf.model, listattr)
-    for attr in keys(uf.varattr)
-        push!(list, attr)
+    for (attr, dict) in uf.varattr
+        if !isempty(dict)
+            push!(list, attr)
+        end
     end
     return list
 end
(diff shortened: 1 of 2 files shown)


src/Utilities/universalfallback.jl
```diff
@@ -486,8 +490,10 @@ function MOI.get(
     listattr::MOI.ListOfVariableAttributesSet,
 )
     list = MOI.get(uf.model, listattr)
-    for attr in keys(uf.varattr)
-        push!(list, attr)
+    for (attr, value) in uf.varattr
+        if value !== nothing
```
blegat, line 494:
```suggestion
        if any(!isnothing, values(value))
```

odow:
This needs some tests

odow:
https://github.com/jump-dev/MathOptInterface.jl/actions/runs/11655223284

odow:
solver-tests.yml shows no issues.

blegat (approved):

blegat:
Another approach would be to simply delete it from the dictionary when it's set to `nothing`

PR #2577, opened 2024-11-02
Promote bound of Interval if they have different types

blegat:
- [x] Add tests
- [x] Do it for `Semiinteger` and `Semicontinuous` as well

Closes https://github.com/jump-dev/MathOptInterface.jl/issues/2576

### DIFF
--- a/src/sets.jl
+++ b/src/sets.jl
@@ -286,6 +286,10 @@ struct Interval{T<:Real} <: AbstractScalarSet
     upper::T
 end
 
+function Interval(lower, upper)
+    return Interval(promote(lower, upper)...)
+end
+
 function Base.:(==)(a::Interval{T}, b::Interval{T}) where {T}
     return a.lower == b.lower && a.upper == b.upper
 end
--- a/src/sets.jl
+++ b/src/sets.jl
@@ -286,7 +286,7 @@ struct Interval{T<:Real} <: AbstractScalarSet
     upper::T
 end
 
-function Interval(lower, upper)
+function Interval(lower::Real, upper::Real)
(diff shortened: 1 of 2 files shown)


src/sets.jl
```diff
@@ -286,6 +286,10 @@ struct Interval{T<:Real} <: AbstractScalarSet
     upper::T
 end
 
+function Interval(lower, upper)
```
odow, line 289:
```suggestion
function Interval(lower::Real, upper::Real)
```

test/sets.jl
```diff
@@ -363,6 +363,19 @@ function runtests()
     end
 end
 
+function test_interval_promote()
+    for S in [MOI.Interval, MOI.Semiinteger, MOI.Semicontinuous]
```
odow, line 367:
```suggestion
    for S in (MOI.Interval, MOI.Semiinteger, MOI.Semicontinuous)
```
It's okay in this case, but I always try to avoid `[]` for these sorts of heterogeneous collections because it promotes to a common argument. `[1, x, 2x + 3]` is a great example because they all get promoted to `AffExpr` so you don't actually test different values.

test/sets.jl
```diff
@@ -363,6 +363,19 @@ function runtests()
...
+        set = S(1.0, π)
+        @test set isa S{Float64}
+        @test set.lower == 1.0
+        @test set.upper ≈ π
+        set = S(big(1), 2)
+        @test set isa S{BigInt}
+        @test set.lower == 1
+        @test set.upper == 2
+    end
```
odow, line 376:
```suggestion
    end
    return
```

odow (approved):

PR #2587, opened 2024-12-04
[Bridges] add ExponentialConeToScalarNonlinearFunctionBridge

odow:
This would enable:

```julia
julia> using JuMP, Ipopt

julia> begin
           set_silent(model)
           @variable(model, x == 1.5)
           @variable(model, z)
           @objective(model, Min, z)
           @constraint(model, [x, 1, z] in MOI.ExponentialCone())
           optimize!(model)
           @assert is_solved_and_feasible(model)
           value(z), exp(1.5)
       end
(4.4816890628439685, 4.4816890703380645)
```

Now, it's an open question whether we _should_ do this, but it comes up from time to time.

Latest is https://discourse.julialang.org/t/lazybridgeoptimizer-not-working-with-copy-to/123366

cc @thibault-vignon

### DIFF
--- a/src/Bridges/Constraint/bridges/ExponentialConeToScalarNonlinearFunctionBridge.jl
+++ b/src/Bridges/Constraint/bridges/ExponentialConeToScalarNonlinearFunctionBridge.jl
@@ -36,7 +36,7 @@ mutable struct ExponentialConeToScalarNonlinearFunctionBridge{
     ci_y::MOI.ConstraintIndex{MOI.ScalarAffineFunction{T},MOI.GreaterThan{T}}
 end
 
-const ExponentialToScalarNonlinearFunction{T,OT<:MOI.ModelLike} =
+const ExponentialConeToScalarNonlinearFunction{T,OT<:MOI.ModelLike} =
     SingleBridgeOptimizer{ExponentialConeToScalarNonlinearFunctionBridge{T},OT}
 
 function bridge_constraint(
--- a/src/Bridges/Constraint/bridges/ExponentialConeToScalarNonlinearFunctionBridge.jl
+++ b/src/Bridges/Constraint/bridges/ExponentialConeToScalarNonlinearFunctionBridge.jl
@@ -1,0 +1,162 @@
+# Copyright (c) 2017: Miles Lubin and contributors
+# Copyright (c) 2017: Google Inc.
+#
+# Use of this source code is governed by an MIT-style license that can be found
+# in the LICENSE.md file or at https://opensource.org/licenses/MIT.
+
+"""
+    ExponentialConeToScalarNonlinearFunctionBridge{T,F} <:
+        Bridges.Constraint.AbstractBridge
+
+`ExponentialConeToScalarNonlinearFunctionBridge` implements the following
+reformulation:
+
+  * ``(x, y, z) \\in \\textsf{ExponentialCone}()`` to
+    ``y \\cdot exp(x / y)) - z \\le 0``, ``y \\ge 0``.
+
+## Source node
+
+`ExponentialConeToScalarNonlinearFunctionBridge` supports:
+
+  * `F` in [`MOI.ExponentialCone`](@ref)
+
+## Target nodes
+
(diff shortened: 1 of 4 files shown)


src/Bridges/Constraint/bridges/ExponentialConeToScalarNonlinearFunctionBridge.jl
```diff
@@ -0,0 +1,159 @@
...
+) where {T,F<:Union{MOI.VectorOfVariables,MOI.VectorAffineFunction{T}}}
+    x, y, z = MOI.Utilities.scalarize(f)
+    g_x_div_y = MOI.ScalarNonlinearFunction(:/, Any[x, y])
+    g_exp_x_div_y = MOI.ScalarNonlinearFunction(:exp, Any[g_x_div_y])
+    g = MOI.ScalarNonlinearFunction(
+        :-,
+        Any[MOI.ScalarNonlinearFunction(:*, Any[y, g_exp_x_div_y]), z],
+    )
+    ci = MOI.add_constraint(model, g, MOI.LessThan(zero(T)))
+    ci_y = MOI.add_constraint(model, one(T) * y, MOI.GreaterThan(zero(T)))
```
blegat, line 56: Can you add a comment to remind why we don't add a variable bound here ?

blegat: I remember that it gets very complicated in case it already has a bound if we implement deletion, we need to do it in `final_touch` and so on. We already discussed this in another bridge but I don't remember which one and what we ended up doing. Ideally we should have a small note in the dev doc of MOI that is referenced from every bridge that does that.

blegat:
I think it's fine to do by default

src/Bridges/Constraint/bridges/ExponentialConeToScalarNonlinearFunctionBridge.jl
```diff
@@ -0,0 +1,159 @@
...
+mutable struct ExponentialConeToScalarNonlinearFunctionBridge{
+    T,
+    F<:Union{MOI.VectorOfVariables,MOI.VectorAffineFunction{T}},
+} <: AbstractBridge
+    f::F
+    ci::MOI.ConstraintIndex{MOI.ScalarNonlinearFunction,MOI.LessThan{T}}
+    ci_y::MOI.ConstraintIndex{MOI.ScalarAffineFunction{T},MOI.GreaterThan{T}}
+end
+
+const ExponentialToScalarNonlinearFunction{T,OT<:MOI.ModelLike} =
```
odow, line 39:
```suggestion
const ExponentialConeToScalarNonlinearFunction{T,OT<:MOI.ModelLike} =
```

src/Bridges/Constraint/bridges/ExponentialConeToScalarNonlinearFunctionBridge.jl
```diff
@@ -0,0 +1,159 @@
...
+) where {T,F<:Union{MOI.VectorOfVariables,MOI.VectorAffineFunction{T}}}
+    x, y, z = MOI.Utilities.scalarize(f)
+    g_x_div_y = MOI.ScalarNonlinearFunction(:/, Any[x, y])
+    g_exp_x_div_y = MOI.ScalarNonlinearFunction(:exp, Any[g_x_div_y])
+    g = MOI.ScalarNonlinearFunction(
+        :-,
+        Any[MOI.ScalarNonlinearFunction(:*, Any[y, g_exp_x_div_y]), z],
+    )
+    ci = MOI.add_constraint(model, g, MOI.LessThan(zero(T)))
+    ci_y = MOI.add_constraint(model, one(T) * y, MOI.GreaterThan(zero(T)))
```
odow, line 56:
```suggestion
    # We add this as a constraint to avoid conflicting with existing variable
    # bounds, of which there can be at most one of.
    ci_y = MOI.add_constraint(model, one(T) * y, MOI.GreaterThan(zero(T)))
```

PR #2590, opened 2024-12-08
[Bridges] enable SOCtoNonconvexQuadBridge by default

odow:
If we're adding #2587, we should think about changing this.


odow:
Conclusion from developer call is that we should not add this by default. Instead, we should properly document how people can turn this on.

PR #2591, opened 2024-12-12
[Bridges] add ComplementsToScalarNonlinearFunctionBridge

### LINKED ISSUE #2589: Bridge from complement to nonlinear
We discussed this with @frapac this week. Given `0 <= f ⟂ g >= 0`, there are two possibilities: add a constraint that the product is zero or add one that the product is nonpositive. According to @frapac , the latter is more appropriate if the solver is interior point. So it's unclear which one should be added by default but we can start by creating a bridge (probably parametrized by a `Bool` indicating which approach is employed).

odow:
Closes #2589

This would be a replacement for Complementarity.jl (cc @chkwon). It's not a perfect replacement because there's no flexibility in how we reformulate, but it's good enough for basic stuff:
```julia
julia> using JuMP, Ipopt

julia> begin
           set_silent(model)
           @variable(model, x)
           @variable(model, z)
           @variable(model, w >= 0)
           @objective(model, Min, (x - 3.5)^2 + (z + 4)^2)
           @constraint(model, z - 3 + 2*z*w == 0)
           @constraint(model, -(z^2 - x) ⟂  w)
           optimize!(model)
           solution_summary(model)
       end
* Solver : Ipopt

* Status
  Result count       : 1
  Termination status : LOCALLY_SOLVED
  Message from the solver:
  "Solve_Succeeded"

* Candidate solution (result #1)
  Primal status      : FEASIBLE_POINT
  Dual status        : FEASIBLE_POINT
  Objective value    : 2.82500e+01

* Work counters
  Solve time (sec)   : 8.78320e-02
  Barrier iterations : 303
```

Current reformulation is really just a proof of concept to check things work. Numerically, we might be better off doing this reformulation though. It just required more lines of code so I didn't do it to begin with.

```julia
y - F == 0  # G-in-EqualTo
if isfinite(l)
    (x - l) * y <= 0.0  # ScalarQuadratic-in-LessThan
else
    y <= 0.0  # VariableIndex-in-LessThan
end
if isfinite(u)
    (x - u) * y <= 0.0  # ScalarQuadratic-in-LessThan
else
    y >= 0.0.  # VariableIndex-in-GreaterThan
end
```

### DIFF
--- a/src/Bridges/Constraint/Constraint.jl
+++ b/src/Bridges/Constraint/Constraint.jl
@@ -113,6 +113,10 @@ function add_all_bridges(bridged_model, ::Type{T}) where {T}
         bridged_model,
         ExponentialConeToScalarNonlinearFunctionBridge{T},
     )
+    MOI.Bridges.add_bridge(
+        bridged_model,
+        ComplementsToScalarNonlinearFunctionBridge{T},
+    )
     return
 end
 
--- a/src/Bridges/Constraint/bridges/ComplementsToScalarNonlinearFunctionBridge.jl
+++ b/src/Bridges/Constraint/bridges/ComplementsToScalarNonlinearFunctionBridge.jl
@@ -5,23 +5,24 @@
 # in the LICENSE.md file or at https://opensource.org/licenses/MIT.
 
 """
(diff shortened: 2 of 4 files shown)


odow:
Add as a package, not a bridge

PR #2595, opened 2024-12-17
Simplify test_conic_linear with forced bridging

blegat:
A lot of these helpers could be simplified with `SingleBridgeOptimizer`. What we are doing is basically hardcoding a bridge so adding a `SingleBridgeOptimizer` would be equivalent since this bridge optimizer layer always applies the bridge, even if the constraint is supported (which is why I call it *forced* bridging).
This PR just does it for one test as a proof of concept.

### DIFF
--- a/src/Test/test_conic.jl
+++ b/src/Test/test_conic.jl
@@ -19,8 +19,8 @@ Opt obj = -11, soln x = 1, y = 0, z = 2
 """
 function test_conic_linear_VectorOfVariables(
     model::MOI.ModelLike,
-    config::Config,
-)
+    config::Config{T},
+) where {T}
     @requires MOI.supports_incremental_interface(model)
     @requires MOI.supports(
         model,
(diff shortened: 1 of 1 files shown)


src/Test/test_conic.jl
```diff
@@ -161,13 +124,14 @@ end
...
+the same problem as [`test_conic_linear_VectorOfVariables`](@ref).
 """
 function test_conic_linear_VectorAffineFunction(
     model::MOI.ModelLike,
-    config::Config,
-)
-    _test_conic_linear_helper(model, config, false)
+    config::Config{T},
+) where {T}
+    test_conic_linear_VectorOfVariables(MOI.Bridges.Constraint.VectorFunctionize{T}(model), config)
```
odow, line 134: No. Because we can't nest `SingleBridgeOptimizer` and `LazyBridgeOptimizer`.

odow (changes requested):
See my comment about nesting `SingleBridgeOptimizer` and `LazyBridgeOptimizer`

odow:
https://github.com/jump-dev/MathOptInterface.jl/actions/runs/12424939329

PR #2596, opened 2024-12-21
[Bridges.Constraint] add bridges in alphabetical order

odow:


src/Bridges/Constraint/Constraint.jl
```diff
@@ -22,97 +22,115 @@ for filename in readdir(joinpath(@__DIR__, "bridges"); join = true)
...
+    add_all_bridges(model, ::Type{T}) where {T}
 
-Add all bridges defined in the `Bridges.Constraint` submodule to
-`bridged_model`. The coefficient type used is `T`.
-"""
-function add_all_bridges(bridged_model, ::Type{T}) where {T}
-    if T <: AbstractFloat
-        MOI.Bridges.add_bridge(bridged_model, GreaterToIntervalBridge{T})
-        MOI.Bridges.add_bridge(bridged_model, LessToIntervalBridge{T})
-    end
```
odow, line 34: @blegat: is it intended behaviour that the bridges depend on the order of calls in this function? That is quite a subtle design choice. But I guess it makes sense if there are two bridges with the same cost.

blegat: Yes it's quite subtle. In `test/Bridges` we have a few tests verifying that for specific model, we have the right bridges that are applied and these may not pass with any order of addition of bridges

odow: Can we tweak the bridge weights instead?

src/Bridges/Constraint/Constraint.jl
```diff
@@ -22,97 +22,115 @@ for filename in readdir(joinpath(@__DIR__, "bridges"); join = true)
...
-    add_all_bridges(bridged_model, ::Type{T}) where {T}
+    add_all_bridges(model, ::Type{T}) where {T}
 
-Add all bridges defined in the `Bridges.Constraint` submodule to
-`bridged_model`. The coefficient type used is `T`.
+Add all bridges defined in the `Bridges.Constraint` submodule to `model`.
+
+The coefficient type used is `T`.
 """
-function add_all_bridges(bridged_model, ::Type{T}) where {T}
```
blegat, line 30: Can you rename `bridged_model` to `model` in a separate PR ? Keeping a clean diff for this PR is important

odow: Will do in https://github.com/jump-dev/MathOptInterface.jl/pull/2598

odow:
I pulled out the `.Variable` stuff into https://github.com/jump-dev/MathOptInterface.jl/pull/2599

I'll update this to focus only on `.Constraint`

odow:
https://github.com/jump-dev/MathOptInterface.jl/actions/runs/12552291548

odow:
https://github.com/jump-dev/MathOptInterface.jl/actions/runs/12573879685

src/Bridges/Constraint/bridges/SOCtoPSDBridge.jl
```diff
@@ -186,6 +191,11 @@ end
 const RSOCtoPSD{T,OT<:MOI.ModelLike} =
     SingleBridgeOptimizer{RSOCtoPSDBridge{T},OT}
 
+# This bridge destorys a lot of structure and adding PSD variables is almost
+# always undesirable. We give this bridge an arbitrarily hight cost so that it
+# is used only if necessary.
+bridging_cost(::Type{<:RSOCtoPSDBridge}) = 10.0
```
odow, line 197: @blegat, I've added these costs here too

src/Bridges/Constraint/bridges/SOCtoPSDBridge.jl
```diff
@@ -77,6 +77,11 @@ end
 const SOCtoPSD{T,OT<:MOI.ModelLike} =
     SingleBridgeOptimizer{SOCtoPSDBridge{T},OT}
 
+# This bridge destorys a lot of structure and adding PSD variables is almost
+# always undesirable. We give this bridge an arbitrarily hight cost so that it
+# is used only if necessary.
+bridging_cost(::Type{<:SOCtoPSDBridge}) = 10.0
```
odow, line 83:
```suggestion
MOI.Bridges.bridging_cost(::Type{<:SOCtoPSDBridge}) = 10.0
```

src/Bridges/Constraint/bridges/SOCtoPSDBridge.jl
odow, line 197:
```suggestion
MOI.Bridges.bridging_cost(::Type{<:RSOCtoPSDBridge}) = 10.0
```

src/Bridges/Constraint/bridges/SOCtoPSDBridge.jl
```diff
@@ -186,6 +191,11 @@ end
 const RSOCtoPSD{T,OT<:MOI.ModelLike} =
     SingleBridgeOptimizer{RSOCtoPSDBridge{T},OT}
 
+# This bridge destorys a lot of structure and adding PSD variables is almost
+# always undesirable. We give this bridge an arbitrarily hight cost so that it
```
blegat, line 195:
```suggestion
# always undesirable. We give this bridge an arbitrarily high cost so that it
```

src/Bridges/Constraint/bridges/SOCtoPSDBridge.jl
```diff
@@ -77,6 +77,11 @@ end
 const SOCtoPSD{T,OT<:MOI.ModelLike} =
     SingleBridgeOptimizer{SOCtoPSDBridge{T},OT}
 
+# This bridge destorys a lot of structure and adding PSD variables is almost
+# always undesirable. We give this bridge an arbitrarily hight cost so that it
```
blegat, line 81:
```suggestion
# always undesirable. We give this bridge an arbitrarily high cost so that it
```

blegat, line 81:
```suggestion
# always undesirable. We give this bridge an arbitrary hight cost so that it
```

blegat: Do you mean arbitrarily high like infinite or high and arbitrary in the sense that the value 10 was chosen arbitrarily. Maybe we could also say "arbitrarily give"

blegat:
It's much cleaner like this, if we desire a priority over bridges, we should rely in explicit bridge cost, not bridge order. It also make sense that the bridges that we needed to add cost for are these soc to sdp ones.

test/Bridges/lazy_bridge_optimizer.jl
```diff
@@ -1751,13 +1750,13 @@ function test_bridge_selection()
         bridged_mock,
         MOI.VectorOfVariables,
         MOI.RotatedSecondOrderCone,
-    ) == MOI.Bridges.Constraint.RSOCtoPSDBridge{
+    ) == MOI.Bridges.Constraint.RSOCtoSOCBridge{
```
odow, line 1753: @blegat this test should have shown us the problem! It bridged RSOC to PSD, even though RSOC to SOC was available. It used to pick toPSD because that was the first bridge that was added.

PR #2599, opened 2024-12-30
[Bridges.Variable] add bridges in alphabetical order and change weight

odow:
Extracted from #2596 

The behaviour should, where possible, not depend on this somewhat arbitrary order of bridges in this function. That is too brittle and the behaviour is undocumented.

### DIFF
--- a/src/Bridges/Variable/bridges/RSOCtoPSDBridge.jl
+++ b/src/Bridges/Variable/bridges/RSOCtoPSDBridge.jl
@@ -56,6 +56,8 @@ end
 const RSOCtoPSD{T,OT<:MOI.ModelLike} =
     SingleBridgeOptimizer{RSOCtoPSDBridge{T},OT}
 
+MOI.Bridges.bridging_cost(::Type{<:RSOCtoPSDBridge}) = 1.5
+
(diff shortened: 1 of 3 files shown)


src/Bridges/Variable/bridges/RSOCtoPSDBridge.jl
```diff
@@ -56,6 +56,8 @@ end
 const RSOCtoPSD{T,OT<:MOI.ModelLike} =
     SingleBridgeOptimizer{RSOCtoPSDBridge{T},OT}
 
+MOI.Bridges.bridging_cost(::Type{<:RSOCtoPSDBridge}) = 1.5
```
odow, line 59: So there's an argument to be made here. Would it be better to globally penalise PSD sets more than RSOC sets? We're starting to muck with bridge costs, and it becomes a bit subjective.

blegat: It's debatable, this one is clearly destructing structure so it make sense to give it a high cost. But this would be inferrable form a higher cost given to PSD variables. You can even give a higher cost than 1, it should really only be used when there is no other choice ^^

PR #2603, opened 2025-01-06
[Utilities] add fallbacks for scalar_type and vector_type

odow:
https://github.com/jump-dev/MathOptInterface.jl/pull/2198 broke DiffOpt because it adds a call to `scalar_type`:
https://github.com/jump-dev/MathOptInterface.jl/pull/2601#issuecomment-2571845947

Ideally, we should have something that works always (but might be sub-optimal), and then DiffOpt can implement `scalar_type` if it wants better performance.

### DIFF
--- a/src/Utilities/functions.jl
+++ b/src/Utilities/functions.jl
@@ -592,7 +592,7 @@ end
 Type of functions obtained by indexing objects obtained by calling `eachscalar`
 on functions of type `F`.
 """
-function scalar_type end
+scalar_type(::Type{F}) where {F} = Any  # A default fallback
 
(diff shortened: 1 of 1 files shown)


blegat:
Given how uncommon it is ti define a new vector function type, I'd rather have an explicit error than silently having `Any`s creeping in. Let's say that `scalar_type` is part of the API so this just exposed a bug in DiffOpt that we can easily fix by defining `scalar_type` in DiffOpt before we release MOI

odow:
Closing in favor of https://github.com/jump-dev/DiffOpt.jl/pull/264

PR #2605, opened 2025-01-08
[Nonlinear] add support for simplifying NonlinearFunction

odow:
Heading towards https://github.com/jump-dev/MathOptInterface.jl/issues/2553

We have a few choices:

 * To leave this standalone
 * To make this the default for `Utilities.canonicalize`
 * To use this in `isapprox`

Not really sure what is a good approach. To be honest, `canonicalize!(::ScalarNonlinearFunction)` already does more than I remember adding.

### DIFF
--- a/src/Nonlinear/SymbolicAD/SymbolicAD.jl
+++ b/src/Nonlinear/SymbolicAD/SymbolicAD.jl
@@ -44,7 +44,7 @@ end
 function simplify(f::MOI.VectorAffineFunction{T}) where {T}
     f = MOI.Utilities.canonical(f)
     if isempty(f.terms)
-        return f.constant
+        return f.constants
     end
     return f
 end
--- a/src/Nonlinear/SymbolicAD/SymbolicAD.jl
+++ b/src/Nonlinear/SymbolicAD/SymbolicAD.jl
@@ -1,0 +1,225 @@
+# Copyright (c) 2017: Miles Lubin and contributors
+# Copyright (c) 2017: Google Inc.
+#
+# Use of this source code is governed by an MIT-style license that can be found
+# in the LICENSE.md file or at https://opensource.org/licenses/MIT.
+
+module SymbolicAD
+
+import MathOptInterface as MOI
+
+"""
+    simplify(f)
+
+Return a simplified version of the function `f`.
+
+!!! warning
+    This function is not type stable by design.
+"""
+simplify(f) = f
+
+function simplify(f::MOI.ScalarAffineFunction{T}) where {T}
+    f = MOI.Utilities.canonical(f)
+    if isempty(f.terms)
+        return f.constant
+    end
+    return f
+end
+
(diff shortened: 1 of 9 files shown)


chriscoey:
Nice, this seems very useful

chriscoey:
If we decide to add more algebraic simplifications in future, they won't be considered breaking right? Definitely looking forward to this!

odow:
Given the various bits I'm having to poke and prod, I'm not sure if this is a good or bad idea to have on by default.

src/Nonlinear/SymbolicAD/SymbolicAD.jl
```diff
@@ -0,0 +1,225 @@
...
+    if isempty(f.terms)
+        return f.constant
+    end
+    return f
+end
+
+function simplify(f::MOI.ScalarQuadraticFunction{T}) where {T}
+    f = MOI.Utilities.canonical(f)
+    if isempty(f.quadratic_terms)
+        return simplify(MOI.ScalarAffineFunction(f.affine_terms, f.constant))
```
blegat, line 32: Calling `simplify` will canonicalize again, we don't need to do the type unstable part of deciding whether it should be a constant.

src/Nonlinear/SymbolicAD/SymbolicAD.jl
```diff
@@ -0,0 +1,225 @@
+# Copyright (c) 2017: Miles Lubin and contributors
+# Copyright (c) 2017: Google Inc.
+#
+# Use of this source code is governed by an MIT-style license that can be found
+# in the LICENSE.md file or at https://opensource.org/licenses/MIT.
+
+module SymbolicAD
```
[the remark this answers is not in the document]
odow: My plan was to move https://github.com/lanl-ansi/MathOptSymbolicAD.jl/pull/39 into here

src/Nonlinear/SymbolicAD/SymbolicAD.jl
odow: I don't understand this comment

src/Nonlinear/SymbolicAD/SymbolicAD.jl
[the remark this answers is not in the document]
blegat: Makes sense then

src/Nonlinear/SymbolicAD/SymbolicAD.jl
blegat: we just did `f = MOI.Utilities.canonical(f)` so `f.affine_terms` are already canonicalized. Here, we create a SAF that we know is canonical and then pass it to `simplify`. The first thing the function will do is canonicalize again which is a bit wasteful

odow:
I went back to making some changes in https://github.com/lanl-ansi/MathOptSymbolicAD.jl/pull/39 for now

odow:
Closing. I'll open something else.

PR #2606, opened 2025-01-09
[FileFormats.MOF] replace OrderedDict by NamedTuple when writing

joaquimg:
All OrderedDicts used are very small. Hence, they allocate unnecessarily. This also creates more GC pressure.

### DIFF
--- a/src/FileFormats/MOF/write.jl
+++ b/src/FileFormats/MOF/write.jl
@@ -32,7 +32,7 @@ function Base.write(io::IO, model::Model)
     return
 end
 
-function write_variables(object, model::Model)
+function write_variables(object::Object, model::Model)
     name_map = Dict{MOI.VariableIndex,String}()
     for index in MOI.get(model, MOI.ListOfVariableIndices())
         variable = moi_to_object(index, model)
@@ -81,10 +81,10 @@ function extract_function_and_set(expr::Expr)
 end
 
 function write_nlpblock(
-    object::T,
+    object::Object,
     model::Model,
     name_map::Dict{MOI.VariableIndex,String},
-) where {T<:Object}
+)
     nlp_block = MOI.get(model, MOI.NLPBlock())
     if nlp_block === nothing
         return
@@ -116,10 +116,10 @@ function write_nlpblock(
 end
 
 function write_objective(
-    object::T,
+    object::Object,
     model::Model,
     name_map::Dict{MOI.VariableIndex,String},
-) where {T<:Object}
+)
     if object["objective"][:sense] != "feasibility"
         return  # Objective must have been written from NLPBlock.
     end
@@ -136,7 +136,7 @@ function write_objective(
 end
 
 function write_constraints(
-    object,
+    object::Object,
     model::Model,
     name_map::Dict{MOI.VariableIndex,String},
 )
--- a/src/FileFormats/MOF/write.jl
+++ b/src/FileFormats/MOF/write.jl
@@ -95,10 +95,10 @@ function write_nlpblock(
         objective = MOI.objective_expr(nlp_block.evaluator)
         objective = _lift_variable_indices(objective)
         sense = MOI.get(model, MOI.ObjectiveSense())
-        object["objective"] = NamedTuple{(:sense, :function)}(
-            (moi_to_object(sense),
-            moi_to_object(Nonlinear(objective), name_map))
-        )
+        object["objective"] = NamedTuple{(:sense, :function)}((
+            moi_to_object(sense),
+            moi_to_object(Nonlinear(objective), name_map),
+        ))
     end
     for (row, bounds) in enumerate(nlp_block.constraint_bounds)
         constraint = MOI.constraint_expr(nlp_block.evaluator, row)
@@ -106,10 +106,10 @@ function write_nlpblock(
         func = _lift_variable_indices(func)
         push!(
             object["constraints"],
-            NamedTuple{(:function, :set)}(
-            (moi_to_object(Nonlinear(func), name_map),
-            moi_to_object(set, name_map),)
-            )
+            NamedTuple{(:function, :set)}((
+                moi_to_object(Nonlinear(func), name_map),
+                moi_to_object(set, name_map),
+            )),
         )
     end
     return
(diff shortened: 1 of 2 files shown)


odow:
Could we get some benchmarks?

src/FileFormats/MOF/write.jl
```diff
@@ -94,21 +95,21 @@ function write_nlpblock(
...
         objective = _lift_variable_indices(objective)
         sense = MOI.get(model, MOI.ObjectiveSense())
-        object["objective"] = T(
-            "sense" => moi_to_object(sense),
-            "function" => moi_to_object(Nonlinear(objective), name_map),
-        )
+        object["objective"] = NamedTuple{(:sense, :function)}((
+            moi_to_object(sense),
+            moi_to_object(Nonlinear(objective), name_map),
+        ))
```
odow, line 101: TIL:
```Julia
julia> (function = 1,)
ERROR: ParseError:
# Error @ REPL[7]:1:11
(function = 1,)
#         ╙ ── unexpected `=`
Stacktrace:
 [1] top-level scope
   @ none:1
```

You can instead do:
```julia
julia> (; :function => 1)
(function = 1,)
```

joaquimg: nice catch, that is so much better

src/FileFormats/MOF/write.jl
```diff
@@ -119,17 +120,19 @@ function write_objective(
...
+    values = Any[moi_to_object(sense)]
     if sense != MOI.FEASIBILITY_SENSE
         F = MOI.get(model, MOI.ObjectiveFunctionType())
         objective_function = MOI.get(model, MOI.ObjectiveFunction{F}())
-        object["objective"]["function"] =
-            moi_to_object(objective_function, name_map)
+        push!(keys, :function)
+        push!(values, moi_to_object(objective_function, name_map))
     end
+    object["objective"] = NamedTuple{Tuple(keys)}(values)
```
odow, line 135: I think I would prefer 
```Julia
    sense = MOI.get(model, MOI.ObjectiveSense())
    if sense == MOI.FEASIBILITY_SENSE
        # Already set in default constructor
    else
        F = MOI.get(model, MOI.ObjectiveFunctionType())
        objective_function = MOI.get(model, MOI.ObjectiveFunction{F}())
        object["objective"] = (;
            :sense = moi_to_object(sense),
            :function = moi_to_object(objective_function, name_map)
        )
    end
```

odow (commented):
This is quite nice

src/FileFormats/MOF/write.jl
```diff
@@ -174,22 +177,28 @@ function moi_to_object(
...
+        push!(keys, :dual_start)
+        push!(values, dual_start)
     end
     if !isnothing(primal_start)
-        object["primal_start"] = primal_start
+        push!(keys, :primal_start)
+        push!(values, primal_start)
     end
-    return object
+    return NamedTuple{Tuple(keys)}(values)
```
odow, line 201: Instead of two separate vectors, you can use:
```julia
julia> pairs = Pair{Symbol,Any}[]
Pair{Symbol, Any}[]

julia> push!(pairs, :name => "abc")
1-element Vector{Pair{Symbol, Any}}:
 :name => "abc"

julia> push!(pairs, :dual_start => 0.0)
2-element Vector{Pair{Symbol, Any}}:
       :name => "abc"
 :dual_start => 0.0

julia> NamedTuple(pairs)
(name = "abc", dual_start = 0.0)
```

odow:
I opened an issue in JuliaFormatter to discuss formatting the named tuples: https://github.com/domluna/JuliaFormatter.jl/issues/895

joaquimg:
A test
```julia
using JuMP

model = Model()
@variable(model, x[1:1000] >= 0)
@constraint(model, y[1:1000], sum(i*x[i] for i in 1:1000) == 1)

for i in 1:5
    GC.gc()
    @time write_to_file(model, "test.mof.json")
end
```
Results:

before PR:
```
  5.528170 seconds (20.16 M allocations: 1.669 GiB, 31.54% gc time)
  5.244265 seconds (20.16 M allocations: 1.669 GiB, 30.18% gc time)
  5.487519 seconds (20.16 M allocations: 1.669 GiB, 29.10% gc time)
  5.344851 seconds (20.16 M allocations: 1.669 GiB, 30.58% gc time)
  5.348303 seconds (20.16 M allocations: 1.669 GiB, 29.13% gc time)
```

after PR:
```
  2.293431 seconds (5.13 M allocations: 936.505 MiB, 13.93% gc time)
  2.281071 seconds (5.13 M allocations: 936.508 MiB, 13.94% gc time)
  2.995470 seconds (5.13 M allocations: 936.454 MiB, 12.53% gc time)
  3.088288 seconds (5.13 M allocations: 936.497 MiB, 12.26% gc time)
  2.971376 seconds (5.13 M allocations: 936.498 MiB, 12.24% gc time)
```

Which is a big win as there is also the fixed cost of writing the json that both incur.

odow:
Nice nice nice.

Note that this could be regarded as technically breaking, but I don't think anyone is using `moi_to_object`: https://juliahub.com/ui/Search?type=code&q=moi_to_object. It's also not mentioned in the docs, so I'm going to say that it is the private API.

odow:
Reduction in memory pressure is a big deal for the larger Sienna models.

cc @jd-lara this should help your 'always write to MOF'

joaquimg:
should we rename `moi_to_object` to `_moi_to_object`?

odow:
Nah leave it for now

joaquimg:
green light from CI

src/FileFormats/MOF/write.jl
```diff
@@ -320,43 +311,40 @@ function moi_to_object(
     foo::MOI.ScalarAffineTerm,
     name_map::Dict{MOI.VariableIndex,String},
 )
-    return OrderedObject(
-        "coefficient" => foo.coefficient,
-        "variable" => name_map[foo.variable],
-    )
+    return (coefficient = foo.coefficient, variable = name_map[foo.variable])
```
odow, line 314: This should make such a difference; we were making a new OrderedDict for every single coefficient in the model 😭

odow (approved):

odow:
@joaquimg do you have a script for the before/after Sienna profile flame graph?

joaquimg:
Is the same script used here (but with PProf):

https://github.com/NREL-Sienna/PowerSimulations.jl/issues/1183

PR #2607, opened 2025-01-10
[FileFormats.MOF] remove all usage of OrderedDict

odow:
Follow-up to https://github.com/jump-dev/MathOptInterface.jl/pull/2606 to see if we can remove all instances of OrderedDict

### DIFF
--- a/src/FileFormats/MOF/write.jl
+++ b/src/FileFormats/MOF/write.jl
@@ -35,7 +35,6 @@ function Base.write(io::IO, model::Model)
 end
 
 function _write_variables(variables::Vector{NamedTuple}, model::Model)
-    variables = NamedTuple[]
     name_map = Dict{MOI.VariableIndex,String}()
     for index in MOI.get(model, MOI.ListOfVariableIndices())
         variable = moi_to_object(index, model)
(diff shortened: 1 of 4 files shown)


odow:
The Ordered/Unordered split was added in https://github.com/jump-dev/MathOptInterface.jl/pull/1111, just so that we could read unordered. But now that we use NamedTuples for writing, we don't need Ordered at all.

src/FileFormats/MOF/write.jl
```diff
@@ -11,33 +11,36 @@ Write `model` to `io` in the MathOptFormat file format.
...
+    variables, constraints = NamedTuple[], NamedTuple[]
+    name_map = _write_variables(variables, model)
+    objective = _write_nlpblock(constraints, model, name_map)
+    if objective === nothing
+        objective = _write_objective(model, name_map)
+    end
+    _write_constraints(constraints, model, name_map)
+    object = (;
+        name = "MathOptFormat Model",
+        version => (
```
odow, line 24:
```suggestion
        version = (
```

src/FileFormats/MOF/write.jl
```diff
@@ -11,33 +11,36 @@ Write `model` to `io` in the MathOptFormat file format.
...
-    write_objective(object, model, name_map)
-    write_constraints(object, model, name_map)
     indent = options.print_compact ? nothing : 2
     Base.write(io, JSON.json(object, indent))
     return
 end
 
-function write_variables(object::Object, model::Model)
+function _write_variables(variables::Vector{NamedTuple}, model::Model)
+    variables = NamedTuple[]
```
odow, line 38:
```suggestion
```

PR #2618, opened 2025-02-09
Add all linear indicator constraints in the Utilities.Model

matbesancon:
For some reason only indicator less than was added to MOI.Utilities.Model, which prevents copying models with that constraint

### DIFF
--- a/src/Utilities/model.jl
+++ b/src/Utilities/model.jl
@@ -771,6 +771,17 @@ const LessThanIndicatorOne{T} =
 const LessThanIndicatorZero{T} =
     MOI.Indicator{MOI.ACTIVATE_ON_ZERO,MOI.LessThan{T}}
 
+const GreaterThanIndicatorOne{T} =
+    MOI.Indicator{MOI.ACTIVATE_ON_ONE,MOI.GreaterThan{T}}
+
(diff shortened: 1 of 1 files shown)


joaquimg:
We should consider a copy test with all constraint types.

matbesancon:
can we do that in a separate PR? I'd love to avoid growing my infamous PR scope explosion legend

odow:
> can we do that in a separate PR? I'd love to avoid growing my infamous PR scope explosion legend

Yes, I can do this. :+1: for no scope explosion.

> which prevents copying models with that constraint

Can you clarify the use-case? You probably want to copy to `MOI.Utilities.UniversalFallback(MOI.Utilities.Model{Float64}())`

odow:
I just added a test here https://github.com/jump-dev/MathOptInterface.jl/pull/2619

matbesancon:
> Can you clarify your use case?

Yes indeed in my case using the universal fallback on top works, but it seems odd to need it on top of MOIU.Model given the description of the latter:

> An implementation of ModelLike that supports all functions and sets defined in MOI. It is parameterized by the coefficient type.

odow:
Haha yeah it's "supports almost all functions and sets"

PR #2620, opened 2025-02-10
[Nonlinear] fix performance of evaluating univariate operators

odow:
With the energy modeling benchmarking that @joaquimg has been doing, we've been talking about how small things never show up in a flame graph until you look at their aggregate across multiple call sites. 

One thing that shows up is `getindex`

image

With the only giveaway being the `ht_keyindex(::Dict{Symbol,Int}, ::Symbol)` method.

It's because we're calling `operators.univariate_operators[node.index]` in multiple places, which is a `Dict` lookup.

These calls don't show much because of inlining etc. Even in the code view, you're kinda like "call is expensive. Sure"

image

## Before

```Julia
julia> @benchmark(
           MOI.Nonlinear.ReverseAD._reverse_mode($(evaluator.backend), x),
           setup=(x=rand(num_variables(model))),
       )
BenchmarkTools.Trial: 44 samples with 1 evaluation per sample.
 Range (min … max):  112.457 ms … 132.677 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     113.303 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):   115.478 ms ±   4.641 ms  ┊ GC (mean ± σ):  0.00% ± 0.00%

   █▂                                                            
  ▇███▆▃▁▄▁▁▁▁▁▃▁▁▁▁▁▃▁▁▁▁▃▃▄▃▁▁▁▁▁▃▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▃▁▁▁▁▁▃ ▁
  112 ms           Histogram: frequency by time          133 ms <

 Memory estimate: 0 bytes, allocs estimate: 0.
```

## After

```Julia
julia> @benchmark(
           MOI.Nonlinear.ReverseAD._reverse_mode($(evaluator.backend), x),
           setup=(x=rand(num_variables(model))),
       )
BenchmarkTools.Trial: 48 samples with 1 evaluation per sample.
 Range (min … max):  102.579 ms … 115.444 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     104.459 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):   105.355 ms ±   2.961 ms  ┊ GC (mean ± σ):  0.00% ± 0.00%

       ███▆█ ▁                                                   
  ▄▁▁▄▄█████▇█▄▇▄▄▁▁▁▁▁▁▁▁▁▁▄▁▁▁▁▁▁▁▁▁▄▁▁▁▁▁▁▁▁▁▁▁▁▁▄▁▁▄▁▄▁▁▁▁▄ ▁
  103 ms           Histogram: frequency by time          115 ms <

 Memory estimate: 0 bytes, allocs estimate: 0.
```

## Script

```Julia
using JuMP, PowerModels, BenchmarkTools
case = "test/data/pglib_opf_case13659_pegase.m"
backend = MOI.Nonlinear.SparseReverseMode()
begin
    pm = PowerModels.instantiate_model(
        case,
        PowerModels.ACPPowerModel,
        PowerModels.build_opf,
    );
    model = pm.model
    nlp = MOI.Nonlinear.Model()
    for (F, S) in list_of_constraint_types(model)
        if F != NonlinearExpr
            continue
        end
        for ci in all_constraints(model, F, S)
            o = constraint_object(ci)
            MOI.Nonlinear.add_constraint(nlp, o.func, o.set)
        end
    end
    evaluator = MOI.Nonlinear.Evaluator(
        nlp,
        backend,
        index.(all_variables(model)),
    )
end;
MOI.initialize(evaluator, [:Grad, :Jac, :Hess])
```

### DIFF
--- a/src/Nonlinear/operators.jl
+++ b/src/Nonlinear/operators.jl
@@ -63,6 +63,33 @@ struct _UnivariateOperator{F,F′,F′′}
     end
 end
 
+function eval_univariate_function(operator::_UnivariateOperator, x::T) where {T}
+    ret = operator.f(x)
+    check_return_type(T, ret)
+    return ret::T
+end
+
+function eval_univariate_gradient(operator::_UnivariateOperator, x::T) where {T}
+    ret = operator.f′(x)
+    check_return_type(T, ret)
+    return ret::T
+end
+
+function eval_univariate_hessian(operator::_UnivariateOperator, x::T) where {T}
+    ret = operator.f′′(x)
+    check_return_type(T, ret)
+    return ret::T
+end
+
+function eval_univariate_function_and_gradient(
+    operator::_UnivariateOperator,
+    x::T,
+) where {T}
+    ret_f = eval_univariate_function(operator, x)
+    ret_f′ = eval_univariate_gradient(operator, x)
+    return ret_f, ret_f′
+end
+
 struct _MultivariateOperator{F,F′,F′′}
     N::Int
     f::F
@@ -564,9 +591,7 @@ function eval_univariate_function(
     end
     offset = id - registry.univariate_user_operator_start
     operator = registry.registered_univariate_operators[offset]
-    ret = operator.f(x)
-    check_return_type(T, ret)
-    return ret::T
+    return eval_univariate_function(operator, x)
 end
 
 """
@@ -619,9 +644,7 @@ function eval_univariate_gradient(
     end
     offset = id - registry.univariate_user_operator_start
     operator = registry.registered_univariate_operators[offset]
-    ret = operator.f′(x)
-    check_return_type(T, ret)
-    return ret::T
+    return eval_univariate_gradient(operator, x)
 end
 
 """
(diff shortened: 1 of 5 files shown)


blegat (approved):
Makes sense to remove this back and forth index mapping

src/Nonlinear/operators.jl
```diff
@@ -568,22 +624,109 @@ function eval_univariate_gradient(
...
+    registry::OperatorRegistry,
+    id::Integer,
+    x::T,
+) where {T}
+    if id <= registry.univariate_user_operator_start
+        return _eval_univariate(id, x)::Tuple{T,T}
+    end
+    offset = id - registry.univariate_user_operator_start
+    operator = registry.registered_univariate_operators[offset]
+    ret_f = operator.f(x)
```
blegat, line 676: `operator.f` and `operator.f'` both trigger a runtime method lookup. I'm wondering if adding a function barrier on `operator` would bring a speed up if there is a lot of user-defined functions

PR #2624, opened 2025-02-13
[Nonlinear] add SymbolicAD submodule

odow:
Twas prototyped in https://github.com/lanl-ansi/MathOptSymbolicAD.jl/pull/39

I'm open to arguments against adding this. The evaluator could arguably remain in the separate MathOptSymbolicAD.jl package. But the simplification and symbolic derivative stuff is more widely useful. I occasionally get questions on the forum asking how to compute the derivative of a JuMP expression.

## TODO

 - [x] Add documentation: https://jump.dev/MathOptInterface.jl/previews/PR2624/submodules/Nonlinear/SymbolicAD/

## Related issues

This doesn't close any issues directly, but it relates to a number of open and previously closed issues:

 * Replaces https://github.com/jump-dev/MathOptInterface.jl/pull/2605
 * Contains the features needed for https://github.com/jump-dev/MathOptInterface.jl/issues/2553
 * Fixes https://github.com/jump-dev/MathOptInterface.jl/issues/2533 (which was previously closed because I moved work to MathOptSymbolicAD.jl)
 * Sparsity detection is complete, so this also addresses https://github.com/jump-dev/MathOptInterface.jl/issues/2527
 * It's a good implementation of a new AD backend, so it is a pointer for https://github.com/jump-dev/MathOptInterface.jl/issues/2526

It currently lacks support for multivariate user-defined functions, because these require https://github.com/jump-dev/MathOptInterface.jl/issues/2402

## Use with JuMP

We require only a few trivial helper methods in JuMP to link this all up:

```Julia
julia> using JuMP

julia> function derivative(model::GenericModel{T}, f, x) where {T}
           df_dx = MOI.Nonlinear.SymbolicAD.derivative(moi_function(f), index(x))
           return jump_function(model, MOI.Nonlinear.SymbolicAD.simplify!(df_dx))
       end
derivative (generic function with 2 methods)

julia> derivative(f, x) = gradient(owner_model(x), f, x)
derivative (generic function with 2 methods)

julia> function gradient(model::GenericModel{T}, f) where {T}
           g = moi_function(f)
           ∇f = Dict{GenericVariableRef{T},Any}()
           for xi in MOI.Nonlinear.SymbolicAD.variables(g)
               df_dx = MOI.Nonlinear.SymbolicAD.simplify!(
                   MOI.Nonlinear.SymbolicAD.derivative(g, xi),
               )
               ∇f[GenericVariableRef{T}(model, xi)] = jump_function(model, df_dx)
           end
           return ∇f
       end
gradient (generic function with 2 methods)

julia> gradient(f) = gradient(owner_model(f), f)
gradient (generic function with 2 methods)

julia> model = Model();

julia> @variable(model, x[1:2]);

julia> f = sin(x[1]) + log(x[2])
sin(x[1]) + log(x[2])

julia> gradient(f)
Dict{VariableRef, Any} with 2 entries:
  x[1] => cos(x[1])
  x[2] => 1.0 / x[2]
```

### DIFF
--- a/docs/src/submodules/Nonlinear/SymbolicAD.md
+++ b/docs/src/submodules/Nonlinear/SymbolicAD.md
@@ -1,0 +1,259 @@
+```@meta
+CurrentModule = MathOptInterface
+DocTestSetup = quote
+    import MathOptInterface as MOI
+end
+DocTestFilters = [r"MathOptInterface|MOI"]
+```
+
+# SymbolicAD
+
+The `Nonlinear.SymbolicAD` submodule contains data structures and functions for
+working with the symbolic derivatives of a nonlinear optimization problem. This
+page explains the API and describes the rationale behind its design.
+
+## Background
+
+The code in `SymbolicAD` is inspired by Hassan Hijazi's work on
+[coin-or/gravity](https://github.com/coin-or/Gravity), a high-performance
+algebraic modeling language in C++.
+
+Hassan made the following observations:
+
+ * For large scale models, symbolic differentiation is slower than other
+   automatic differentiation techniques, such as the reverse mode algorithm
+   implemented in `MOI.Nonlinear.ReverseAD`.
+ * However, most large-scale nonlinear programs have a lot of structure.
+ * Gravity asks the user to provide structure in the form of
+   _template constraints_, where the user gives the symbolic form of the
+   constraint as well as a set of data to convert from a symbolic form to the
+   numerical form.
+ * Instead of differentiating each constraint in its numerical form, we can
+   compute one symbolic derivative of the constraint in symbolic form, and then
+   plug in the data in to get the numerical derivative of each function.
+ * As a final step, if users don't provide the structure, we can still infer it
+   --perhaps with less accuracy--by comparing the expression tree of each
+   constraint.
+
+The symbolic differentiation approach of Gravity works well when the problem is
+large with few unique constraints. For example, a model like:
+```julia
+model = Model()
+@variable(model, 0 <= x[1:10_000] <= 1)
+@constraint(model, [i=1:10_000], sin(x[i]) <= 1)
+@objective(model, Max, sum(x))
+```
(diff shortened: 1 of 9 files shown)


docs/src/submodules/Nonlinear/SymbolicAD.md
```diff
@@ -0,0 +1,259 @@
...
+@variable(model, x[1:2])
+@objective(model, Min, (1 - x[1])^2 + 100 * (x[2] - x[1]^2)^2)
+optimize!(model)
+```
+
+To revert back to the default sparse reverse mode algorithm, set the
+[`AutomaticDifferentiationBackend`](@ref) attribute to
+[`Nonlinear.SparseReverseMode`](@ref).
+
+## `simplilfy`
```
chriscoey, line 89: ## `simplify`

odow:
Yeah open question.

For now I've kept things separate from where we `canonicalize` affine/quadratic stuff, because we generally expect callers to use `canonicalize` in various places, for it to be type stable, and for it to have O(N) performance. We don't for example, canonicalize `ScalarQuadraticFunction` to `ScalarAffineFunction` if there are no quadratic terms.

`simplify` doesn't have these properties, and I don't think we want to start simplifying every nonlinear expression provided by the user (e.g., from JuMP). This really needs to be opt-in. And if it's opt-in, it kinda wants to be near the rest of the tools for working with symbolic expressions.

> Perhaps in some cases the rewrites could simplify an SNF to a SAF/SQF

Yes, this is on my TODO list. We can certainly do better at detecting linear subexpressions within a SNF.

joaquimg:
This would be **very** useful for: https://github.com/jump-dev/ParametricOptInterface.jl/issues/144
So, POI can support parameters in quadratic objectives.

POI only needs cubic function simplifcation, ideally in: sum of  cubic terms + SQF.

Then, POI would be used in DiffOpt so that we can easily differentiate wrt coefficients in the quadratic objective function.

PR #2626, opened 2025-02-14
VectorAffineFunction easy constructor

matbesancon:
VectorAffineFunction has always been a massive pain to construct.
This added constructor builds one from a vector of scalar affine functions, which is the mental model most people will have from a VAF.

This can drastically simplify a lot of code

### DIFF
--- a/src/Utilities/functions.jl
+++ b/src/Utilities/functions.jl
@@ -2207,14 +2207,7 @@ Returns the vector of scalar affine functions in the form of a
 `MOI.VectorAffineFunction{T}`.
 """
 function vectorize(funcs::AbstractVector{MOI.ScalarAffineFunction{T}}) where {T}
-    nterms =
-        mapreduce(func -> number_of_affine_terms(T, func), +, funcs, init = 0)
-    out_dim = mapreduce(func -> output_dim(T, func), +, funcs, init = 0)
-    terms = Vector{MOI.VectorAffineTerm{T}}(undef, nterms)
-    constant = zeros(T, out_dim)
-    fill_vector(terms, T, fill_terms, number_of_affine_terms, funcs)
-    fill_vector(constant, T, fill_constant, output_dim, funcs)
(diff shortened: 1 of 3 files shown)


odow:
I'm in favor of this. It hurts me _every_ time I try to create a VAF.

Thoughts on VectorQuadraticFunction as well?

matbesancon:
> Thoughts on VectorQuadraticFunction as well?

I'll take a shot in the next days, it will be in a separate PR

odow:
I can do it

blegat:
I'm not sure we need this since we already have `MOI.Utilities.vectorize`

matbesancon:
the constructor is easier to discover and search for, it has zero cost, gives you a way to construct the VAF that is close to the way it's printed, and the mental model makes sense -> constructing a vector affine expression takes a vector of scalar affine expressions.

MOI.Utilities.vectorize or MOI.Utilities.operate are both awkward to use constructors in disguise IMO

blegat:
We could consider having two names for the same thing but we shouldn't duplicate the implementation. One of the two should call the other one

odow:
We could document `vectorize` in the default constructor?

But case in point: I forget about `vectorize`

blegat:
Yes, and vectorize should call the constructor instead of duplicating the cose

joaquimg:
I like this PR, I almost wrote it twice recently.

odow:
Nah we should go the other way around: vectorize use this constructor. I can update the PR.

blegat:
Now the code is a bit weird. `vectorize` was sharing code between creating `VQF` and `VAF`. I would expect the code to also be faster as it preallocate the right dimension. I'd move that code to `src/function` for both the `VQF` and `VAF` constructor and then we can keep `vectorize` in `MOI.Utilities` that just calls the constructor.

odow:
Closing in favor of https://github.com/jump-dev/MathOptInterface.jl/pull/2636

PR #2664, opened 2025-02-26
[Bridges.Variable] change ParameterToEqualToBridge to <: SetMapBridge

odow:
@blegat it doesn't look like a `SetMapBridge{T,<:MOI.AbstractScalarSet}` bridge was ever implemented or tested?

### DIFF
--- a/src/Bridges/Variable/bridges/ParameterToEqualToBridge.jl
+++ b/src/Bridges/Variable/bridges/ParameterToEqualToBridge.jl
@@ -23,121 +23,38 @@
 
  * One variable node: [`MOI.VariableIndex`](@ref) in [`MOI.EqualTo{T}`](@ref)
 """
-struct ParameterToEqualToBridge{T} <: AbstractBridge
-    x::MOI.VariableIndex
-    ci::MOI.ConstraintIndex{MOI.VariableIndex,MOI.EqualTo{T}}
+struct ParameterToEqualToBridge{T} <:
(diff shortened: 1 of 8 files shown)


odow:
Okay, one is implemented here in `test`:
https://github.com/jump-dev/MathOptInterface.jl/blob/3bbca2a14bae37e4cecd8985969ccef06982b66e/test/Bridges/identity_bridge.jl#L12-L32

but it doesn't get properly tested, so we don't hit the relevant methods:

https://github.com/jump-dev/MathOptInterface.jl/blob/3bbca2a14bae37e4cecd8985969ccef06982b66e/test/Bridges/bridge_optimizer.jl#L713-L729

There is, for example, no check for `NumberOfVariables` or `ListOfVariableIndices`.

blegat (approved):

src/Bridges/Variable/bridges/ParameterToEqualToBridge.jl
```diff
@@ -23,121 +23,38 @@
 
  * One variable node: [`MOI.VariableIndex`](@ref) in [`MOI.EqualTo{T}`](@ref)
 """
-struct ParameterToEqualToBridge{T} <: AbstractBridge
-    x::MOI.VariableIndex
-    ci::MOI.ConstraintIndex{MOI.VariableIndex,MOI.EqualTo{T}}
+struct ParameterToEqualToBridge{T} <:
+       SetMapBridge{T,MOI.EqualTo{T},MOI.Parameter{T}}
```
blegat, line 27: This is much better

PR #2666, opened 2025-02-26
[Test] improve test coverage

odow:
We might need to check solver-tests.yml for this one. I don't think it's possible in MOI to return a tuple here if there are 0 constraints.

### DIFF
--- a/src/Bridges/Bridges.jl
+++ b/src/Bridges/Bridges.jl
@@ -178,7 +178,7 @@ function _test_structural_identical(
         Test.@test (F, S) in b_constraint_types || MOI.get(a, attr) == 0
     end
     for (F, S) in b_constraint_types
-        @test haskey(constraints, (F, S))
+        Test.@test haskey(constraints, (F, S))
         # Check that the same number of constraints are present
         attr = MOI.NumberOfConstraints{F,S}()
         Test.@test MOI.get(a, attr) == MOI.get(b, attr)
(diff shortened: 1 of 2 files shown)


src/Bridges/Bridges.jl
```diff
@@ -178,16 +178,10 @@ function _test_structural_identical(
         Test.@test (F, S) in b_constraint_types || MOI.get(a, attr) == 0
     end
     for (F, S) in b_constraint_types
+        @test haskey(constraints, (F, S))
```
odow, line 181:
```suggestion
        Test.@test haskey(constraints, (F, S))
```

src/Bridges/Bridges.jl
```diff
@@ -402,9 +396,7 @@ function _test_delete(Bridge, model, inner)
         MOI.delete.(model, MOI.get(model, MOI.ListOfConstraintIndices{F,S}()))
     end
     #  * So now there should be no constraints in the problem
-    for (F, S) in MOI.get(inner, MOI.ListOfConstraintTypesPresent())
-        Test.@test MOI.get(inner, MOI.NumberOfConstraints{F,S}()) == 0
-    end
+    @test isempty(MOI.get(inner, MOI.ListOfConstraintTypesPresent()))
```
odow, line 399:
```suggestion
    Test.@test isempty(MOI.get(inner, MOI.ListOfConstraintTypesPresent()))
```

PR #2682, opened 2025-03-05
[Utilities] fix operate(vcat for VectorNonlinearFunction

odow:
https://github.com/jump-dev/MathOptInterface.jl/issues/2553 got complicated to debug because of this :smile:

Without this PR, this code:

https://github.com/jump-dev/MathOptInterface.jl/blob/927c58bb2e157813c3d71de754c8bd07374963b6/src/Bridges/Constraint/bridges/NormInfinityBridge.jl#L61-L68

modifies the user's input function.

### DIFF
--- a/src/Utilities/operate.jl
+++ b/src/Utilities/operate.jl
@@ -991,7 +991,11 @@ function operate(
             append!(out, scalarize(a))
         end
     end
-    return MOI.VectorNonlinearFunction(out)
+    # We need to `copy` the rows hre, because if `a` are mutable
+    # AbstractScalarFunction then mutating the return value will mutate the
+    # inputs. This _is_ what Base.vcat does, but it doesn't fit with the general
+    # assumption that `Utilities.operate(` returns a new object.
+    return MOI.VectorNonlinearFunction(copy.(out))
 end
 
 ### 6a: operate(::typeof(imag), ::Type{T}, ::F)
--- a/src/Utilities/operate.jl
+++ b/src/Utilities/operate.jl
@@ -992,7 +992,7 @@ function operate(
         end
     end
     _to_new_snf(f::MOI.ScalarNonlinearFunction) = copy(f)
-    _to_new_snf(f) = convert(MMOI.ScalarNonlinearFunction, f)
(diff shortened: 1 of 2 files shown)


src/Utilities/operate.jl
```diff
@@ -991,7 +991,11 @@ function operate(
             append!(out, scalarize(a))
```
blegat, line 991: You don't need to copy if it's the output of `scalarize`, I would rather copy in the two `push!` above

src/Utilities/operate.jl
```diff
@@ -991,7 +991,11 @@ function operate(
...
         end
     end
-    return MOI.VectorNonlinearFunction(out)
+    # We need to `copy` the rows hre, because if `a` are mutable
+    # AbstractScalarFunction then mutating the return value will mutate the
+    # inputs. This _is_ what Base.vcat does, but it doesn't fit with the general
+    # assumption that `Utilities.operate(` returns a new object.
```
blegat, line 997: Yes `operate` needs you to return something that can be mutated with `operate!`

src/Utilities/operate.jl
odow: > You don't need to copy if it's the output of scalarize

Ah but you do if you hit this method `MOI.Utilities.scalarize(f::MOI.VectorNonlinearFunction) = f.rows`.

The question of whether `scalarize` and `vectorize` are mutable hasn't really come up before, because they always produced unique objects.

src/Utilities/operate.jl
```diff
@@ -991,7 +991,15 @@ function operate(
             append!(out, scalarize(a))
         end
     end
-    return MOI.VectorNonlinearFunction(out)
+    _to_new_snf(f::MOI.ScalarNonlinearFunction) = copy(f)
+    _to_new_snf(f) = convert(MMOI.ScalarNonlinearFunction, f)
```
odow, line 995:
```suggestion
    _to_new_snf(f) = convert(MOI.ScalarNonlinearFunction, f)
```

src/Utilities/operate.jl
blegat: :fearful:

PR #2691, opened 2025-03-11
[Test] update version_added

odow:
Closes https://github.com/jump-dev/MathOptInterface.jl/issues/2689


joaquimg (commented):
If this is something forgetable, we could have a ci check.

odow:
I added a checklist item in #2690

PR #2698, opened 2025-03-26
[docs] improve the docstrings of the various enum values

odow:
In relation to https://discourse.julialang.org/t/approaching-a-linear-program-correctly/127354/12?u=odow

I'll post a preview link once it's up: https://jump.dev/MathOptInterface.jl/previews/PR2698/reference/models/#MathOptInterface.TerminationStatusCode

### DIFF
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -2124,8 +2124,8 @@ struct ListOfSupportedNonlinearOperators <: AbstractOptimizerAttribute end
     """
     The algorithm concluded that no dual bound exists for the problem.
 
-    If the problem is a conic optimization problem, this status means the dual
-    problem is infeasible..
+    If the problem is a conic optimization problem (thus also a linear program), this status means the dual
+    problem is infeasible.
 
     If a primal feasible solution exists, this status typically implies that the
     problem is unbounded, with some technical exceptions (for example, if the
--- a/src/attributes.jl
+++ b/src/attributes.jl
@@ -1435,7 +1435,7 @@ results will typically be alternate solutions that the solver found during the
 search for the first result.
 
 If a (local) optimal solution is available, that is, [`TerminationStatus`](@ref)
-is [`OPTIMAL`](@ref) or [`LOCALLY_SOLVED`](@Ref), the first result must
+is [`OPTIMAL`](@ref) or [`LOCALLY_SOLVED`](@ref), the first result must
 correspond to the (locally) optimal solution. Other results may be alternative
 optimal solutions, or they may be other suboptimal solutions; use
 [`ObjectiveValue`](@ref) to distinguish between them.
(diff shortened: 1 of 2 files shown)


src/attributes.jl
```diff
@@ -2090,64 +2104,169 @@ struct ListOfSupportedNonlinearOperators <: AbstractOptimizerAttribute end
...
+    """
+    The algorithm proved that no primal feasible solution exists.
+    """,
     INFEASIBLE,
-    "The algorithm concluded that no dual bound exists for the problem. If, additionally, a feasible (primal) solution is known to exist, this status typically implies that the problem is unbounded, with some technical exceptions.",
+    """
+    The algorithm concluded that no dual bound exists for the problem.
+
+    If the problem is a conic optimization problem, this status means the dual
+    problem is infeasible..
```
blegat, line 2128:
```suggestion
    problem is infeasible.
```

src/attributes.jl
```diff
@@ -2090,64 +2104,169 @@ struct ListOfSupportedNonlinearOperators <: AbstractOptimizerAttribute end
...
-    "The algorithm concluded that no feasible solution exists.",
+    """
+    The algorithm proved that no primal feasible solution exists.
+    """,
     INFEASIBLE,
-    "The algorithm concluded that no dual bound exists for the problem. If, additionally, a feasible (primal) solution is known to exist, this status typically implies that the problem is unbounded, with some technical exceptions.",
+    """
+    The algorithm concluded that no dual bound exists for the problem.
+
+    If the problem is a conic optimization problem, this status means the dual
```
blegat, line 2127:
```suggestion
    If the problem is a conic optimization problem (thus also a linear program), this status means the dual
```

src/attributes.jl
```diff
@@ -2090,64 +2104,169 @@ struct ListOfSupportedNonlinearOperators <: AbstractOptimizerAttribute end
...
+    problem is infeasible..
+
+    If a primal feasible solution exists, this status typically implies that the
+    problem is unbounded, with some technical exceptions (for example, if the
```
blegat, line 2131: Not sure why we need strong duality. If we have a both a feasible point and a feasible ray of the primal then we can just start at this feasible point and follow the ray and the objective will be infinite no ?

blegat: I think the technical exception mentioned before was just about the existence of the feasible point

odow: What about a problem like:
```
model = Model()
@variable(model, X[1:2, 1:2], Symmetric)
@constraint(model, X >= 0, set = PSDCone())
@constraint(model, X[1,1] == 0)
@objective(model, Min, 2 * X[1, 2])
```
The primal has a finite optimal primal solution with solution `X = 0`. But the dual is infeasible.

PR #2703, opened 2025-03-27
Add support for ForwardDiff@1

odow:
Replaces #2702. (Because I don't know why CI didn't run in that PR...)


franckgaga:
@odow That was cleary the cause for the `!=` vs `!==` bug in the [user-defined operators with vector outputs](https://jump.dev/JuMP.jl/stable/tutorials/nonlinear/tips_and_tricks/#User-defined-operators-with-vector-outputs) tutorial! We can probably left it like that for the retro-compatibility.

Issue #2696, opened 2025-03-25
MOI.is_valid can be falsely positive for bridged constraints


odow:
Found as part of #2695. But it does't need to hold that up merging.

The last `MOI.is_valid` should return `false`:

```Julia
julia> import MathOptInterface as MOI

julia> include("test/Bridges/sdpa_models.jl")

julia> model = MOI.instantiate(StandardSDPAModel{Float64}; with_bridge_type = Float64);

julia> x = MOI.add_variables(model, 2);

julia> c = MOI.add_constraint(model, MOI.VectorOfVariables(x), MOI.Nonpositives(2));

julia> MOI.is_valid(model, c)
true

julia> MOI.is_valid(model, typeof(c)(c.value + 1))
true
```

odow:
```Julia
julia> using Revise

julia> import MathOptInterface as MOI

julia> MOI.Utilities.@model(
           Model,
           (), (MOI.EqualTo,), (MOI.Nonnegatives,), (),
           (), (MOI.ScalarAffineFunction,), (MOI.VectorOfVariables,), (),
       )
MathOptInterface.Utilities.GenericModel{T, MathOptInterface.Utilities.ObjectiveContainer{T}, MathOptInterface.Utilities.VariablesContainer{T}, ModelFunctionConstraints{T}} where T

julia> MOI.supports_constraint(::Model, ::Type{MOI.VectorOfVariables}, ::Type{MOI.Reals}) = false

julia> MOI.supports_add_constrained_variables(::Model, ::Type{MOI.Nonnegatives}) = true

julia> MOI.supports_add_constrained_variables(::Model, ::Type{MOI.Reals}) = false

julia> model = MOI.instantiate(Model{Float64}; with_bridge_type = Float64);

julia> x = MOI.add_variables(model, 2);

julia> c = MOI.add_constraint(model, MOI.VectorOfVariables(x), MOI.Nonpositives(2));

julia> MOI.is_valid(model, c)
true

julia> ci = typeof(c)(c.value + 1)
MathOptInterface.ConstraintIndex{MathOptInterface.VectorOfVariables, MathOptInterface.Nonpositives}(-2)

julia> MOI.is_valid(model, ci)
true
```

PR #2705, opened 2025-04-02
[Bridges] add a test_broken test for #2696

odow:
@blegat I don't understand https://github.com/jump-dev/MathOptInterface.jl/issues/2696.

As far as I can tell, the only issue is that we may return a false positive (`is_valid -> true`) if there is a `VectorOfVariables` constraint on a vector of variables that were bridged, and the bridge happened to add a constraint with the same index.


blegat:
I'll take a look

odow:
I'll add this test in case we make changes in the future that fix it, but I'll leave the issue open for you to take a look.

PR #2708, opened 2025-04-03
[Nonlinear] fix _UnsafeVectorView with ForwardDiff@1.0.0

odow:
ForwardDiff@1 changed to use `CartesianIndex`, which we hadn't supported.

We also never tested this, because we tested only that the operator could be called, but never end-to-end through the ReverseAD system. Nasty.

https://github.com/jump-dev/MathOptInterface.jl/actions/runs/14232574450


odow:
https://github.com/jump-dev/MathOptInterface.jl/actions/runs/14233113304

odow:
There's a surprising amount of nuance in this.

PR #2710, opened 2025-04-04
Allow using Test.runtests in other packages

blegat:
Would be useful for https://github.com/blegat/LowRankOpt.jl/pull/6 for instance

### DIFF
--- a/src/Test/Test.jl
+++ b/src/Test/Test.jl
@@ -173,6 +173,7 @@ version_added(::F) where {F} = v"0.10.5"  # The default for any unlabeled tests.
         warn_unsupported::Bool = false,
         exclude_tests_after::VersionNumber = v"999.0.0",
         verbose::Bool = false,
+        test_module = @__MODULE__,
     )
 
 Run all tests in `MathOptInterface.Test` on `model`.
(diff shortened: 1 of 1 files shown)


src/Test/Test.jl
```diff
@@ -173,6 +173,7 @@ version_added(::F) where {F} = v"0.10.5"  # The default for any unlabeled tests.
         warn_unsupported::Bool = false,
         exclude_tests_after::VersionNumber = v"999.0.0",
         verbose::Bool = false,
+        test_module = @__MODULE__,
     )
 
 Run all tests in `MathOptInterface.Test` on `model`.
```
odow, line 179: This needs updating

odow:
I'm not necessarily for or against this. One massive downside is that we'd be committing to the API Innside this function. I don't know that it's worth it to save a single package (LowRankOpt) copy-pasting a single function. Especially if the long-term goal is to merge some form of LowRankOpt back into MOI.

blegat:
I was wondering about how much more we're committing to if the API (name of the function, name of the keywords, behavior) is already public anyway. What would be breaking for a use of `test_module` that wouldn't be also breaking now anyway ?

odow (approved):
I'll merge this if https://github.com/jump-dev/MathOptInterface.jl/actions/runs/14327216427 looks okay

PR #2711, opened 2025-04-06
Sync Indicator docstring with implementation

blegat:


odow:
Shouldn't we fix the set instead? This looks like a mistake in the set definition, not a typo in the docstring.

blegat:
But it's not a subtype of the enum type, it's a value of it so I don't know how to make it work

odow:
Closing in favor of #2712

PR #2713, opened 2025-04-07
Fix Bridge test for non-invertible constraint bridge

blegat:
This is needed for https://github.com/blegat/LowRankOpt.jl/pull/10. The getters for ConstraintFunction and ConstrainPrimalStart are not so important as there is often a cache so bridges corresponding to non-invertible transformations are fine and being able to use these tests for these bridges would be useful.

### DIFF
--- a/src/Bridges/Bridges.jl
+++ b/src/Bridges/Bridges.jl
@@ -251,9 +251,11 @@ Run a series of tests that check the correctness of `Bridge`.
 `input_fn` and `output_fn` are functions such that `input_fn(model)`
 and `output_fn(model)` load the corresponding model into `model`.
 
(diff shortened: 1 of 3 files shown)


src/Bridges/Bridges.jl
```diff
@@ -328,7 +329,19 @@ function runtests(
...
+                        MOI.get(model, attr, ci)
+                    catch err
+                        # For a Constraint bridge for which the map is not invertible, the constraint primal cannot
+                        # be inverted
+                        _runtests_error_handler(
+                            err,
+                            Bridge <: MOI.Bridges.Constraint.AbstractBridge &&
+                                cannot_unbridge,
+                        )
+                        continue
```
odow, line 342: See the coverage. By my new rule, you must add a test case that catches these lines.

PR #2718, opened 2025-04-09
[FileFormats.NL] add support for reading binary format

odow:
Closes https://github.com/jump-dev/MathOptInterface.jl/issues/2717

<s>I still need to add much better tests for this</s>


odow:
I've pushed another commit, but it hasn't shown up. I just see:

image

PR #2721, opened 2025-04-10
Add more packages to solver-tests.yml

odow:
https://github.com/jump-dev/MathOptInterface.jl/actions/runs/14391068832


odow:
I have been fixing a number of downstream packages over the past few days.

I've submitted a fix for Clarabel: https://github.com/oxfordcontrol/Clarabel.jl/pull/198

So then we just have the SumOfSquares/PolyJuMP issue to fix and all tests are back to being green.

PR #2724, opened 2025-04-15
Add bridge from Hermitian PSD to complex function in Symmetric PSD

blegat:
See https://github.com/jump-dev/SeDuMi.jl/pull/28#issuecomment-2804975387

## Basic

 - [x] Create a new file in `src/Bridges/XXX/bridges` named after the type of
       the bridge
       struct must end in `Bridge`
       the `Bridge` suffix

## Tests

       named after the type of the bridge
       bridge
       untested lines.

## Documentation

## Final touch

the model:

 - [x] Implement `MOI.Utilities.needs_final_touch(::Bridge)`
 - [x] Implement `MOI.Utilities.final_touch(::Bridge, ::MOI.ModelLike)`
 - [x] Ensure that `final_touch` can be called multiple times in a row


odow:
What's the reason not to define the `SingleBridgeOptimizer`?

odow (approved):
I haven't looked in detail, but if you're happy I'm happy.

PR #2726, opened 2025-04-15
Add testset to Bridges.runtests

blegat:
It's helpful when debugging


odow:
I'll take over and fix this. It annoys me all the time, but I never thought to do this.

PR #2730, opened 2025-04-20
Simplify _eval_hessian_inner

blegat:
A large of code was almost copy-pasted so I think a sub-function is a win for readability.
Doing the `_reinterpret` magic inside `_hessian_slice_inner` along with the other `_reinterpret` calls also allows to simplify logic quite a bit by keeping these contained inside this function only.

### DIFF
--- a/src/Nonlinear/ReverseAD/forward_over_reverse.jl
+++ b/src/Nonlinear/ReverseAD/forward_over_reverse.jl
@@ -61,7 +61,13 @@ function _eval_hessian_inner(
     # leftover chunk
     remaining = num_products - CHUNK * num_chunks
     if remaining > 0
-        _eval_hessian_chunk(d, ex, CHUNK * num_chunks + 1, remaining, Val(CHUNK))
+        _eval_hessian_chunk(
+            d,
+            ex,
+            CHUNK * num_chunks + 1,
+            remaining,
+            Val(CHUNK),
+        )
     end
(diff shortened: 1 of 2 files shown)


src/Nonlinear/ReverseAD/forward_over_reverse.jl
```diff
@@ -127,7 +85,40 @@ function _eval_hessian_inner(
...
+    chunk::Int,
+    ::Val{CHUNK},
+) where {CHUNK}
+    for r in eachindex(ex.rinfo.local_indices)
+        # set up directional derivatives
+        @inbounds idx = ex.rinfo.local_indices[r]
+        # load up ex.seed_matrix[r,k,k+1,...,k+remaining-1] into input_ϵ
+        for s in 1:chunk
+            # If `chunk < CHUNK`, leaves junk in the unused components
+            d.input_ϵ[(idx-1)*CHUNK+s] = ex.seed_matrix[r, offset+s-1]
```
blegat, line 101: @odow Any reason why we don't use `@inbounds` here ?

odow: I would be in favor of removing most `@inbounds`. They made more of a difference in earlier Julia versions.

blegat: Makes sense, that would be a different PR then anyway

PR #2732, opened 2025-04-21
[Nonlinear] Merge forward_storage_ϵ with reverse_storage_ϵ

blegat:
They seem to only be used internally by `_forward_eval_ϵ` and `reverse_eval_ϵ` so the same vector is sufficient for both forward and reverse as well as for all subexpressions and expressions

### DIFF
--- a/src/Nonlinear/ReverseAD/mathoptinterface_api.jl
+++ b/src/Nonlinear/ReverseAD/mathoptinterface_api.jl
@@ -138,10 +138,10 @@ function MOI.initialize(d::NLPEvaluator, requested_features::Vector{Symbol})
     end
     # 10 is hardcoded upper bound to avoid excess memory allocation
     max_chunk = min(max_chunk, 10)
-    max_expr_with_sub_length =
-        maximum(d.subexpression_order; init = max_expr_length) do k
-            return length(d.subexpressions[k].nodes)
-        end
+    max_expr_with_sub_length = max(
+        max_expr_length,
+        maximum(x -> length(x.nodes), d.subexpressions; init = 0),
+    )
     if d.want_hess || want_hess_storage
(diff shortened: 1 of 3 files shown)


src/Nonlinear/ReverseAD/mathoptinterface_api.jl
```diff
@@ -138,27 +138,25 @@ function MOI.initialize(d::NLPEvaluator, requested_features::Vector{Symbol})
     end
     # 10 is hardcoded upper bound to avoid excess memory allocation
     max_chunk = min(max_chunk, 10)
+    max_expr_with_sub_length =
+        maximum(d.subexpression_order; init = max_expr_length) do k
```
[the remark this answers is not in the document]
blegat: Oh what a strange choice, I never noticed

src/Nonlinear/ReverseAD/mathoptinterface_api.jl
```diff
@@ -138,27 +138,25 @@ function MOI.initialize(d::NLPEvaluator, requested_features::Vector{Symbol})
     end
     # 10 is hardcoded upper bound to avoid excess memory allocation
     max_chunk = min(max_chunk, 10)
+    max_expr_with_sub_length = max(
+        max_expr_length,
+        maximum(x -> length(x.nodes), d.subexpressions; init = 0),
```
blegat, line 143: It's `undef` if the expression is not used in any constraint or the objective

blegat: That's why I did [707ab59](https://github.com/jump-dev/MathOptInterface.jl/pull/2732/commits/707ab59aef5f1775fa725c735e1927a90eec2834)

PR #2736, opened 2025-04-25
Remove arguments of _forward_eval_ϵ

blegat:
They are always the same anyway. I wonder if there is any performance benefit of saving these interpret calls. Is there a standard benchmark used for checking for regressions ?
Maybe https://github.com/lanl-ansi/rosetta-opf/blob/main/jump.jl ?

### DIFF
--- a/src/Nonlinear/ReverseAD/forward_over_reverse.jl
+++ b/src/Nonlinear/ReverseAD/forward_over_reverse.jl
@@ -126,22 +126,10 @@ function _hessian_slice_inner(d, ex, ::Val{CHUNK}) where {CHUNK}
         subexpr_forward_values_ϵ[i] = _forward_eval_ϵ(
             d,
             subexpr,
-            _reinterpret_unsafe(T, d.storage_ϵ),
             _reinterpret_unsafe(T, subexpr.partials_storage_ϵ),
-            input_ϵ,
-            subexpr_forward_values_ϵ,
-            d.data.operators,
         )
     end
-    _forward_eval_ϵ(
-        d,
-        ex,
-        _reinterpret_unsafe(T, d.storage_ϵ),
-        _reinterpret_unsafe(T, d.partials_storage_ϵ),
-        input_ϵ,
-        subexpr_forward_values_ϵ,
-        d.data.operators,
-    )
+    _forward_eval_ϵ(d, ex, _reinterpret_unsafe(T, d.partials_storage_ϵ))
     # do a reverse pass
     subexpr_reverse_values_ϵ =
         _reinterpret_unsafe(T, d.subexpression_reverse_values_ϵ)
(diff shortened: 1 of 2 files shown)


odow:
Yes that, with a much larger instance, say https://github.com/power-grid-lib/pglib-opf/blob/master/pglib_opf_case30000_goc.m

odow:
Or build that model and then manually evaluate the Hessian with `MOI.eval_hessian_lagrangian` etc.

odow:
When I refactored this out of JuMP, I tried to make the fewest possible changes. The code can be cleaned up lots.

blegat:
> Or build that model and then manually evaluate the Hessian with `MOI.eval_hessian_lagrangian` etc.

Yes, we could add a reduced version to `MOI.Benchmarks`

odow:
Using
```
using Revise
import Ipopt
import JuMP
import PGLib
import PowerModels

model = JuMP.direct_model(Ipopt.Optimizer())
pm = PowerModels.instantiate_model(
    PGLib.pglib("pglib_opf_case10000_goc"),
    PowerModels.ACPPowerModel,
    PowerModels.build_opf;
    jump_model = model,
);

ipopt = JuMP.backend(model)
x = MOI.get(ipopt, MOI.ListOfVariableIndices())
m, n = length(ipopt.nlp_model.constraints), length(x)

evaluator = MOI.Nonlinear.Evaluator(
    ipopt.nlp_model,
    MOI.Nonlinear.SparseReverseMode(),
    x,
)
MOI.initialize(evaluator, [:Grad, :Jac, :Hess])

H_struct = MOI.hessian_lagrangian_structure(evaluator)
H = zeros(length(H_struct))
mu = rand(m)
sigma = 0.0
x_v = rand(n)
@time MOI.eval_hessian_lagrangian(evaluator, H, x_v, sigma, mu)
```

Before
```julia
julia> x_v = rand(n);

julia> @time MOI.eval_hessian_lagrangian(evaluator, H, x_v, sigma, mu)
  6.617456 seconds (158.26 k allocations: 11.273 MiB)

julia> x_v = rand(n);

julia> @time MOI.eval_hessian_lagrangian(evaluator, H, x_v, sigma, mu)
  6.661574 seconds (158.26 k allocations: 11.273 MiB)
```

After 

```julia
julia> x_v = rand(n);

julia> @time MOI.eval_hessian_lagrangian(evaluator, H, x_v, sigma, mu)
  6.644427 seconds (158.26 k allocations: 11.273 MiB)

julia> x_v = rand(n);

julia> @time MOI.eval_hessian_lagrangian(evaluator, H, x_v, sigma, mu)
  6.850976 seconds (158.26 k allocations: 11.273 MiB)
```

Seems okay to me. No difference.

PR #2738, opened 2025-04-25
Disable linearity detection for subexpression if no hessian

blegat:
This was done for `_FunctionStorage` but not here

### DIFF
--- a/test/Nonlinear/ReverseAD.jl
+++ b/test/Nonlinear/ReverseAD.jl
@@ -639,7 +639,8 @@ function test_linearity_no_hess()
     MOI.initialize(evaluator, [:Grad, :Jac])
     # We initialized without the need for the hessian so
     # the linearity shouldn't be computed.
-    @test only(evaluator.backend.subexpressions).linearity == ReverseAD.NONLINEAR
+    @test only(evaluator.backend.subexpressions).linearity ==
+          ReverseAD.NONLINEAR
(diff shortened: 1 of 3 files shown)


test/Nonlinear/ReverseAD.jl
```diff
@@ -630,6 +630,19 @@ function test_linearity()
...
+    x = MOI.VariableIndex(1)
+    model = Nonlinear.Model()
+    ex = Nonlinear.add_expression(model, :($x + 1))
+    Nonlinear.set_objective(model, ex)
+    evaluator = Nonlinear.Evaluator(model, Nonlinear.SparseReverseMode(), [x])
+    MOI.initialize(evaluator, [:Grad, :Jac])
+    # We initialized without the need for the hessian so
+    # the linearity shouldn't be computed.
+    @test only(evaluator.backend.subexpressions).linearity ==
+          ReverseAD.NONLINEAR
```
odow, line 643:
```suggestion
          ReverseAD.NONLINEAR
    return
```

odow (approved):
So this is okay. But more generally, we should consider:

 * Is it worth making code changes for edge cases?

This requires Hessians to be disabled AND the model to have many subexpressions AND the linearity check to take a non-trivial amount of time.

I'd be in favor of fewer special cases, even if it meant a small decrease in performance.

PR #2740, opened 2025-04-28
Remove dynamic dispatch in Hessian evaluation

blegat:
As `chunk` is maximum 10, we can reuse the `_create_binary_switch` we use for type-stable evaluation of our list of operators to call `_eval_hessian_inner` in a type-stable way.
On this benchmark https://github.com/jump-dev/MathOptInterface.jl/pull/2739#issuecomment-2833683884
Before this PR
```
5.959426 seconds (158.26 k allocations: 11.273 MiB)
5.973144 seconds (158.26 k allocations: 11.273 MiB)
6.078342 seconds (158.26 k allocations: 11.273 MiB)
6.031799 seconds (158.26 k allocations: 11.273 MiB, 0.23% gc time)
6.022742 seconds (158.26 k allocations: 11.273 MiB)
6.013003 seconds (158.26 k allocations: 11.273 MiB)
6.021324 seconds (158.26 k allocations: 11.273 MiB)
```
After this PR:
```
5.925775 seconds
5.820862 seconds
5.865539 seconds
5.823728 seconds
5.848671 seconds
5.864283 seconds
5.876498 seconds
5.911389 seconds
5.894025 seconds
```

### DIFF
--- a/src/Nonlinear/ReverseAD/forward_over_reverse.jl
+++ b/src/Nonlinear/ReverseAD/forward_over_reverse.jl
@@ -108,31 +108,16 @@ function _eval_hessian_chunk(
 end
 
 # A wrapper function to avoid dynamic dispatch.
-function _hessian_slice_inner(d, ex, chunk::Int)
-    @assert 1 <= chunk <= MAX_CHUNK
-    @assert MAX_CHUNK == 10
-    if chunk == 1
-        _hessian_slice_inner(d, ex, ForwardDiff.Partials{1,Float64})
-    elseif chunk == 2
-        _hessian_slice_inner(d, ex, ForwardDiff.Partials{2,Float64})
-    elseif chunk == 3
-        _hessian_slice_inner(d, ex, ForwardDiff.Partials{3,Float64})
-    elseif chunk == 4
-        _hessian_slice_inner(d, ex, ForwardDiff.Partials{4,Float64})
-    elseif chunk == 5
-        _hessian_slice_inner(d, ex, ForwardDiff.Partials{5,Float64})
-    elseif chunk == 6
-        _hessian_slice_inner(d, ex, ForwardDiff.Partials{6,Float64})
-    elseif chunk == 7
-        _hessian_slice_inner(d, ex, ForwardDiff.Partials{7,Float64})
-    elseif chunk == 8
-        _hessian_slice_inner(d, ex, ForwardDiff.Partials{8,Float64})
-    elseif chunk == 9
-        _hessian_slice_inner(d, ex, ForwardDiff.Partials{9,Float64})
-    else
-        _hessian_slice_inner(d, ex, ForwardDiff.Partials{10,Float64})
+function _generate_hessian_slice_inner()
+    exprs = map(1:MAX_CHUNK) do id
+        return :(_hessian_slice_inner(d, ex, ForwardDiff.Partials{$id,Float64}))
     end
-    return
+    return _create_binary_switch(1:MAX_CHUNK, exprs)
+end
+
+@eval function _hessian_slice_inner(d, ex, id::Int)
+    $(_generate_hessian_slice_inner())
+    return error("Invalid chunk size: $id")
 end
 
 function _hessian_slice_inner(d, ex, ::Type{T}) where {T}
--- a/src/Nonlinear/ReverseAD/forward_over_reverse.jl
+++ b/src/Nonlinear/ReverseAD/forward_over_reverse.jl
@@ -6,6 +6,13 @@
 
 const TAG = :ReverseAD
 
+"""
+    const MAX_CHUNK::Int = 10
+
+An upper bound on the chunk sie for forward-over-reverse. Increasing this could
+improve performance at the cost of extra memory allocation. It has been 10 for a
+long time, and nobody seems to have complained.
+"""
 const MAX_CHUNK = 10
 
 """
--- a/src/Nonlinear/ReverseAD/forward_over_reverse.jl
+++ b/src/Nonlinear/ReverseAD/forward_over_reverse.jl
@@ -6,35 +6,41 @@
 
 const TAG = :ReverseAD
 
-"""
-    _eval_hessian(
+function _generate_eval_hessian()
+    exprs = map(1:10) do chunk
+        return :(return _eval_hessian_inner(d, f, H, λ, offset, Val($chunk)))
+    end
+    return Nonlinear._create_binary_switch(1:10, exprs)
+end
+
+@eval begin
+    """
+        _eval_hessian(
+            d::NLPEvaluator,
+            f::_FunctionStorage,
+            H::AbstractVector{Float64},
+            λ::Float64,
+            offset::Int,
+        )::Int
+
+    Evaluate the hessian matrix of the function `f` and store the result, scaled by
+    `λ`, in `H`, beginning at element `offset+1`. This function assumes that
+    `_reverse_mode(d, x)` has already been called.
+
+    Returns the number of non-zeros in the computed Hessian, which will be used to
+    update the offset for the next call.
+    """
+    function _eval_hessian(
         d::NLPEvaluator,
         f::_FunctionStorage,
         H::AbstractVector{Float64},
         λ::Float64,
         offset::Int,
     )::Int
-
-Evaluate the hessian matrix of the function `f` and store the result, scaled by
-`λ`, in `H`, beginning at element `offset+1`. This function assumes that
-`_reverse_mode(d, x)` has already been called.
-
-Returns the number of non-zeros in the computed Hessian, which will be used to
-update the offset for the next call.
-"""
-function _eval_hessian(
-    d::NLPEvaluator,
-    f::_FunctionStorage,
-    H::AbstractVector{Float64},
-    λ::Float64,
-    offset::Int,
-)::Int
-    chunk = min(size(f.seed_matrix, 2), d.max_chunk)
-    # As a performance optimization, skip dynamic dispatch if the chunk is 1.
-    if chunk == 1
-        return _eval_hessian_inner(d, f, H, λ, offset, Val(1))
-    else
-        return _eval_hessian_inner(d, f, H, λ, offset, Val(chunk))
+        id = min(size(f.seed_matrix, 2), d.max_chunk)
+        # As a performance optimization, skip dynamic dispatch if the chunk is 1.
+        $(_generate_eval_hessian())
+        error("Invalid chunk `$id`. Please report this.")
     end
 end
 
--- a/src/Nonlinear/ReverseAD/forward_over_reverse.jl
+++ b/src/Nonlinear/ReverseAD/forward_over_reverse.jl
@@ -112,7 +112,7 @@ function _generate_hessian_slice_inner()
     exprs = map(1:MAX_CHUNK) do id
         return :(_hessian_slice_inner(d, ex, ForwardDiff.Partials{$id,Float64}))
     end
-    return _create_binary_switch(1:MAX_CHUNK, exprs)
+    return MOI.Nonlinear._create_binary_switch(1:MAX_CHUNK, exprs)
 end
 
 @eval function _hessian_slice_inner(d, ex, id::Int)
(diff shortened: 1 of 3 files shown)


odow:
I'm very very very hesitant about this. At minimum, I think we should manually write out the switch. Macros and `@eval` are just too complicated and don't justify the extra 20 lines of code. It's a different story when there are 70 something operators.

It would also mean we need tests for every possible chunk size to ensure 100% code coverage.

odow:
It's nice, there are already tests that check all of the different chunk sizes.

src/Nonlinear/ReverseAD/forward_over_reverse.jl
```diff
@@ -90,32 +83,59 @@ function _eval_hessian_chunk(
...
+        _hessian_slice_inner(d, ex, ForwardDiff.Partials{5,Float64})
+    elseif chunk == 6
+        _hessian_slice_inner(d, ex, ForwardDiff.Partials{6,Float64})
+    elseif chunk == 7
+        _hessian_slice_inner(d, ex, ForwardDiff.Partials{7,Float64})
+    elseif chunk == 8
+        _hessian_slice_inner(d, ex, ForwardDiff.Partials{8,Float64})
+    elseif chunk == 9
+        _hessian_slice_inner(d, ex, ForwardDiff.Partials{9,Float64})
+    else
```
joaquimg, line 132: Isn't 10 the most frequent value? Should it be the first?

odow: I don't think it really matters. The branch predictor can probably figure it out if we end up calling this method multiple times.

```julia
const MAX_CHUNK = 10

function foo_inner(::Type{Array{Int,N}}) where {N}
    return N^2
end

function foo(chunk::Int)
    @assert 1 <= chunk <= MAX_CHUNK
    @assert MAX_CHUNK == 10
    return if chunk == 1
        foo_inner(Array{Int,1})
    elseif chunk == 2
        foo_inner(Array{Int,2})
    elseif chunk == 3
        foo_inner(Array{Int,3})
    elseif chunk == 4
        foo_inner(Array{Int,4})
    elseif chunk == 5
        foo_inner(Array{Int,5})
    elseif chunk == 6
        foo_inner(Array{Int,6})
    elseif chunk == 7
        foo_inner(Array{Int,7})
    elseif chunk == 8
        foo_inner(Array{Int,8})
    elseif chunk == 9
        foo_inner(Array{Int,9})
    else
        foo_inner(Array{Int,10})
    end
end

struct Foo
    x::Int
end

main(f::Foo) = foo(f.x)

using BenchmarkTools
```

```Julia
julia> for i in 1:10
           f = Foo(i)
           display(@benchmark main($f))
       end
BenchmarkTools.Trial: 10000 samples with 1000 evaluations per sample.
 Range (min … max):  3.470 ns … 30.515 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     3.574 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   3.948 ns ±  1.046 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  █▇  ▄▁▁  ▃   ▁▂   ▃▃   ▂▄▁   ▃▃    ▄▁    ▅▄▁    ▄▂     ▅▅▃ ▂
  ██▃▁███▁███▇███▇▄▁██▇▄▁███▄▁▅██▅▃▁▇██▆▁▁▁███▁▁▁▆██▆▃▁▁▁███ █
  3.47 ns      Histogram: log(frequency) by time     4.77 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.
BenchmarkTools.Trial: 10000 samples with 1000 evaluations per sample.
 Range (min … max):  3.470 ns … 23.178 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     3.496 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   3.859 ns ±  0.968 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  █▅  ▂▂▁  ▂   ▁▃   ▁▂    ▃▁   ▂▃    ▂▄▁    ▄▃    ▁▄▂     ▄▃ ▂
  ██▅▁███▃▇██▅████▁▄██▇▁▁▇██▆▄▁███▅▁▁███▅▁▁███▆▁▁▁███▅▁▁▁▁██ █
  3.47 ns      Histogram: log(frequency) by time     4.75 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.
BenchmarkTools.Trial: 10000 samples with 1000 evaluations per sample.
 Range (min … max):  3.469 ns … 47.020 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     3.501 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   3.869 ns ±  1.194 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  █▆  ▄▃▁  ▄▂  ▁▂   ▂▃    ▂    ▃▄    ▃▂    ▂▄▂    ▂▄▂     ▄▃ ▂
  ██▄▁███▁▇██▆███▇▅▃███▅▄▇██▃▃▁███▁▁▁██▆▃▁▁███▄▁▁▁███▄▁▁▄▆██ █
  3.47 ns      Histogram: log(frequency) by time     4.76 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.
BenchmarkTools.Trial: 10000 samples with 1000 evaluations per sample.
 Range (min … max):  3.463 ns … 90.157 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     3.495 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   3.834 ns ±  1.315 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  █▆  ▂▄▁  ▄▄   ▂    ▃▁   ▃▂   ▁▃▁    ▄▃    ▅▃     ▄▃      ▂ ▂
  ███▁███▅▁██▆▇▇██▅▅███▄▄▁██▆▁▁███▄▁▁███▅▁▁▁██▅▁▁▁▁███▄▁▁▁██ █
  3.46 ns      Histogram: log(frequency) by time     4.75 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.
BenchmarkTools.Trial: 10000 samples with 1000 evaluations per sample.
 Range (min … max):  3.880 ns … 58.926 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     4.553 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   4.542 ns ±  0.957 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

                    ▁      ▅▂     ▂       █▁                  
  ▃▃▂▁▁▃▄▃▂▁▁▃▃▃▂▂▁▂█▄▂▂▂▂▃██▃▂▂▂▂█▇▃▃▂▂▂▅██▃▂▂▂▂▂▂▂▂▁▁▁▁▁▂▂ ▃
  3.88 ns        Histogram: frequency by time        5.09 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.
BenchmarkTools.Trial: 10000 samples with 1000 evaluations per sample.
 Range (min … max):  3.462 ns … 39.041 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     3.567 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   3.922 ns ±  1.637 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  █                                                           
  █▅▂▃▃▂▃▂▃▃▂▃▂▃▃▁▃▁▃▂▁▂▁▂▂▂▁▂▂▁▁▁▂▁▂▁▁▂▁▁▂▁▁▂▁▁▂▁▁▂▁▂▂▂▂▂▂▂ ▂
  3.46 ns        Histogram: frequency by time        7.38 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.
BenchmarkTools.Trial: 10000 samples with 1000 evaluations per sample.
 Range (min … max):  3.469 ns … 22.645 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     3.498 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   3.797 ns ±  0.785 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  █▆  ▁▂   ▄▃ ▁ ▂    ▂    ▄▃    ▃▂    ▃▃    ▃▃▁    ▄▃▁    ▁▃ ▂
  ██▇▁███▄▇██▅███▇▁▄███▅▄▁██▆▁▁▇██▅▁▁▇██▅▁▁▁███▃▁▁▁███▁▁▁▁██ █
  3.47 ns      Histogram: log(frequency) by time     4.74 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.
BenchmarkTools.Trial: 10000 samples with 1000 evaluations per sample.
 Range (min … max):  3.469 ns … 57.400 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     3.495 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   3.788 ns ±  1.015 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  █▆▁  ▂▂   ▃▄    ▂▁   ▄     ▂▃     ▃▂    ▂▃▃     ▄▃     ▁▄▃ ▂
  ███▁▃███▄▁███▇▇███▅▁▃██▅▅▆▁███▅▄▁███▅▁▁▁███▅▁▁▁████▃▁▁▁███ █
  3.47 ns      Histogram: log(frequency) by time     4.59 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.
BenchmarkTools.Trial: 10000 samples with 1000 evaluations per sample.
 Range (min … max):  3.464 ns … 24.399 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     3.483 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   3.749 ns ±  0.745 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  █▅   ▃    ▄▃    ▂    ▁▃     ▃     ▃▂     ▃▃      ▄      ▁▄ ▂
  ██▁▁▁██▃▁▁██▅▁▁▇█▄▁▁▃██▆▃▁▁██▆▁▁▁▃██▃▁▁▁▁██▁▁▁▁▁▇██▁▁▁▁▁██ █
  3.46 ns      Histogram: log(frequency) by time     4.57 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.
BenchmarkTools.Trial: 10000 samples with 1000 evaluations per sample.
 Range (min … max):  3.464 ns … 47.211 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     3.574 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   3.885 ns ±  1.219 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  █                                                           
  █▃▁▁▇▅▂▁▂▄▂▂▁▂▃▁▁▁▂▃▁▂▁▁▄▂▂▁▁▂▃▂▂▁▁▂▃▂▁▁▁▁▃▄▂▂▁▁▂▃▃▂▁▁▁▁▂▄ ▂
  3.46 ns        Histogram: frequency by time        4.73 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.
```

joaquimg: good to know

odow: At minimum, it's more efficient than dynamic dispatch.

blegat:
> It would also mean we need tests for every possible chunk size to ensure 100% code coverage.

I don't agree. We're using a binary switch that is already heavily tested, what's the risk ?
If you hardcode 10 different `if-else` then you might make a typo but not with the binary switch.
Also, the advantage of the switch is that you can modify that constant without the need to modify the code

odow:
I've changed to use the binary switch. I wouldn't say that it is "heavily" tested. It's used in one other place.

> If you hardcode 10 different if-else then you might make a typo but not with the binary switch.

Not if we have every line tested :smile:

blegat:
> I wouldn't say that it is "heavily" tested. It's used in one other place.

Maybe that was an over-exageration :smile: But if the binary switch had a bug I prefer MOI to be completely broken rather than having a hidden bug. In that sense, giving it even more coverage is good :)

PR #2744, opened 2025-05-01
[Nonlinear.ReverseAD] Fix use of reinterpret instead _reinterpret_unsafe

odow:
Introduced by https://github.com/jump-dev/MathOptInterface.jl/pull/2736, specifically:

https://github.com/jump-dev/MathOptInterface.jl/pull/2736/files#diff-fd2e1ad4a6c5ea5294c6e338afffbb76c837b6c4a868d949007a0229ad846ec0R187

image

I found the test case by taking the failed Juniper run, and reducing.


odow:
https://github.com/jump-dev/MathOptInterface.jl/actions/runs/14768059895

blegat:
I copy-pasted from here but I guess it's correct because we use a chunk of on there

PR #2745, opened 2025-05-01
Add docstring to _reinterpret_unsafe

blegat:


docs/make.jl
```diff
@@ -123,6 +123,13 @@ end
 # link checking. Inn production we replace this by running the LaTeX build.
 write(joinpath(@__DIR__, "src", "MathOptInterface.pdf"), "")
 
+DocMeta.setdocmeta!(
+    MathOptInterface,
+    :DocTestSetup,
+    :(import MathOptInterface as MOI);
```
blegat, line 129: I have to add it here because `_reinterpret_unsafe` is not added on any page.

blegat: Maybe we can remove it from all the pages now by the way

odow: This should be a separate PR. Keep `import MathOptInterface as MOI` in the docstring for now.

PR #2750, opened 2025-05-09
Implement adjacency_matrix and expression in _FunctionStorage

blegat:
EAGO is accessing the `objective` and `constraints` fields of `ReverseAD.NLPEvaluator` which is ok since it's not starting with a `_`.
But then it is accessing internal fields of `_FunctionStorage` and `_SubexpressionStorage` which it shouldn't since these start with an `_`.
Actually, all they need are the `nodes` and `const_values` which are part of `MOI.Nonlinear.Expression` and the adjacency matrix so by implementing `adjacency_matrix` and adding a `expression` function that it implements, it allows EAGO to stop using internal fields and stick to the public API.

See the corresponding EAGO PR here : https://github.com/PSORLab/EAGO.jl/pull/147

### DIFF
--- a/src/Nonlinear/ReverseAD/types.jl
+++ b/src/Nonlinear/ReverseAD/types.jl
@@ -43,10 +43,7 @@ struct _SubexpressionStorage
 end
 
 function MOI.Nonlinear.expression(expr::_SubexpressionStorage)
-    return MOI.Nonlinear.Expression(
-        expr.nodes,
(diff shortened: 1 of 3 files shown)


odow:
I'd prefer we just closed #2747

odow:
Close this as well?

blegat:
Yes, let's close

PR #2753, opened 2025-05-10
Use throw_if_scalar_and_constant_not_zero when appropriate

blegat:
I think these lines were written before that helper existed

### DIFF
--- a/src/Utilities/mockoptimizer.jl
+++ b/src/Utilities/mockoptimizer.jl
@@ -182,7 +182,7 @@ function MOI.add_constraint(
     if !mock.add_con_allowed
         throw(MOI.AddConstraintNotAllowed{typeof(func),typeof(set)}())
     elseif mock.scalar_function_constant_non_zero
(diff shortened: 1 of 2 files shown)


src/Utilities/mockoptimizer.jl
```diff
@@ -181,12 +181,8 @@ function MOI.add_constraint(
...
     if !mock.add_con_allowed
         throw(MOI.AddConstraintNotAllowed{typeof(func),typeof(set)}())
-    elseif mock.scalar_function_constant_non_zero && !iszero(func.constant)
-        throw(
-            MOI.ScalarFunctionConstantNotZero{T,typeof(func),typeof(set)}(
-                func.constant,
-            ),
-        )
+    elseif mock.scalar_function_constant_non_zero
+        MOI.throw_if_scalar_and_constant_not_zero(func, typeof(S))
```
blegat, line 185:
```suggestion
        MOI.throw_if_scalar_and_constant_not_zero(func, typeof(set))
```

Issue #2765, opened 2025-05-28
[FileFormats] support other number types other than Float64


hdavid16:
Hi, I am trying to read this model file so that it uses `Int` in order to use it with MiniZinc/Chuffed. However, I get an error on the types. Is it possible to use `MOI.read_from_file` so that values are not converted from Int to Float?

[feasibility.mof.json](https://github.com/user-attachments/files/20492746/feasibility.mof.json)

```julia
m = MOI.Utilities.Model{Int}()
MOI.read_from_file(m, "feasibility.mof.json")

ERROR: UnsupportedConstraint: `MathOptInterface.ScalarAffineFunction{Float64}`-in-`MathOptInterface.EqualTo{Float64}` constraints are not supported by the
solver you have chosen, and we could not reformulate your model into a
form that is supported.

To fix this error you must choose a different solver.
```

```julia
pkg> st JuMP
Status `~/Documents/rai-supply-chain-demo/Project.toml`
  [4076af6c] JuMP v1.26.0
```

odow:
This needs some features to be added to the `MOI.FileFormats.MOF` submodule. Currently it assumes `Float64`.

chriscoey:
@odow relatedly, we also seemed to have trouble writing `Int` models to LP format. I didn't see Int tests in the LP format tests, and I see https://github.com/jump-dev/MathOptInterface.jl/blob/9aba539f227f22c90b66a9c1d476d46c56acb78f/src/FileFormats/LP/LP.jl#L96 seems to hardcode Float64. Perhaps the title of this issue can be generalized, or a separate issue can be created?

odow:
I've changed the title

odow:
https://github.com/jump-dev/MathOptInterface.jl/actions/runs/15457235366

odow:
With `master`, this now yields:

```julia
julia> m = MOI.Utilities.Model{Int}()
MOIU.Model{Int64}
├ ObjectiveSense: FEASIBILITY_SENSE
├ ObjectiveFunctionType: MOI.ScalarAffineFunction{Int64}
├ NumberOfVariables: 0
└ NumberOfConstraints: 0

julia> MOI.read_from_file(m, "feasibility.mof.json"; coefficient_type = Int)

julia> m
MOIU.Model{Int64}
├ ObjectiveSense: FEASIBILITY_SENSE
├ ObjectiveFunctionType: MOI.ScalarAffineFunction{Int64}
[...]

PR #2768, opened 2025-06-04
[FileFormats.LP] add support for other number types

odow:
Part of #2765

### DIFF
--- a/test/FileFormats/LP/LP.jl
+++ b/test/FileFormats/LP/LP.jl
@@ -1130,6 +1130,69 @@ function test_unsupported_variable_types()
     return
 end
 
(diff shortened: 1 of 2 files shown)


test/FileFormats/LP/LP.jl
```diff
@@ -1130,6 +1130,65 @@ function test_unsupported_variable_types()
     return
 end
 
+function _test_int_round_trip(src)
+    model = MOI.FileFormats.LP.Model()#; coefficient_type = Int)
```
odow:
```suggestion
    model = MOI.FileFormats.LP.Model(; coefficient_type = Int)
```

PR #2774, opened 2025-06-11
Optionally disable warn on `PenaltyRelaxation`

joaquimg:
This is a massive pain in:
- https://github.com/jump-dev/ModelAnalyzer.jl
- https://github.com/jump-dev/MathOptConflictSolver.jl

As we keep skipping bounds and such.

The behaviour is opt-out, so users remain with the warning.

### DIFF
--- a/src/Utilities/penalty_relaxation.jl
+++ b/src/Utilities/penalty_relaxation.jl
@@ -189,7 +189,7 @@ cannot be modified in-place.
 To modify variable bounds, rewrite them as linear constraints.
 
 If a constraint type can not be modified, a warning is logged and the
-constraint is skipped. This can be disabled by setting `warn = false`.
+constraint is skipped. The warning can be disabled by setting `warn = false`.
 
 ## Example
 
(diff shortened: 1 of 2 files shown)


src/Utilities/penalty_relaxation.jl
```diff
@@ -142,6 +142,7 @@ end
     PenaltyRelaxation(
         penalties = Dict{MOI.ConstraintIndex,Float64}();
         default::Union{Nothing,T} = 1.0,
+        no_warning_skip_constraint::Bool = false,
```
odow:
```suggestion
        warn::Bool = true,
```
It'd be so much simpler if people opted out with `warn = false`.

odow:
I'll take over from here

odow (approved):

joaquimg:
Good for me!

src/Utilities/penalty_relaxation.jl
```diff
@@ -187,6 +188,9 @@ cannot be modified in-place.
 
 To modify variable bounds, rewrite them as linear constraints.
 
+If a constraint type can not be modified, a warning is logged and the
```
blegat:
```suggestion
If a constraint cannot be modified, a warning is logged and the
```

PR #2775, opened 2025-06-11
Add ConflictCount and conflict_index to ConstraintConflictStatus

joaquimg:
This is a request for comments.

While implementing:
- https://github.com/jump-dev/MathOptAnalyzer.jl
- https://github.com/jump-dev/MathOptIIS.jl

It became clear that returning multiple conflicts is very useful, for instance, when returning multiple bound conflicts.

Moreover, I know that Xpress supports returning multiple conflicts.

The `ListOfConstraintIndicesInConflict` can be more controversial and left for another PR.


src/attributes.jl
```diff
@@ -1844,6 +1844,16 @@ struct ConflictStatus <: AbstractModelAttribute end
 
 attribute_value_type(::ConflictStatus) = ConflictStatusCode
 
+"""
+    ConflictCount()
```
joaquimg, line 1890: I think this is better than overloading ResultCount.

src/attributes.jl
```diff
@@ -2677,13 +2687,78 @@ end
...
+```julia
+MOI.get(
+    ::Optimizer,
+    ::MOI.ListOfConstraintIndicesInConflict,
+)::Vector{MOI.ConstraintIndex}
+```
+A ineficient fallback is available for `get`.
+They should not implement [`set`](@ref) or [`supports`](@ref).
+"""
+struct ListOfConstraintIndicesInConflict <: MOI.AbstractModelAttribute
```
joaquimg: Useful to avoid looping through the entire model all the time, becomes even more important if we have multiple conflicts.

This leads to questions:
- Only one attribute for both IN_CONFLICT and MAYBE_IN_CONFLICT?
- The type should be a parameter such as: `ListOfConstraintIndicesForConflictType(IN_CONFLICT )`?
- Mix constraint types or do `ListOfConstraintIndicesInConflict{F,S}`?

blegat: We need to do `ListOfConstraintIndicesInConflict{F,S}`

blegat: What about `ListOfConstraintIndicesWithConflictStatus{F,S}(status, conflict_index)` ?

joaquimg: > ListOfConstraintIndicesWithConflictStatus{F,S}(status, conflict_index)

does make sense to me

joaquimg: If we are parametrizing that, should we have: `ListOfConstraintTypesWithConflictStatusPresent(conflict_index)`?

blegat: We could always add it later if useful, we don't need to add it just for the sake of consistency

odow: I'm not sure I like the direction of this attribute. Gurobi et al don't have an efficient way to implement it. And looping over the constraints isn't that bad. I don't really see why we need it.

src/attributes.jl
odow: Fine by me

src/attributes.jl
joaquimg: I will remove this from the PR and add in a future one, is necessary. 

Also, Xpress gives the list: https://www.fico.com/fico-xpress-optimization/docs/latest/solver/optimizer/HTML/XPRSgetiisdata.html

odow: My feeling is that the computational cost is quite large. We don't need to complicate things with extra attributes.

MathOptInfeasibleSubsystem.jl could define this attribute, and solvers could opt-in to it if they use the package an have an efficient way of handling it? I don't think it needs to be in MOI

odow:
What's the plan with this? I think we are agreed that returning multiple IIS is useful. But another ListOf... is not needed.

joaquimg:
I will cleanup and remove the ListOf...

odow:
Let's discuss this tomorrow

odow:
Do we need a corresponding error like `MOI.ResultIndexBoundsError`?

odow (approved):

odow:
Good to merge?

joaquimg:
Good for me

odow:
I just tweaked the docstring

PR #2779, opened 2025-07-02
Fix expression comparisons on nightly

odow:


odow:
Upstream culprit is https://github.com/JuliaLang/julia/pull/58661

I need to think if there is a better way to fix this.

odow:
This also impacted JuMP: https://github.com/jump-dev/JuMP.jl/pull/4023

odow:
Closing in favor of #2780

PR #2787, opened 2025-07-18
[FileFormats.MPS] test that duplicate entries are summed

### LINKED ISSUE #2786: [FileFormats.MPS] test repeated coefficients in MPS file
x-ref https://github.com/ERGO-Code/HiGHS/pull/2470

odow:
Closes #2786

It's not obvious what the right thing to do here is.

Gurobi ignores the repeated coefficients:
```
(base) oscar@Oscars-MacBook-Pro MathOptInterface % gurobi_cl test/FileFormats/MPS/duplicate_coefficient.mps
Set parameter LicenseID to value 890341
Set parameter LogFile to value "gurobi.log"
Using license file /Users/oscar/gurobi.lic

Gurobi Optimizer version 11.0.1 build v11.0.1rc0 (mac64[x86] - Darwin 24.1.0 24B83)
Copyright (c) 2024, Gurobi Optimization, LLC

Warning: duplicate nonzeros for column x and row c.
Read MPS format model from file test/FileFormats/MPS/duplicate_coefficient.mps
Reading time = 0.00 seconds
DUPLICATE_COEFFICIENT: 1 rows, 1 columns, 1 nonzeros

CPU model: Intel(R) Core(TM) i5-8259U CPU @ 2.30GHz
Thread count: 4 physical cores, 8 logical processors, using up to 8 threads

Optimize a model with 1 rows, 1 columns and 1 nonzeros
Model fingerprint: 0x0866ac82
Coefficient statistics:
  Matrix range     [1e+00, 1e+00]
  Objective range  [1e+00, 1e+00]
  Bounds range     [0e+00, 0e+00]
  RHS range        [1e+00, 1e+00]
Presolve removed 1 rows and 1 columns
Presolve time: 0.01s
Presolve: All rows and columns removed
Iteration    Objective       Primal Inf.    Dual Inf.      Time
       0    1.0000000e+00   0.000000e+00   0.000000e+00      0s

Solved in 0 iterations and 0.01 seconds (0.00 work units)
Optimal objective  1.000000000e+00
```

Cbc says:
```
Coin:import test/FileFormats/MPS/duplicate_coefficient.mps
At line 1 NAME          DUPLICATE_COEFFICIENT
At line 2 ROWS
At line 5 COLUMNS
Duplicate row c at line 8 <     x         c         1 >
At line 9 RHS
At line 11 BOUNDS
At line 13 ENDATA
Problem DUPLICATE_COEFFICIENT has 1 rows, 1 columns and 1 elements
Coin0008I DUPLICATE_COEFFICIENT read with 1 errors
There were 1 errors on input
```

So should we keep the current behavior or change to an error? Gurobi throwing a warning but solving an arbitrary (which one did it ignore!!!) problem seems like a massively wrong choice. (cc @simonbowly)

### DIFF
--- a/test/FileFormats/MPS/MPS.jl
+++ b/test/FileFormats/MPS/MPS.jl
@@ -1661,6 +1661,20 @@ function test_obj_constant_max_to_max()
     return
 end
 
+function test_duplicate_coefficient()
+    model = MPS.Model()
+    MOI.read_from_file(model, joinpath(@__DIR__, "duplicate_coefficient.mps"))
+    dest = MOI.Utilities.Model{Float64}()
+    MOI.copy_to(dest, model)
+    F, S = MOI.ScalarAffineFunction{Float64}, MOI.EqualTo{Float64}
+    x = only(MOI.get(dest, MOI.ListOfVariableIndices()))
+    c = only(MOI.get(dest, MOI.ListOfConstraintIndices{F,S}()))
+    f = MOI.get(dest, MOI.ConstraintFunction(), c)
+    @test isapprox(f, 2.0 * x)
+    @test MOI.get(dest, MOI.ConstraintSet(), c) == MOI.EqualTo(1.0)
+    return
+end
+
 end  # TestMPS
 
 TestMPS.runtests()
--- a/test/FileFormats/MPS/duplicate_coefficient.mps
+++ b/test/FileFormats/MPS/duplicate_coefficient.mps
@@ -1,0 +1,13 @@
+NAME          DUPLICATE_COEFFICIENT
+ROWS
+ N  obj
+ E  c
+COLUMNS
+    x         obj       1
+    x         c         1
+    x         c         1
+RHS
+    rhs       c        1
+BOUNDS
+ FR bounds    x
+ENDATA


simonbowly:
Yep, seems weird, but I wasn't at the basement meeting where MPS was figured out. I found some old CPLEX manuals stating "a warning will be issued and the most recently assigned value will be used", which sounds about right for a punch card format. I'll see if anyone here remembers.

For an old format it seems maybe logical to follow the oldest of solvers ... is Cbc erroring out, or just printing messages?

odow:
Cbc properly errors:
```
(base) oscar@Oscars-MacBook-Pro MPS % cbc duplicate_coefficient.mps 
Welcome to the CBC MILP Solver 
Version: 2.10.8 
Build Date: May  7 2022 

command line - cbc duplicate_coefficient.mps (default strategy 1)
At line 1 NAME          DUPLICATE_COEFFICIENT
At line 2 ROWS
At line 5 COLUMNS
Duplicate row c at line 8 <     x         c         1 >
At line 9 RHS
At line 11 BOUNDS
At line 13 ENDATA
Problem DUPLICATE_COEFFICIENT has 1 rows, 1 columns and 1 elements
Coin0008I DUPLICATE_COEFFICIENT read with 1 errors
There were 1 errors on input
Total time (CPU seconds):       0.00   (Wallclock seconds):       0.01
```

The Gurobi docs don't mention anything about repeated entries: https://docs.gurobi.com/projects/optimizer/en/current/reference/fileformats/modelformats.html#columns-section

I think for now I stand by the decision to sum. I can't think of any situation in which a user is intentionally adding repeated entries to use only the last one, or where they expect an error. We will also never write repeated entries, so this applies only to models that we're reading from external source, and it seems like files with repeated entries are not well formed.

simonbowly:
> We will also never write repeated entries, so this applies only to models that we're reading from external source

Neither will Gurobi, but reading MPS files from other sources is already a bit of a minefield with inconsistencies galore.

> and it seems like files with repeated entries are not well formed

Isn't that a stronger argument for erroring out? Seems like the best way to avoid ambiguous comparisons where the field isn't consistent.

odow:
> Isn't that a stronger argument for erroring out?

Maybe you're right. I'll take a look.

I hate file formats without a schema. If you want to push something inside Gurobi... we'd love for people to adopt https://github.com/jump-dev/MathOptFormat

odow:
I added a check for the duplicates, and it broke some other tests. Making it error would be a breaking change (although it could be regarded as a bug fix). I think I'm willing to leave the summing behavior as-is. There's also a performance hit to have to check for duplicates. 

People writing MPS files with duplicate columns already do so at their own risk.

PR #2799, opened 2025-08-04
[Nonlinear] parse x*x as x^2 in Nonlinear.Model

odow:
Issue raised in https://github.com/PSORLab/EAGO.jl/issues/150

I'll run `solver-tests` to see if there's any impact.

https://github.com/jump-dev/MathOptInterface.jl/actions/runs/16733837980

### DIFF
--- a/src/Nonlinear/parse.jl
+++ b/src/Nonlinear/parse.jl
@@ -345,18 +345,40 @@ function parse_expression(
     x::MOI.ScalarQuadraticTerm,
     parent_index::Int,
 )
-    id_mul = data.operators.multivariate_operator_to_id[:*]
-    push!(expr.nodes, Node(NODE_CALL_MULTIVARIATE, id_mul, parent_index))
-    mul_parent = length(expr.nodes)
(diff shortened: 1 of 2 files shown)


odow:
JuMP failure is because of #2775

odow:
It's an open question whether we should merge this. You might be tempted to say that this is an EAGO issue, not a JuMP/MOI issue.

They should detect structure rather than relying on the precise way that JuMP users write models or MOI lowers a quadratic.

odow:
https://github.com/jump-dev/MathOptInterface.jl/actions/runs/16794100302

blegat:
I also think that they should probably detect the structure, the user may still be able to write `x * x`, e.g., using `@force_nonlinear` ? On the other hand, this make to MOI makes sense. If we have a square in the `ScalarQuadraticFunction`, we should probably detect is as square.

blegat (approved):

odow:
Yeah there's no right or wrong answer here. Very much open for debate

PR #2803, opened 2025-08-06
Don't copy ScalarNonlinearFunction in VectorOfConstraints

blegat:
Extracted from https://github.com/jump-dev/MathOptInterface.jl/pull/2788

I'm not too convinced by the current approach.

For the `copy` in `ConstraintFunction`, I think we add it mostly to be extra-sure the user does not alter the function but we don't really need it in the context of `MOI.copy_to` since `add_constraint` specifies that the model should have its own copy. Maybe we should also have `UnsafeConstraintFunction` that is used inside `MOI.copy_to`, it should also safe wasteful copies for `ScalarAffineFunction` and `ScalarQuadraticFunction`. So if we have tests that checks that solvers have their own copies in `add_constraint` then it should be safe.

### DIFF
--- a/src/Utilities/vector_of_constraints.jl
+++ b/src/Utilities/vector_of_constraints.jl
@@ -103,7 +103,14 @@ function MOI.get(
 ) where {F,S}
     MOI.throw_if_not_valid(v, ci)
     f, _ = v.constraints[ci]::Tuple{F,S}
-    return copy(f)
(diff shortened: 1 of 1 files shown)


odow:
I think the first step before we go making any changes to this is to come up with a large suite of benchmarks and tests. There are a number of subtle behaviors here.

PR #2804, opened 2025-08-07
Fix duplicate test with same name

blegat:
Julia was complaining we were overwriting an existing function.


odow:
I need to figure out a CI check for this. It happens somewhat often

PR #2809, opened 2025-08-12
Fix ConstraintDual transformation for SplitHyperRectangleBridge

blegat:
Fix for the issues detected in the SolverTests run of https://github.com/jump-dev/MathOptInterface.jl/pull/2808
I did not add any test since the test is in https://github.com/jump-dev/MathOptInterface.jl/pull/2810
I did https://github.com/jump-dev/MathOptInterface.jl/pull/2810 in a separate PR not only to show in CI that it would have caught this one but also to avoid delaying this PR and hence https://github.com/jump-dev/MathOptInterface.jl/pull/2808 in case https://github.com/jump-dev/MathOptInterface.jl/pull/2810 requires some discussion

### DIFF
--- a/src/Bridges/Constraint/bridges/SplitHyperRectangleBridge.jl
+++ b/src/Bridges/Constraint/bridges/SplitHyperRectangleBridge.jl
@@ -229,7 +229,7 @@ function MOI.set(
     set = bridge.set
     new_values = vcat(
         T[max(T(0), v) for (v, l) in zip(values, set.lower) if isfinite(l)],
-        T[min(T(0), v) for (v, u) in zip(values, set.upper) if isfinite(u)],
+        T[max(T(0), -v) for (v, u) in zip(values, set.upper) if isfinite(u)],
     )
     MOI.set(model, attr, bridge.ci, new_values)
     return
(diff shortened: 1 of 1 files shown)


odow:
I don't like that all the `solver-tests` are now going through this bridge. Is that what we want? Or do we need to change the bridge weight?

odow:
https://github.com/jump-dev/MathOptInterface.jl/actions/runs/16919956500

odow:
It looks like this fixed some errors, but there are a couple of other issues

blegat:
I don't think it's an issue that this bridge is now used. If you want to go from Interval to Nonnegatives, it makes sense

blegat:
If you go `Interval` -> `LessThan` and `GreaterThan` then you still need 2 bridges to create `Nonnegatives`. If you go `Interval` to `Hyperrectangle`, only one bridge is left. What do you think it's an issue that we use this bridge ?

odow:
It makes sense that this bridge is chosen. It's just a large change (evidenced by the pretty broad failure of solver tests). But maybe that's okay. It's good if Hyperrectange gets more use

PR #2810, opened 2025-08-12
Test ConstraintDual for bridges

blegat:
I always thought it a bit silly to only define the setter for `ConstraintDualStart` and not with `ConstraintDual`. I sometimes want to test things mixing a `MockOptimizer` and a bridge and I can't because bridges don't define the setters for `ConstraintDual`.

I suggest the following non-breaking change: We should now encourage bridges to supports `ConstraintDual`. When they do, a new test will be enabled that will do an important consistency check. So the new test will not be run for existing bridge which is nice since it could be breaking but after the transition, it will be run for all bridges that don't disable it by doing `dual = nothing`.

This would have caught the bug in https://github.com/jump-dev/MathOptInterface.jl/pull/2809 as the CI should show

### DIFF
--- a/src/Bridges/Bridges.jl
+++ b/src/Bridges/Bridges.jl
@@ -283,7 +283,7 @@ julia> MOI.Bridges.runtests(
            end,
        )
 Test Summary:    | Pass  Total  Time
-Bridges.runtests |   32     32  0.8s
+Bridges.runtests |   33     33  0.8s
 ```
 """
 function runtests(args...; kwargs...)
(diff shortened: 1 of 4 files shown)


src/Bridges/Constraint/bridges/SplitHyperRectangleBridge.jl
```diff
@@ -174,7 +174,11 @@ end
 
 function MOI.supports(
     model::MOI.ModelLike,
-    attr::Union{MOI.ConstraintPrimalStart,MOI.ConstraintDualStart},
+    attr::Union{
+        MOI.ConstraintPrimalStart,
+        MOI.ConstraintDualStart,
+        MOI.ConstraintDual,
```
odow, line 202: This is a big dislike from me. ConstraintDual is a result attribute. We shouldn't make bridges "support" it.

blegat: Since `MockOptimizer` supports `ConstraintDual`, then it's quite useful for testing. `ConstraintDual` and `ConstraintDualStart` have different semantics for solvers but for bridges that only care about how things are transformed, they are completely equivalent.

odow:
I'm still against this. Bridges could/should write better tests regardless. Testing the dual is tricky. And even our `Start` is a bad test because it checks only get/set doesn't error. It doesn't really have a way to test that the logic is correct.

blegat:
Yes, we just check that get and set are inverse of each other and that's precisely what this PR is fixing. Now it also checks that it is the adjoint of the bridge transformation so it's starting to get difficult to get the test passing with an incorrect dual

src/Utilities/results.jl
```diff
@@ -88,9 +88,17 @@ function _dual_objective_value(
...
+_lower(::MOI.ZeroOne, ::Type{T}) where {T} = zero(T)
+_upper(::MOI.ZeroOne, ::Type{T}) where {T} = one(T)
+
 function _dual_objective_value(
     model::MOI.ModelLike,
-    ci::MOI.ConstraintIndex{<:MOI.AbstractScalarFunction,<:MOI.Interval},
+    ci::MOI.ConstraintIndex{
+        <:MOI.AbstractScalarFunction,
+        <:Union{MOI.ZeroOne,MOI.Interval},
+    },
```
odow: This needs to be a separate PR. What is `ZeroOne` doing in a PR about testing constraint duals?

blegat: This is an issue detected by running the dual tests through all bridges. This is therefore tested in this PR as well

odow: Which bridge? Why is ZeroOne involved? How can there be a dual if the ZeroOne set is involved?

blegat: The `AbstractFunctionConversionBridge` already supports setting `ConstraintDual`, they treat it the same as `ConstraintDualStart` so it got hit by the new tests.

> How can there be a dual if the ZeroOne set is involved?

From the perspective of `_dual_objective_value`, since `ZeroOne` is equivalent to `Interval` + `Integer`, and `Integer` is ignored by `_dual_objective_value` then it's ok to treat `ZeroOne` as an `Interval`. If the solver doesn't have any dual, we will fail when we query the dual from the solver anyway with the expected error. Here, at the level of bridges or `_dual_objective_value`, we are just taking the transpose of linear maps so we shouldn't worry about whether some variables are integer or not.

odow: Can this be a separate PR? (With an explicit test)

odow:
Let me take over from here and split it up.

src/Utilities/results.jl
blegat: We can disable the tests for the failing ones yes

odow:
So you added `ZeroOne` to the fallback for dual objective value because some tests tried to call `constant(::ZeroOne)`? Why is `IndicatorSOS1Bridge` even attempting to test the duals? 

This design feels wrong. Could dual tests be opt-in?

Is a separate bug that we don't have a generic fallback for `constant`? The test error violates the MethodError principle.

odow:
#2826 removes the use of `constant`.

src/Bridges/Bridges.jl
```diff
@@ -293,12 +293,60 @@ function runtests(args...; kwargs...)
...
+    mock = MOI.Utilities.MockOptimizer(inner)
+    model = _bridged_model(Bridge{eltype}, mock)
+    input_fn(model)
+    final_touch(model)
+    # Should be able to call final_touch multiple times.
+    final_touch(model)
+    # If the bridges does not support `ConstraintDualStart`, it probably won't
+    # support `ConstraintDual` so we skip these tests
+    list_of_constraints = MOI.get(model, MOI.ListOfConstraintTypesPresent())
+    attr = MOI.ConstraintDual()
```
odow, line 323: Did you mean
```suggestion
    attr = MOI.ConstraintDualStart()
```

blegat: No, if the bridge don't support `ConstraintDual`, we skip the test. Because `get_fallback` don't work with `ConstraintDualStart`, we need to use `ConstraintDual`

PR #2811, opened 2025-08-12
Add a check that bridge is used

blegat:
Most of the time when I debug a new bridge, it starts by not being used and I sometimes loose some time checking it. Having it explicitly checked would be helpful.

### DIFF
--- a/src/Bridges/Bridges.jl
+++ b/src/Bridges/Bridges.jl
@@ -303,16 +303,21 @@ function _runtests(
     model_eltype = eltype,
     print_inner_model::Bool = false,
     cannot_unbridge::Bool = false,
+    no_bridge_used::Bool = false,
 )
     # Load model and bridge it
     inner = MOI.Utilities.UniversalFallback(MOI.Utilities.Model{model_eltype}())
     model = _bridged_model(Bridge{eltype}, inner)
     input_fn(model)
+    _check_bridged(model; no_bridge_used)
     final_touch(model)
     # Should be able to call final_touch multiple times.
     final_touch(model)
     if print_inner_model
         print(inner)
+    end
+    if no_bridge_used
+        return
     end
     Test.@testset "Test outer bridged model appears like the input" begin       # COV_EXCL_LINE
         test =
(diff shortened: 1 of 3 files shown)


odow:
> Most of the time when I debug a new bridge, it starts by not being used

I don't understand this. Can you point to an example?

test/Bridges/Constraint/IndicatorSOS1Bridge.jl
```diff
@@ -345,6 +345,7 @@ function test_runtests()
         [z, 2.0 * x * x] in Indicator{ACTIVATE_ON_ONE}(LessThan(2.0))
         z in ZeroOne()
         """,
+        no_bridge_used = true,
```
odow, line 348: This one has an explicit comment that the model is unchanged. Is this PR really needed?

blegat:
When you don't implement support properly for the bridge then the bridge isn't used and you get complicated errors saying that the constraint types of the expected model don't match. You might think that the expected model you wrote is wrong or that your bridging isn't done correctly until you realize that the bridge just isn't being used. I thought that starting with a test that says : hey you bridge isn't even being used helps you gain time in many cases.

odow:
This has just never happened to me :smile:

src/Bridges/Bridges.jl
```diff
@@ -303,17 +303,24 @@ function _runtests(
     model_eltype = eltype,
     print_inner_model::Bool = false,
     cannot_unbridge::Bool = false,
+    no_bridge_used::Bool = false,
 )
     # Load model and bridge it
     inner = MOI.Utilities.UniversalFallback(MOI.Utilities.Model{model_eltype}())
     model = _bridged_model(Bridge{eltype}, inner)
     input_fn(model)
+    @testset "Bridge used" begin
```
odow:
```suggestion
    Test.@testset "Bridge used" begin
```

src/Bridges/Bridges.jl
```diff
@@ -512,6 +519,13 @@ function _general_bridge_tests(bridge::B) where {B<:AbstractBridge}
     return
 end
 
+function _check_bridged(model; no_bridge_used)
+    unused = isempty(MOI.Bridges.Variable.bridges(model)) &&
+        isempty(MOI.Bridges.Constraint.bridges(model)) &&
+        isempty(MOI.Bridges.Objective.bridges(model))
+    Test.@test unused == no_bridge_used
```
odow, line 527:
```suggestion
    Test.@test unused == no_bridge_used
    return
```

test/Bridges/Objective/VectorSlackBridge.jl
```diff
@@ -103,7 +103,8 @@ function test_runtests()
         """
         variables: x
         maxobjective: [x]
-        """,
+        """;
+        no_bridge_used = true,
```
odow, line 107: This is also technically a breaking change because the tests fail without this change

odow:
I think I want to close this as won't fix.

test/Bridges/Objective/VectorSlackBridge.jl
blegat: It's only breaking for tests so it won't really break user code. And this breakage would be quite rare

Issue #2813, opened 2025-08-12
[FileFormats] error properly on vector-valued objectives


SouthEndMusic:
When I take bi-objective example from [here](https://jump.dev/JuMP.jl/stable/tutorials/linear/multi_objective_examples/) and try to use `JuMP.write_to_file`, I get (with JuMP v1.28.0 MultiObjectiveAlgorithms v.1.6.0)

```
ERROR: MethodError: no method matching _write_function(::IOStream, ::MathOptInterface.Utilities.GenericModel{…}, ::MathOptInterface.VectorAffineFunction{…}, ::Dict{…})
The function `_write_function` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  _write_function(::IO, ::MathOptInterface.Utilities.GenericModel{T, MathOptInterface.Utilities.ObjectiveContainer{T}, MathOptInterface.Utilities.VariablesContainer{T}, MathOptInterface.FileFormats.LP.ModelFunctionConstraints{T}} where T, ::MathOptInterface.ScalarQuadraticFunction, ::Dict{MathOptInterface.VariableIndex, String}; print_half, kwargs...)
   @ MathOptInterface C:\ProgramData\DEVDrives\.julia\packages\MathOptInterface\vK6dk\src\FileFormats\LP\LP.jl:167
  _write_function(::IO, ::MathOptInterface.Utilities.GenericModel{T, MathOptInterface.Utilities.ObjectiveContainer{T}, MathOptInterface.Utilities.VariablesContainer{T}, MathOptInterface.FileFormats.LP.ModelFunctionConstraints{T}} where T, ::MathOptInterface.ScalarAffineFunction, ::Dict{MathOptInterface.VariableIndex, String}; print_one, kwargs...)
   @ MathOptInterface C:\ProgramData\DEVDrives\.julia\packages\MathOptInterface\vK6dk\src\FileFormats\LP\LP.jl:137
  _write_function(::IO, ::MathOptInterface.Utilities.GenericModel{T, MathOptInterface.Utilities.ObjectiveContainer{T}, MathOptInterface.Utilities.VariablesContainer{T}, MathOptInterface.FileFormats.LP.ModelFunctionConstraints{T}} where T, ::MathOptInterface.VariableIndex, ::Dict{MathOptInterface.VariableIndex, String}; kwargs...)
[...]

PR #2814, opened 2025-08-12
[FileFormats] fix and test MOI.supports of MOI.ObjectiveFunction

odow:
x-ref #2813


test/FileFormats/SDPA/SDPA.jl
```diff
@@ -402,6 +402,23 @@ function test_example_A_integer_coefficients()
     return
 end
 
+function test_unsupported_objectives()
+    model = MPS.Model()
```
odow:
```suggestion
    model = SDPA.Model()
```

PR #2821, opened 2025-08-19
[Bridges] pass variable names through ParameterToEqualToBridge

### LINKED ISSUE #2762: Printing models with parameters
Consider:
```julia
model = Model()
@variable(model, y == 18)
@variable(model, p in Parameter(1.3))
@constraint(model, c1, 2y + 3p >= 0)
@objective(model, Min, 5y + 6p)

write_to_file(model, "test_param.mps")

print(read("test_param.mps", String))
```
The resulting file is:
```julia
NAME
ROWS
 N  OBJ
 G  c1
COLUMNS
    y         c1        2
    y         OBJ       5
    x2        c1        3
    x2        OBJ       6
RHS
    rhs       c1        0
RANGES
BOUNDS
 FX bounds    y         18
 FX bounds    x2        1.3
ENDATA
```
It seems that, once the parameter goes to the `ParameterToEqualToBridge` the name information is lost

odow:
Closes #2762

But I'm not 100% sure we should merge this. It's really just a single special case. It isn't obvious what naming convention we should use if the map is `y := Ax + b`.

### DIFF
--- a/src/Bridges/bridge_optimizer.jl
+++ b/src/Bridges/bridge_optimizer.jl
@@ -1757,6 +1757,20 @@ function MOI.get(
     end
 end
 
+function _get_variable_if_equivalent(b, x)
+    return _get_variable_if_equivalent(bridged_variable_function(b, x))
+end
+
+function _get_variable_if_equivalent(f::MOI.ScalarAffineFunction)
+    if length(f.terms) == 1
+        term = only(f.terms)
+        if isone(term.coefficient)
+            return term.variable
+        end
+    end
+    return nothing
+end
+
 function MOI.set(
     b::AbstractBridgeOptimizer,
     attr::MOI.VariableName,
@@ -1766,6 +1780,11 @@ function MOI.set(
     if is_bridged(b, vi)
         b.var_to_name[vi] = name
         b.name_to_var = nothing # Invalidate the name map.
+        # Set the internal variable name iff the bridged variable is equivalent
+        # to `y := 1 * x`.
+        if (x = _get_variable_if_equivalent(b, vi)) !== nothing
+            MOI.set(b.model, attr, x, name)
+        end
     else
         MOI.set(b.model, attr, vi, name)
     end
--- a/test/Bridges/bridge_optimizer.jl
+++ b/test/Bridges/bridge_optimizer.jl
@@ -1570,6 +1570,28 @@ function test_issue_2817c()
     return
 end
 
+function test_issue_2762_success()
+    inner = MOI.Utilities.Model{Float64}()
+    model = MOI.Bridges.Variable.ParameterToEqualTo{Float64}(inner)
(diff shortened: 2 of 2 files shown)


odow:
This is actually not as easy as I thought.

src/Bridges/bridge_optimizer.jl
```diff
@@ -1757,6 +1757,20 @@ function MOI.get(
     end
 end
 
+function _get_variable_if_equivalent(b, x)
+    return _get_variable_if_equivalent(bridged_variable_function(b, x), x)
+end
```
odow, line 1765:
```suggestion
end

_get_variable_if_equivalent(b, ::Nothing) = nothing
```

src/Bridges/bridge_optimizer.jl
```diff
@@ -1757,6 +1757,28 @@ function MOI.get(
     end
 end
 
+function _get_variable_if_equivalent(
+    b::AbstractBridgeOptimizer,
+    x::VariableIndex,
```
odow:
```suggestion
    x::MOI.VariableIndex,
```

odow:
I'm not convinced we should merge this, but if we close this, we should also close #2762

odow:
I'll let @blegat decide what we should do

src/Bridges/bridge_optimizer.jl
```diff
@@ -1839,6 +1866,9 @@ function MOI.get(
     end
     vi_bridged = get(b.name_to_var, name, nothing)
     MOI.Utilities.throw_if_multiple_with_name(vi_bridged, name)
+    if _get_variable_if_equivalent(b, vi_bridged) !== nothing
```
blegat, line 1869: What is this for ?

blegat: Maybe add a comment explaining

blegat:
I think that in general, bridge could optionally implement names, just the same way they can optionally implement any attribute.
The bridge optimizer could check if they support constraint names (for constraint bridges) or variable names (for variable bridges) and if so, it will set it.
For SetMapBridge, we know variable bridges create a unique variable in the scalar case so it's clear we can pass the name of that variable. But actually, when there is a nontrivial mapping, we like to give it a different name so we actually don't really want to pass the name along.
So I think that we should only pass the name along if it's mapped identically so this PR seems to do exactly what we should do.

odow:
> I think that in general, bridge could optionally implement names, just the same way they can optionally implement any attribute.

This seems reasonable.

blegat:
Sorry if I wasn't clear, I meant that since it only makes sense to implement it for a bridge that maps a function identically, this PR is the way to go

odow:
I went for the explicit support option in https://github.com/jump-dev/MathOptInterface.jl/pull/2822

odow:
Closing in favor of #2822

PR #2822, opened 2025-08-25
[Bridges] add support for VariableName in variable bridges

odow:
Alternative to https://github.com/jump-dev/MathOptInterface.jl/pull/2821


joaquimg (approved):
This was easier to understant

blegat (approved):
Indeed much simpler ^^

PR #2825, opened 2025-08-27
Update set_dot for sets with non-zero constants

odow:
I don't know if this is the right fix, but isn't one part of the problem with https://github.com/jump-dev/MathOptInterface.jl/pull/2810 that some `set_dot` don't take into account the set constants? It seems better to accept for them here rather than in `dual_objective_value`.


odow:
Closing in favor of #2826

PR #2834, opened 2025-09-07
Add ToScalarNonlinearBridge 

LukasBarner:
This is based on https://github.com/jump-dev/MathOptInterface.jl/issues/2833. 

Do you think this is a reasonable approach? 
One problematic thing (in my opinion) is that I had to remove the nonlinear bridge from a few tests, because it is not bound to a type. 
However, this probably is an issue for constraints already, right?


odow:
> because it is not bound to a type

Yeah. I think this may have been a mistake. We could somehow restrict to only supporting `Float64`?

odow:
I pushed a couple of tweaks. I'll see if there are any test issues. I don't think we have sufficient tests to catch the constraint case. But I'll worry about that separately.

odow (approved):

Issue #2351, opened 2023-11-16
[FileFormats.LP] write a proper parser


odow:
The LP reader is a source of many problems because it was designed as a bit of a hack and doesn't follow any formal grammar. It's probably time we rewrote a proper implementation.

https://discourse.julialang.org/t/sos-constraints-in-lp-file-throwing-malformed-sos-constraint-error/106324

See also: https://github.com/jump-dev/MathOptInterface.jl/issues?q=is%3Aissue+FileFormats.LP+is%3Aclosed

cc @hdavid16

pedromxavier:
@rafabench I think this might be interesting for some of our applications, right?

odow:
To clarify, you can read LP files currently. But the reader is not robust to the wide variety of LP files in the wild.

pedromxavier:
We have already used this LP reader in some apps. I remember to have discussed it with my coworkers some months ago.

odow:
See also https://github.com/jump-dev/MathOptInterface.jl/issues/2563

The parser should not read line-by-line.

PR #2837, opened 2025-09-09
[FileFormats.LP] fix for regex match on very long lines

### LINKED ISSUE #2836: [FileFormats.LP] Size Limit for read_from_file
Hi,

I tried to use JuMP to read a file from QPLIB, case 10005, (https://qplib.zib.de/QPLIB_10005.html), the LP file. But I got the following error:

Image

So does the error come from the size limit of read_from_file or other packages? Though this case is large, I remember I tried some other larger instances from MIPLIB.

Best,
YD

odow:
This fixed #2836 for me. But it isn't very satisfactory. And it still takes 2 minutes to read the file. This really needs #2351.

### DIFF
--- a/src/FileFormats/LP/LP.jl
+++ b/src/FileFormats/LP/LP.jl
@@ -809,7 +809,7 @@ function _parse_section(
             cache.constraint_name = "R$(cache.num_constraints)"
         end
     end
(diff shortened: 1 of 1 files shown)


odow:
Closing because I'm working on a proper fix.

PR #2846, opened 2025-09-12
[FileFormats.LP] remove lowercase usage to reduce allocations

odow:
Before
```Julia

julia> function main()
           model = MOI.FileFormats.LP.Model{Float64}()
           MOI.read_from_file(model, "/Users/odow/Downloads/QPLIB_10005.lp")
           return model
       end

julia> @time main();
  1.065072 seconds (43.71 M allocations: 1.511 GiB, 22.66% gc time)
```

After
```julia

julia> function main()
           model = MOI.FileFormats.LP.Model{Float64}()
           MOI.read_from_file(model, "/Users/odow/Downloads/QPLIB_10005.lp")
           return model
       end

julia> @time main();
  1.062581 seconds (30.88 M allocations: 1.156 GiB, 19.92% gc time)
```

### DIFF
--- a/src/FileFormats/LP/read.jl
+++ b/src/FileFormats/LP/read.jl
@@ -138,7 +138,7 @@ every identifier.
 This function tries to be a little cleverer and doesn't allocate.
 """
 function _case_insenstive_identifier_to_keyword(input::String)
-    if !(3 <= length(input) <= 8)
+    if !(2 <= length(input) <= 8)
         return nothing  # identifiers outside these lengths are not recognized
     elseif _compare_case_insenstive(input, 'm', _MAXIMIZE_KEYWORDS)
         return "MAXIMIZE"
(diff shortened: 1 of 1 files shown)


src/FileFormats/LP/read.jl
```diff
@@ -106,52 +106,62 @@ function Base.read!(io::IO, model::Model{T}) where {T}
...
-    "bin" => :BINARY,
-    "binary" => :BINARY,
-    "binaries" => :BINARY,
-    # SOS
-    "sos" => :SOS,
-    # END
-    "end" => :END,
-)
+function _case_insenstive_identifier_to_keyword(input::String)
+    if !(3 <= length(input) <= 8)
```
odow:
```suggestion
    if !(2 <= length(input) <= 8)
```

PR #2848, opened 2025-09-16
[FileFormats.LP] add a documentation page and refactor some names

odow:
And fall down the rabbit hole with respect to the many differences between solvers.

Now that I've played around with Xpress, I wonder if I need to re-re-think how we go about new line handling. Gurobi's parser is pretty weak in comparison.

My justification for spending so much time on this is that people reading LP files actually happens quite a lot, and historically, we have had many issues related to it (and, I assume, many more that aren't reported).

I think a single, great, open-source implementation that publicly documents all the weird differences in solvers is actually pretty useful.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -67,6 +67,7 @@ const _PAGES = [
         "FileFormats" => [
             "Overview" => "submodules/FileFormats/overview.md",
             "API Reference" => "submodules/FileFormats/reference.md",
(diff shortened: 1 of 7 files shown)


odow:
Okay. I'm stopping here. There's a limit of how hard we should work to parse bad LP files. We at least give good error messages, and we can parse all properly formatted files.

PR #2858, opened 2025-09-28
[docs] fix output nodes of `SquareBridge`

joaquimg:


odow (approved):
It's off topic, but I wonder why we didn't collect these into a vector and have `[g] in Zeros()`. Presumably a solver supporting triangle matrices is more likely to support the vector orthant than scalar equality?

Edit: I immediately answered my own question. My assumption is wrong. As one example see CSDP: https://github.com/jump-dev/CSDP.jl

PR #2860, opened 2025-10-02
Add VectorNonlinearOracle set

odow:
- [ ] Still need to test a few things, and check that it's okay once I update Ipopt

## Basic

 - [x] If `isbitstype(S) == false`, implement `Base.copy(set::S)`
 - [x] <s>If `isbitstype(S) == false`, implement `Base.:(==)(x::S, y::S)`</s>
 - [x] If an `AbstractVectorSet`, implement `dimension(set::S)`, unless the
       dimension is given by `set.dimension`.

## Utilities

 - [ ] If an `AbstractVectorSet`, implement `Utilities.set_dot`,
       `LinearAlgebra.dot`
 - [ ] If an `AbstractVectorSet`, implement `Utilities.set_with_dimension` in

## Documentation

## Tests

## MathOptFormat

 - [x] <s>Open an issue at `https://github.com/jump-dev/MathOptFormat` to add
       support for the new set {{ replace with link to the issue }}</s>

## Optional

 - [x] <s>Implement `dual_set(::S)` and `dual_set_type(::Type{S})`</s>
 - [x] <s>Add new bridges to convert your set into more commonly used sets</s>


odow:
Any objections?

PR #2865, opened 2025-10-07
Prep for v1.46.0

odow:
## Basic

 - [x] `version` field of `Project.toml` has been updated
       - If a breaking change, increment the MAJOR field and reset others to 0
       - If adding new features, increment the MINOR field and reset PATCH to 0
       - If adding bug fixes or documentation changes, increment the PATCH field

## Documentation

 - [x] Add a new entry to `docs/src/changelog.md`, following existing style

## Tests

 - [x] <s>https://github.com/jump-dev/MathOptInterface.jl/actions/runs/18326452732</s>
 - [x] <s>https://github.com/jump-dev/MathOptInterface.jl/actions/runs/18362251645</s>
 - [ ] https://github.com/jump-dev/MathOptInterface.jl/actions/runs/18511698884
 - [x] If new tests were added, ensure that `MOI.Test.version_added` is implemented.

### DIFF
--- a/docs/src/changelog.md
+++ b/docs/src/changelog.md
@@ -7,7 +7,7 @@ CurrentModule = MathOptInterface
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## v1.46.0 (October 17, 2025)
+## v1.46.0 (October 21, 2025)
 
 ### Added
 


odow:
The Juniper and Pavito failures are super interesting and they are caused by https://github.com/jump-dev/MathOptInterface.jl/pull/2857

 * They have `z --> {f(x) in S}, z in {0, 1}`
 * They have an outer solver that covers `z in {0, 1}`. They pass `z --> {f(x) in S}, z == a` to the inner solver. The inner solver says, sure, I know how to reformulate `-->` assuming `z` is binary. But `z` is fixed and isn't `{0, 1}`. The reformulation is still valid though, which is why the tests were passing before.

Maybe we should revert https://github.com/jump-dev/MathOptInterface.jl/pull/2857 and add a check at the JuMP level?

Or maybe the bridge needs to say: I added a `ZeroOne` constraint even though I didn't just so we validate that an indicator can't be bridged by a continuous solver.

Or maybe we need to convert the `error()` `AddConstraintNotAllowed`.

odow:
https://github.com/jump-dev/MathOptInterface.jl/actions/runs/18362251645

odow:
- [x] BARON: emailed Nick
- [x] DSDP: https://github.com/jump-dev/DSDP.jl/issues/45
- [x] PolyJuMP: known
- [x] SDPA: https://github.com/JuliaPackaging/Yggdrasil/pull/12251
- [x] SumOfSquares: known
- [x] BlockDecomposition: https://github.com/atoptima/BlockDecomposition.jl/pull/111
- [x] Coluna: https://github.com/atoptima/BlockDecomposition.jl/pull/111
- [x] EAGO: some generated function/Cassette thing
- [x] FrankWolfe: https://github.com/ZIB-IOL/FrankWolfe.jl/pull/621
- [x] SetProg: https://github.com/kalmarek/SymbolicWedderburn.jl/issues/97

docs/src/changelog.md
```diff
@@ -7,6 +7,26 @@ CurrentModule = MathOptInterface
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## v1.46.0 (October 8, 2025)
```
odow:
```suggestion
## v1.46.0 (October 10, 2025)
```

docs/src/changelog.md
```diff
@@ -7,6 +7,26 @@ CurrentModule = MathOptInterface
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## v1.46.0 (October 10, 2025)
```
odow:
```suggestion
## v1.46.0 (October 14, 2025)
```

odow:
Hearing no objections?

blegat:
> The Juniper and Pavito failures are super interesting and they are caused by #2857
> 
>     * They have `z --> {f(x) in S}, z in {0, 1}`
> 
>     * They have an outer solver that covers `z in {0, 1}`. They pass `z --> {f(x) in S}, z == a` to the inner solver. The inner solver says, sure, I know how to reformulate `-->` assuming `z` is binary. But `z` is fixed and isn't `{0, 1}`. The reformulation is still valid though, which is why the tests were passing before.
> 
> 
> Maybe we should revert #2857 and add a check at the JuMP level?
> 
> Or maybe the bridge needs to say: I added a `ZeroOne` constraint even though I didn't just so we validate that an indicator can't be bridged by a continuous solver.
> 
> Or maybe we need to convert the `error()` `AddConstraintNotAllowed`.

Pavito and Juniper shouldn't support Indicator constraints so that it's bridged before they pass it down. It doesn't make sense for them to pass down an indicator constraints to a non-MIP solver. It's not like there is a chance for these non-MIP solvers to support indicator constraints.

blegat:
Maybe this should be fixed before we release ?

odow:
Already done in https://github.com/jump-dev/MathOptInterface.jl/pull/2868

odow:
Latest failures in Alpine and PowerModels are because of https://github.com/jump-dev/HiGHS.jl/pull/302

docs/src/changelog.md
```diff
@@ -7,6 +7,26 @@ CurrentModule = MathOptInterface
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## v1.46.0 (October 14, 2025)
```
odow:
```suggestion
## v1.46.0 (October 16, 2025)
```

odow:
@blegat can you confirm you're okay with my changes with the ToMILPBridge?

docs/src/changelog.md
```diff
@@ -7,6 +7,26 @@ CurrentModule = MathOptInterface
...
+ - Fixed [`Bridges.Constraint.IndicatorToMILPBridge`](@ref) when the `z` variable
+   was not binary (#2857), (#2868)
+ - Fixed an `error` to [`GetAttributeNotAllowed`](@ref) in
+   [`Bridges.Variable.ZerosBridge`](@ref) (#2863)
+ - Fixed various bridges to throw MOI error subtypes instead of `Base.error()`
+   (#2866)
+
+### Other
+
+ - Documentation updates (#2855), (#2858)
```
odow:
```suggestion
 - Documentation updates (#2855), (#2858), (#2864)
```

docs/src/changelog.md
```diff
@@ -7,6 +7,26 @@ CurrentModule = MathOptInterface
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## v1.46.0 (October 16, 2025)
```
odow:
```suggestion
## v1.46.0 (October 17, 2025)
```

docs/src/changelog.md
```diff
@@ -7,6 +7,26 @@ CurrentModule = MathOptInterface
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## v1.46.0 (October 17, 2025)
```
odow:
```suggestion
## v1.46.0 (October 21, 2025)
```

blegat:
> Already done in #2868

Ok so the issue was that Juniper said that it supported indicator because it was redirecting to the inner solver which said it was supporting it. Now with https://github.com/jump-dev/MathOptInterface.jl/pull/2868/files, the inner solver says it does not support indicator so Juniper says the same so it is bridged in the bridge outer of Juniper, that's an elegant fix.

odow:
> that's an elegant fix.

Yip, although it isn't a perfect fix.

On the outside, Juniper fixes `z` to either `0` or `1`. This flows to the inner solver, so the MIP reformulation _would_ have been valid even without `z in {0, 1}`. But detecting that either `z` is `ZeroOne`, or that it has fixed bound of 0 or 1, or that it is integer and has bounds `[0, 1]`, or that it is a constant, or that there is some other equality constraint that `1.0 * z == 1` (or that there's a constraint like `z * (1 - z) = 0`) seems like it is out-of-scope for the bridge.

PR #2868, opened 2025-10-08
[Bridges] fix supports_constraint for IndicatorToMILPBridge

### LINKED ISSUE #2867: supports_constraint can be wrong
Check this out:
```Julia
julia> import Ipopt; import MathOptInterface as MOI

julia> model = MOI.instantiate(Ipopt.Optimizer; with_bridge_type=Float64)
MOIB.LazyBridgeOptimizer{IpoptMOIExt.Optimizer}
├ Variable bridges: none
├ Constraint bridges: none
├ Objective bridges: none
└ model: IpoptMOIExt.Optimizer
  ├ ObjectiveSense: FEASIBILITY_SENSE
  ├ ObjectiveFunctionType: MOI.ScalarAffineFunction{Float64}
  ├ NumberOfVariables: 0
  └ NumberOfConstraints: 0

julia> MOI.supports_constraint(
           model,
           MOI.VectorOfVariables,
           MOI.Indicator{MOI.ACTIVATE_ON_ONE,MOI.EqualTo{Float64}},
       )
true
```

This is wrong, because it assumes that the variable will be binary, which is not part of the constraint.

odow:
Closes #2867 

This needs solver tests. I don't know if I understand the full implication.

### DIFF
--- a/src/Bridges/Constraint/bridges/IndicatorToMILPBridge.jl
+++ b/src/Bridges/Constraint/bridges/IndicatorToMILPBridge.jl
@@ -81,7 +81,7 @@ end
 function MOI.Bridges.added_constrained_variable_types(
     ::Type{<:IndicatorToMILPBridge},
 )
-    return Tuple{Type}[(MOI.Reals,)]
+    return Tuple{Type}[(MOI.Reals,), (MOI.ZeroOne,)]
 end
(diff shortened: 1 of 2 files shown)


odow:
Failures are unrelated

PR #2873, opened 2025-10-28
Add the package UnoSolver.jl in solver_tests.yml

amontoison:


odow:
Actually, I guess we can add it more generally to the JuMP docs as well. I'll make a PR.

amontoison:
I think we can also uncomment the line for `MadNLP.jl` but I started to upgrade its MOI interface: https://github.com/MadNLP/MadNLP.jl/pull/509
I will open a PR for the next major release of `MadNLP.jl` (`v0.9.0`).

PR #2876, opened 2025-10-30
[docs] improve docstring of VectorNonlinearOracle

odow:
x-ref this discussion https://discourse.julialang.org/t/computing-hessian-of-the-langrangian-efficiently-with-di-jl/133462/11

### DIFF
--- a/src/sets.jl
+++ b/src/sets.jl
@@ -2702,7 +2702,11 @@ eval_jacobian(ret::AbstractVector, x::AbstractVector)::Nothing
 which fills the sparse Jacobian ``\\nabla f(x)`` into `ret`.
 
 The one-indexed sparsity structure must be provided in the `jacobian_structure`
-argument.
+argument and it must be independent of the point ``x``.
+
+`jacobian_structure` is not required to be sorted and it may contain duplicates,
+in which case the solver will combine the corresponding non-zero values in `ret`
+by adding them together.
 
 ## Hessian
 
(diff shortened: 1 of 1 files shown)


src/sets.jl
```diff
@@ -2702,7 +2702,11 @@ eval_jacobian(ret::AbstractVector, x::AbstractVector)::Nothing
 which fills the sparse Jacobian ``\\nabla f(x)`` into `ret`.
 
 The one-indexed sparsity structure must be provided in the `jacobian_structure`
-argument.
+argument and it must be independent of the point ``x``.
+
+`jacobian_structure` is not required to be sorted and it may contain duplicates,
+in which case the solver will combine the corresponding non-zero values in `ret`
+by adding them together.
```
franckgaga, line 2709: Did you meant "averaging them together" ? 

"Adding them together" feels like giving `[(1,1), (1,1), (1,1)]` and `[1.0, 1.0, 1.0]` would fill the Hessian with a `3.0` instead of a `1.0` at the index `1,1`.

src/sets.jl
```diff
@@ -2723,15 +2727,23 @@ which fills the sparse Hessian of the Lagrangian ``\\sum \\mu_i \\nabla^2 f_i(x)
 into `ret`.
 
 The one-indexed sparsity structure must be provided in the
-`hessian_lagrangian_structure` argument.
+`hessian_lagrangian_structure` argument and it must be independent of the point
+``x``.
+
+`hessian_lagrangian_structure` is not required to be sorted and it may contain
+duplicates, in which case the solver will combine the corresponding non-zero
+values in `ret` by adding them together.
```
franckgaga, line 2735: idem

src/sets.jl
```diff
@@ -2723,15 +2727,23 @@ which fills the sparse Hessian of the Lagrangian ``\\sum \\mu_i \\nabla^2 f_i(x)
...
+
+Any mix of lower and upper-triangular indices is valid. Elements `(i, j)` and
+`(j, i)`, if both present, are treated as duplicates.
```
franckgaga, line 2738: Great, it answers my question!

src/sets.jl
odow: No, I mean adding them together. It would fill in `3.0` at the index `1, 1`.

PR #2886, opened 2025-11-29
[docs] Fix LaTeX alignement in example of `VectorNonlinearOracle` docstring

franckgaga:
It was not displayed correctly in the generated HTML doc.


franckgaga:
Oh I made a mistake, I correct the alignement on the old equation before @odow docstring improvements. Let me correct this.

PR #2890, opened 2025-12-02
Add MOI.LagrangeMultiplier attribute

### LINKED ISSUE #2887: The dual of VectorNonlinearOracle
See https://discourse.julialang.org/t/lagrange-multipliers-of-a-vectornonlinearoracle-after-solve/134236

```Julia
julia> using JuMP, Ipopt

julia> set = MOI.VectorNonlinearOracle(;
           dimension = 2,
           l = [-Inf],
           u = [1.0],
           eval_f = (ret, x) -> (ret[1] = x[1]^2 + x[2]^2),
           jacobian_structure = [(1, 1), (1, 2)],
           eval_jacobian = (ret, x) -> ret .= 2.0 .* x,
           hessian_lagrangian_structure = [(1, 1), (2, 2)],
           eval_hessian_lagrangian = (ret, x, u) -> ret .= 2.0 .* u[1],
       )
VectorNonlinearOracle{Float64}(;
    dimension = 2,
    l = [-Inf],
    u = [1.0],
    ...,
)

julia> model = Model(Ipopt.Optimizer)
A JuMP Model
├ solver: Ipopt
├ objective_sense: FEASIBILITY_SENSE
├ num_variables: 0
├ num_constraints: 0
└ Names registered in the model: none

julia> set_silent(model)

julia> @variable(model, x[1:2])
2-element Vector{VariableRef}:
 x[1]
 x[2]

julia> @objective(model, Max, sum(x))
x[1] + x[2]

julia> @constraint(model, c, x in set)
c : [x[1], x[2]] ∈ VectorNonlinearOracle{Float64}(;
    dimension = 2,
    l = [-Inf],
    u = [1.0],
    ...,
)

julia> optimize!(model)

julia> value(x)
2-element Vector{Float64}:
 0.7071067834685273
 0.7071067834685273

julia> dual(c)
2-element Vector{Float64}:
 -0.9999999999982961
 -0.9999999999982961
```

What should `dual(c)` here be?

I based it on:
```Julia
julia> using JuMP, Ipopt

julia> model = Model(Ipopt.Optimizer)
A JuMP Model
├ solver: Ipopt
├ objective_sense: FEASIBILITY_SENSE
├ num_variables: 0
├ num_constraints: 0
└ Names registered in the model: none

julia> set_silent(model)

julia> @variable(model, x[1:2])
2-element Vector{VariableRef}:
 x[1]
 x[2]

julia> @objective(model, Max, sum(x))
x[1] + x[2]

julia> @constraint(model, c, sum(x.^2) <= 1)
c : x[1]² + x[2]² ≤ 1

julia> optimize!(model)

julia> value(x)
2-element Vector{Float64}:
 0.7071067834685273
 0.7071067834685273

julia> dual(c)
-0.7071067789033629

julia> dual(c)' * (2 * value(x))
2-element Vector{Float64}:
 -0.9999999999982961
 -0.9999999999982961
```

but the user doesn't have an easy way to access the multipliers on the individual rows.

cc @Robbybp

odow:
Closes #2887

x-ref https://github.com/jump-dev/Ipopt.jl/pull/521

Questions:

 - [x] should we have `get_fallback`?

### DIFF
--- a/src/Test/test_nonlinear.jl
+++ b/src/Test/test_nonlinear.jl
@@ -2459,7 +2459,7 @@ end
 function version_added(
     ::typeof(test_VectorNonlinearOracle_LagrangeMultipliers_MAX_SENSE),
 )
-    return v"1.47.0"
+    return v"1.48.0"
 end
 
 function test_VectorNonlinearOracle_LagrangeMultipliers_MIN_SENSE(
@@ -2538,5 +2538,5 @@ end
 function version_added(
     ::typeof(test_VectorNonlinearOracle_LagrangeMultipliers_MIN_SENSE),
 )
-    return v"1.47.0"
+    return v"1.48.0"
 end


Robbybp:
Is there any reason this is LagrangeMultiplier**s** (plural)? Just seems inconsistent with ConstraintDual. Should there also be a `LagrangeMutlipliersStart` attribute for initializing these multipliers?

odow:
I was wondering about the start. And yes, we could go with singular.

ghost:
Warm start is a nice-to-have but less important IMO, since we can always do the conversion manually with $\mu^\top J(x)$, which is not the case in the opposite direction

odow:
Nah, we've added an explicit warm-start for ``\mu``, not for ``y = \\mu^top J(x)``. We can't guarantee we can get a start for ``\mu`` based on a start for ``y``.

blegat:
If we only have solvers for which the attribute is easy to implement, I'd rather not have `get_fallback`. The attributes that have a fallback are the ones for which if we had to implement it in the solvers we would need to copy paste the code of the fallback. Here, the solver have more information than the fallback so they can directly give these multipliers.

odow:
Okay, I removed the fallback.

src/attributes.jl
```diff
@@ -3272,6 +3272,75 @@ function get_fallback(
...
+## Implementation
+
+Optimizers should implement the following methods:
+```
+MOI.get(::Optimizer, ::MOI.LagrangeMultiplier, ::MOI.ConstraintIndex)
+```
+They should not implement [`set`](@ref) or [`supports`](@ref).
+
+Solvers should implement [`LagrangeMultiplier`](@ref) only if they also
+implement the [`ConstraintDual`](@ref), and only if if the two values are
```
odow:
```suggestion
implement the [`ConstraintDual`](@ref), and only if the two values are
```

src/attributes.jl
```diff
@@ -3272,6 +3272,75 @@ function get_fallback(
...
 
+"""
+    LagrangeMultiplier(result_index::Int = 1)
+
+An [`AbstractConstraintAttribute`](@ref) for the Lagrange multiplier associated
+with a constraint.
+
+## Relationship to `ConstraintDual`
+
+In most cases, the value of this attribute is equivalent to
```
blegat: Remove

src/Test/test_nonlinear.jl
```diff
@@ -2382,3 +2382,161 @@ function setup_test(
...
+        end,
+    )
+    model.eval_variable_constraint_dual = false
+    return () -> model.eval_variable_constraint_dual = true
+end
+
+function version_added(
+    ::typeof(test_VectorNonlinearOracle_LagrangeMultipliers_MIN_SENSE),
+)
+    return v"1.47.0"
```
odow:
```suggestion
    return v"1.48.0"
```

src/Test/test_nonlinear.jl
```diff
@@ -2382,3 +2382,161 @@ function setup_test(
...
+        end,
+    )
+    model.eval_variable_constraint_dual = false
+    return () -> model.eval_variable_constraint_dual = true
+end
+
+function version_added(
+    ::typeof(test_VectorNonlinearOracle_LagrangeMultipliers_MAX_SENSE),
+)
+    return v"1.47.0"
```
odow:
```suggestion
    return v"1.48.0"
```

franckgaga:
@odow, just to be sure, I assume that's it not possible right now with MOI v1.48 to fetch the optimal Langrange multiplier on a real problem solved by `Ipopt.jl` (i.e. the attribute will need to be implemented in `Ipopt.jl` package first) ?

odow:
Give me a minute! Yes, we needed to release v1.48 first. Then I need to make changes to Ipopt.jl and KNITRO.jl. Only after they are released will you be able to call `get_attribute(model, MOI.LagrangeMultiplier(), c)`

franckgaga:
Yeah no problem, no hurries! I was just lost in the "attribute" concept (if it's somethings that should just work without any changes required on NLP solver side or not).

odow:
MOI can be confusing. Its main job is to be a thin API layer that just defines the attributes. This lets external Julia packages implement multiple dispatch without type piracy. In any other Julia community we could have created MathOptInterfaceBase.jl. 

But there's also a bunch of other heavy lifting code in MOI for doing related things, so it can be hard to know from the outside what are new API definitions that need work in other packages, and what are new features being added to MOI that are instantly accessible from every solver package.

Issue #2888, opened 2025-12-02
[FileFormats.MPS] slowing reading of .mps file


JunchengLiamLi:
The `read_from_file()` function is quite slow on my .mps file. It takes 33 seconds to build a model from the .mps file, while the C_API of HiGHS (call from Julia) takes only about the 8 seconds. Please see the below:

`@time model_JuMP = read_from_file("2025-1126-SCUC-noncon.mps");`
`33.343285 seconds (236.12 M allocations: 14.443 GiB, 25.32% gc time)`

`@time HiGHS.Highs_readModel(model_HiGHS,"2025-1126-SCUC-noncon.mps");`
` 7.832455 seconds (1.50 k allocations: 78.234 KiB, 0.05% compilation time)`

Just wondering if there is any way to build a JuMP.Model faster from .mps files.

Many thanks!

odow:
There are known overheads associated with `read_from_file`. See my recent talk on MOI: https://youtu.be/M31xoZGyj9w?si=Ma60aWbEAV4RAePt&t=1667

We could perhaps be a bit faster, but the times aren't really directly comparable between JuMP and HiGHS because JuMP is doing significantly more work.

A better question to ask is: why do you care that reading this file takes a few seconds? In most cases this shouldn't be a significant runtime part of any optimization-related workflows.

JunchengLiamLi:
Using `MOI.readModel()` is indeed faster than `JuMP.read_from_file()`. As shown in the following: 
`julia> @time dest = MOI.FileFormats.Model(format = MOI.FileFormats.FORMAT_MPS);`

`julia> @time MOI.read_from_file(dest, "2025-1126-SCUC-noncon.mps");`
` 22.695831 seconds (150.83 M allocations: 9.657 GiB, 29.85% gc time)`

>  A better question to ask is: why do you care that reading this file takes a few seconds? In most cases this shouldn't be a significant runtime part of any optimization-related workflows.

That is because in this scenario, the reading of mps files is used in a self-defined primal heuristic that aims to find a good feasible solution as soon as possible, instead of building a model, passing to solver and waiting for the result. Moreover, this heuristic could be revoked multiple times in one solution procedure. Therefore, a 20-second improvement in the speed of reading from .mps file every time is not trivial. 

[...]

PR #2892, opened 2025-12-02
[FileFormats.MPS] allow AbstractString when parsing MPS files

odow:
x-ref #2888

Before
```julia
julia> @time main();
  6.228992 seconds (67.45 M allocations: 4.029 GiB, 16.71% gc time)
```
After
```julia
julia> @time main();
  5.684912 seconds (33.12 M allocations: 2.878 GiB, 16.25% gc time)
```

Code
```Julia
function main()
    model = MOI.FileFormats.MPS.Model()
    MOI.read_from_file(model, "/Users/odow/Downloads/supportcase19.mps")
end
```

### DIFF
--- a/src/FileFormats/MPS/read.jl
+++ b/src/FileFormats/MPS/read.jl
@@ -34,7 +34,7 @@
 
 @enum(Sense, SENSE_N, SENSE_G, SENSE_L, SENSE_E, SENSE_UNKNOWN)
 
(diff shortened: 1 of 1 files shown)


odow:
Here's a small win. A better approach might be to completely refactor the parser into something that didn't read one line by line, split into string tokens, and then parse.

PR #2902, opened 2025-12-09
Allow creating model from parts

blegat:
We allow the users to create the model without using the MOI interface which might seem scary.
But this will probably only be used either by advanced users or using something like https://github.com/jump-dev/MatrixOptInterface.jl/pull/35

Needed for https://github.com/jump-dev/MatrixOptInterface.jl/pull/35

### DIFF
--- a/src/Utilities/matrix_of_constraints.jl
+++ b/src/Utilities/matrix_of_constraints.jl
@@ -306,7 +306,8 @@ MOI.is_empty(v::MatrixOfConstraints) = MOI.is_empty(v.sets)
 function _reset_caches!(v::MatrixOfConstraints{T}) where {T}
     v.caches =
         [Tuple{_affine_function_type(T, S),S}[] for S in set_types(v.sets)]
-    return v.are_indices_mapped = [BitSet() for _ in eachindex(v.caches)]
+    v.are_indices_mapped = [BitSet() for _ in eachindex(v.caches)]
+    return
 end
 
 function MOI.empty!(v::MatrixOfConstraints)
--- a/src/Utilities/sparse_matrix.jl
+++ b/src/Utilities/sparse_matrix.jl
@@ -184,12 +184,14 @@ function Base.convert(
 end
 
 _indexing(A::MutableSparseMatrixCSC) = A.indexing
+
 _indexing(::SparseArrays.SparseMatrixCSC) = OneBasedIndexing()
 
-const _SparseMatrixCSC{Tv,Ti} =
-    Union{MutableSparseMatrixCSC{Tv,Ti},SparseArrays.SparseMatrixCSC{Tv,Ti}}
-
-function _first_in_column(A::_SparseMatrixCSC, row::Integer, col::Integer)
+function _first_in_column(
+    A::Union{MutableSparseMatrixCSC,SparseArrays.SparseMatrixCSC},
+    row::Integer,
+    col::Integer,
+)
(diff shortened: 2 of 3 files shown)


odow:
Tests please

blegat:
I'll first make https://github.com/jump-dev/MatrixOptInterface.jl/pull/35 and then take a miniature version of https://github.com/jump-dev/MatrixOptInterface.jl/pull/35 as test here

src/Utilities/matrix_of_constraints.jl
```diff
@@ -292,13 +303,17 @@ function rows end
...
+end
+
+function MOI.empty!(v::MatrixOfConstraints)
     MOI.empty!(v.coefficients)
     empty!(v.constants)
     MOI.empty!(v.sets)
-    v.caches =
-        [Tuple{_affine_function_type(T, S),S}[] for S in set_types(v.sets)]
-    v.are_indices_mapped = [BitSet() for _ in eachindex(v.caches)]
+    _reset_caches!(v)
```
blegat, line 317: These caches actually only make sense for `MutableSparseMatrixCSC`, it doesn't make sense for a `SparseMatrixCSC` nor for a `MutableSparseMatrixCSR` but let's not complicate this PR too much

src/Utilities/sparse_matrix.jl
```diff
@@ -183,19 +183,21 @@ function Base.convert(
...
     range = SparseArrays.nzrange(A, col)
-    row = _shift(row, OneBasedIndexing(), A.indexing)
+    row = _shift(row, OneBasedIndexing(), _indexing(A))
     idx = searchsortedfirst(view(A.rowval, range), row)
     return get(range, idx, last(range) + 1)
 end
 
 function extract_function(
-    A::MutableSparseMatrixCSC{T},
+    A::_SparseMatrixCSC{T},
```
blegat: We can either implement it for `SparseMatrixCSC` here or implement it in `MatrixOptInterface`.
The issue of the second option is that it's type piracy and it's just a copy-paste.
One might then ask for `exctract_function` for `Base.Matrix` or `Transpose{SparseMatrixCSC{Tv,Ti}}` etc... But let's keep this PR simple. I think, implementing `Matrix`, `SparseMatrixCSC` and their adjoint/transpose should be good enough for 99% of the cases of this use case which is already niche ^^

blegat:
I don't understand the codecov missing lines, it's reporting the signature of function as not covered but then its body is covered...

odow:
Yeah the codecov thing is https://github.com/jump-dev/JuMP.jl/issues/4088 I don't understand

blegat:
No objection to merge then ?

src/Utilities/matrix_of_constraints.jl
```diff
@@ -292,13 +303,17 @@ function rows end
 
 MOI.is_empty(v::MatrixOfConstraints) = MOI.is_empty(v.sets)
 
-function MOI.empty!(v::MatrixOfConstraints{T}) where {T}
+function _reset_caches!(v::MatrixOfConstraints{T}) where {T}
+    v.caches =
+        [Tuple{_affine_function_type(T, S),S}[] for S in set_types(v.sets)]
+    return v.are_indices_mapped = [BitSet() for _ in eachindex(v.caches)]
```
odow:
```suggestion
    v.are_indices_mapped = [BitSet() for _ in eachindex(v.caches)]
    return
```

PR #2903, opened 2025-12-14
Add DISABLE_AMEND_COVERAGE_FROM_SRC: yes

odow:
Seeing if this is the cause of https://github.com/jump-dev/JuMP.jl/issues/4088


odow:
This fixes some but not all of the problem, at a significant disadvantage: https://github.com/JuliaCI/Coverage.jl?tab=readme-ov-file#a-note-for-advanced-users.

Closing for now. I'll leave the issue open because I need to actually dig into the coverage issue.

PR #2908, opened 2026-01-14
Fix various errors found by JET

odow:
x-ref https://github.com/jump-dev/MathOptInterface.jl/issues/2907

None of these are bugs or actual run-time errors. They're just things that JET doesn't like.


odow:
Let's see if anything happened: https://github.com/jump-dev/MathOptInterface.jl/actions/runs/21018607992

src/constraints.jl
```diff
@@ -95,6 +95,10 @@ An error indicating that the constant part of the function in the constraint
 """
 struct ScalarFunctionConstantNotZero{T,F,S} <: Exception
     constant::T
+
+    function ScalarFunctionConstantNotZero{T,F,S}(constant::T) where {T,F,S}
```
blegat, line 99: What's wrong with the default constructor ?

Issue #2912, opened 2026-01-18
Distance to VectorNonlinearOracle


Robbybp:
So we can use `primal_feasibility_report`. E.g., I want the following to run:
```julia
import MathOptInterface as MOI

function eval_f(ret::AbstractVector, x::AbstractVector)
    ret[1] = sum(x)
    ret[2] = x[1]^2 + x[2]^2 - x[3]^2
    ret[3] = x[1]^2 - x[2] * x[3]
    return
end

jacobian_structure = [
    (1, 1),
    (1, 2),
    (1, 3),
    (2, 1),
    (2, 2),
    (2, 3),
    (3, 1),
    (3, 2),
    (3, 3),
]

function eval_jacobian(ret::AbstractVector, x::AbstractVector)
    ret[1] = 1.0
    ret[2] = 1.0
    ret[3] = 1.0
    ret[4] = 2.0 * x[1]
    ret[5] = 2.0 * x[2]
    ret[6] = -2.0 * x[3]
    ret[7] = 2.0 * x[1]
    ret[8] = -1.0 * x[3]
    ret[9] = -1.0 * x[2]
    return
end

set = MOI.VectorNonlinearOracle(;
    dimension = 3,
    l = [1.0, -Inf, -Inf],
    u = [1.0, 0.0, 0.0],
    eval_f,
    jacobian_structure,
    eval_jacobian,
)

d = MOI.Utilities.distance_to_set([0, 0, 0], set)
```

However, finding the distance from `x` to `{x: l <= f(x) <= u}` is not generally possible. What I think we intuitively expect is the distance from `f(x)` to the box `[l,u]`. But this violates our convention (rule?) where distance is computed in `x`-space. Maybe we define a new kind of distance?

PR #2913, opened 2026-01-18
Distance to VectorNonlinearOracle

Robbybp:
Kind of addresses #2912. Here I just leave the "type of distance" undefined. This works for `primal_feasibility_report`.


odow:
Tricky, tricky, tricky.

I tend to think we make an exception. There's no way we can report the distance in `x` space.

We could potentially scale by `∇f`, so some sort of "projection of linearised-f"?

Or we just make an exception for this set.

PR #2914, opened 2026-01-18
[Utilities] add distance_to_set for VectorNonlinearOracle

### LINKED ISSUE #2912: Distance to VectorNonlinearOracle

odow:
Closes #2912
Closes #2913

### DIFF
--- a/src/Utilities/distance_to_set.jl
+++ b/src/Utilities/distance_to_set.jl
@@ -557,3 +557,25 @@ function distance_to_set(
     eigvals .= min.(zero(T), eigvals)
     return LinearAlgebra.norm(eigvals, 2)
 end
+
+"""
+    distance_to_set(
+        ::ProjectionUpperBoundDistance,
+        x::AbstractVector{T},
+        set::MOI.VectorNonlinearOracle,
+    ) where {T<:Real}
+
+!!! warning
+    This is not an upper bound on the projection in `x` space; instead, it
+    is a projection in `f(x)`-space.
+"""
+function distance_to_set(
+    ::ProjectionUpperBoundDistance,
+    x::AbstractVector{T},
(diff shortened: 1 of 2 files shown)


odow:
Thoughts @blegat?

PR #2918, opened 2026-01-26
[FileFormats.MOF] replace JSON3 by JSON.jl

### LINKED ISSUE #2917: Revert to JSON from JSON3?
Shall we revert 70b3c83c60fae7aa029cd6ed2aad891cb5b9c9ee?
As JSON3 is deprecated and JSON v1 is out. See https://github.com/quinnj/JSON3.jl/blob/main/README.md

odow:
Closes #2917

Note that JSON.jl@1.0 is effectively rewrite of JSON3.jl

### DIFF
--- a/test/FileFormats/MOF/MOF.jl
+++ b/test/FileFormats/MOF/MOF.jl
@@ -43,7 +43,7 @@ function _validate(filename::String)
         "r",
         MOI.FileFormats.AutomaticCompression(),
     ) do io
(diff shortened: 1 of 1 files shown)


odow:
Hmm. This would require us to drop <v1.9 support. Which is probably okay.

mlubin:
Fine with me. 1.10 has been LTS since October 2024

odow:
Did a benchmark, and the runtimes are identical.

PR #2919, opened 2026-01-26
Allow JSON@0.21

odow:
A follow-up to #2918

I just realised that Clarabel currently depends on JSON@0.21: https://github.com/oxfordcontrol/Clarabel.jl/pull/218

So we wouldn't be able to build the JuMP docs with this version of MOI.

I'll run solver-tests: https://github.com/jump-dev/MathOptInterface.jl/actions/runs/21344108668


odow:
We probably need the downgrade CI job to actually test this properly

PR #2923, opened 2026-01-28
[Utilities] add distance_to_set for MOI.Indicator

### LINKED ISSUE #2922: `primal_feasibility_report` fails with indicator constraints in `distance_to_set`
I am not sure whether this is a JuMP.jl problem calling the wrong method or a MathOptInterface.jl problem missing the correct method. I guess the latter one, so I report it here.

I want to find the constraint which violates a solution which should be valid according to what I want, but seems to be invalid according to what I have written. However, it crashes. I boiled it down to this:

```julia
julia> using JuMP

julia> model = Model();

julia> @variable(model, b, Bin);

julia> @variable(model, x);

julia> @constraint(model, b --> {x >= 0});

julia> primal_feasibility_report(model, Dict(x => 1.0, b => 1.0), atol=10^-9)
ERROR: distance_to_set using the distance metric MathOptInterface.Utilities.ProjectionUpperBoundDistance() for set type MathOptInterface.Indicator{MathOptInterface.ACTIVATE_ON_ONE, MathOptInterface.GreaterThan{Float64}} has not been implemented yet.
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:44
  [2] distance_to_set(d::MathOptInterface.Utilities.ProjectionUpperBoundDistance, ::Vector{…}, set::MathOptInterface.Indicator{…})
    @ MathOptInterface.Utilities ~/.julia/packages/MathOptInterface/03Qtw/src/Utilities/distance_to_set.jl:75
  [3] distance_to_set(point::Vector{…}, set::MathOptInterface.Indicator{…})
    @ MathOptInterface.Utilities ~/.julia/packages/MathOptInterface/03Qtw/src/Utilities/distance_to_set.jl:71
  [4] _distance_to_set(point::Vector{…}, set::MathOptInterface.Indicator{…}, ::Type)
    @ JuMP ~/.julia/packages/JuMP/7eD71/src/feasibility_checker.jl:214
  [5] _add_infeasible_constraints(model::Model, ::Type{…}, ::Type{…}, violated_constraints::Dict{…}, point_f::Function, atol::Float64)
    @ JuMP ~/.julia/packages/JuMP/7eD71/src/feasibility_checker.jl:143
  [6] primal_feasibility_report(point::Function, model::Model; atol::Float64, skip_missing::Bool)
    @ JuMP ~/.julia/packages/JuMP/7eD71/src/feasibility_checker.jl:107
  [7] primal_feasibility_report
    @ ~/.julia/packages/JuMP/7eD71/src/feasibility_checker.jl:99 [inlined]
  [8] #primal_feasibility_report#114
    @ ~/.julia/packages/JuMP/7eD71/src/feasibility_checker.jl:58 [inlined]
  [9] top-level scope
    @ REPL[6]:1
 [10] top-level scope
    @ REPL:1
Some type information was truncated. Use `show(err)` to see complete types.
```

odow:
Closes #2922

### DIFF
--- a/src/Utilities/distance_to_set.jl
+++ b/src/Utilities/distance_to_set.jl
@@ -579,3 +579,32 @@ function distance_to_set(
     set.eval_f(y, x)
     return LinearAlgebra.norm(y .- clamp.(y, set.l, set.u))
 end
+
+# This is the minimal L2-norm.
+function distance_to_set(
+    distance::ProjectionUpperBoundDistance,
+    x::AbstractVector{T},
+    set::MOI.Indicator{MOI.ACTIVATE_ON_ONE},
+) where {T<:Real}
+    _check_dimension(x, set)
+    return min(
+        # Distance of x[1] from 0
+        abs(x[1]),
+        # Distance of x[1] from 1 + distance to set
+        sqrt((1 - x[1])^2 + distance_to_set(distance, x[2], set.set)^2),
+    )
+end
+
+function distance_to_set(
+    distance::ProjectionUpperBoundDistance,
+    x::AbstractVector{T},
+    set::MOI.Indicator{MOI.ACTIVATE_ON_ZERO},
+) where {T}
+    _check_dimension(x, set)
+    return min(
+        # Distance of x[1] from 1
+        abs(one(T) - x[1]),
+        # Distance of x[1] from 0 + distance to set
+        sqrt(x[1]^2 + distance_to_set(distance, x[2], set.set)^2),
+    )
+end
--- a/test/Utilities/distance_to_set.jl
+++ b/test/Utilities/distance_to_set.jl
@@ -376,6 +376,30 @@ function test_vectornonlinearoracle()
     return
 end
 
+function test_indicator()
+    _test_set(
+        MOI.Indicator{MOI.ACTIVATE_ON_ZERO}(MOI.EqualTo(1.0)),
+        [0.0, 1.0] => 0.0,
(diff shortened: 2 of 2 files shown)


MagicMuscleMan:
Thanks a lot, @odow, to implement this so quickly!

Just do double check: Is it correct, that the distance is now limited to the interval `[0, 1]` if the indicator variable behaves well? I.e. independent of how large the distance of `set.set` is, due to the `min`, this will never exceed 1.
This feels incorrect, but might be irrelevant, if it's guaranteed that distances are always below 1 in all relevant cases.

Additionally: Is it on purpose that `x` is defined differently in both new methods? I would have expected either `x::AbstractVector{<:Real}` or `x::AbstractVector` for both methods, but neither of the two variants with `where` which you used, as `T` is not used and I currently don't see which extra specialization hint could be needed.

My thoughts might be incorrect as I don't know the package well enough, but maybe a comment would help the uninformed reader if the code stays as it is.

odow:
> Is it correct, that the distance is now limited to the interval [0, 1] if the indicator variable behaves well?

Yes. That follows from the definition of the set, since we can trivially satisfy feasibility by fixing the `z` variable to the "off" state. Our distance is "what is the L2 distance we can change the input vector by to satisfy feasibility, not what is the distance to the indicated constraint, conditional on the binary"

>  Is it on purpose that x is defined differently in both new methods?

Oops. No. I'l make a follow-up PR. I meant to add tests for the complex case, but I obviously forgot.

PR #2931, opened 2026-01-30
[Bridges] Fix computing Q=U'U in QuadtoSOCBridge when Q has 0 eigenvalue

### LINKED ISSUE #1971: Sparse LDL for QuadtoSOC
QuadtoSOC computes a cholesky factorization, which fails when the Q matrix is positive semidefinite but not positive definite:

https://github.com/jump-dev/MathOptInterface.jl/blob/8c8636f47cbcd0bda8494ac9dcb852df75e2bcc5/src/Bridges/Constraint/bridges/quad_to_soc.jl#L71-L85

Per @dpo we could use [LDLFactorizations](https://github.com/JuliaSmoothOptimizers/LDLFactorizations.jl) with a pivot tolerance of zero (or HSL/MUMPS, if we want external binary dependencies).

odow:
Closes #1971 

There aren't many good options here.

My trivial test case is `(x + y)^2 / 2`.

The permissive options all fail:
```julia
julia> using LDLFactorizations, LinearAlgebra, QDLDL, SparseArrays

julia> A = [1.0 1.0; 1.0 1.0]
2×2 Matrix{Float64}:
 1.0  1.0
 1.0  1.0

julia> Q = SparseArrays.sparse(A);
julia> LinearAlgebra.cholesky(Q)
ERROR: PosDefException: matrix is not positive definite; Factorization failed.
Stacktrace:

julia> LinearAlgebra.ldlt(Q)
ERROR: ZeroPivotException: factorization encountered one or more zero pivots. Consider switching to a pivoted LU factorization.
Stacktrace:

julia> QDLDL.qdldl(Q)
ERROR: Zero entry in D (matrix is not quasidefinite)
Stacktrace:
```

LDLFactorizations.jl works:
```Julia
julia> LDLFactorizations.ldl(Q)
LDLFactorizations.LDLFactorization{Float64, Int64, Int64, Int64}(true, false, false, 2, [2, -1], [1, 0], [2, 2], [1, 2], [1, 2], [1, 2, 2], Int64[], Int64[], [2], [1.0], [1.0, 0.0], [0.0, 0.0], [1, 1], 0.0, 0.0, 0.0, 2)
```
but it requires us to add a GPL dependency.

This PR implements an option I don't think we've previously considered: do something that isn't upper triangular...

Am I missing something obviously wrong?

### DIFF
--- a/src/Bridges/Constraint/bridges/QuadtoSOCBridge.jl
+++ b/src/Bridges/Constraint/bridges/QuadtoSOCBridge.jl
@@ -73,8 +73,20 @@ function bridge_constraint(
     if !less_than
         LinearAlgebra.rmul!(Q, -1)
     end
-    F = try
-        LinearAlgebra.cholesky(LinearAlgebra.Symmetric(Q))
+    # Construct the VectorAffineFunction. We're aiming for:
+    #  |          1  |
+    #  | -a^T x + ub | ∈ RotatedSecondOrderCone()
+    #  |     Ux + 0  |
+    # Start with the -a^T x terms...
+    vector_terms = MOI.VectorAffineTerm{T}[
+        MOI.VectorAffineTerm(
+            2,
+            MOI.ScalarAffineTerm(scale * term.coefficient, term.variable),
+        ) for term in func.affine_terms
+    ]
+    # Compute a sparse U such that U' × U == Q.
+    I, J, V = try
+        _compute_sparse_U(Q)
     catch
         throw(
             MOI.UnsupportedConstraint{typeof(func),typeof(set)}(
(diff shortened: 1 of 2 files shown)


mlubin:
I don't think the eivenvalue decomposition is wrong, just potentially expensive and, as a result, surprising to users.

odow:
The alternative here is that we're going to error. I don't know if people solve problems with super large Q matrices? And need to use this bridge, and worry about performance?

odow:
The suggestion from #1971 is to add LDLFactorizations as a package extension.

src/Bridges/Constraint/bridges/QuadtoSOCBridge.jl
```diff
@@ -129,6 +125,31 @@ function bridge_constraint(
...
+        J, I, V = SparseArrays.findnz(L)
+        # Then, we want to permute the columns of L'. The rows stay in the same
+        # order.
+        return I, p[J], V
+    end
+    # Cholesky failed. Use instead an eigen value decomposition. This is not
+    # triangular, but nothing says that it has to be.
+    E = LinearAlgebra.eigen(LinearAlgebra.Symmetric(Matrix(Q)))
+    if !all(isreal, E.values) || minimum(E.values) < 0
+        error("Matrix is not PSD")
```
blegat, line 147: Maybe give the minimal eigenvalue

blegat:
Using eigenvalues as a fallback sounds like a good idea, maybe we should allow a small tolerance. Historically, we have allowed `1e-10` to be considered as zero:
https://github.com/jump-dev/MathOptInterface.jl/blob/05cb2261b664404451e88d22bd5713901d3f049f/src/Bridges/Constraint/bridges/SquareBridge.jl#L110-L123
We can use a very tight tolerance and only loosen it upon request. Maybe starting with `0` is good but it should be easy to find example of reasonable PSD matrices having negative eigenvalues like `-1e-15`

PR #2932, opened 2026-01-30
Use ParallelTestRunner.jl to parallelize the tests

odow:
Previously, the longest job took about 25 minutes... let's see what this does

I also need to figure out how to bring back the check for duplicated test names. But maybe we can do that statically.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -70,13 +70,14 @@ testsuite = Dict{String,Expr}(
     file in joinpath.(root, filter(is_test_file, files))
 )
 
-@static if Sys.WORD_SIZE == 64
-    import MathOptInterface
-    import ParallelTestRunner
+import MathOptInterface
+import ParallelTestRunner
+import Test
+
+if Sys.WORD_SIZE == 64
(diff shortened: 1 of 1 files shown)


odow:
One failure is because of https://github.com/jump-dev/MathOptInterface.jl/pull/2934

test/runtests.jl
```diff
@@ -44,27 +44,39 @@ function. Thus, some warnings may appear in the wrong place.
...
+    file in joinpath.(root, filter(is_test_file, files))
+)
+
+@static if Sys.WORD_SIZE == 64
+    import MathOptInterface
+    import ParallelTestRunner
+    ParallelTestRunner.runtests(MathOptInterface, ARGS; testsuite, init_code)
+else
+    import Test
+    Test.@test "$filename" for (filename, _) in testsuite
```
odow:
```suggestion
    Test.@testset "$filename" for (filename, _) in testsuite
```

odow:
I'm going to merge this for now and keep experimenting. We're bottlenecked by x86 not parallelising in ParallelTestRunner.jl.

I wonder if we could just remove `Julia 1 - ubuntu-latest-x64`. It's already covered by both the Windows build and the x86 build.

PR #2936, opened 2026-02-05
Add an explicit test for method redefinitions instead of parsing logs

odow:
Hopefully this works better with parallelism than redirecting stderr.

### DIFF
--- a/test/General/test_errors.jl
+++ b/test/General/test_errors.jl
@@ -444,7 +444,7 @@ end
 
 # This function tests that all files in `/test` do not have redefined methods.
 function test_method_redefinition()
-    for (root, dirs, files) in walkdir("test")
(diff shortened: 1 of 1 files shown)


test/General/test_errors.jl
```diff
@@ -427,6 +427,31 @@ function test_GetAttributeNotAllowed_showerror()
...
+            end
+            push!(functions, fn_name)
+        end
+    end
+    return
+end
+
+# This function tests that all files in `/test` do not have redefined methods.
+function test_method_redefinition()
+    for (root, dirs, files) in walkdir("test")
```
odow:
```suggestion
    for (root, dirs, files) in walkdir(dirname(@__DIR__))
```

PR #2956, opened 2026-02-27
Fix error message for LDL factorization failures

odow:
Found by solver tests: https://github.com/jump-dev/MathOptInterface.jl/actions/runs/22469911691/job/65084327414

### DIFF
--- a/ext/MathOptInterfaceLDLFactorizationsExt.jl
+++ b/ext/MathOptInterfaceLDLFactorizationsExt.jl
@@ -21,7 +21,11 @@ function MOI.Bridges.Constraint.compute_sparse_sqrt_fallback(
 ) where {F<:MOI.ScalarQuadraticFunction,S<:MOI.AbstractSet}
     n = LinearAlgebra.checksquare(Q)
     factor = LDLFactorizations.ldl(Q)
-    if !LDLFactorizations.factorized(factor) || minimum(factor.D) < 0
+    # Ideally we should use `LDLFactorizations.factorized(factor)` here, but it
+    # has some false negatives. Instead we check that the factorization appeared
+    # to work. This is a heuristic. There might be other cases where check is
+    # insufficient.
+    if minimum(factor.D) < 0 || any(issubnormal, factor.D)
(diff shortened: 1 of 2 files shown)


odow:
I'm not sure what's going on here. `LDLFactorizations.factorized` returns `false`, but there is a valid factorisation:

```julia
julia> import LDLFactorizations

julia> import LinearAlgebra

julia> Q = [1.0 1.0; 1.0 1.0]
2×2 Matrix{Float64}:
 1.0  1.0
 1.0  1.0

julia> f = LDLFactorizations.ldl(Q)
LDLFactorizations.LDLFactorization{Float64, Int64, Int64, Int64}(true, false, false, 2, [2, -1], [1, 0], [2, 2], [1, 2], [1, 2], [1, 2, 2], Int64[], Int64[], [2], [1.0], [1.0, 0.0], [0.0, 0.0], [1, 1], 0.0, 0.0, 0.0, 2)

julia> @assert f.P == [1, 2]

julia> L = Matrix(f.L) .+ LinearAlgebra.I(2)
2×2 Matrix{Float64}:
 1.0  0.0
 1.0  1.0

julia> D = Matrix(f.D)
2×2 Matrix{Float64}:
 1.0  0.0
 0.0  0.0

julia> LDLFactorizations.factorized(f)
false

julia> L * D * L'
2×2 Matrix{Float64}:
 1.0  1.0
 1.0  1.0
```

cc @amontoison any ideas?

amontoison:
@odow I checked the code and we stop the factorization when we encounter `D[k] == 0` and we return `false` for "factorized".

But if `k == n`, we reached the end of the factorization and we can return `true`.

odow:
Is that intentional or a bug?

amontoison:
It was intentional because we can't do `ldiv!` or `\` if we have `D[k] == 0`.
But we can still do `mul!`.
I added another attribute to specify if we have a singularity of not.

odow:
Merging because this is an improvement for now. We can update when a new version of LDLFactorizations is released

PR #2961, opened 2026-03-02
[Bridges] wrap compute_sparse_sqrt in a try-catch

odow:
https://github.com/jump-dev/MathOptInterface.jl/pull/2960 found a flakey error. I assume it's because of our heuristic for checking whether LDLFactorizations succeeded. But we can fix it properly once a new version is released.

### DIFF
--- a/src/Bridges/Constraint/bridges/QuadtoSOCBridge.jl
+++ b/src/Bridges/Constraint/bridges/QuadtoSOCBridge.jl
@@ -80,7 +80,7 @@ function compute_sparse_sqrt_fallback(Q, ::F, ::S) where {F,S}
 end
 
 function compute_sparse_sqrt(Q, func, set)
-    # There's a big try-catch here because `cholesky` can fail even if
+    # There's a big try-catch here because Cholesky can fail even if
     # `check = false`. As one example, it currently (v1.12) fails with
     # `BigFloat`. Similarly, we want to guard against errors in
     # `compute_sparse_sqrt_fallback`.


joaquimg:
This might be good to comment well in the code. Or it can become a great candidate for Chesterton's fence.

test/Bridges/Constraint/test_QuadtoSOCBridge.jl
```diff
@@ -389,6 +389,7 @@ function test_compute_sparse_sqrt_edge_cases()
         BigFloat[1.0 0.0; 0.0 2.0],
         BigFloat[1.0 1.0; 1.0 1.0],
     ]
+        @show A
```
odow:
```suggestion
```

src/Bridges/Constraint/bridges/QuadtoSOCBridge.jl
```diff
@@ -80,22 +80,32 @@ function compute_sparse_sqrt_fallback(Q, ::F, ::S) where {F,S}
 end
 
 function compute_sparse_sqrt(Q, func, set)
-    factor = try
-        LinearAlgebra.cholesky(Q; check = false)
-    catch
-        msg = "There was an error computing a Cholesky decomposition"
+    # There's a big try-catch here because `cholesky` can fail even if
```
odow:
```suggestion
    # There's a big try-catch here because Cholesky can fail even if
```

PR #2965, opened 2026-03-06
Pivoted sparse Cholesky for QuadtoSOCBridge.

samuelsonric:
This PR adds the pivoted sparse Cholesky factorization algorithm in [CliqueTrees.jl](https://github.com/AlgebraicJulia/CliqueTrees.jl) as a fallback in QuadtoSOCBridge. This algorithm can handle singular matrices, and it is MIT licensed.

### Example

```julia-repl
julia> using CliqueTrees.Multifrontal, LinearAlgebra, LDLFactorizations

julia> A = [1 1 0; 1 1 0; 0 0 1]
3×3 Matrix{Int64}:
 1  1  0
 1  1  0
 0  0  1
```

The CliqueTrees factorization correctly reconstructs the matrix.

```julia-repl
julia> F = cholesky!(ChordalCholesky(A), RowMaximum());

julia> F.L * F.D * F.L'
3×3 Matrix{Float64}:
 1.0  1.0  0.0
 1.0  1.0  0.0
 0.0  0.0  1.0
```

The LDLFactorizations factorization does not.

```julia-repl
julia> G = ldl(A);

julia> UnitLowerTriangular(G.L) * G.D * UnitLowerTriangular(G.L)'
3×3 SparseArrays.SparseMatrixCSC{Int64, Int64} with 5 stored entries:
 1  1           ⋅
 1  1           ⋅
 ⋅  ⋅  5873643920
 ```
 
 ### Related Pull Requests
 
- https://github.com/jump-dev/MathOptInterface.jl/pull/2931
- https://github.com/jump-dev/MathOptInterface.jl/pull/2933

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -4,6 +4,7 @@ version = "1.49.0"
 
 [deps]
 BenchmarkTools = "6e4b80f9-dd63-53aa-95a3-0cdb28fa8baf"
+CliqueTrees = "60701a23-6482-424a-84db-faee86b9b1f8"
 CodecBzip2 = "523fee87-0ab8-5b00-afb7-3ecf72e48cfd"
 CodecZlib = "944b1d66-785c-5afd-91f1-9de20f533193"
 ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
@@ -22,10 +23,12 @@ Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 LDLFactorizations = "40e66cde-538c-5869-a4ad-c39174c6795b"
 
 [extensions]
+MathOptInterfaceCliqueTreesExt = "CliqueTrees"
 MathOptInterfaceLDLFactorizationsExt = "LDLFactorizations"
 
 [compat]
 BenchmarkTools = "1"
+CliqueTrees = "1.17"
 CodecBzip2 = "0.6, 0.7, 0.8"
 CodecZlib = "0.6, 0.7"
 ForwardDiff = "1"
@@ -45,9 +48,10 @@ Test = "1"
 julia = "1.10"
 
 [extras]
-LDLFactorizations = "40e66cde-538c-5869-a4ad-c39174c6795b"
+CliqueTrees = "60701a23-6482-424a-84db-faee86b9b1f8"
 JSONSchema = "7d188eb4-7ad8-530c-ae41-71a32a6d4692"
+LDLFactorizations = "40e66cde-538c-5869-a4ad-c39174c6795b"
 ParallelTestRunner = "d3525ed8-44d0-4b2c-a655-542cee43accc"
 
 [targets]
-test = ["LDLFactorizations", "JSONSchema", "ParallelTestRunner"]
+test = ["CliqueTrees", "LDLFactorizations", "JSONSchema", "ParallelTestRunner"]
--- a/ext/MathOptInterfaceCliqueTreesExt.jl
+++ b/ext/MathOptInterfaceCliqueTreesExt.jl
@@ -1,0 +1,35 @@
+# Copyright (c) 2017: Miles Lubin and contributors
+# Copyright (c) 2017: Google Inc.
+#
+# Use of this source code is governed by an MIT-style license that can be found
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by odow** [comment]
I need to think about this more. This will conflict if someone loads both CliqueTrees and LDLFactoriztions.

**Comment by samuelsonric** [comment]
Ah good point.

**Comment by odow** [comment]
I'm also not very keen on adding dependencies. I wonder if we can just document this somehow. It's such a small edge-case. Very, very few people will ever need this.

**Comment by odow** [comment]
We haven't released the LDL stuff yet (https://github.com/jump-dev/MathOptInterface.jl/pull/2955), so we have a small window to get this right. I'll take a look.

**Comment by samuelsonric** [comment]
Also takes `BigFloat`.

https://github.com/jump-dev/MathOptInterface.jl/pull/2959

**Comment by odow** [comment]
Are there situations when LDLFactorizations is preferable to CliqueTrees?

**Comment by samuelsonric** [comment]
Yes: if the matrix is extremely sparse it can be ~2x faster.

**Comment by odow** [comment]
Is there a better way to check if the factorisation failed other than testing `Q = U' * U`?

**Comment by odow** [comment]
For example, your `LinearAlgebra.issuccess` seems wrong here:

```Julia
julia> Q = [0 -1; -1 0]
2×2 Matrix{Int64}:
  0  -1
 -1   0

julia> G = LinearAlgebra.cholesky!(
           CliqueTrees.Multifrontal.ChordalCholesky(Q),
           LinearAlgebra.RowMaximum(),
       )
2×2 FChordalCholesky{:L, Float64, Int64} with 3 stored entries:
 0.0   ⋅ 
 0.0  0.0

julia> LinearAlgebra.issuccess(G)
true

julia> U = SparseArrays.sparse(G.U) * G.P
2×2 SparseArrays.SparseMatrixCSC{Float64, Int64} with 3 stored entries:
 0.0  0.0
  ⋅   0.0

julia> U' * U
2×2 SparseArrays.SparseMatrixCSC{Float64, Int64} with 4 stored entries:
 0.0  0.0
 0.0  0.0
```

samuelsonric:
LDLFactorizations halts early if it hits a zero pivot, so I think that you can just scan the diagonal.

CliqueTrees (with `RowMaximum`) does not halt early. However depending on the condition number etc. the factorization can be more or less accurate.

You could try solving Fx = b for some random b and compare Qx to b.

samuelsonric:
@odow

We truncate negative pivots to 0 when computing a pivoted factorization. I wasn't thinking about non-PSD matrices. I can change the behavior to something else.

odow:
So is `all(iszero, G.D)` sufficient to check for failure? Or I guess you might have some positive values and some negative values projected to zero so you really can tell only by comparing the matrices?

samuelsonric:
Currently, you cannot check for failure without solving a problem or reconstructing or something. However, I can push a patch tonight that sets `issuccess` to false if the algorithm encounters a sufficiently large negative pivot.

Also -- beware. `G.D` is the identity if you perform a cholesky factorization (`cholesky!`). It is nontrivial if you perform an LDLt factorization (`ldlt!`).

odow:
We just want a method that produces `A = B * B'` where `A` is PSD. Robustness is #1 priority. Sparsity is great. Performance is a lesser concern.

samuelsonric:
> We just want a method that produces `A = B * B'` where `A` is PSD. Robustness is #1 priority. Sparsity is great. Performance is a lesser concern.

Right. The current status is this: CliqueTrees is robust for PSD (moreso than LDLFactorizations) but might produce strange results for non-PSD.

Your breaking example (`Q = [0 -1; -1 0]`) is non-PSD.

odow:
Closing for now in favour of #2967. I need to discuss our options on the next JuMP developer call.

samuelsonric:
@odow makes sense. Is there something you need from me?

odow:
Nope. I think I understand the issues.

I'm also going to revert the LDLFactorizations stuff for the immediate future as well: https://github.com/jump-dev/MathOptInterface.jl/pull/2972

PR #2967, opened 2026-03-12
Add an extension for CliqueTrees.jl

odow:
Replacement for #2965

I'm really not sure about this. This is about the 10th thing I have prototyped locally. It's just a mess. And I don't know how to test the situation where the packages aren't loaded without something brittle like "run this file before running this one".

### DIFF
--- a/src/Bridges/Constraint/bridges/QuadtoSOCBridge.jl
+++ b/src/Bridges/Constraint/bridges/QuadtoSOCBridge.jl
@@ -177,7 +177,7 @@ function bridge_constraint(
     ]
     sqrt_ret = _compute_sparse_sqrt(LinearAlgebra.Symmetric(Q))
     if sqrt_ret === nothing
-        msg = _get_sqrt_error_message(is_defined(CliqueTrees()))
+        msg = _get_sqrt_error_message(is_defined(_CliqueTrees()))
         return throw(MOI.UnsupportedConstraint{typeof(func),typeof(set)}(msg))
     end
     for (i, j, v) in zip(sqrt_ret[1], sqrt_ret[2], sqrt_ret[3])


Project.toml
```diff
@@ -19,13 +19,16 @@ SpecialFunctions = "276daf66-3868-5448-9aa4-cd146d93841b"
...
+CliqueTrees = "60701a23-6482-424a-84db-faee86b9b1f8"
 LDLFactorizations = "40e66cde-538c-5869-a4ad-c39174c6795b"
 
 [extensions]
+MathOptInterfaceCliqueTreesExt = "CliqueTrees"
 MathOptInterfaceLDLFactorizationsExt = "LDLFactorizations"
 
 [compat]
 BenchmarkTools = "1"
+CliqueTrees = "1.17"
```
odow, line 30:
```suggestion
CliqueTrees = "1.13"
```

src/Utilities/sparse_square_root.jl
```diff
@@ -0,0 +1,72 @@
...
+
+By default, this function attempts to use a `LinearAlgebra.cholesky`. If that
+fails, it may optionally use various extension packages.
+
+These extension packages must be loaded before calling `compute_sparse_sqrt`.
+
+The extensions currently supported are:
+
+ * The LDL routine in `LDLFactorizations.jl`
+ * The pivoted cholesky in `CliqueTrees.jl`
```
odow:
```suggestion
 * The pivoted Cholesky in `CliqueTrees.jl`
```

src/Utilities/sparse_square_root.jl
```diff
@@ -0,0 +1,72 @@
...
+## Return value
+
+If successful, this function returns an `(I, J, V)` triplet of the sparse `A`
+matrix.
+
+If unsuccessful, this function returns `nothing`.
+
+## Extensions
+
+By default, this function attempts to use a `LinearAlgebra.cholesky`. If that
```
odow:
```suggestion
By default, this function attempts to use a Cholesky decomposition. If that
```

odow:
I now have maintain access to CliqueTrees (thanks @samuelsonric). So the question remains: should we use only CliqueTrees? Or should we also include LDLFactorizations? Using only CliqueTrees would be simpler in the long term.

mlubin:
I vote for using only CliqueTrees unless we have a specific reason to add LDLFactorizations. Being a bit faster in some cases isn't a sufficient reason unless this code path is shown to be a bottleneck.

odow:
Okay let me remove the LDL stuff. We can always add it back later.

PR #2972, opened 2026-03-17
Revert LDLFactorizations package extension

odow:
There is a bunch of stuff piling up that deserves a release: https://github.com/jump-dev/MathOptInterface.jl/pull/2955

But I'm not happy with the LDLFactorizations extension. Rather than releasing it and then needing to make a breaking change, I propose that we remove it temporarily while we come up with a better design.

There are currently upstream issues with LDLFactorizations.jl (https://github.com/JuliaSmoothOptimizers/LDLFactorizations.jl/pull/142), and we also need to decide whether we should accept multiple package extensions (https://github.com/jump-dev/MathOptInterface.jl/pull/2967), and if so, how they would work together.


odow:
We can discuss the options in next week's developer call.

PR #2981, opened 2026-03-26
[Bridges] change graph to favor Constraint bridges over Variable bridges

odow:
x-ref https://github.com/jump-dev/ParametricOptInterface.jl/issues/201#issuecomment-4125905135

- [x] I need to change a bunch of the associated tests
- [x] I need to run `solver-tests.yml`
    - [x] https://github.com/jump-dev/MathOptInterface.jl/actions/runs/23575853864

I wonder if we should go further: should we use variable bridges iff we cannot do anything else? This would apply for CSDP and free variables, and ParameterToEqualTo. But not much else.

```Julia
using Clarabel
begin
    inner = MOI.instantiate(Clarabel.Optimizer; with_cache_type = Float64)
    model = MOI.Bridges.full_bridge_optimizer(inner, Float64)
    x, _ = MOI.add_constrained_variable(model, MOI.GreaterThan(1.0))
    MOI.Bridges.print_active_bridges(model)
    print(inner)
end

 * Supported objective: MOI.ScalarAffineFunction{Float64}
 * Unsupported constraint: MOI.VariableIndex-in-MOI.GreaterThan{Float64}
 |  bridged by:
 |   MOIB.Constraint.VectorizeBridge{Float64, MOI.VectorAffineFunction{Float64}, MOI.Nonnegatives, MOI.VariableIndex}
 |  may introduce:
 |   * Supported constraint: MOI.VectorAffineFunction{Float64}-in-MOI.Nonnegatives
Feasibility

Subject to:

VectorAffineFunction{Float64}-in-Nonnegatives
 ┌               ┐
 │-1.0 + 1.0 v[1]│
 └               ┘ ∈ Nonnegatives(1)
```

### DIFF
--- a/src/Bridges/debug.jl
+++ b/src/Bridges/debug.jl
@@ -664,7 +664,7 @@ function print_active_bridges(
             # The constraint can be both variable bridged and constraint
             # bridged. Which is cheaper?
             v_node = b.variable_node[(S,)]
-            if bridging_cost(b.graph, v_node) <= bridging_cost(b.graph, c_node)
+            if MOI.Bridges.is_variable_edge_best(b.graph, v_node)
                 return print_active_bridges(io, b, S, offset)
             end
         else
--- a/src/Bridges/graph.jl
+++ b/src/Bridges/graph.jl
@@ -330,7 +330,22 @@ constrained on creation, or as a free variable followed by a constraint.
 """
 function is_variable_edge_best(graph::Graph, node::VariableNode)
     _compute_bellman_ford(graph)
-    return graph.variable_dist[node.index] == _dist(graph, node)
+    # This is the cost of adding a constrained variable
+    dist_as_variable = graph.variable_dist[node.index]
+    # This is the cost of adding the constraint, if we were to add it.
+    dist_as_constraint = INFINITY
+    # If free variables are bridged but the functionize bridge was not
+    # added, constraint_node is `ConstraintNode(INVALID_NODE_INDEX)`.
+    constraint_node = graph.variable_constraint_node[node.index]
+    if constraint_node.index != INVALID_NODE_INDEX
+        dist_as_constraint = _dist(graph, constraint_node)
+        if dist_as_constraint != INFINITY
+            dist_as_constraint += graph.variable_constraint_cost[node.index]
+        end
+    end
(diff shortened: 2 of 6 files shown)


odow:
@blegat this seems to have worked okay

blegat:
I don't think we can go further, there a few cones for which there exists both a variable bridge and a constraint bridge. You can always use the constraint bridge but it is going to be super-inefficient. In conic programming, variable bridge are really necessary.
Here, we added a `+1` if we were to, instead of creating a variable bridge, add it as free variables and then add a normal constraint.
This actually feels correct to me now that I see it. I remember why I did it like this.
The reason this will usually not matter is that conic bridges will usually support either constrained variables or affine conic constraints but not both so one bridge cost will be much larger than the other.
This won't be the case though for Mosek or if you use a MOI.Utilities.Model that support both.
There, if you have a variable bridge that can exploit the fact that the constraint is added as constrained variable to do something better, it would be preferable to add it as variable bridge.
But in practice, for these cases, we maybe usually only have a variable bridge, but I can't guarantee it 100% without looking at all bridges we have.
Another aspect is that sometimes for the same set we have both a constraint and variable bridges. In that case, we also have a variable bridge because applying the constraint bridge for a solver in kernel form like CSDP would be inefficient, it's the example of `HermitianToSymmetricPSDBridge` for instance. For these one, this change is ok because for those solver the bridge cost will be much larger for the constraint bridge. So this PR might be ok.
I am actually close to finish updating SumOfSquares so if we could wait for the new version to be released to check it does not break SumOfSquares, I would prefer it though, SumOfSquares relies heavily on variable bridges.

odow:
> I am actually close to finish updating SumOfSquares so if we could wait for the new version to be released to check it does not break SumOfSquares

Yeah, this can wait.

odow:
https://github.com/jump-dev/MathOptInterface.jl/actions/runs/25088292759

odow:
@blegat SumOfSquares is failing with this PR: https://github.com/jump-dev/MathOptInterface.jl/actions/runs/25088292759/job/73508535218

blegat:
That's what I feared, I'll investigate

odow:
https://github.com/jump-dev/MathOptInterface.jl/actions/runs/29546714242

PR #2985, opened 2026-04-01
Add bridge attribute for choosing the Cholesky backend

blegat:
Inspired by https://github.com/blegat/ComplementOpt.jl/pull/29

### DIFF
--- a/src/Bridges/Constraint/bridges/QuadtoSOCBridge.jl
+++ b/src/Bridges/Constraint/bridges/QuadtoSOCBridge.jl
@@ -5,6 +5,29 @@
 # in the LICENSE.md file or at https://opensource.org/licenses/MIT.
 
 """
+    QuadtoSOCSquareRoot <: MOI.AbstractConstraintAttribute
+
+Constraint attribute to override the square root method used by
+[`QuadtoSOCBridge`](@ref) for a specific quadratic constraint.
+
+When set, this takes precedence over the default behavior of trying all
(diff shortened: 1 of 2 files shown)


odow:
I vote no. Let's wait a while to see if we need this. There is no benefit to making the bridge more complicated just for the sake of it.

src/Bridges/Constraint/bridges/QuadtoSOCBridge.jl
```diff
@@ -175,10 +255,15 @@ function bridge_constraint(
             MOI.ScalarAffineTerm(scale * term.coefficient, term.variable),
         ) for term in func.affine_terms
     ]
-    sqrt_ret = _compute_sparse_sqrt(LinearAlgebra.Symmetric(Q))
+    Q_sym = LinearAlgebra.Symmetric(Q)
+    sqrt_ret = if bridge.method !== nothing
+        _compute_sparse_sqrt(bridge.method, Q_sym)
```
odow, line 260: For example, adding this attribute would also require us to make these methods public and commit to their API, etc.

odow:
It took 4 years to close https://github.com/jump-dev/MathOptInterface.jl/issues/1971 with few, if any, people complaining. I don't think it is a high priority.

mlubin:
Agreed this additional complexity seems premature.

odow:
Closing as won't fix. It's nice to know that there is an approach that can work with the bridges, but I don't think that now is the time to add it.

PR #2988, opened 2026-04-03
Throw GetAttributeNotAllowed in AbstractBridge getter

blegat:
If the user sets `cannot_unbridge = true` in `MOI.Bridges.runtests` then it is allowed to not implement `MOI.ConstraintSet` and `MOI.ConstraintFunction` (since these are actually mostly useless to implement).
This was working for `SetMapBridge` thanks to
https://github.com/jump-dev/MathOptInterface.jl/blob/f604fc1b36847dc28556c4299156087097c591b5/src/Bridges/Constraint/set_map.jl#L116-L119
but actually not for classical `AbstractBridge` because it was throwing an `ArgumentError`. This PR fixes this.


odow:
https://github.com/jump-dev/MathOptInterface.jl/actions/runs/24013567368

PR #2989, opened 2026-04-08
Add model function

blegat:
This allows AD backend to request another `Model`. Needed for https://github.com/blegat/ArrayDiff.jl/pull/39

### DIFF
--- a/src/Nonlinear/model.jl
+++ b/src/Nonlinear/model.jl
@@ -78,14 +78,14 @@ function set_objective(model::Model, ::Nothing)
 end
 
 """
-    nonlinear_model(backend::AbstractAutomaticDifferentiation)
+    model(backend::AbstractAutomaticDifferentiation)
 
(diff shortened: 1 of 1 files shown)


odow:
This needs some serious discussion…

blegat:
Conclusion after JuMP-dev call: let's first figure out if we can modify `MOI.Nonlinear.Model` so that it can accomodate `ArrayDiff` as well. In the meantime, ArrayDiff will continue with a forked version to freely experiment and the PRs https://github.com/JuliaSmoothOptimizers/NLPModelsJuMP.jl/pull/229 and https://github.com/jump-dev/NLopt.jl/pull/273 will be experimental and remain unmerged

odow:
Conclusion from JuMP-dev call: We'll accept this if Ipopt can error when the AD backend changes and the inner model is a different type and it is not empty.

PR #3001, opened 2026-05-06
[Bridges] fix use of BridgeCost when computing supports in LazyBridgeOptimizer

blegat:
## Abstract description

Consider layers: `LazyBridgeOptimizer(Dualization(LazyBridgeOptimizer))` (with possible caching layers in between).
If you have a set `S` that does not implement `dual_set` but that has bridges, then a constraint `F`-in-`S` with this set are not supported by the dualization layer so it will be bridged by the outer layer. Now, all the other constraints that ##are supported by dualization will just go through the dualization layer will bridging cost of 0, even though they may be bridged by the inner bridges. Because of this, the bridge choice for `F`-in-`S` will be completely blind and might be sub-optimal.

## Concrete example

The whole idea detailed in https://github.com/jump-dev/SumOfSquares.jl/issues/205 explains that we have a `KernelBridge` and an `ImageBridge` and the right one will be picked thanks to the bridge cost

Written by Claude, ~~I haven't checked yet~~. I have now iterated, looks good.

### DIFF
--- a/src/Bridges/lazy_bridge_optimizer.jl
+++ b/src/Bridges/lazy_bridge_optimizer.jl
@@ -255,7 +255,9 @@ function node(
     @nospecialize(b::LazyBridgeOptimizer),
     @nospecialize(S::Type{<:MOI.AbstractSet}),
 )
-    # If we support the set, the node is 0.
+    # If we support the set, the node is 0 unless the inner model reports a
+    # non-zero `VariableBridgingCost` (which can happen when the inner model is
+    # itself a bridge optimizer that needs to bridge `S`).
     if (
         S <: MOI.AbstractScalarSet &&
         MOI.supports_add_constrained_variable(b.model, S)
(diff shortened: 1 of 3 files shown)


odow (commented):
Missing a test when the CachingOptimizer has `NO_OPTIMIZER`.

I'll run solver-tests:

https://github.com/jump-dev/MathOptInterface.jl/actions/runs/25891420946

odow:
This breaks a lot of packages

blegat:
Looking at the breaking, some are already failing: https://github.com/jump-dev/MathOptInterface.jl/actions/runs/25967424335
For the other ones, it seems that layers were incorrectly implementing the bridging cost attributes.

- [x] https://github.com/jump-dev/Dualization.jl/pull/227
- [x] https://github.com/jump-dev/ParametricOptInterface.jl/pull/240
- [x] https://github.com/jump-dev/PolyJuMP.jl/pull/146
- [x] https://github.com/jump-dev/Convex.jl/pull/729

We should release patch version of these packages before releasing this but then it should be fine. It might break MOI layers we are not testing with solver-tests though :/
The thing is that they are not implementing the bridging cost correctly so in a sense we're catching an existing bug.
What we could do is first to a release that adds tests to catch such bugs.

odow:
Is this ready to go?

odow:
https://github.com/jump-dev/MathOptInterface.jl/actions/runs/26076144297

odow:
Okay, this is now just the LazyBridgeOptimizer stuff.

odow:
Thoughts on merging this and tagging a new release?

PR #3003, opened 2026-05-17
Fix basic tests with power cone

blegat:
Detected when doing https://github.com/jump-dev/MathOptInterface.jl/pull/3002


odow (approved):
So approve, but this was somewhat of an open question. In theory even a solver supporting other `T` could support `Float64` here.

blegat:
Yes but `Utilities.Model{BigFloat}` did not support it

PR #3004, opened 2026-05-19
Fix various getters for VariableBridgingCost and ConstraintBridgingCost

blegat:
The tests were failing in https://github.com/jump-dev/MathOptInterface.jl/pull/3002, this is the fix by Claude


odow:
Thief 😄 i Already fixed this in #3001

blegat:
Ahah ok, but I though we would first make a release without https://github.com/jump-dev/MathOptInterface.jl/pull/3001, so we need to add the fixes for UniversalFallback and CachingOptimizer on master and leave the PR that changes LazyBridgeOptimizer for later

PR #3008, opened 2026-06-05
Add canonical

blegat:
Extracted from https://github.com/jump-dev/MathOptInterface.jl/pull/2803

I'm not too convinced by the current approach.

For `canonical`, I think we need to implement `canonical` with a cache that checks that it creates aliases when it sees a second time an alias to the same sub-expression.

### DIFF
--- a/src/Utilities/functions.jl
+++ b/src/Utilities/functions.jl
@@ -1056,6 +1056,32 @@ function canonical(f::MOI.AbstractFunction)
     return g
 end
 
+# Workaround: both `is_canonical` and `canonicalize!` would be slow otherwise
+canonical(f::MOI.ScalarNonlinearFunction) = f
+
+# This is maybe the right fix, but commented out to be sure to isolate
+# JuMP-side of profiling in https://github.com/jump-dev/JuMP.jl/pull/4032
+# We still need to figure out what's the right fix here anyway
+#function canonical(f::MOI.ScalarNonlinearFunction)
+#    cache = Dict{MOI.AbstractScalarFunction,MOI.AbstractScalarFunction}()
+#    # Don't use recursion here. This gets called for all scalar nonlinear
+#    # constraints.
+#    stack = Any[arg for arg in f.args]
+#    while !isempty(stack)
+#        arg = pop!(stack)
+#        if arg isa MOI.ScalarNonlinearFunction
+#            for a in arg.args
+#                push!(stack, a)
+#            end
+#        else
+#            if !is_canonical(arg)
+#                return false
+#            end
+#        end
+#    end
+#    return true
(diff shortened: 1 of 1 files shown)


src/Utilities/functions.jl
```diff
@@ -1056,6 +1056,29 @@ function canonical(f::MOI.AbstractFunction)
     return g
 end
 
+# Workaround: both `is_canonical` and `canonicalize!` would be slow otherwise
+canonical(f::MOI.ScalarNonlinearFunction) = f
```
odow, line 1060: Ask Claude why they would add this method when there is the one below 😆

blegat: I did it pre-Claude actually 😅 this line is just a temporary hack to isolate things in profiling

src/Utilities/functions.jl
```diff
@@ -1056,6 +1056,32 @@ function canonical(f::MOI.AbstractFunction)
...
+#                push!(stack, a)
+#            end
+#        else
+#            if !is_canonical(arg)
+#                return false
+#            end
+#        end
+#    end
+#    return true
+#end
```
odow, line 1083: Delete this code? It isn't correct regardless, because the function is `canonical` but it returns a `::Bool` instead of the canonicalised function.

odow: Why not change the `is_canonical` function? With this change it can be that `is_canonical(canonical(f)) == false`

odow: What's the underlying problem? That we walk the expression graph?

blegat: I wrote this in July 2025, I don't remember much, I will need to look at it again but let's first focus on merging https://github.com/jump-dev/JuMP.jl/pull/4032, then we'll look at the MOI level

src/Utilities/functions.jl
```diff
@@ -1056,6 +1056,32 @@ function canonical(f::MOI.AbstractFunction)
...
 end
 
+# Workaround: both `is_canonical` and `canonicalize!` would be slow otherwise
+canonical(f::MOI.ScalarNonlinearFunction) = f
+
+# This is maybe the right fix, but commented out to be sure to isolate
+# JuMP-side of profiling in https://github.com/jump-dev/JuMP.jl/pull/4032
+# We still need to figure out what's the right fix here anyway
+#function canonical(f::MOI.ScalarNonlinearFunction)
+#    cache = Dict{MOI.AbstractScalarFunction,MOI.AbstractScalarFunction}()
```
blegat, line 1066: The goal of this was to write a version with this cache but I first want to check that this is really necessary

PR #3009, opened 2026-06-09
[Utilities] fix use of recursion in canonicalize!(::ScalarNonlinearFunction)

odow:
Alternative to #3008

### DIFF
--- a/src/Utilities/functions.jl
+++ b/src/Utilities/functions.jl
@@ -1072,9 +1072,17 @@ function canonicalize!(
 end
 
 function canonicalize!(f::MOI.ScalarNonlinearFunction)
-    for (i, arg) in enumerate(f.args)
-        if !is_canonical(arg)
-            f.args[i] = canonicalize!(arg)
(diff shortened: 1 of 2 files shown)


blegat:
Why is it an alternative ? Do you preserve aliases ?

odow:
Your one doesn't do any. This PR fixes the use of recursion. There's an open question about aliases etc.

I think this PR would automatically account for aliases because if it reaches duplicate `ScalarAffineFunction` that need canonicalisation, then the first will update in place and we won't need to check the second.

blegat:
Good point, `canonical!` is alias-friendly. I opened https://github.com/jump-dev/MathOptInterface.jl/pull/3008 just to make that branch visible as it was needed in https://github.com/jump-dev/JuMP.jl/pull/4032 but it's indeed not implementing the alias-preservation at all yet. I think this PR is an improvement nevertheless so we can merge it

blegat (approved):

odow:
I'll fix the formatting in a separate PR.

PR #3022, opened 2026-07-08
[Utilities] fix ProductOfSets to handle zero-dimensional rows

### LINKED ISSUE #3019: ProductOfSets is wrong when a constraint has zero rows
This took me a while to track down. It is the cause of the failure in https://github.com/jump-dev/JuMP.jl/pull/4188#issuecomment-4849909737

The chain goes:

 * PiecewiseLinearOpt attempts to solve a convex MINLP
 * It uses Pajarito, with HiGHS + COSMO
 * Pajarito doesn't have special handling for zero-dim rows. It tries to modify them as if they're normal: https://github.com/jump-dev/Pajarito.jl/blob/83048f2225eac4b8d4226a9c686658b885b2b6d7/src/models.jl#L209-L214
 * COSMO doesn't support modification, so we throw away the optimizer and try to modify the cache
 * The cache says it doesn't have this constraint, even though we previously added it via `add_constraint`

Phew!

Here's a MWE:

```Julia

julia> MOI.Utilities.@product_of_sets(_Cones, MOI.Zeros)

julia> const OptimizerCache{T} = MOI.Utilities.GenericModel{
           T,
           MOI.Utilities.ObjectiveContainer{T},
           MOI.Utilities.VariablesContainer{T},
           MOI.Utilities.MatrixOfConstraints{
               T,
               MOI.Utilities.MutableSparseMatrixCSC{
                   T,
                   Int,
                   MOI.Utilities.OneBasedIndexing,
               },
               Vector{T},
               _Cones{T},
           },
       };

julia> model = OptimizerCache{Float64}();

julia> f = MOI.VectorAffineFunction(MOI.VectorAffineTerm{Float64}[], Float64[]);

julia> c = MOI.add_constraint(model, f, MOI.Zeros(0))
MathOptInterface.ConstraintIndex{MathOptInterface.VectorAffineFunction{Float64}, MathOptInterface.Zeros}(1)

julia> MOI.Utilities.final_touch(model, nothing)

julia> MOI.get(model, MOI.ListOfConstraintTypesPresent())
Tuple{Type, Type}[]

julia> F, S = MOI.VectorAffineFunction{Float64}, MOI.Zeros
(MathOptInterface.VectorAffineFunction{Float64}, MathOptInterface.Zeros)

julia> MOI.get(model, MOI.NumberOfConstraints{F,S}())
0

julia> MOI.get(model, MOI.ListOfConstraintIndices{F,S}())
MathOptInterface.ConstraintIndex{MathOptInterface.VectorAffineFunction{Float64}, MathOptInterface.Zeros}[]
```

odow:
Closes #3019

### DIFF
--- a/src/Utilities/product_of_sets.jl
+++ b/src/Utilities/product_of_sets.jl
@@ -170,25 +170,14 @@ macro product_of_sets(name, set_types...)
         mutable struct $(esc_name){$(T)} <:
                        $MOI.Utilities.OrderedProductOfSets{$(T)}
             """
-            `dimension[i][j]` is the dimension of `ConstraintIndex().value` `j`
-            of set type `i`.
-            """
-            dimension::Vector{Vector{Int}}
+            `rows[i][j]` corresponds to constraint `j` of set type `i`.
 
+            The value depends on `final_touch`:
+             * Before `final_touch`, these are `1:dimension` of the constraint
+             * After `final_touch`, these are the 1-indexed rows of the full
+               constraint matrix
             """
-            `offset[i][j]` is the 0-indexed row offset of constraint `j` of set
-            type `i`. The rows are therefore `offset[i][j] + 1:dimension[i][j]`.
-
-            The `offset` vector gets created during `final_touch`.
-            """
-            offset::Vector{Vector{Int}}
-
-            """
-            The total number of rows in the sets.
-
-            This value gets set during `final_touch`.
-            """
-            num_rows::Int
+            rows::Vector{Vector{UnitRange{Int}}}
 
             """
             A sanity bit to check that we don't call functions out-of-order.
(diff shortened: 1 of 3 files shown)


odow:
Okay, I'm happy with this. It's just simpler all round. It now has one `UnitRange{Int}` for each constraint. No more dictionary lookups, and we handle zero-dimensional rows.

test/Utilities/test_product_of_sets.jl
```diff
@@ -290,7 +290,94 @@ function test_vector_ListOfConstraintIndices2()
     S = MOI.Utilities.set_types(sets)[2]
     VAF = MOI.VectorAffineFunction{Float64}
     indices = MOI.get(sets, MOI.ListOfConstraintIndices{VAF,S}())
-    @test indices == MOI.ConstraintIndex{VAF,S}.([1, 3, 6, 8])
+    @test indices == MOI.ConstraintIndex{VAF,S}.([1, 2, 3, 4])
```
odow, line 293: The downside is that you might argue this is breaking, because we change the constraint indices. But I don't think we ever document that the index values are part of the public API. We might re-order them in any way. They just need to make sure that we iterate over them in the order that they were added (within a constraint type).

src/Utilities/product_of_sets.jl
```diff
@@ -170,91 +170,74 @@ macro product_of_sets(name, set_types...)
...
-function MOI.dimension(sets::OrderedProductOfSets)
+function MOI.dimension(sets::OrderedProductOfSets)::Int
     @assert sets.final_touch
-    if isempty(sets.num_rows)
-        # There is no set type
-        return 0
-    else
-        return sets.num_rows[end]
-    end
+    return sum(num_rows(sets, S) for S in set_types(sets); init = 0)
```
blegat: Isn't it simply like `rows[end][end][end]` ?

odow: That might be empty. I guess we could iterate backwards to find the last nonempty. But this gets called only once it isn’t very important

odow: Fixed.

odow:
https://github.com/jump-dev/MathOptInterface.jl/actions/runs/29452497977

odow:
https://github.com/jump-dev/MathOptInterface.jl/actions/runs/29458111448

odow:
SCS is using the non-public API: https://github.com/jump-dev/SCS.jl/blob/a2674aa5f6adf4692ffb87670f603363f3592fd5/src/MOI_wrapper/MOI_wrapper.jl#L442-L443

https://github.com/jump-dev/SCS.jl/pull/358

odow:
We can fix SCS, but I've also added a backward compatible `getproperty` call to this PR.

https://github.com/jump-dev/MathOptInterface.jl/actions/runs/29459997140

odow:
Sigh. 

DiffOpt has: https://github.com/jump-dev/DiffOpt.jl/blob/master/src/product_of_sets.jl

This is why we shouldn't have all these tricky little helpers everywhere. Consumers use them in all sorts of unexpected ways and we get boxed into how we can support them.

odow:
@blegat we should chat about this. I don't really know how to get these simplifications in MOI without breaking DiffOpt.

We probably need to keep the current fields and just add some additional bookkeeping for the zero-dim rows. And then also go and fix this in DiffOpt.

odow:
> We probably need to keep the current fields and just add some additional bookkeeping for the zero-dim rows. And then also go and fix this in DiffOpt.

But that won't work because DiffOpt won't have those fields...

I've opened a PR in DiffOpt to remove the usage: https://github.com/jump-dev/DiffOpt.jl/pull/372

PR #3025, opened 2026-07-18
Fix final_touch order

### LINKED ISSUE #1980: Ddding need_final_touch bridges in final_touch
If a bridge adds a constraint that needs a bridge that `need_final_touch` then it will modify the `needs_final_touch` being iterated over
https://github.com/jump-dev/MathOptInterface.jl/blob/95a73a074c08a2e46820cae4a6522b34de6bb460/src/Bridges/Constraint/map.jl#L321-L334
In that case, we might not call `final_touch` on that bridge.

blegat:
Written by claude, still need to check

Closes https://github.com/jump-dev/MathOptInterface.jl/issues/1980

### DIFF
--- a/src/Bridges/Constraint/map.jl
+++ b/src/Bridges/Constraint/map.jl
@@ -20,14 +20,23 @@ struct Map <: AbstractDict{MOI.ConstraintIndex,AbstractBridge}
     # of creation so we need `OrderedDict` and not `Dict`.
     # For `VariableIndex` constraints: (variable, set type) -> bridge
     single_variable_constraints::OrderedDict{Tuple{Int64,Type},AbstractBridge}
-    needs_final_touch::OrderedDict{Type,OrderedSet}
+    # The bridges that need a final touch, mapped to the "creation order" of the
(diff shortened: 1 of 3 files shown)


src/Bridges/Constraint/map.jl
```diff
@@ -310,41 +321,69 @@ function add_key_for_bridge(
...
-function _final_touch(bridges, model)
-    for bridge in bridges
-        MOI.Bridges.final_touch(bridge, model)
-    end
-    return
+# Return the bridges that need a final touch, sorted by increasing creation
+# order (the order in which their `bridge_constraint` was called).
+function _sorted_needs_final_touch(map::Map)
+    bridges = collect(keys(map.needs_final_touch))
+    sort!(bridges; by = bridge -> map.needs_final_touch[bridge])
```
odow, line 364: This probably needs to be the counter, not whether it needs final touch

odow: Now `needs_final_touch` is an integer, so this part is correct

PR #3027, opened 2026-07-24
[Bridges] fix support for nested final_touch bridges

### LINKED ISSUE #1980: Ddding need_final_touch bridges in final_touch

odow:
Now that I understand the fix in #3025, this is a hopefully simpler fix.

Closes #1980!
Closes #3025

### DIFF
--- a/src/Bridges/bridge_optimizer.jl
+++ b/src/Bridges/bridge_optimizer.jl
@@ -1968,8 +1968,10 @@ function add_bridged_constraint(b, BridgeType, f, s)
     # To make this happen, we remember how many bridges are already in the
     # vector, and then we insert the bridge into the vector. This is an O(N)
     # operation in the number of new bridges, but it's a pretty rare edge-case,
-    # and the number of new bridges should be small. (And in most common bridges
-    # that need final touch, like the ToMILP bridges, N=0.)
+    # and the number of new bridges should be small. (And most of the time, N=0,
+    # because most bridges that need final touch don't add any bridges that
+    # themselves need final touch. Good examples are the ToMILP bridges that
+    # need variable bounds.)
     n_final_touch = length(map.needs_final_touch)
     bridge = Constraint.bridge_constraint(BridgeType, recursive_model(b), f, s)
     if MOI.Bridges.needs_final_touch(bridge)
--- a/src/Bridges/bridge_optimizer.jl
+++ b/src/Bridges/bridge_optimizer.jl
@@ -1961,14 +1961,27 @@ function MOI.supports_constraint(
 end
 
 function add_bridged_constraint(b, BridgeType, f, s)
+    map = Constraint.bridges(b)::Constraint.Map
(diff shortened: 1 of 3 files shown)


src/Bridges/bridge_optimizer.jl
```diff
@@ -1961,14 +1961,27 @@ function MOI.supports_constraint(
...
 function add_bridged_constraint(b, BridgeType, f, s)
+    map = Constraint.bridges(b)::Constraint.Map
+    # Okay, this part is a bit tricky. `Constraint.bridge_constraint` might add
+    # new constraints which also need `final_touch`. But we need the return
+    # value `bridge` to come _before_ them in the `needs_final_touch` vector.
+    # To make this happen, we remember how many bridges are already in the
+    # vector, and then we insert the bridge into the vector. This is an O(N)
+    # operation in the number of new bridges, but it's a pretty rare edge-case,
+    # and the number of new bridges should be small. (And in most common bridges
+    # that need final touch, like the ToMILP bridges, N=0.)
```
blegat: Since it's a `Vector{Any`, can't we just do `push!(map.needs_final_touch, nothing)` here and then replace it ?

src/Bridges/bridge_optimizer.jl
```diff
@@ -1961,14 +1961,27 @@ function MOI.supports_constraint(
...
+    n_final_touch = length(map.needs_final_touch)
     bridge = Constraint.bridge_constraint(BridgeType, recursive_model(b), f, s)
+    if MOI.Bridges.needs_final_touch(bridge)
+        insert!(map.needs_final_touch, n_final_touch + 1, bridge)
```
blegat, line 1978: I prefer doing these things in `Constraint/map.jl`, `bridge_optimizer.jl`

src/Bridges/bridge_optimizer.jl
blegat: Why is `N=0` for MILP bridge ?

odow: Because we might not need to

odow: They don’t add any bridges that themselves need final touch

src/Bridges/bridge_optimizer.jl
odow: Feel free to make a PR with some changes. This seemed simplest place to put it for me but I might be missing something.

odow:
https://github.com/jump-dev/MathOptInterface.jl/actions/runs/30219991948

PR #3034, opened 2026-07-27
Use placeholder in needs_final_touch

blegat:
If this works then we're sure we won't hit any `O(N)` complexity.

### DIFF
--- a/src/Bridges/bridge_optimizer.jl
+++ b/src/Bridges/bridge_optimizer.jl
@@ -1965,17 +1965,15 @@ function add_bridged_constraint(b, BridgeType, f, s)
     # Okay, this part is a bit tricky. `Constraint.bridge_constraint` might add
     # new constraints which also need `final_touch`. But we need the return
     # value `bridge` to come _before_ them in the `needs_final_touch` vector.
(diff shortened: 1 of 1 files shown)


src/Bridges/bridge_optimizer.jl
```diff
@@ -1965,17 +1965,15 @@ function add_bridged_constraint(b, BridgeType, f, s)
...
-    # vector, and then we insert the bridge into the vector. This is an O(N)
-    # operation in the number of new bridges, but it's a pretty rare edge-case,
-    # and the number of new bridges should be small. (And most of the time, N=0,
-    # because most bridges that need final touch don't add any bridges that
-    # themselves need final touch. Good examples are the ToMILP bridges that
-    # need variable bounds.)
-    n_final_touch = length(map.needs_final_touch)
+    # To make this happen, we insert a placeholder before calling
+    # `bridge_constraint` and we replace it by the bridge afterwards.
+    if MOI.Bridges.needs_final_touch(bridge)
```
odow, line 1970: It doesn’t because bridge is not known here yet

PR #3036, opened 2026-07-29
Improve error messages of various errors

odow:
I might break out the UnsupportedConstraint into a separate PR, but I'll see the damage here first.

### DIFF
--- a/src/Bridges/lazy_bridge_optimizer.jl
+++ b/src/Bridges/lazy_bridge_optimizer.jl
@@ -544,6 +544,22 @@ function is_variable_bridged(b::LazyBridgeOptimizer, S::Type{<:MOI.AbstractSet})
     return is_bridged(b, S) && is_variable_edge_best(b.graph, node(b, S))
 end
 
+function MOI.UnsupportedConstraint{F,S}(
+    model::LazyBridgeOptimizer,
+) where {F<:MOI.AbstractFunction,S<:MOI.AbstractSet}
+    msg = """
+    ## More information
+
+    Bridges are enabled, but there are none that can rewrite the \
+    constraint into a form supported by the solver.
+
+    The model that we were unable to add the constraint to is:
+
+    $(sprint(show, model))
+    """
+    return MOI.UnsupportedConstraint{F,S}(msg)
+end
+
 function bridge_type(b::LazyBridgeOptimizer, S::Type{<:MOI.AbstractSet})
     bt = get(b.cached_bridge_type, (S,), nothing)
     if bt !== nothing
@@ -552,7 +568,7 @@ function bridge_type(b::LazyBridgeOptimizer, S::Type{<:MOI.AbstractSet})
     index = bridge_index(b, S)
     if iszero(index)
         F = MOI.Utilities.variable_function_type(S)
-        throw(MOI.UnsupportedConstraint{F,S}())
+        throw(MOI.UnsupportedConstraint{F,S}(b))
     end
     new_bt = Variable.concrete_bridge_type(b.variable_bridge_types[index], S)
     b.cached_bridge_type[(S,)] = new_bt
@@ -570,7 +586,7 @@ function bridge_type(
     end
     index = bridge_index(b, F, S)
     if iszero(index)
-        throw(MOI.UnsupportedConstraint{F,S}())
+        throw(MOI.UnsupportedConstraint{F,S}(b))
     end
     new_bt =
         Constraint.concrete_bridge_type(b.constraint_bridge_types[index], F, S)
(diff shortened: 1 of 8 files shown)


odow:
Motivated by https://github.com/jump-dev/Dualization.jl/issues/232

```Julia
julia> using LinearAlgebra

julia> using Hypatia

julia> using Dualization

julia> using JuMP

julia> function test_dual()
           model = Model()
           @variable(model, M[1:2, 1:2], Symmetric)
           @constraint(model, [1, M[1, 1], sqrt(2) * M[1, 2], M[2, 2]] in Hypatia.EpiNormSpectralTriCone{Float64,Float64}(1 + 3))
           @objective(model, Max, real(tr(M)))
           #set_optimizer(model, Hypatia.Optimizer)
           set_optimizer(model, Dualization.dual_optimizer(Hypatia.Optimizer))
           optimize!(model)
           return value(M)
       end
test_dual (generic function with 1 method)

julia> test_dual()
ERROR: UnsupportedConstraint{
    MathOptInterface.VectorAffineFunction{Float64},
    EpiNormSpectralTriCone{Float64, Float64},
}

This constraint type is not supported by the solver.

To fix this error you must choose a different solver.

## More information

Bridges are enabled, but there are none that can rewrite the constraint into a form supported by the solver.

The model that we were unable to add the constraint to is:

MOIB.LazyBridgeOptimizer{MOIU.CachingOptimizer{DualOptimizer{Float64, MOIB.LazyBridgeOptimizer{MOIU.CachingOptimizer{Hypatia.Optimizer{Float64}, MOIU.UniversalFallback{MOIU.Model{Float64}}}}}, MOIU.UniversalFallback{MOIU.Model{Float64}}}}
├ Variable bridges: none
├ Constraint bridges: none
├ Objective bridges: none
└ model: MOIU.CachingOptimizer
  ├ state: EMPTY_OPTIMIZER
  ├ mode: AUTOMATIC
  ├ model_cache: MOIU.UniversalFallback{MOIU.Model{Float64}}
  │ ├ ObjectiveSense: MAX_SENSE
  │ ├ ObjectiveFunctionType: MOI.ScalarAffineFunction{Float64}
  │ ├ NumberOfVariables: 3
  │ └ NumberOfConstraints: 0
  └ optimizer: DualOptimizer{Float64, MOIB.LazyBridgeOptimizer{MOIU.CachingOptimizer{Hypatia.Optimizer{Float64}, MOIU.UniversalFallback{MOIU.Model{Float64}}}}}
    ├ ObjectiveSense: unknown
    ├ ObjectiveFunctionType: unknown
    ├ NumberOfVariables: unknown
    └ NumberOfConstraints: unknown

Stacktrace:
```

odow:
cc @araujoms

src/Bridges/lazy_bridge_optimizer.jl
```diff
@@ -544,6 +544,22 @@ function is_variable_bridged(b::LazyBridgeOptimizer, S::Type{<:MOI.AbstractSet})
...
+) where {F<:MOI.AbstractFunction,S<:MOI.AbstractSet}
+    msg = """
+    ## More information
+
+    Bridges are enabled, but there are none that can rewrite the \
+    constraint into a form supported by the solver.
+
+    The model that we were unable to add the constraint to is:
+
+    $(sprint(show, model))
```
araujoms, line 558:
```suggestion
    constraint into a form supported by the solver $(sprint(show, MOI.get(model, MOI.SolverName())))
```
I'd just print the solver name instead of the whole model, since the solver is what makes the constraint unsupported.

odow: So this isn't `print(model)`, but it's `show(model)`. Which is the high level summary.

test/Bridges/General/test_lazy_bridge_optimizer.jl
```diff
@@ -2703,6 +2704,47 @@ function test_custom_cost_model_bridge_selection()
...
+    MOIB.LazyBridgeOptimizer{MOIU.Model{Float64}}
+    ├ Variable bridges: none
+    ├ Constraint bridges: none
+    ├ Objective bridges: none
+    └ model: MOIU.Model{Float64}
+      ├ ObjectiveSense: FEASIBILITY_SENSE
+      ├ ObjectiveFunctionType: MOI.ScalarAffineFunction{Float64}
+      ├ NumberOfVariables: 1
+      └ NumberOfConstraints: 0
+    """
```
odow, line 2743: @araujoms here you go

src/Bridges/lazy_bridge_optimizer.jl
araujoms: Which still has a lot of irrelevant information.

odow: `SolverName` drops a lot of relevant information.

```Julia
julia> MOI.get(b, MOI.SolverName())
"Dual model with Hypatia attached"
```

It looks through the bridging layer and the cache.

araujoms: I know this is the result, it's exactly the information I wanted to have, which is why I suggested it.

It's just a suggestion, feel free to ignore it.

odow:
https://github.com/jump-dev/MathOptInterface.jl/actions/runs/30490632040

odow:
One consideration is that this is going to break the JuMP tests, but that's okay. We shouldn't be relying on string equality from a dependent package.

blegat (approved):
Nice improvements

Issue #3037, opened 2026-08-01
Add dual cones for cones without them


araujoms:
It seems that the problem is that MOI doesn't define its dual. Is that intentional or would the addition of the cone be welcome?

joaquimg:
The initial recommendation is to create a separate package and play with the idea. Once stable, it might make sense to propose an addition to the MOI repo.

You would need to define the new cone and overload the function MOI.dual_set. Then Dualization would be able to dualize.

To connect that with solving, you need either: 1) a solver supporting the new cone (is there one?) or 2) bridges to convert the new cone into something supported by some solver (that can live in the separate package).

araujoms:
I'm not creating any package, I'm just happy to help if you're interested in making the dualisation work.

As for solving, it's easy to bridge it to the geometric mean cone itself. The only point in adding it is to enable dualisation. In any case, Hypatia supports it directly.

odow:
I'm going to say that this issue is out-of-scope for Dualization. There are few sets in MOI that we don't have dual sets for. My quick list: NormCone, GeometricMeanCone, RelativeEntropyCone, LogDetConeTriangle, LogDetConeSquare, RootDetConeTriangle, RootDetConeSquare, PositiveSemidefiniteConeSquare.

There's an argument about whether we _should_ have sets for those, but we seem to be doing okay without them. We also typically require multiple solvers to support the set before we add it to MOI.

If a solver like Hypatia supports specific dual sets, then it can already implement `dual_set` and `dual_set_type` and work with Dualization.jl, so I don't know if there's anything we need to do here.

araujoms:
How is an user supposed to know that they need to use `Hypatia.HypoGeoMeanCone` instead of `MOI.GeometricMeanCone` for dualisation to work? This is a terrible experience for no benefit. I think that yes, clearly if a cone is in MOI its dual should also be. Except for  LogDetConeSquare, RootDetConeSquare, and PositiveSemidefiniteConeSquare; it was a mistake to introduce them so you shouldn't dig yourselves deeper in the hole.

Furthermore, the dual of these cones can be easily be bridged into the primal ones together with the nonnegative cone, so it's not as if you're demanding any extra solver support.

blegat:
> How is an user supposed to know that they need to use `Hypatia.HypoGeoMeanCone` instead of `MOI.GeometricMeanCone` for dualisation to work? This is a terrible experience for no benefit.

[...]

PR #3041, opened 2026-08-07
add DualGeometricMeanCone and bridge

araujoms:
Part of #3037

I've added the set, related tests, and the bridge. Still missing are the bridge tests; I'm submitting the PR prematurely because it's already getting big and I'm getting nervous about losing stuff.

I also didn't add support in the file format, I don't know if that's desired.

## Basic

 - [x] If `isbitstype(S) == false`, implement `Base.copy(set::S)`
 - [x] If `isbitstype(S) == false`, implement `Base.:(==)(x::S, y::S)`
 - [x] If an `AbstractVectorSet`, implement `dimension(set::S)`, unless the
       dimension is given by `set.dimension`.
 - [x] Ensure the set does not contain references to any variables or constraints

## Utilities

 - [x] If an `AbstractVectorSet`, implement `Utilities.set_dot`,
       `LinearAlgebra.dot`
 - [x] If an `AbstractVectorSet`, implement `Utilities.set_with_dimension` in

## Documentation

## Tests

## MathOptFormat

 - [x] Open an issue at `https://github.com/jump-dev/MathOptFormat` to add
       support for the new set https://github.com/jump-dev/MathOptFormat/issues/37

## Optional

 - [x] Implement `dual_set(::S)` and `dual_set_type(::Type{S})`
 - [x] Add new bridges to convert your set into more commonly used sets

### DIFF
--- a/src/Test/test_conic.jl
+++ b/src/Test/test_conic.jl
@@ -3370,7 +3370,8 @@ function setup_test(
             mock,
             T[-3, 1, 1, 1]::Vector{T},
             (MOI.ScalarAffineFunction{T}, MOI.LessThan{T}) => [-T(1)],
-            (MOI.VectorAffineFunction{T}, MOI.DualGeometricMeanCone) => [ones(T, 4)],
+            (MOI.VectorAffineFunction{T}, MOI.DualGeometricMeanCone) =>
+                [ones(T, 4)],
         ),
     )
     return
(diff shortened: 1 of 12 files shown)


src/Test/test_conic.jl
```diff
@@ -2850,6 +2850,7 @@ function _test_conic_GeometricMeanCone_helper(
         )
         @test ≈(MOI.get(model, MOI.ConstraintPrimal(), c), n, config)
         if _supports(config, MOI.ConstraintDual)
+            display(MOI.get(model, MOI.ConstraintDual(), gmc))
```
blegat: Let's not forget to remove this

araujoms: ops, thanks for noticing

odow:
I guess you see now why we've only added sets when they became needed. It's a lot of work!

araujoms:
Indeed. I intend to add the dual vector relative entropy cone as well, but I don't use the other cones so I'll leave them for when they're needed.

PR #3045, opened 2026-08-10
[docs] make a single public API page

odow:
I've tried this a couple of times. Opening as a draft for now.

https://jump.dev/MathOptInterface.jl/previews/PR3045/api/MathOptInterface/

The tricky part about this is defining exactly what our public API is...


odow:
Closing because this loses a lot of structure that the docs currently have.

PR #3046, opened 2026-08-12
AD interface compatible with both ArrayDiff and ExaModels

blegat:
In https://github.com/jump-dev/MathOptInterface.jl/pull/2989, I tried doing a simple change to the AD interface for ArrayDiff to serve as an AD backend.
The idea was to have `model` return a custom model type when given the AD backend as argument so that ArrayDiff could deviate from `MOI.Nonlinear.Model`.
The conclusion at the JuMP-dev call was to improve `MOI.Nonlinear.Model` so that we don't need to deviate.
The issue with this is that it forces us to commit the changes to MOI so it means we can't do breaking changes which makes it hard to play around and try things. ArrayDiff is then blocked in a state where it cannot be tried by user until we just commit to an interface that just cannot be changed anymore.
Another proof that the AD interface is too rigid is the fact that ExaModels couldn't interface itself as a new AD backend and had to define itself as a meta-solver.

With AI getting more powerful, I think the features of MOI that allows people to extend it independently in different package is quite important. Trying to have a centralized `MOI.Nonlinear.Model` that would cover all need is probably too strict.
So this PR attempt to create a new interface so that both ArrayDiff and ExaModels can be defined as AD backends.

On major difference between ExaModels and ReverseAD is that ReverseAD is that Ipopt and NLPModelsJuMP don't give the quadratic constraints to ReverseAD while ExaModels takes the quadratic constraints in its AD.
So for the Ipopt wrapper to work for both, it would need to give the quadratic constraints to both and ReverseAD should also handle quadratic constraints.
The way this PR does it is to refactor the handling of quadratic constraints and `VectorNonlinearOracle` of Ipopt as layers that we can stack on top of an existing AD.
This is continuing the approach started in https://github.com/jump-dev/Ipopt.jl/pull/520
As a side effect, this reduces the large code duplication of the handling of quadratic constraints and nonlinear oracles in Ipopt, NLopt, MadNLP, NLPModelsJuMP, etc... Again, with AI, we might see a lot of new solvers appear, so having all of them copy the Ipopt wrapper is probably not a good idea. Because when we'll modify the Ipopt wrapper, they won't follow and it will create inconsistent behavior across solvers. Even if many vibe-coded solvers won't be any good, natural selection will do its job and natural selection will work better if they are all easily accessible through JuMP.

So Claude Fable 5 create this PR which seems reasonable although I haved looked at all details yet. 
We could probably open a PR for 459c20a2484333cb94294ab0f13b70293e54087a and merge it independently if we decide we want to go for this.

We needed linearity information for NLPModelsJuMP and Ipopt so I reused the linearity of MOI.Nonlinear.ReverseAD, just added quadratic in the mix. The quadratic detection might have some false negative at the moment but it's find since the quadratic given as quadratic in JuMP are already reported by the WithQuad layer.

Let's try it with the downstream packages to see if it works

- [ ] https://github.com/blegat/ArrayDiff.jl/pull/95
- [ ] https://github.com/madsuite-org/ExaModels.jl/pull/307
- [ ] https://github.com/jump-dev/Ipopt.jl/pull/520 (maybe the changes should be made to a new branch)
- [ ] https://github.com/jump-dev/NLopt.jl/pull/273
- [ ] https://github.com/madsuite-org/MadNLP.jl/pull/639
- [ ] https://github.com/JuliaSmoothOptimizers/NLPModelsJuMP.jl/pull/229

### DIFF
--- a/docs/src/submodules/Nonlinear/reference.md
+++ b/docs/src/submodules/Nonlinear/reference.md
@@ -74,6 +74,15 @@ Nonlinear.SparseReverseMode
 Nonlinear.SymbolicMode
 ```
 
+## Linearity
+
+```@docs
+Nonlinear.Linearity
+Nonlinear.num_constraints
+Nonlinear.constraint_linearity
+Nonlinear.objective_linearity
+```
+
 ## Data-structure
 
 ```@docs
--- a/docs/src/submodules/Nonlinear/reference.md
+++ b/docs/src/submodules/Nonlinear/reference.md
@@ -81,6 +81,18 @@ Nonlinear.Linearity
 Nonlinear.num_constraints
 Nonlinear.constraint_linearity
 Nonlinear.objective_linearity
+Nonlinear.constraint_bounds
+```
+
+## Model layers
+
+```@docs
+Nonlinear.model
+Nonlinear.ModelWithQuad
+Nonlinear.ModelWithOracles
+Nonlinear.EvaluatorWithQuad
+Nonlinear.EvaluatorWithOracles
+Nonlinear.QPBlockData
 ```
 
 ## Data-structure
--- a/src/Nonlinear/Nonlinear.jl
+++ b/src/Nonlinear/Nonlinear.jl
@@ -45,5 +45,7 @@ include("SymbolicAD/SymbolicAD.jl")
 # Depends on the `Linearity` enum defined in `ReverseAD`, so must be included
 # after it.
 include("linearity.jl")
+include("qp_block_data.jl")
+include("layers.jl")
 
 end  # module
--- a/src/Nonlinear/Nonlinear.jl
+++ b/src/Nonlinear/Nonlinear.jl
@@ -42,4 +42,8 @@ include("evaluator.jl")
 include("ReverseAD/ReverseAD.jl")
 include("SymbolicAD/SymbolicAD.jl")
 
+# Depends on the `Linearity` enum defined in `ReverseAD`, so must be included
+# after it.
+include("linearity.jl")
+
 end  # module
--- a/src/Nonlinear/ReverseAD/graph_tools.jl
+++ b/src/Nonlinear/ReverseAD/graph_tools.jl
@@ -106,7 +106,7 @@ function _classify_linearity(
             op =
                 get(Nonlinear.DEFAULT_UNIVARIATE_OPERATORS, node.index, nothing)
             if op == :+ || op == :-
-                # Unary plus and minus preserve the linearity of the child.
+                # A unary plus or minus preserves the linearity of the child.
                 linearity[k] = worst
             elseif op == :abs && worst <= PIECEWISE_LINEAR
                 linearity[k] = PIECEWISE_LINEAR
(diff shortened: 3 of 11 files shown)


odow:
Without looking at the details, it's worth investigating, but I'm pretty strongly against this.

My experience with https://github.com/madsuite-org/ExaModels.jl/pull/290 would suggest that writing a new `ExaModels.Optimizer` works fine. And even if you wanted an AD-backend, lowering from the current `MOI.Nonlinear.Model` would be okay.

We should have a very, very high bar for making the nonlinear interface in MOI any more complicated.

amontoison:
I like the `Nonlinear.QPBlockData`. It will simplify a lot the extensions for MOI in `Ipopt.jl`, `UnoSolver.jl`, and `MadNLP.jl`.

For the other components, I don't know enough MOI to provide feedbacks. Oscar is the expert 🙂

odow:
It was always my intention to move something like https://github.com/jump-dev/Ipopt.jl/blob/master/ext/IpoptMathOptInterfaceExt/utils.jl into MOI.

I'd accept that as a separate PR.

blegat:
> I'd accept that as a separate PR.

That's a good first step

> We should have a very, very high bar for making the nonlinear interface in MOI any more complicated.

I agree we shouldn't complicate it too much but I feel that at the moment, `MOI.Nonlinear.Model` is just too rigid for people to experiment with new AD backends. NLP solvers don't really need the nonlinear model to be `MOI.Nonlinear.Model`, they just use the `MOI.Nonlinear` callback interface so forcing the model type is adding a lot of friction for no benefit I feel.

> And even if you wanted an AD-backend, lowering from the current `MOI.Nonlinear.Model` would be okay.

Actually it won't. How would you make https://github.com/madsuite-org/ExaModels.jl/pull/237 work ? GenOpt generator functions can be directly given to ExaModels without needing to be expanded. If they have to go through `MOI.Nonlinear.Model`, then they will need to be expanded. We just completely loose the extensibility of MOI and the ability to try new things on separate package that haven't reached version v1 yet.

> My experience with https://github.com/madsuite-org/ExaModels.jl/pull/290 would suggest that writing a new `ExaModels.Optimizer` works fine.

It's fine as a one time thing but then once you add something new like `VectorNonlinearOracle`, all the MOI wrappers need to be updated. We loose a decoupling. It's not the end of the world but my feeling is that a slight modification of the interface would make the `MOI.Nonlinear.AbstractAutomaticDifferentiation` much more useful. Now we'll have to see if a slight modification is enough so I'll see where this PR goes.

amontoison:
> It was always my intention to move something like https://github.com/jump-dev/Ipopt.jl/blob/master/ext/IpoptMathOptInterfaceExt/utils.jl into MOI.
> 
> I'd accept that as a separate PR.

@blegat If you open a new PR for that, I added the support of J * v, J' * v, and H * v in the `utils.jl` / `MOI_utils.jl` for MadNLP.jl and UnoSolver.jl:
- https://github.com/madsuite-org/MadNLP.jl/blob/master/ext/MadNLPMOI/MOI_utils.jl
-https://github.com/cvanaret/Uno/blob/main/interfaces/Julia/ext/UnoSolverMathOptInterfaceExt/MOI_utils.jl

Please use the variant with the operators such that we can remove this file for all packages.

PR #3048, opened 2026-08-13
Add QPBlockData

blegat:
This is essentially a copy-paste from @odow 's implementation in Ipopt : https://github.com/jump-dev/Ipopt.jl/blob/b5b0e62b11aff4ae4bf0c5f0bc852c55b8953b4f/ext/IpoptMathOptInterfaceExt/utils.jl#L62 with these 3 functions added by @frapac in MadNLP https://github.com/madsuite-org/MadNLP.jl/blob/9947ee116559e9a2ab684a2006d37951f0bb9a6f/ext/MadNLPMOI/MOI_utils.jl#L654-L691
The changes I made on top of it were: adding `_` in front of a few functions like `eval_function`-> `_eval_function`.
The returned value of `eval_constraint_jacobian/eval_hessian_lagrangian` was returning `nnz + 1`, not it returns `nnz`.
`_is_parameter(x) = x.value >= 0x00f0000000000000` is replaced with `haskey(block.parameters, x.value)` which means the parameters need to be registered before being used.

Downstream PRs:

- https://github.com/madsuite-org/MadNLP.jl/pull/640
- https://github.com/jump-dev/Ipopt.jl/pull/548
- https://github.com/JuliaSmoothOptimizers/NLPModelsJuMP.jl/pull/237
- https://github.com/jump-dev/NLopt.jl/pull/274
- https://github.com/cvanaret/Uno/pull/863

### DIFF
--- a/src/Nonlinear/qp_block_data.jl
+++ b/src/Nonlinear/qp_block_data.jl
@@ -538,7 +538,6 @@ function MOI.jacobian_structure(block::QPBlockData)
     return J
 end
 
-# Returns the number of entries written to `J`.
 function MOI.eval_constraint_jacobian(
     block::QPBlockData{T},
     J::AbstractVector{T},
@@ -549,7 +548,7 @@ function MOI.eval_constraint_jacobian(
         ∇f = view(J, (i+1):length(J))
         i += _eval_sparse_gradient(∇f, constraint, x, block.parameters)
     end
-    return i
+    return
 end
 
 function MOI.hessian_lagrangian_structure(block::QPBlockData)
@@ -561,7 +560,6 @@ function MOI.hessian_lagrangian_structure(block::QPBlockData)
     return H
 end
 
-# Returns the number of entries written to `H`.
 function MOI.eval_hessian_lagrangian(
     block::QPBlockData{T},
     H::AbstractVector{T},
@@ -574,7 +572,7 @@ function MOI.eval_hessian_lagrangian(
         ∇²f = view(H, (i+1):length(H))
         i += _eval_sparse_hessian(∇²f, constraint, μ[row], block.parameters)
     end
-    return i
+    return
 end
 
 # The product evaluators below ACCUMULATE into their output vector, so that
--- a/src/Nonlinear/qp_block_data.jl
+++ b/src/Nonlinear/qp_block_data.jl
@@ -431,14 +431,14 @@ function MOI.get(
 ) where {T,F,S}
     row = c.value
     if block.bound_type[row] == _kBoundTypeEqualTo
-        return MOI.EqualTo(block.g_L[row])
+        return MOI.EqualTo(block.g_L[row])::S
     elseif block.bound_type[row] == _kBoundTypeLessThan
-        return MOI.LessThan(block.g_U[row])
+        return MOI.LessThan(block.g_U[row])::S
     elseif block.bound_type[row] == _kBoundTypeGreaterThan
-        return MOI.GreaterThan(block.g_L[row])
+        return MOI.GreaterThan(block.g_L[row])::S
     else
         @assert block.bound_type[row] == _kBoundTypeInterval
-        return MOI.Interval(block.g_L[row], block.g_U[row])
+        return MOI.Interval(block.g_L[row], block.g_U[row])::S
     end
 end
 
--- a/src/Nonlinear/qp_block_data.jl
+++ b/src/Nonlinear/qp_block_data.jl
@@ -4,12 +4,6 @@
 # in the LICENSE.md file or at https://opensource.org/licenses/MIT.
 
 # This file is adapted from `Ipopt.jl/ext/IpoptMathOptInterfaceExt/utils.jl`.
-#
-# Unlike the Ipopt version, a variable is treated as a parameter if and only
-# if its index is a key of the `parameters` dictionary, instead of an
-# index-offset convention. Parameters must therefore be registered in
-# `parameters` before any structure query, but their values may be updated
-# freely between function evaluations.
 
 @enum(
     _FunctionType,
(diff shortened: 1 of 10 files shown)


src/Nonlinear/qp_block_data.jl
```diff
@@ -0,0 +1,739 @@
...
+
+"""
+    QPBlockData{T}()
+
+A data structure holding an affine or quadratic objective and a block of
+affine and quadratic constraints, together with methods to evaluate them
+following the [`MOI.AbstractNLPEvaluator`](@ref) callback conventions.
+
+This is the storage behind [`ModelWithQuad`](@ref); it is not typically used
+directly.
```
odow: Update

src/Nonlinear/qp_block_data.jl
```diff
@@ -0,0 +1,739 @@
...
+    c::MOI.ConstraintIndex{F,S},
+) where {T,F,S}
+    return convert(F, block.constraints[c.value])
+end
+
+function MOI.get(
+    block::QPBlockData{T},
+    ::MOI.ConstraintSet,
+    c::MOI.ConstraintIndex{F,S},
+) where {T,F,S}
```
odow, line 446: This function probably needs some checks/asserts on `S`? We're not guaranteed to return `::S` here.

src/Nonlinear/qp_block_data.jl
```diff
@@ -0,0 +1,739 @@
...
+
+function MOI.jacobian_structure(block::QPBlockData)
+    J = Tuple{Int,Int}[]
+    for (row, constraint) in enumerate(block.constraints)
+        _append_sparse_gradient_structure!(constraint, J, row, block.parameters)
+    end
+    return J
+end
+
+# Returns the number of entries written to `J`.
```
odow: This might need discussion or explicit documentation. It's not part of the MOI spec.

src/Nonlinear/qp_block_data.jl
```diff
@@ -0,0 +1,739 @@
...
+function MOI.hessian_lagrangian_structure(block::QPBlockData)
+    H = Tuple{Int,Int}[]
+    _append_sparse_hessian_structure!(block.objective, H, block.parameters)
+    for constraint in block.constraints
+        _append_sparse_hessian_structure!(constraint, H, block.parameters)
+    end
+    return H
+end
+
+# Returns the number of entries written to `H`.
```
odow: Ditto here

src/Nonlinear/qp_block_data.jl
```diff
@@ -0,0 +1,739 @@
...
+    for (row, constraint) in enumerate(block.constraints)
+        ∇²f = view(H, (i+1):length(H))
+        i += _eval_sparse_hessian(∇²f, constraint, μ[row], block.parameters)
+    end
+    return i
+end
+
+# The product evaluators below ACCUMULATE into their output vector, so that
+# they compose with the products of the other layers. Zero the output before
+# the first call.
```
odow: This needs to be something other than a comment.

src/Nonlinear/qp_block_data.jl
```diff
@@ -0,0 +1,739 @@
...
+end
+
+function MOI.eval_constraint_jacobian_product(
+    block::QPBlockData{T},
+    y::AbstractVector{T},
+    x::AbstractVector{T},
+    w::AbstractVector{T},
+) where {T}
+    for (i, constraint) in enumerate(block.constraints)
+        _eval_Jv_product(constraint, y, x, w, block.parameters, i)
```
odow: For example, we don't zero `y` before this first call.

src/Nonlinear/qp_block_data.jl
```diff
@@ -0,0 +1,739 @@
...
+    return
+end
+
+function MOI.eval_constraint_jacobian_transpose_product(
+    block::QPBlockData{T},
+    y::AbstractVector{T},
+    x::AbstractVector{T},
+    w::AbstractVector{T},
+) where {T}
+    for (i, constraint) in enumerate(block.constraints)
```
odow, line 769: Ditto here

src/Nonlinear/qp_block_data.jl
```diff
@@ -0,0 +1,739 @@
...
+
+function MOI.eval_hessian_lagrangian_product(
+    block::QPBlockData{T},
+    H::AbstractVector{T},
+    x::AbstractVector{T},
+    v::AbstractVector{T},
+    σ::T,
+    μ::AbstractVector{T},
+) where {T}
+    _eval_Hv_product(block.objective, H, x, v, σ, block.parameters)
```
odow: Ditto here

src/Nonlinear/qp_block_data.jl
odow: You could argue we shouldn't return anything here. It's the caller's responsibility to know how many non-zeros will be written because they also construct the sparsity pattern and must prepare a vector to hold the results, etc. I think I just returned the value to simplify some code in Ipopt.

src/Nonlinear/qp_block_data.jl
blegat: Removed

src/Nonlinear/qp_block_data.jl
blegat: Good point, I was just staying closed to Ipopt here but it's cleaner to just remove

src/Nonlinear/qp_block_data.jl
blegat: Yes, this is the source of bugs like https://github.com/madsuite-org/MadNLP.jl/issues/641

blegat: I'll just rename them to `add_...`. These `eval_...` need to not zero out so that we can combine the NL part with the quadratic part so I'll just create new add functions here. We could make them internal functions if we just used this `QPBlockData` internally and only expose the `EvaluatorWithQuad`. But if we want to first release an `QPBlockData` used by wrappers, we have to expose these new `add_...` functions. That's probably an argument toward `EvaluatorWithQuad` :innocent:

blegat:
I merged this PR into https://github.com/jump-dev/MathOptInterface.jl/pull/3053 and did the same for downstream PRs. Replying to https://github.com/jump-dev/MathOptInterface.jl/pull/3053#issuecomment-5362704252:

> I'm not overly enthused by having another 1000 lines of code in MOI...

I agree, I try to think of it as avoiding to duplicate 1000 lines in 5 downstream packages. We've waiting a bit since the new `MOI.Nonlinear` so that its downstream packages stabilizes so we should be ready now to commit to merging the handling of quadratic constraints in MOI.

PR #3053, opened 2026-08-15
Add ModelWithQuad and privatize the product functions of QPBlockData

blegat:
As discussed in https://github.com/jump-dev/MathOptInterface.jl/pull/3048#discussion_r3775360381, it might be better to directly go for `ModelWithQuad` and have the solver use that directly so that we don't have to make the interface to `QPBlockData` public and we can change it later.

Github is suggesting me to "Start a pull request stack", I didn't know this feature, let's try.

- https://github.com/madsuite-org/MadNLP.jl/pull/643
- https://github.com/jump-dev/Ipopt.jl/pull/549
- https://github.com/JuliaSmoothOptimizers/NLPModelsJuMP.jl/pull/238
- https://github.com/jump-dev/NLopt.jl/pull/276
- https://github.com/cvanaret/Uno/pull/863

### DIFF
--- a/src/Nonlinear/model_with_quad.jl
+++ b/src/Nonlinear/model_with_quad.jl
@@ -145,6 +145,42 @@ before the rows of the inner model in the corresponding evaluator.
 """
 Base.length(model::ModelWithQuad) = length(model.qp)
 
+# Replace the parameters of `f`, encoded as `MOI.VariableIndex`es offset by
+# [`_PARAMETER_OFFSET`](@ref), by the corresponding [`ParameterIndex`](@ref),
+# which the inner model understands. An affine or quadratic function that
+# contains a parameter is converted to `MOI.ScalarNonlinearFunction`, because
+# the inner model parses such functions with their variable indices verbatim.
+_replace_parameters(f) = f
+
+function _replace_parameters(f::MOI.VariableIndex)
+    if _is_parameter(f)
+        return ParameterIndex(f.value - _PARAMETER_OFFSET)
+    end
+    return f
+end
+
+function _replace_parameters(f::MOI.ScalarAffineFunction)
+    if any(_is_parameter, f.terms)
+        return _replace_parameters(convert(MOI.ScalarNonlinearFunction, f))
+    end
+    return f
+end
+
+function _replace_parameters(f::MOI.ScalarQuadraticFunction)
+    if any(_is_parameter, f.affine_terms) ||
+       any(_is_parameter, f.quadratic_terms)
+        return _replace_parameters(convert(MOI.ScalarNonlinearFunction, f))
+    end
+    return f
+end
+
+function _replace_parameters(f::MOI.ScalarNonlinearFunction)
+    for (i, arg) in enumerate(f.args)
+        f.args[i] = _replace_parameters(arg)
+    end
+    return f
+end
+
 # Methods forwarded to the inner model.
 
 function add_parameter(model::ModelWithQuad, value::Real)
@@ -194,6 +230,9 @@ end
 function set_objective(model::ModelWithQuad{T}, obj) where {T}
     F = MOI.ScalarAffineFunction{T}
     MOI.set(model.qp, MOI.ObjectiveFunction{F}(), zero(F))
+    if !isempty(model.qp.parameters)
+        obj = _replace_parameters(obj)
+    end
     set_objective(model.inner, obj)
     model.objective_sink = obj === nothing ? :none : :inner
     return
@@ -225,6 +264,9 @@ function add_constraint(
 end
 
 function add_constraint(model::ModelWithQuad, func, set)
+    if !isempty(model.qp.parameters)
+        func = _replace_parameters(func)
+    end
     return add_constraint(model.inner, func, set)
 end
 
--- a/src/Nonlinear/model_with_quad.jl
+++ b/src/Nonlinear/model_with_quad.jl
@@ -13,9 +13,7 @@
 
 A model layer that stores affine and quadratic objectives and constraints in
 a [`QPBlockData`](@ref) and forwards everything else to the `inner` model,
-typically a [`Model`](@ref). Solvers with their own storage can pass an
-existing `qp` and set `inner` to `nothing` when they manage the nonlinear
-model themselves.
+typically a [`Model`](@ref).
 
 `ModelWithQuad(inner)` and `ModelWithQuad{T}(inner)` create an empty
 [`QPBlockData`](@ref), with `T` defaulting to `Float64`.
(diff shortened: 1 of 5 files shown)


odow:
Some questions:

 * Do we care about duplicated code in Ipopt.jl and Uno.jl etc?
 * Should we add yet-another-complicated-helper to MOI?
 * Do any of the solvers need any tweaks to this wrapper? Do they handle modifications etc the same?
 * Could we add explicit linear/quadratic support to the existing `MOI.Nonlinear.Model`?
 * Is VectorNonlinearOracle supported?

src/Nonlinear/model_with_quad.jl
```diff
@@ -0,0 +1,545 @@
...
+`ModelWithQuad(inner)` and `ModelWithQuad{T}(inner)` create an empty
+[`QPBlockData`](@ref), with `T` defaulting to `Float64`.
+
+Add variables with `MOI.add_variable`: the layer guarantees that the variable
+indices are `1:n`, like `MOI.Utilities.MatrixOfConstraints`. Add parameters
+with `MOI.add_constrained_variable(model, ::MOI.Parameter)`: parameters get
+indices offset by [`_PARAMETER_OFFSET`](@ref), and their values are stored in
+the inner model through [`add_parameter`](@ref). The inner model must expose
+that storage as `parameters::Vector{T}`, like [`Model`](@ref) does:
+`qp.parameters` aliases it, so a parameter update is visible to both blocks.
```
odow, line 30: Is this `parameters::Vector{T}` requirement supported by all solvers?

src/Nonlinear/model_with_quad.jl
```diff
@@ -0,0 +1,545 @@
...
+
+Add constraints with [`add_constraint`](@ref) or `MOI.add_constraint`, and
+set the objective with [`set_objective`](@ref): affine and quadratic
+functions are routed to the QP block, everything else to the inner model.
+`objective_sink` records where the objective currently lives (`:none`,
+`:quad` or `:inner`).
```
odow: Don't use `Symbol` for things like enums.

src/Nonlinear/model_with_quad.jl
```diff
@@ -0,0 +1,545 @@
...
+    return
+end
+
+function MOI.jacobian_structure(d::EvaluatorWithQuad)
+    J = MOI.jacobian_structure(d.model.qp)
+    offset = length(d.model.qp)
+    # An evaluator is only required to implement `jacobian_structure` if it
+    # supports `:Jac`. If the inner evaluator does not (it then must not have
+    # any rows for the stack to be usable), append nothing.
+    if :Jac in MOI.features_available(d.inner)
```
odow: Is this right? Surely if `d.inner` can't return anything, then I shouldn't return the `J` from model.qp?

src/Nonlinear/model_with_quad.jl
```diff
@@ -0,0 +1,545 @@
...
+
+function MOI.eval_constraint_jacobian(d::EvaluatorWithQuad, J, x)
+    MOI.eval_constraint_jacobian(d.model.qp, J, x)
+    MOI.eval_constraint_jacobian(d.inner, view(J, (d.qp_nnzj+1):length(J)), x)
+    return
+end
+
+function MOI.hessian_lagrangian_structure(d::EvaluatorWithQuad)
+    H = MOI.hessian_lagrangian_structure(d.model.qp)
+    if :Hess in MOI.features_available(d.inner)
```
odow: Ditto here

src/Nonlinear/model_with_quad.jl
```diff
@@ -0,0 +1,545 @@
...
+        x,
+        w,
+    )
+    _add_constraint_jacobian_product(d.model.qp, y, x, w)
+    return
+end
+
+# Both blocks accumulate into the same variable-dimensional output. Call the
+# inner evaluator FIRST because implementations are allowed to overwrite the
+# output, and accumulate the QP block afterwards.
```
odow: Which implementations?

And isn't _implementations are allowed to overwrite_ -> _implementations overwrite_? I don't think we require the vector to be zeroed going in, so `+=` is not correct.

blegat:
> * Do we care about duplicated code in Ipopt.jl and Uno.jl etc?
> * Do any of the solvers need any tweaks to this wrapper? Do they handle modifications etc the same?

Of course my other motivation is https://github.com/jump-dev/MathOptInterface.jl/pull/3046 but I think we also do care about copy-paste. It's a bit like how `MatrixOfConstraints` allowed us to simplify conic wrappers but here I feel it is simplifying even more. All solvers that just want callbacks will essentially reinvent the wheel here, at least we're going to verify that here by trying to use this new features in all solvers before we merge, we can add Uno to the list

> Should we add yet-another-complicated-helper to MOI?

The fact that it's complicated is also kind of the reason to do it right just once, a bit like `MatrixOfConstraints`. At some point, I was even thinking we should just ask for new solvers to just interface NLPModels and then use `NLPModelsJuMP` and it would fix the duplication of complexity. But CCOpt has special support for complementarity constraints so it should have an MOI wrapper. So I think this PR is useful.

> Could we add explicit linear/quadratic support to the existing `MOI.Nonlinear.Model`?

This is essentially what this PR is doing except that we do it with composition. The advantage is that `ArrayDiff` can be a replacement for `MOI.Nonlinear.Model` and not have to redo the quadratic part.

 > Is VectorNonlinearOracle supported?

We can do it in a follow-up PR. It can be either a third layer or added to `ModelWithQuad` or added to `MOI.Nonlinear.Model`. Given that ExaModels doesn't support VectorNonlinearOracle it may make sense to make it a separate layer but let's leave that decision for later.

src/Nonlinear/model_with_quad.jl
blegat: This `parameters::Vector{T}` is just something both `ArrayDiff` and `MOI.Nonlinear.Model` should have. Solvers actually don't care about parameters. Ipopt for instance don't even support parameters, the MOI wrapper was just making them disappear. That's kind of why all this code is easy to share between solvers. Solvers just want simple callbacks. All of this is mostly just transforming all the complicated MOI things into these simple callbacks. We can almost reach a state where the solvers just forward all MOI things to their AD model and we don't even need to find `Parameter` when we grep their MOI wrapper.

blegat:
CI is now green on this PR and also all 5 other downstream PRs (except MadNLP but there seems to be an issue with the ci there). @odow any objection to merge ?

odow:
If you're set on this ModelWithQuad thing, then can you merge each thing into one PR per package? It's kinda hard to review at the moment.

I'm not overly enthused by having another 1000 lines of code in MOI...

PR #3063, opened 2026-08-23
[Bridges.Constraint] change conversion cost to 0.5 for ScalarAffineFunction

odow:
x-ref https://github.com/jump-dev/MathOptInterface.jl/pull/3062

https://github.com/jump-dev/MathOptInterface.jl/actions/runs/32667229192


odow:
This needs @blegat to think about the implications of.

blegat (approved):
I think it makes sense. It surprising we didn't face this issue reported by @araujoms before

PR #3068, opened 2026-09-06
[Bridges] fix SquareBridge with differing constants in off-diagonal

odow:
# `SquareBridge` `ConstraintFunction` getter has a sign error when a symmetrization constraint has a nonzero constant

`src/Bridges/Constraint/bridges/SquareBridge.jl`, `MOI.get(::MOI.ModelLike, ::MOI.ConstraintFunction, ::SquareBridge)`, line 270.

When entries `(i, j)` and `(j, i)` of the square matrix differ by a nonzero
constant `c` (e.g. `f_ij = 3 + x`, `f_ji = 2 + x`), `bridge_constraint` adds a
constraint `(f_ij - f_ji) - c == -c` (the constant is moved into the set by
`normalize_and_add_constraint`). When reconstructing `f_ji` in the
`ConstraintFunction` getter, the code subtracts `rhs` a second time instead of
adding it back, so the recovered `f_ji` is off by `2c`.

Repro (`c = 1`):
```julia
inner = MOI.Utilities.Model{Float64}()
model = MOI.Bridges.Constraint.Square{Float64}(inner)
x = MOI.add_variable(model)
y = MOI.add_variable(model)
f = MOI.Utilities.operate(vcat, Float64, 1.0x .+ 1.0, 1.0x .+ 2.0, 1.0x .+ 3.0, 2.0y)
ci = MOI.add_constraint(model, f, MOI.PositiveSemidefiniteConeSquare(2))
MOI.get(model, MOI.ConstraintFunction(), ci)
# entry (2,1) comes back as `4 + x` instead of the original `2 + x`
```

This only shows up when the off-diagonal expressions differ by a nonzero
constant (e.g. `1 + x` vs `2 + x`); the common case of numerically-identical
or purely-linear-difference-with-zero-constant entries doesn't trigger it,
which is presumably why the existing test suite (`test/Bridges/Constraint/test_SquareBridge.jl`)
didn't catch it.

## Fix

Add `rhs` instead of subtracting it:

```julia
f_ji = MOI.Utilities.operate!(+, T, f_ji, rhs)
```

See `issue-001.patch`. Verified: the repro above returns the original
function exactly after the fix, and `test/Bridges/Constraint/test_SquareBridge.jl`
still passes in full (all testsets, 367 `@test`s total).

Note: `issue-003.patch` touches a different hunk of the same file
(`SquareBridge.jl`, the `ConstraintPrimal`/`ConstraintPrimalStart` getter and
setter rather than the `ConstraintFunction` getter); they're independent
fixes and both apply cleanly together.

### DIFF
--- a/test/Bridges/Constraint/test_SquareBridge.jl
+++ b/test/Bridges/Constraint/test_SquareBridge.jl
@@ -330,6 +330,24 @@ function test_square_bridge_with_constant()
     return
 end
 
+function test_constraint_primal_start()
+    inner = MOI.Utilities.UniversalFallback(MOI.Utilities.Model{Float64}())
+    model = MOI.Bridges.Constraint.Square{Float64}(inner)
+    x = MOI.add_variables(model, 4)
+    f = MOI.Utilities.vectorize(1.0 .* x .+ (1.0:4.0))
+    ci = MOI.add_constraint(model, f, MOI.PositiveSemidefiniteConeSquare(2))
+    bridge = MOI.Bridges.bridge(model, ci)
+    start = [11.0, 21.0, 12.0, 22.0]
+    MOI.set(model, MOI.ConstraintPrimalStart(), ci, start)
+    @test isapprox(
+        MOI.get(inner, MOI.ConstraintPrimalStart(), bridge.triangle),
+        [11.0, 12.0, 22.0]
+    )
+    @test ≈(MOI.get(inner, MOI.ConstraintPrimalStart(), bridge.sym[1][2]), -9.0)
+    @test ≈(MOI.get(model, MOI.ConstraintFunction(), ci), f)
+    return
+end
+
 end  # module
 
 TestConstraintSquare.runtests()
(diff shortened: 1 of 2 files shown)


test/Bridges/Constraint/test_SquareBridge.jl
```diff
@@ -314,6 +314,40 @@ function test_VectorNonlinearFunction_mixed_type()
...
+    model = MOI.Bridges.Constraint.Square{Float64}(inner)
+    x = MOI.add_variables(model, 4)
+    f = MOI.Utilities.vectorize(1.0 .* x .+ (1.0:4.0))
+    ci = MOI.add_constraint(model, f, MOI.PositiveSemidefiniteConeSquare(2))
+    bridge = MOI.Bridges.bridge(model, ci)
+    start = [11.0, 21.0, 12.0, 22.0]
+    MOI.set(model, MOI.ConstraintPrimalStart(), ci, start)
+    @test isapprox(
+        MOI.get(inner, MOI.ConstraintPrimalStart(), bridge.triangle),
+        [11.0, 12.0, 22.0]
```
odow:
```suggestion
        [11.0, 12.0, 22.0],
```

PR #3078, opened 2026-09-08
Enforce dimension > 0 for SOS1 and SOS2

### LINKED ISSUE #3074: Dimension of SOS sets
Should we forbid dimension=0 for SOS sets? What does this even mean?

You could argue that throwing an error on `MOI.SOS1(Float64[])` is a breaking change, but I don't really set a good reason why anyone would ever want to do it. There's also an argument for making `dimension > 1` for SOS1, and `dimension > 2` for SOS2...

```Julia
julia> using JuMP

julia> model = Model();

julia> @constraint(model, VariableRef[] in SOS1())
[] ∈ MathOptInterface.SOS1{Float64}(Float64[])
```

Discuss...

odow:
Closes #3074

I probably need to run solver tests because this could arguably be breaking. But equally. Why would anyone do this.

https://github.com/jump-dev/MathOptInterface.jl/actions/runs/34176516796

### DIFF
--- a/src/sets.jl
+++ b/src/sets.jl
@@ -1853,7 +1853,7 @@ MathOptInterface.ConstraintIndex{MathOptInterface.VectorOfVariables, MathOptInte
 struct SOS1{T<:Real} <: AbstractVectorSet
     weights::Vector{T}
 
-   function SOS1{T}(weights::Vector{T}) where {T<:Real}
+    function SOS1{T}(weights::Vector{T}) where {T<:Real}
         if isempty(weights)
             throw(DimensionMismatch("Dimension of SOS1 must be > 0."))
         end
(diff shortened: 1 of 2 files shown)


src/sets.jl
```diff
@@ -1852,8 +1852,19 @@ MathOptInterface.ConstraintIndex{MathOptInterface.VectorOfVariables, MathOptInte
 """
 struct SOS1{T<:Real} <: AbstractVectorSet
     weights::Vector{T}
+
+   function SOS1{T}(weights::Vector{T}) where {T<:Real}
```
odow:
```suggestion
    function SOS1{T}(weights::Vector{T}) where {T<:Real}
```

